PR #49, opened 2016-04-28
New tilts

rcooke-ast:
Fixed a bug in the arc line tilts
Implemented new spline tilt algorithm
several minor bugs fixed


src/artrace.py
```diff
@@ -804,59 +805,116 @@ def refine_traces(binarr, outpar, extrap_cent, extrap_diff, extord, orders, disp
...
+    detuse = arcdet[aduse]
+    idxuse = np.arange(arcdet.size)[aduse]
+    olduse = aduse.copy()
+    for s in xrange(nuse):
+        w = np.where((np.abs(arcdet-detuse[s]) <= ncont) & (np.abs(arcdet-detuse[s]) >= 1.0))[0]
+        for u in xrange(w.size):
+            if ampl[w[u]] > ampl[olduse][s]:
+                aduse[idxuse[s]] = 0
+                break
+
```
profxj: Is this loop removing all lines that has at least one other within ncont or keeping the brightest one?  It seems to me the latter.

src/artrace.py
```diff
@@ -804,59 +805,116 @@ def refine_traces(binarr, outpar, extrap_cent, extrap_diff, extord, orders, disp
...
+    # Remove lines that are within ncont pixels
+    nuse = np.sum(aduse)
+    detuse = arcdet[aduse]
+    idxuse = np.arange(arcdet.size)[aduse]
+    olduse = aduse.copy()
+    for s in xrange(nuse):
+        w = np.where((np.abs(arcdet-detuse[s]) <= ncont) & (np.abs(arcdet-detuse[s]) >= 1.0))[0]
+        for u in xrange(w.size):
+            if ampl[w[u]] > ampl[olduse][s]:
+                aduse[idxuse[s]] = 0
```
profxj: Should this be False instead of 0?

src/artrace.py
```diff
@@ -1350,159 +1054,131 @@ def model_tilt(slf, det, msarc, guesstilts=None, censpec=None, plotQA=False, ref
...
-                                               maxv=msarc.shape[slf._dispaxis]-1)
-                    tilts = tiltval[:,np.newaxis].repeat(tiltang.shape[1],axis=1)
-                else:
-                    msgs.warn("There are still not enough detections to obtain a reliable tilt trace")
-                    msgs.info("Assuming there is no tilt")
-                    tilts = np.zeros_like(slf._lordloc)
+            if np.max(np.abs(zwght[1:]-zwght[:-1])) != 0.0:
+                wtilt[xint:xint+2*sz+1, j] = 1.0/np.max(np.abs(zwght[1:]-zwght[:-1]))
+
+            # Extrapolate off the chip
```
profxj: Maybe add an additional comment here explaining why this step is necessary.

src/artrace.py
```diff
@@ -934,412 +978,72 @@ def model_tilt_test(slf, det, msarc, guesstilts=None, censpec=None, plotQA=False
...
-#                plt.plot(xtfit[wmask], ytfit[wmask], 'bx')
-#                plt.show()
-#               try:
-#                   tcoeff = np.polynomial.polynomial.polyfit(xtfit[wmask],ytfit[wmask],1,w=1.0/mt)
-#               except:
-#                   tcoeff = np.polynomial.polynomial.polyfit(xtfit[wmask],ytfit[wmask],1)
-            if guesstilts is not None: ytfit[wmask] -= np.median(ytfit[wmask])
+            # Perform a robust polynomial fit to the traces
             wmsk, mcoeff = arutils.robust_polyfit(xtfit[wmask], ytfit[wmask]/(msarc.shape[0]-1.0),
                                                   slf._argflag['trace']['orders']['tiltorder'],
```
profxj: Why do both the polyscan and this polyfit?

src/artrace.py
```diff
@@ -934,412 +978,72 @@ def model_tilt_test(slf, det, msarc, guesstilts=None, censpec=None, plotQA=False
...
-    msgs.work("Detecting lines..")
-    tampl, tcent, twid, w, satsnd, _ = ararc.detect_lines(slf, det, msarc, censpec=censpec)
-    satval = slf._spect['det'][det-1]['saturation']*slf._spect['det'][det-1]['nonlinear']
-    fitxy = [slf._argflag['trace']['orders']['tiltorder'], 8]  # order of the polynomials to be used when fitting the tilts.
-    arcdet = (tcent[w]+0.5).astype(np.int)
-    maskrows = np.ones(msarc.shape[0], dtype=np.int) # Start by masking every row, then later unmask the rows with usable arc lines
-    totnum = 0
-    if np.size(w[0])>totnum:
-        totnum = np.size(w[0])
+            ytfit = (2.0*model[sz]-ytfit)
```
profxj: This is the magical new line.  ;)

src/artrace.py
```diff
@@ -1350,159 +1054,131 @@ def model_tilt(slf, det, msarc, guesstilts=None, censpec=None, plotQA=False, ref
...
+
+        if slf._argflag['trace']['orders']['tilts'].lower() == "spline":
+            xspl = np.linspace(0.0, 1.0, msarc.shape[1])
+            yspl = np.append(0.0, np.append(arcdet[np.where(aduse)]/(msarc.shape[0]-1.0), 1.0))
+            zspl = np.zeros((msarc.shape[1], np.sum(aduse)+2))
+            zspl[:, 1:-1] = mtilt[:, np.where(aduse)[0]]
+            zspl[:, 0] = zspl[:, 1] + polytilts[0, :] - polytilts[arcdet[np.where(aduse)[0][0]], :]
+            zspl[:, -1] = zspl[:, -2] + polytilts[-1, :] - polytilts[arcdet[np.where(aduse)[0][-1]], :]
+
+            tiltspl = interp.RectBivariateSpline(xspl, yspl, zspl, kx=3, ky=3)
```
profxj: I fear this still isn't quite right.  See conversation comment and let's discuss.

profxj:
I think we need to discuss the final step.  
I believe the correct approach is to do a 2D fit to:

x = slit position (as defined currently)
y = position of the arc line on the detector, i.e. model
z = position of the arc line at slit center, i.e. model[sz]

This would mean we can't use a RectBivariateSpline.
I suspect a SmoothBivariateSpline might work fine.  
In LowRedux, we use a two dimensional Legendre surface fit.  2nd or 3rd order, I think.

What we are doing isn't too far from this which is why it will
work ok, but let's do it 'right' if we can.

src/artrace.py
rcooke-ast: You're correct

src/artrace.py
rcooke-ast: It's searches the neighborhood of each detected line to see if there are other detected lines within ncont pixels. If there is, it will only take the brightest line. I'll need to think about this some more...

src/artrace.py
rcooke-ast: polyscan is used to accurately follow the trace of each arc line. These traces allow us to interpolate between the arc lines. The coefficients of polyfit are used in the PCA, and are needed to extrapolate to the edges of the chip (in the dispersion direction). See lines 1145-1146 below.

src/artrace.py
rcooke-ast: Yes, but I've since realized that we get the (almost) right result for the wrong reasons. I'll fix this in the next PR.

rcooke-ast:
You're right, we should chat about the main issue.

I'm not convinced that a polynomial fitting function is the
correct approach, since it doesn't accurately reproduce
the tilts in the spectral direction. In any case, this is pretty
much what is already available with the PCA option.

The spline (which is in fact a scanning polynomial in
the spatial direction and an interpolating spline in the
spectral direction) overcomes the fitting simplification.

However, if we get a bad trace of just one spectral
line, we're screwed with the spline approach. I thought
this might be overcome by only selecting the best lines
(see code at the beginning of the model_tilt function).

I'll try the SmoothBivariateSpline. If it's not too
computationally expensive, it might be best. I'll
keep you posted!

src/artrace.py
```diff
@@ -1350,159 +1054,144 @@ def model_tilt(slf, det, msarc, guesstilts=None, censpec=None, plotQA=False, ref
...
+            zspl = np.zeros((msarc.shape[1], np.sum(aduse)+2))
+            zspl[:, 1:-1] = mtilt[:, np.where(aduse)[0]]
+            zspl[:, 0] = zspl[:, 1] + polytilts[0, :] - polytilts[arcdet[np.where(aduse)[0][0]], :]
+            zspl[:, -1] = zspl[:, -2] + polytilts[-1, :] - polytilts[arcdet[np.where(aduse)[0][-1]], :]
+            # Make sure the endpoints are set to 0.0 and 1.0
+            zspl[:, 0] -= zspl[ordcen[0, 0], 0]
+            zspl[:, -1] = zspl[:, -1] - zspl[ordcen[-1, 0], -1] + 1.0
+            # Prepare the spline variables
+            xsbs = np.outer(xspl, np.ones(yspl.size)).flatten()
+            ysbs = np.outer(np.ones(xspl.size), yspl).flatten()
```
profxj: For the spline, I believe you want:

   yspl = np.append(0.0, polytilts[arcdet[np.where(aduse)[0][sz]], :], 1.0))
   ysbs = np.outer(np.ones(xspl.size), yspl).flatten()

Or something like that.  You want it to be the best estimate of the arc line centroid at the center of the slit in row arcdet[np.where(aduse)]

src/artrace.py
```diff
@@ -1350,159 +1054,144 @@ def model_tilt(slf, det, msarc, guesstilts=None, censpec=None, plotQA=False, ref
...
-    # Plot the model tilts
-
-#guesstilts[arcdet[j],ordcen[arcdet[j],0]-k]*float(msarc.shape[0]-1.0)
-
-
-    tiltsplot = tilts[arcdet,:].T
+            polytilts += guesstilts
+
+        if slf._argflag['trace']['orders']['tilts'].lower() in ["spline", "interp"]:
+            xspl = np.linspace(0.0, 1.0, msarc.shape[1])
```
profxj: I think this should be over a limited x range, not the full detector,
e.g.
[xint:xint+2*sz+1]

src/artrace.py
```diff
@@ -1350,159 +1054,144 @@ def model_tilt(slf, det, msarc, guesstilts=None, censpec=None, plotQA=False, ref
...
+            tilts = polytilts.copy()
+        if tt == 0:
+            ztilto = ztilt.copy()
+        elif tt < nitertilts-1:
+            # Provide a starting guess of the tilts for the next iteration
+            guesstilts = tilts.copy()
+
+    # Now do the QA
+    msgs.info("Preparing arc tilt QA data")
+    tiltsplot = tilts[arcdet, :].T
```
profxj: The QA will be goofy unless you use polyfits here.

src/artrace.py
```diff
@@ -1350,159 +1054,144 @@ def model_tilt(slf, det, msarc, guesstilts=None, censpec=None, plotQA=False, ref
...
-
-
-    tiltsplot = tilts[arcdet,:].T
+            polytilts += guesstilts
+
+        if slf._argflag['trace']['orders']['tilts'].lower() in ["spline", "interp"]:
+            xspl = np.linspace(0.0, 1.0, msarc.shape[1])
+            yspl = np.append(0.0, np.append(arcdet[np.where(aduse)]/(msarc.shape[0]-1.0), 1.0))
+            zspl = np.zeros((msarc.shape[1], np.sum(aduse)+2))
+            zspl[:, 1:-1] = mtilt[:, np.where(aduse)[0]]
```
profxj: On KastR there are some lines which are giving junk due to a small number of bad values in ytfit which is wrecking 
arcyutils.polyfit_scan_intext

Either that Cython algorithm needs to include some rejection or we may need to avoid it.

src/artrace.py
rcooke-ast: I agree. I've made a minor change, so that the yspl values are taken from the center of the slit, rather than the center of the detector

src/artrace.py
rcooke-ast: Agreed, but perhaps we should just fit the original xtilt, ytilt, and ztilt
values. At the moment, the spline is fit to the model of the tilts, but we
should be fitting the spline to the measured tilts (i.e. the original
result of the cross correlation). I'll look into this shortly...

profxj: There were a bunch of junk things in mtilt
that were crashing the Spline for Kastr and LRISr.
That pushed me on this path which looks a lot better anyhow.
I guess it feels like we are double smoothing, but I’m ok with that.


> On Apr 29, 2016, at 4:42 PM, Ryan Cooke <notifications@github.com> wrote:
> 
> In src/artrace.py <https://github.com/PYPIT/PYPIT/pull/49#discussion_r61658351>:
> 
> > -#    if refine_tilts:
> > -#        debugger.set_trace()
> > -
> > -    #xx, yy = np.meshgrid(np.arange(msarc.shape[1])/(msarc.shape[1]-1.0), np.arange(msarc.shape[0])/(msarc.shape[0]-1.0))
> > -    #coeff = arutils.polyfit2d(xx, yy, tilts, [8,8])
> > -    #tilts = arutils.polyval2d(coeff, np.arange(msarc.shape[1])/(msarc.shape[1]-1.0), np.arange(msarc.shape[0])/(msarc.shape[0]-1.0))
> > -    # Plot the model tilts
> > -
> > -#guesstilts[arcdet[j],ordcen[arcdet[j],0]-k]*float(msarc.shape[0]-1.0)
> > -
> > -
> > -    tiltsplot = tilts[arcdet,:].T
> > +            polytilts += guesstilts
> > +
> > +        if slf._argflag['trace']['orders']['tilts'].lower() in ["spline", "interp"]:
> > +            xspl = np.linspace(0.0, 1.0, msarc.shape[1])
> Agreed, but perhaps we should just fit the original xtilt, ytilt, and ztilt values. At the moment, the spline is fit to the model of the tilts, but we should be fitting the spline to the measured tilts (i.e. the original result of the cross correlation). I'll look into this shortly...
> … <x-msg://95/#>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> 
> !DSPAM:826,5723f18622588138197847! <https://github.com/PYPIT/PYPIT/pull/49#discussion_r61565690> <https://github.com/PYPIT/PYPIT/pull/49/files/7f6e721a24457cd7164550e1e1677a27d8702468#r61565690>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub <https://github.com/PYPIT/PYPIT/pull/49/files/7f6e721a24457cd7164550e1e1677a27d8702468#r61658351>
> !DSPAM:826,5723f18622588138197847!

J. Xavier Prochaska
UCO/Lick Observatory
1156 High St.
UC Santa Cruz
Santa Cruz, CA 95064

http://imps.ucolick.org/
http://www.qpqsurvey.org/
http://k1dm3.ucolick.org/

831-459-2135 (Direct)
831-459-2991 (UCO/Lick Main)
831-459-5244 (Fax)

profxj:
Merging

Issue #127, opened 2016-07-25
Syntax in .pypit input file


ntejos:
I think it would be nice to relax the syntax for the input .pypit file. Currently, I tried the following:

[...]
data read
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0082.fits #bias
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0083.fits #bias
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0084.fits #bias
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0085.fits #bias
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0086.fits #bias
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0087.fits #bias
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0088.fits #bias
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0089.fits #bias
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0028.fits # science 1413_1
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0029.fits # science 1413_1
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0020.fits # arc
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0021.fits # arc
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0022.fits # arc
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0066.fits # twilight flat
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0067.fits # twilight flat
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0068.fits # twilight flat
    /media/ntejos/disk1/fields/J1413/LRIS_mos/data/b160405_0046.fits # standard
data end
[...]

and PYPIT complains:

[ERROR]   :: There must be no spaces when specifying the datafile.

I think it is useful to allow comments at the end of lines like in my case, which helps with bookkeeping.

PR #128, opened 2016-07-26
Syntax input: comments in lines pointing to raw data.

ntejos:
syntax_input: allow for comments in the lines pointing to the raw data.

This PR addresses Issue #127, but feel free to reject it if you prefer not to have such flexibility.

Cheers,


rcooke-ast:
I like it - thanks!

Merging....

PR #129, opened 2016-07-26
syntax_input: minor bug solved.

ntejos:
There was a bug when the line was all commented out, so becomes `dfname=""`, for which `dfname[0]` does not make sense. This commit solves it.


rcooke-ast:
Thanks, Nicolas!  I've just spotted this one too, and have a slightly different (and more general) solution. I'll submit a PR now.

PR #138, opened 2016-08-24
updated SIP avoidance instructions

adwasser:

PR #139, opened 2016-08-24
Updates on installing.rst from recent installation on El Capitan

banados:


adwasser:
Given this is the same as #138, let's close?

PR #148, opened 2016-11-09
new kast red settings file

tiffanyhsyu:


rcooke-ast:
I'll make some changes to the branch (if that's OK?) before merging to make the settings file consistent with the new parser. Otherwise, you will need to compare the old kast settings file with the new one to see which keyword arguments have been changed.

rcooke-ast:
Thanks for making the updates Tiffany! This looks good to me.... merging!

By the way, it would be good to have an example provided in the development suite, just to make sure future changes to the code support the new KASTr detector.

PR #192, opened 2017-03-02
Have made some changes to accomdate the 600/10000 grating

AlexaVillaume:


pypit/ararc.py
```diff
@@ -211,11 +211,11 @@ def setup_param(slf, sc, det, fitsdict):
         else:
             msgs.error('Not ready for this disperser {:s}!'.format(disperser))
     elif sname=='lris_blue':
-        lamps = ['ZnI','CdI','HgI']
+        lamps = ['NeI', 'ArI', 'CdI', 'KrI', 'XeI', 'ZnI','CdI','HgI']
```
profxj: I don't think this is s a good idea as the default for LRISb.
I think we'll need to allow an optional list.  @AlexaVillaume , do
you want to try to negotiate arparse? :)    Probably not.

I suggest we meet on Friday to do this together.

pypit/ararc.py
```diff
@@ -538,6 +543,7 @@ def calib_with_arclines(slf, det, get_poly=False, use_basic=False):
     msgs.work("Detecting lines..")
     tampl, tcent, twid, w, satsnd, spec = detect_lines(slf, det, slf._msarc[det-1])
 
+    use_basic = True
```
profxj: This should not be set to True here.

pypit/arflux.py
```diff
@@ -87,7 +87,7 @@ def bspline_magfit(wave, flux, var, flux_std, nointerp=False, **kwargs):
     -------
     """
     from pypit import arutils
-    invvar = (var > 0.)/(var + (var <= 0.))
+    invvar = (var > 0.)/(np.max(var,0))
```
profxj: Well done!  I was trying to think of a cleaner way to do this.

rcooke-ast: I agree that this is better, but I think we might also need to account for the case when `var=0` (i.e. in this case `invar=0/0`)? Rarely will this be the case, but perhaps we can have another line straight after this with: `invvar[np.where(var==0.0)] = 0.0`

pypit/ararc.py
AlexaVillaume: I can meet on Friday. Let me know what time works for you.

pypit/arflux.py
AlexaVillaume: Credit for the new line goes to @tiffanyhsyu. 

Even after this change I'm still having issues with the bspline with my red side data so I think this part of the code still needs some work. I'm still working on understanding what's going wrong but maybe @rcooke-ast suggestion might help?

tiffanyhsyu: It was X's idea :P

I'd be interested to sit in on any discussion about this /generating the sensitivity function, if it happens!

AlexaVillaume:
I've made the changes you requested and committed them. I think you have to review them now but correct me if there's something else I need to do.

We made the changes for argparser to accept arc information on your local version of PYPIT.

profxj:
Thanks!

X

profxj:
merging.

PR #212, opened 2017-05-01
Update installing.rst

adwasser:
Note for installing for a python distribution installed where root permission is required.


rcooke-ast:
Seems fine to me - thanks!  Merging...

Issue #242, opened 2017-05-31
Co-addition error


SunilSimha:
I'm trying to co-add the spectra of two objects but I'm hitting a Pdb flag and continuation leads it to crash. I'm including the fits files and the .yaml file [here](https://www.dropbox.com/sh/a8bo04afwpm24u3/AADagyU1L5EwhgR6RV4l1WRga?dl=0). Here is the ouput from the terminal.
`
$pypit_coadd_1dspec qso_add_1d.yaml 
[INFO]    :: Loading extension 6 of spectrum /home/sunil/Desktop/WorkingDir/Keck/LRIS/2017mar19/Run_PYPIT/lris_red_setup_A/Science/spec1d_J092_J0927+2530_LRISr_2017Mar19T053722.fits
> /home/sunil/anaconda3/lib/python3.6/site-packages/linetools/spectra/xspectrum1d.py(230)__init__()
-> self.data[key][kk][gdsigval[-1]+1:].mask = True
(Pdb) c
Traceback (most recent call last):
  File "/home/sunil/anaconda3/bin/pypit_coadd_1dspec", line 11, in <module>
    coadd_1dspec.main(args)
  File "/home/sunil/anaconda3/lib/python3.6/site-packages/pypit/scripts/coadd_1dspec.py", line 105, in main
    spectra = arcoadd.load_spec(gdfiles, iextensions=extensions, extract=ex_value, flux=flux_value)
  File "/home/sunil/anaconda3/lib/python3.6/site-packages/pypit/arcoadd.py", line 504, in load_spec
    spectrum = arload.load_1dspec(fname, exten=extensions[ii], extract=extract, flux=flux)
  File "/home/sunil/anaconda3/lib/python3.6/site-packages/pypit/arload.py", line 408, in load_1dspec
    spec = XSpectrum1D.from_file(fname, exten=exten, **rsp_kwargs)
  File "/home/sunil/anaconda3/lib/python3.6/site-packages/linetools/spectra/xspectrum1d.py", line 67, in from_file
    slf = tio.readspec(ifile, **kwargs)
  File "/home/sunil/anaconda3/lib/python3.6/site-packages/linetools/spectra/io.py", line 116, in readspec
    xspec1d = parse_FITS_binary_table(hdulist, exten=exten, **kwargs)
  File "/home/sunil/anaconda3/lib/python3.6/site-packages/linetools/spectra/io.py", line 567, in parse_FITS_binary_table
[...]

PR #243, opened 2017-06-02
Coadd_extract

SunilSimha:
A simple amendment to automatically check if optimal extraction was not present and hence use boxcar while warning the user of the action.

### DIFF
--- a/pypit/scripts/coadd_1dspec.py
+++ b/pypit/scripts/coadd_1dspec.py
@@ -6,7 +6,7 @@ Wrapper to the linetools XSpecGUI
 from pypit import pyputils
 from pypit import armsgs
 msgs = pyputils.get_dummy_logger()
-
+from numpy import isnan
 
 def parser(options=None):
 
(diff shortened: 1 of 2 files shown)


pypit/scripts/coadd_1dspec.py
```diff
@@ -95,6 +95,12 @@ def main(args, unit_test=False, path=''):
             if mtch_obj is None:
                 print("No object {:s} in file {:s}".format(iobj, fkey))
             elif len(mtch_obj) == 1:
+                #Check if optimal extraction is present in all  objects. If not, warn the user and set ex_value to 'box'.
```
profxj: Is it whether the object is present or whether it is full of
NAN?  Or, is NAN what we get when optimal extraction
is not performed?

pypit/scripts/coadd_1dspec.py
```diff
@@ -95,6 +95,12 @@ def main(args, unit_test=False, path=''):
...
+                hdulist = fits.open(fkey)
+                obj_opt_flam = hdulist[mtch_obj[0]].data['OPT_FLAM']
+                if any(isnan(obj_opt_flam)):
```
profxj: Can you use np.any here?  I'm not familiar with
Python's any.

profxj:
Thanks for this @SunilSimha .  Have put in 2 minor suggestions
for edits.

Addresses Issue #242

SunilSimha:
@profxj, np.isnan returns a boolean array. It is true if the corresponding
array element is NaN. any() returns true if at least one element of an
array is true. So the code checks if there is at least one NaN in the opt
extraction and changes the method to boxcar if true. Is np.any
different/faster?

Sunil

profxj:
Ok, Python's *any* sounds fine.

But how about my main query is whether 
the optimal extraction should be missing 
altogether instead of being NAN ?

SunilSimha:
I don't really know how the 1dspec files are generated. As I understand,
the files will have a NaN array in OPT_FLAM in case the optimal extract was
not produced. I haven't come across a case where the OPT_FLAM array doesn't
exist at all. Is this possible? If so, I can modify the code suitably.

Sunil

profxj:
Looking at the code in arsave.py, it seems possible to not have
either array (box or optimal).  So, since we are here, please add
in a 'catch' for cases without one or the other.    Thanks!

SunilSimha:
@profxj, I've added a try/except statement to catch that error. Is this alright?

profxj:
Looks great.  Am merging

PR #305, opened 2018-03-04
removing cython

kbwestfall:
This removes a few cython functions:

- removed dependency on pypit/arcyproc.pyx cr_screen, added new_cr_screen function to pypit/arproc.py
- removed dependency on pypit/arcytrace.pyx ignore_orders, added new_ignore_orders to pypit/artrace.py
- removed dependency on pypit/arcytrace.pyx find_objects, added new_find_objects to pypit/artrace.py

Codes passes 

python3 pypit_tests all

in PYPIT-development-suite.  Python 2 not tested (but should pass I think).

### DIFF
--- a/pypit/arproc.py
+++ b/pypit/arproc.py
@@ -22,6 +22,9 @@ from pypit import ardebug as debugger
 # Logging
 msgs = armsgs.get_logger()
 
+#KBW TESTING
+import time
+
 
 def background_subtraction(slf, sciframe, varframe, slitn, det, refine=0.0):
     """ Generate a frame containing the background sky spectrum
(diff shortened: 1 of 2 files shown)


rcooke-ast:
This looks good to me. Just a few things to check.

(1) I presume that the returned arrays were identical in python and cython?

(2) Do you have an idea of how the runtimes suffered (I see that you benchmarked the routines)

(3) Before merging, we should delete the cython code, and remove the "new_" prefixes to the python functionality.

Thanks for the new implementations, everything looks fine by me.

kbwestfall:
(1) I only checked 2 of the 9 test reductions to make sure that the returned arrays.  I can check this tomorrow.
(2) The comparisons vary depending on the test reduction, here are some representative examples:

    Old cr_screen: 4.102992 seconds
    New cr_screen: 0.3256830000000015 seconds

    Old ignore_orders: 0.010796000000000916 seconds
    New ignore_orders: 0.17365799999999965 seconds

    Old find_objects: 5.9000000000253294e-05 seconds
    New find_objects: 0.00022400000000288856 seconds

So there are actually gains in the new cr_smooth, but about order of magnitude losses in the other two.

(3) Agreed, but I wanted to wait until I had all the replacement code in place and tested before renaming the pure python functions and removing the cython code.  I can just keep working in my fork until I'm ready for that.  I'll bring this up with X when I talk to him tomorrow.

pypit/arproc.py
```diff
@@ -1721,6 +1752,50 @@ def lacosmic(slf, fitsdict, det, sciframe, scidx, maxiter=1, grow=1.5, maskval=-
...
+        is robustly determined using the median absolute deviation,
+        :math:`sigma_j = 1.4826 MAD`.
+
+    Raises:
+        ValueError: Raised if the input array is not two dimensional or
+            if the selected axis is neither 0 nor 1.
+    """
+    # Check input
+    if len(a.shape) != 2:
+        raise ValueError('Input array must be two-dimensional.')
```
rcooke-ast: In three cases of this PR, a ValueError is raised. Should we instead take advantage of the msgs.error() function?

pypit/arproc.py
```diff
@@ -1721,6 +1752,50 @@ def lacosmic(slf, fitsdict, det, sciframe, scidx, maxiter=1, grow=1.5, maskval=-
...
+    if spatial_axis not in [0,1]:
+        raise ValueError('Spatial axis must be 0 or 1.')
+
+    # Mask the pixels equal to mask value: should use np.isclose()
+    _a = np.ma.MaskedArray(a, mask=(a==mask_value))
+    # Get the median along the spatial axis
+    meda = np.ma.median(_a, axis=spatial_axis)
```
rcooke-ast: Interesting... The original reason I wrote cython was because numpy's masked median was buggy and *very* slow on large 3D arrays. This was my only hesitation for keeping some cython. They must have fixed it  :-)

pypit/artrace.py
```diff
@@ -698,6 +706,111 @@ def trace_object(slf, det, sciframe, varframe, crmask, trim=2,
     return tracedict
 
 
+def new_ignore_orders(edgdet, fracpix, lmin, lmax, rmin, rmax):
```
rcooke-ast: Not a big deal, but a lot of PYPIT's functions are written in alphabetical order, for somewhat easier navigation. I can think of reasons why we should rethink this approach though (i.e. group certain functions together).

pypit/arproc.py
kbwestfall: Not a problem.  I'm not yet up to speed on pypit's exception handling code, but this is an easy change.

pypit/arproc.py
kbwestfall: Yeah, they must have fixed this (I've only been coding in python for about 3 years).  I'm not sure I ever benchmarked it in detail, but I've never noticed a significant difference between masked and unmasked arrays.

pypit/artrace.py
kbwestfall: I didn't really put any thought into the placement of these functions yet, but I can follow suit on this convention.

rcooke-ast:
Once your happy with how the tests pass, feel free to merge - thanks again!

kbwestfall:
I'm closing this pull request because it's outdated.  I'll be making another one soon to pull the additional cython removals made in my kbwdev branch.

PR #306, opened 2018-03-22
Python replacements for most cython code

kbwestfall:
This follows my initial pull request from my fork.  Here's a summary of what I've done with the cython code so far:

Obsolete with new python functions:
 - arcyarc.pyx
 - arcycomb.pyx
 - arcyproc.pyx
 - arcyutils.pyx

Contains functions that were not rewritten:
 - arcytrace.pyx 
    
Functions called but not touched by pypit_test:
 - arcytrace.close_edges
 - arcytrace.close_slits
 - arcytrace.dual_edge
 - arcytrace.expand_slits

arutils.gauss_lsqfit and arutils.gauss_fit call arcyarc.fit_gauss, but this cython function doesn't exist.

```
cython function             python function                     Check
---------------------------------------------------------------------
arcyarc.order_saturation    ararc.new_order_saturation          exact
arcyarc.saturation_mask     ararc.new_saturation_mask           exact
...                         ararc.determine_saturation_region   ...
...                         ararc.search_for_saturation_edge    ...

arcycomb.masked_limitget    ...                                 ... 
arcycomb.masked_limitset    ...                                 ... 
arcycomb.masked_limitsetarr ...                                 ... 
arcycomb.masked_mean        ...                                 ...
arcycomb.masked_median      ...                                 ...
arcycomb.masked_replace     ...                                 ...
arcycomb.masked_weightmean  ararc.new_masked_weightmean         numerical
arcycomb.maxnonsat          ararc.new_maxnonsat                 (1)
arcycomb.mean               ...                                 ...
arcycomb.median             ...                                 ...
arcycomb.minmax             ...                                 ...

arcyproc.cr_screen          arproc.new_cr_screen                (2)

arcytrace.find_between      artrace.new_find_between            exact
arcytrace.find_objects      artrace.new_find_objects            exact
arcytrace.find_peak_limits  artrace.new_find_peak_limits        exact
arcytrace.find_shift        artrace.new_find_shift              exact
arcytrace.ignore_orders     artrace.new_ignore_orders           exact
arcytrace.limit_yval        artrace.new_limit_yval              exact
arcytrace.locate_order      arproc.new_locate_order             exact
arcytrace.match_edges       artrace.new_match_edges             exact   *
arcytrace.minbetween        artrace.new_minbetween              exact
arcytrace.phys_to_pix       artrace.new_phys_to_pix             exact
arcytrace.tilts_image       artrace.new_tilts_image             exact   *

arcyutils.func2d_fit_val    arextract.new_func2d_fit_val        (3)
arcyutils.grow_masked       arproc.new_grow_masked              exact   *
arcyutils.mean_weight       artrace.new_mean_weight             (4)
arcyutils.order_pixels      arproc.new_order_pixels             exact
arcyutils.smooth_x          artrace.new_smooth_x                (4)
...                         filter.BoxcarFilter                 (4)
```

Checks:
---------
 - 'exact': the consistency check use == to check if the data returned by the cython and python functions are the same.
 - 'numerical': the check is to numerical accuracy.
 - '...': the function was removed and/or replaced with a numpy function.

Notes:
--------

1. There's a bug in the old arcycomb.maxnonsat function.  Because of this, I turned off the exact check of the difference between arcycomb.maxnonsat and arcomb.new_maxnonsat.

2. The arcyproc.cr_screen function can yield nan pixels, but arproc.new_cr_screen does not.  This means there are differences between the returned arrays.  However, in all the cases checked so far, the two arrays are identical for all non-nan pixels.  May need to keep checking this differences.

3. There can be substantial differences between the cython and python versions of func2d_fit_val.  This needs to be tested to see if the results of the new function are satisfactory.

4.  I first recoded arcyutils.smooth_x into artrace.new_smooth_x, but it is very slow (~10 seconds).  I switched to using a function from the MaNGA DAP that I wrote, which is much faster.  Instead of copying over the class (BoxcarFilter), I just copied the entire file for the time being.  There are license differences that we should be aware of.  BoxcarFilter can handle masked arrays, weighting, and rejection. However, it will provide different results from smooth_x.  One main difference is that BoxcarFilter can't do the same rejection procedure that's done in smooth_x (rejection of the highest and lowest n values in the boxcar window).  It does a sigma rejection instead.  We need to test to make sure that the differences between the two algorithms are acceptable and determine what rejection sigma is best when it's needed.

Execution times:
-------------------

For most of these functions, the python equivalent is about an order of magnitude slower than its cython counterpart.  In most cases, this means an execution time that goes from a few hundredths to a few tenths of a second.  The exceptions to this are

artrace.new_match_edges
artrace.new_tilts_image
arproc.new_grow_masked

where the execution time can be more like 100 times slower, going from a few hundredths of a second to ~1 second.  I tried to speed these up, but the specifics of the operations can be difficult to capture without the detailed looping and conditional statements.  We can revisit these when I better understand the limitations of algorithmic changes that might better enable a matrix speed up.

To do in a future PR:
------------------------
 - Ryan says I need DEIMOS data to test `arcytrace.close_edges`, `arcytrace.close_slits`, and `arcytrace.dual_edge`.  I've left these in until I can test replacements.
 - Ryan says I should be able to test `arcytrace.expand_slits` with the `apf_levy.pypit` in the dev suite, but I get an error (issue to be submitted).
 - I've left all the cython code in place, and left the python replacements to have names starting with `new_` just for the time being.  As soon as all of the cython is replaced and the replacements have been tested, I can remove the cython code and dependencies.

### DIFF
--- a/pypit/ararc.py
+++ b/pypit/ararc.py
@@ -77,23 +77,23 @@ def detect_lines(slf, det, msarc, censpec=None, MK_SATMASK=False):
     if MK_SATMASK:
         ordwid = 0.5*np.abs(slf._lordloc[det-1] - slf._rordloc[det-1])
         msgs.info("Generating a mask of arc line saturation streaks")
-        print('calling saturation_mask')
-        t = time.clock()
-        _satmask = arcyarc.saturation_mask(msarc, slf._nonlinear[det-1])
-        print('Old saturation_mask: {0} seconds'.format(time.clock() - t))
-        t = time.clock()
+#        print('calling saturation_mask')
+#        t = time.clock()
+#        _satmask = arcyarc.saturation_mask(msarc, slf._nonlinear[det-1])
+#        print('Old saturation_mask: {0} seconds'.format(time.clock() - t))
+#        t = time.clock()
         satmask = new_saturation_mask(msarc, slf._nonlinear[det-1])
-        print('New saturation_mask: {0} seconds'.format(time.clock() - t))
-        assert np.sum(_satmask != satmask) == 0, 'Difference between old and new saturation_mask'
+#        print('New saturation_mask: {0} seconds'.format(time.clock() - t))
+#        assert np.sum(_satmask != satmask) == 0, 'Difference between old and new saturation_mask'
 
-        print('calling order_saturation')
-        t = time.clock()
-        _satsnd = arcyarc.order_saturation(satmask, ordcen, (ordwid+0.5).astype(np.int))
-        print('Old order_saturation: {0} seconds'.format(time.clock() - t))
-        t = time.clock()
+#        print('calling order_saturation')
+#        t = time.clock()
+#        _satsnd = arcyarc.order_saturation(satmask, ordcen, (ordwid+0.5).astype(np.int))
+#        print('Old order_saturation: {0} seconds'.format(time.clock() - t))
+#        t = time.clock()
         satsnd = new_order_saturation(satmask, ordcen, (ordwid+0.5).astype(np.int))
-        print('New order_saturation: {0} seconds'.format(time.clock() - t))
-        assert np.sum(_satsnd != satsnd) == 0, 'Difference between old and new order_saturation'
+#        print('New order_saturation: {0} seconds'.format(time.clock() - t))
+#        assert np.sum(_satsnd != satsnd) == 0, 'Difference between old and new order_saturation'
     else:
         satsnd = np.zeros_like(ordcen)
     # Detect the location of the arc lines
--- a/pypit/ararc.py
+++ b/pypit/ararc.py
@@ -10,6 +10,7 @@ from pypit import ararclines
 from pypit import arqa
 from matplotlib import pyplot as plt
 import os
+import time
 
 from pypit import ardebug as debugger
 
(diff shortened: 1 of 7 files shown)


pypit/ararc.py
```diff
@@ -76,8 +77,23 @@ def detect_lines(slf, det, msarc, censpec=None, MK_SATMASK=False):
     if MK_SATMASK:
         ordwid = 0.5*np.abs(slf._lordloc[det-1] - slf._rordloc[det-1])
         msgs.info("Generating a mask of arc line saturation streaks")
-        satmask = arcyarc.saturation_mask(msarc, slf._nonlinear[det-1])
-        satsnd = arcyarc.order_saturation(satmask, ordcen, (ordwid+0.5).astype(np.int))
+#        print('calling saturation_mask')
```
rcooke-ast: In the description of the PR, you mention that you'll remove cython code completely before merging and remove the prefix 'new_' for all the python functions. I think it's also a good idea to remove the commented code that benchmarks the routines, which I'm sure you had planned to do!

kbwestfall: For this PR, I was planning to commit it as is.  I wanted to remove all the cython code in one fell swoop, meaning that I want to wait until I can tests the remaining few functions.  I thought it would also be useful to have some period of time where people can use the new python code and see if there are any problems that weren't encountered by the tests in the dev suite.  So it wasn't my intention to remove the cython code, rename the python functions, and clean the code of the commented test code in this PR, but I was leaving this for one down the road.  Does that make sense, or should I go ahead and do all of this at least for the cython code that already have replacements?

rcooke-ast:
OK - that all sounds good to me!

Let's merge as is (if you're happy with it!) and we can make the final minor changes to names and comments later. Thanks again for a solid PR!

kbwestfall:
Great.  Thanks, Ryan.  I've now merged my kbwdev branch into master.

PR #366, opened 2018-06-29
Initial step for refactoring settings

kbwestfall:
This is an initial step toward refactoring settings.  This introduces `PypitPar`, but the code itself doesn't actually use `PypitPar` yet.  This PR is mainly just to introduce it into the code so that we can iterate on its structure (or scrap it).  I've included a `PypitPar.ipynb` notebook that you can look through to get a sense of what this does.  For the PR, you really only need to care about that notebook and the two main python files: `pypit/par/parset.py` (a copy from the MaNGA DAP) and `pypit/par/pypitpar.py`.

To see what this might look like in action, I've created a new branch `traceslitpar_test` and changed the `TraceSlits` class so that it now uses `TraceSlitsPar`.  Have a look [here](https://github.com/PYPIT/PYPIT/blob/traceslitpar_test/pypit/traceslits.py).  Putting together this demo meant that I had to propagate some changes to `pypit/core/artraceslits.py` and `pypit/arms.py`, which you can also have a look at.  I haven't tested these changes at all yet...

If this looks like a path we want to follow, we should merge this branch into master and start integrating.  For example, I would actually move the definition of `TraceSlitsPar` into `pypit/traceslits.py`.  Then there's a lot more work to integrate this into the front end...

The main thing still missing that `PypitPar` doesn't handle is the definition of where the data.  This should not be difficult to add, and I can still keep it to one user-level pypit configuration file.


profxj:
@rcooke-ast and @jhennawi -- Have had a close look at this and am quite keen on it.
I will proceed with making specific edit requests shortly and I ask that you also sign off
before we really begin implementing it.

profxj:
Here is a link to the Notebook, @jhennawi 

https://github.com/PYPIT/PYPIT/blob/kbwdev/doc/nb/PypitPar.ipynb

profxj:
The 'basespect' and 'basearg' settings would disappear,
replaced by the construct in par/pypitpar.py

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+        return iter(self.data.values())
+
+
+    def __repr__(self):
+        """Return a string representation of the parameters."""
+        return self._output_string(header=self.cfg_section)
+
+    
+    def _output_string(self, header=None):
+        additional_par_strings = []
```
profxj: Please add a doc string.

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+        # Delimiter
+        row_string[1] = '-'*len(row_string[0])
+        for i in range(2,nrows+1):
+            row_string[i] = '  '.join([ data_table[i-1,j].rjust(col_width[j]) 
+                                                                        for j in range(ncols)])
+        return '\n'.join(row_string)+'\n'
+
+
+    @staticmethod
+    def _data_string(data):
```
profxj: Short doc strings for these methods too.
thanks.

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+
+
+class ParDatabase(object):
+    """
+    Class used as a list of ParSets in a glorified structured numpy
+    array.
+
+    Very similar to yanny when converted to a numpy array.
+
+    Can be initialized using a list of ParSet objects, or an SDSS
```
profxj: SDSS?  :)

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
+"""
+Defines parameter sets used to set the behavior for core pypit
+functionality.
+
+There are two main parameter set groups, those used for reducing the
+data:
+
+    - RunPar: Parameters specific to a given execution of PypIt
+
+    - ReducePar: Parameters the define how PypIt should perform the
```
profxj: I like this spelling of PypIt.

We should settle on one.

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+
+    - WavelengthSolutionPar: Parameters used for constructing the wavelength
+      solution
+
+    - TraceSlitPar: Parameters for tracing slits
+        - PCAPar: PCA parameters
+
+    - TraceTiltsPar: Parameters for tracing tilts of arc lines
+
+    - TraceObjectsPar: Parameter for tracing objects
```
profxj: Parameters

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+from __future__ import unicode_literals
+
+import sys
+import os
+import glob
+import warnings
+import numpy
+import inspect
+
+from configobj import ConfigObj
```
profxj: Make sure to add these dependency to our Install
and requirements.txt

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+# Helper functions
+
+def _pypit_root_directory():
+    """
+    Determine the root directory for the PYPIT source code.
+
+    .. todo::
+        - Set this in __init__.py
+    """
+    for p in sys.path:
```
profxj: pretty sure this works without fail:

```
from pkg_resources import resource_filename
p = resource_filename('pypit', '')
```

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+    raise OSError('Could not find PYPIT in system path.')
+
+
+def _eval_ignore():
+    """Provides a list of strings that should not be evaluated."""
+    return [ 'open', 'file', 'dict' ]
+
+
+def _recursive_dict_evaluate(d):
+    """
```
profxj: Not too sure what this method does.
Please add more comments (and define d)

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        else:
+            try:
+                d[k] = eval(d[k]) if d[k] not in ignore else d[k]
+            except:
+                pass
+
+    return d
+
+
+def _get_parset_list(cfg, pk, parsetclass):
```
profxj: doc string.

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+            srt = numpy.argsort(order)
+            if numpy.any(numpy.array(order)[srt]-1 != numpy.arange(order[srt[-1]])):
+                raise ValueError('Parameter set series must be sequential and 1-indexed.')
+            return [par[i] for i in srt]
+        return None
+
+#-----------------------------------------------------------------------------
+# Reduction ParSets
+
+class FrameGroupPar(ParSet):
```
profxj: Am not sure we need this Parset anymore.
@rcooke-ast can confirm.

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        """
+        Return the list of valid frame types.
+        """
+        return [ 'bias', 'pixelflat', 'arc', 'pinhole', 'trace', 'standard', 'science' ]
+
+    def validate(self):
+        pass
+
+
+class CombineFramesPar(ParSet):
```
profxj: I've also managed to munge this capability in PYPIT.
But we could bring it back.

profxj: I should say the matching.
Not the combining.

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        dtypes['method'] = str
+        descr['method'] = 'Method used to flat field the data; use None to skip flat-fielding.  ' \
+                          'Options are: None, {0}'.format(', '.join(options['method']))
+
+        defaults['params'] = 20
+        dtypes['params'] = [int, list]
+        descr['params'] = 'Flat-field method parameters.  For \'PolyScan\', set params = order, ' \
+                          'numPixels, repeat ; for bspline, set params = spacing '
+
+        # TODO:  How is twodpca used?  Is it just another method that is
```
profxj: Good question for @rcooke-ast

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+    @classmethod
+    def from_dict(cls, cfg):
+        k = cfg.keys()
+        parkeys = [ 'method', 'maxshift', 'spectrum' ]
+        kwargs = {}
+        for pk in parkeys:
+            kwargs[pk] = cfg[pk] if pk in k else None
+        return cls(**kwargs)
+
+    @staticmethod
```
profxj: These should be removed, right?

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+
+        options['pipeline'] = ReducePar.valid_pipelines()
+        dtypes['pipeline'] = str
+        descr['pipeline'] = 'Pipeline options that pypit can use for reductions.  ' \
+                            'Options are: {0}'.format(', '.join(options['pipeline']))
+
+        dtypes['detnum'] = int
+        descr['detnum'] = 'Restrict reduction to a single detector with this index'
+
+        options['masters'] = ReducePar.allowed_master_options()
```
profxj: Let's make Masters its own Parset.

See the MasterFrame class for what I had in mind.

And I need to distinguish between Setup A which specifies
the overall setup and setup = A_01_aa which specifies detector
and the specific calibration set.  Maybe

SetupID

and 

setup

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        options['method'] = WavelengthSolutionPar.valid_methods()
+        dtypes['method'] = str
+        descr['method'] = 'Method to use to fit the individual arc lines.  ' \
+                          '\'fit\' is likely more accurate, but \'simple\' uses a polynomial ' \
+                          'fit (to the log of a gaussian) and is fast and reliable.  ' \
+                          '\'arclines\' uses the arclines python package.' \
+                          'Options are: {0}'.format(', '.join(options['method']))
+
+        options['lamps'] = WavelengthSolutionPar.valid_lamps()
+        dtypes['lamps'] = [str, list]
```
profxj: Probably should insist on a list.
Even if only a single lamp.

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+
+        # Fill out parameter specifications.  Only the values that are
+        # *not* None (i.e., the ones that are defined) need to be set
+
+        defaults['idsonly'] = False
+        dtypes['idsonly'] = bool
+        descr['idsonly'] = 'Only use the arc lines that have an identified wavelength to trace ' \
+                           'tilts'
+
+        defaults['method'] = 'spline'
```
profxj: The spline and PCA have been turned off.

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        defaults['order'] = 1
+        dtypes['order'] = int
+        descr['order'] = 'Order of the polynomial function to be used for the tilt of an ' \
+                         'individual arc line.  Must be 1 for eschelle data (ARMED pipeline).'
+
+        defaults['disporder'] = 1
+        dtypes['disporder'] = int
+        descr['disporder'] = 'Order of the polynomial function to be used to fit the tilts ' \
+                             'along the dispersion direction.  Only used by ARMED pipeline.'
+
```
profxj: See WaveTilts for a few more to add in here.

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        """
+        Return the valid methods to use for tilt tracing.
+        """
+        return [ 'pca', 'spca', 'spline', 'interp', 'perp', 'zero' ]
+
+    def validate(self):
+        pass
+
+
+class TraceObjectsPar(ParSet):
```
profxj: Object tracing is currently being overhauled  ;)

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        defaults['method'] = 'pca'
+        options['method'] = TraceObjectsPar.valid_methods()
+        dtypes['method'] = str
+        descr['method'] = 'Method to use for tracing each object; only used with ARMED pipeline.' \
+                          '  Options are: {0}'.format(', '.join(options['method']))
+
+        defaults['params'] = [1, 0]
+        dtypes['params'] = [int, list]
+        descr['params'] = 'Parameters for the requested method.  For pca, params is a list ' \
+                          'containing the order of the polynomials that should be used to fit ' \
```
profxj: I suppose camera specific, but it is nice to have this
in the DetectorPar

profxj: Not sure we can have meaningful defaults here.

profxj: It will be useful to add at the top of this module a short HOWTO
list on how to add a new parameter, or a new default, or a new 
Parset.

profxj:
I like @jhennawi 's suggestion to match Par sets with the Calibration classes (i.e. calibration steps).
I think that can be done.

I don't, however, follow his suggestion on FramePar.  While I find that construct to be a bit klunky, it is working ok to parse the Header.  Not sure a code-oriented approach would be any cleaner.  The LRIS and DEIMOS code don't touch the header at all (only to read the data).

doc/nb/PypitPar.ipynb
```diff
@@ -0,0 +1,1643 @@
...
+   ],
+   "source": [
+    "pypitpar.TraceObjectsPar()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Object extraction\n",
```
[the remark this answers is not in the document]
kbwestfall: Definitely.  Many changes will be needed once we integrate this into the existing code.  I'm happy to help make the appropriate changes.

doc/nb/PypitPar.ipynb
```diff
@@ -0,0 +1,1643 @@
...
+    "print('Number of detectors: {0}'.format(len(p['detector'])))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Instrument fits files\n",
+    "\n",
+    "The `FrameFitsPar` provides the list of instrument-specific fits file parameters needed to read the data for *any* fits file from the instrument."
```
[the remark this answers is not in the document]
kbwestfall: I don't know the data or the details of what data is needed for pypit well enough to know if this approach is "not viable."  I'm of two minds in terms of having instrument-specific code vs. having instrument-specific configurations that interface with a general reader.  More on this in other threads.

doc/nb/keck_lris_blue_long_400_3400_d560.cfg
```diff
@@ -0,0 +1,37 @@
+[run]
+    ncpus = 1
+    verbosity = 2
+    overwrite = True
+    sortdir = lris_blue_long_400_3400_d560
+
+[rdx]
```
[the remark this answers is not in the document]
kbwestfall: The reason I made this 'rdx' instead of 'reduce' is because there is a python method 'reduce', and I wanted to avoid any complications.  I can change this back to reduce to test if any complications actually exist, but I try to avoid giving anything the same name as a python system object.

pypit/arparse.py
```diff
@@ -314,7 +314,7 @@ def ingest(dct, upd):
         # Update the master dictionary
         self._argflag = ingest(dstr, udct).copy()
 
-    def arc_combine_match(self, v=-1.0):
+    def arc_combine_match(self, v=-1.0):    #
```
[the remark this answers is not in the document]
kbwestfall: Yeah, sorry, that's exactly right.  I was just trying to keep track of what I had and had not included in the new parameter sets.  I will remove these.

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+import warnings
+import textwrap
+import sys
+if sys.version > '3':
+    long = int
+
+
+import numpy
+from astropy.table import Table
+import pandas
```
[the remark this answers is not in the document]
kbwestfall: Nope.  I was testing using this as a way of formatting the parset output without me having to roll my own.  In the end, I just wrote my own.

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+
+        # Set the calling flags
+        self.can_call = dict([ (p, t) for p, t in zip(pars, _can_call) ])
+
+        # Set the calling flags
+        self.descr = dict([ (p, t) for p, t in zip(pars, _descr) ])
+
+        # Set the data dictionary using the internal functions
+        self.data = {}
+        for p, d, v in zip(pars, _defaults, _values):
```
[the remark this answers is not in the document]
kbwestfall: Done.

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+            if v is None:
+                self.__setitem__(p, d)
+                continue
```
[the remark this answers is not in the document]
kbwestfall: If you see this as easier to read, I can add the change.  I do this kind thing --- using `continue`, `break`, `return` to interrupt logic as a convenience to minimizing the need for `elif` and `else` --- all the time.  I find it cleaner, but will relent if you think it affects readability.

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+            self.__setitem__(p, v)
+
+        # Save the configuration file section details
```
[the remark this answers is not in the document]
kbwestfall: The writer --- the combination of `config_lines` and `to_config` --- works recursively, naming the section of the configuration after its keyword.  This doesn't work at the very top level, so these two things add the top-level section name and comment, if they're needed.  See the `to_config` function.

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+            self.data[key] = value
+            return
+
+        if isinstance(value, list):
+            is_parset_or_dict = [ isinstance(v, (ParSet, dict)) for v in value ]
+            if numpy.any(is_parset_or_dict) and not numpy.all(is_parset_or_dict):
+                warnings.warn('List includes a mix of ParSet and dicts with other types.  '
+                              'Displaying and writing the ParSet will not be correct!')
+
+        if self.options[key] is not None and value not in self.options[key]:
```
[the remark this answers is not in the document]
kbwestfall: Yep, that's right.  E.g., you don't want to have to specify all the float values a given parameter can have :)

pypit/par/parset.py
```diff
@@ -0,0 +1,649 @@
...
+            raise ValueError('Input value for {0} invalid: {1}.\nOptions are: {2}'.format(
+                                                                    key, value, self.options[key]))
+        if self.dtype[key] is not None \
+                and not any([ isinstance(value, d) for d in self.dtype[key]]):
+            raise TypeError('Input value for {0} has incorrect type: {1}.'.format(key, value) +
+                            '\nValid types are: {0}'.format(self.dtype[key]))
+
+        if self.can_call[key] and not callable(value):
```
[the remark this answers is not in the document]
kbwestfall: Yeah, I added this to the DAP when I wanted to pass callable functions/classes as parameters.

pypit/par/parset.py
kbwestfall: Done.

pypit/par/parset.py
kbwestfall: Done.

pypit/par/parset.py
kbwestfall: Ha!  ParDatabase actually isn't used in pypit, but I left it in.

pypit/par/pypitpar.py
kbwestfall: :)  I started doing it that way to see what response I got.  Then I went back to Pypit, but I like PypIt, too...

pypit/par/pypitpar.py
kbwestfall: Good point.  Everything except configobj and numpy is provided with the system.  I added `configobj>=5.0.6` to the requirements.txt file.

pypit/par/pypitpar.py
kbwestfall: Thanks!  I've changed the code to use this instead.

pypit/par/pypitpar.py
kbwestfall: I've edited the docstring for this.  Have a look an let me know what you think.

pypit/par/pypitpar.py
kbwestfall: Okay, I'll leave it for now, and then we can simultaneously remove the keyword and the functionality later.

pypit/par/pypitpar.py
kbwestfall: As Joe also noted, these parameter sets need to be updated to whatever's current.  Also a integration/hackday issue...

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        Return the valid flat-field methods
+        """
+        return [ 'vacuum', 'air' ]
+
+    @staticmethod
+    def valid_reference_frames():
+        """
+        Return the valid frame types.
+        """
+        return [ 'heliocentric', 'barycentric' ]
```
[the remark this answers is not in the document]
kbwestfall: Any of the parameters can be None if there's no default.  For now, I've changed the default to None (it's now up to the code to know what to do when the parameter is None).

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        # set
+        defaults = dict.fromkeys(pars.keys())
+        options = dict.fromkeys(pars.keys())
+        dtypes = dict.fromkeys(pars.keys())
+        descr = dict.fromkeys(pars.keys())
+
+        # Fill out parameter specifications.  Only the values that are
+        # *not* None (i.e., the ones that are defined) need to be set
+
+        # TODO: Change pcatype to spacing='irregular' or
```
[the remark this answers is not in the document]
kbwestfall: I think this suggestion was based on me trying to understand the difference between the two types based on their docstrings.  I think my main point is that it's difficult to associate 'pixel' vs. 'order' to what the approach actually does.  Happy to leave it as is.

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+        dtypes['stopcheck'] = bool
+        descr['stopcheck'] = 'If True, pypit will stop and require a user carriage return at ' \
+                             'every quality control check'
+
+        defaults['useIDname'] = False
+        dtypes['useIDname']   = bool
+        descr['useIDname']    = 'If True, file sorting will ensure that the idname is made'
+
+       
+        # TODO: Is this needed?  Should be used when instantiating msgs
```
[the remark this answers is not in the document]
kbwestfall: Poor commenting...  What I meant is that the verbosity is set on the command line for pypit, and it's the command-line option that sets `msgs` for the remainder of pypit.  If that' true, does verbosity also need to be a parameter?

pypit/par/pypitpar.py
kbwestfall: Let's work this out during the hackday.

pypit/par/pypitpar.py
kbwestfall: Okay, done.  I added a couple of lines of code to deal with a config file that only lists one lamp

pypit/par/pypitpar.py
kbwestfall: Okay, leaving it in.

pypit/par/parset.py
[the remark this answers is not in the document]
profxj: Nah, yours is find.  I do it sometimes myself.

pypit/par/parset.py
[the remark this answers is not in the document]
profxj: Got it.

pypit/par/pypitpar.py
kbwestfall: Added.  The comments say that method and params are defunct, but I at least found 'method' used somewhere in the code...

pypit/par/pypitpar.py
kbwestfall: Yep, will need to update these...

pypit/par/pypitpar.py
```diff
@@ -0,0 +1,2486 @@
...
+      fits files from the given instrument should have.
+
+    - FrameIDPar: Defines a set of parameters used to identify fits
+      files as a certain frame type.
+
+These are collected into the main, high-level paramter set, called
+PypitPar.
+
+.. todo::
+    - Warn the user if a default value is being used?
```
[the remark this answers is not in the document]
profxj: Nah, people will accept that they are using defaults.
They will likely prefer it.

pypit/par/pypitpar.py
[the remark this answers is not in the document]
kbwestfall: Done.  We'll want to keep an eye on parameters that are None, but the code doesn't yet know how to deal with that...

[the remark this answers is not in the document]
profxj: Agreed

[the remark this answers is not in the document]
kbwestfall: fixed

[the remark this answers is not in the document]
kbwestfall: I agree.  Maybe all we need is the `valid_spectrographs` method?

kbwestfall: I like Ryan's suggestion, but I'm going to suggest we come back to this.

kbwestfall: I added comments on this in the file heading.

pypit/par/pypitpar.py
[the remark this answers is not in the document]
rcooke-ast: I think it is a good idea to save a single file (or two) that contains the complete list of parameters used for the reduction. That way, whenever a user comes back to a reduction they can see exactly what parameters (including the version of PypIt) they used if they want to reproduce their results. On this note, we need to decide when to reorganise the branch structure of PypIt so that we can save versions, as @kbwestfall once suggested  :-)

profxj:
This looks great, @kbwestfall 

Merge when you feel comfortable.

profxj:
(and when tests are passing Travis)

kbwestfall:
Does anyone know a way out of python2 unicode hell?  I'm having trouble making the code python2 compatible because of how unicode is different between python 2 and 3.

kbwestfall:
tests passing now, so I'm mergin' it!

Issue #329, opened 2018-04-26
Exception handling


kbwestfall:
Using some discussion in #328 as a jumping off point... I'm not crazy about the exception handling in pypit.  I like the fact that `msgs.error` now raises an exception instead of calling `sys.exit` (I don't think the code should ever call exit unless for debugging purposes), but it will always raise PypitError.  Is there a reason to declare this exception instead of using the existing suite of python exceptions (like ValueError, TypeError, KeyError, FileNotFoundError, etc)?

I also want to understand the rationale behind having `run_pypit` hide the Traceback from the general user.  If it were me, I would essentially always be running pypit in develop mode when reducing data.  Otherwise, if it crashed, I wouldn't know if it was because of something simple, like a typo in a file name, or something deeper in the code.  In either of these case, without running in develop mode, I'd be given where the code failed and told to contact the authors.  More relevant to us the developers, if a user actually does contact us about some failure, I personally would first want to see the Traceback to see if was something quick that I can provide immediate advice on or a deeper problem that I need to dig into.  As it is, I would need to have them rerun it in develop mode and/or have them send me their data, etc, so that I could try to reproduce the error.

Sorry, this got a bit ranty, but hopefully you see my point.

rcooke-ast:
I think the original goal was to keep the printout "pretty", and include some traceback messages, but it never got implemented, probably because everyone uses the develop flag without thinking past that!

kbwestfall:
Similar to #405, but maybe this raises the larger question of when/if we should be using `raise ...` vs. `msgs.error()`.

PR #400, opened 2018-07-09
ARMS now using ParSets

kbwestfall:
For now this is more of a "heads-up" PR than something that's ready to merge.  I've pushed through and removed the use of settings from all of `arms.py` and the code it touches.  It now uses a `PypitPar` instance instead.  For expediency and testing, I have a `PypitPar.from_settings` method as a stopgap to changing `PypitSetup`.  As of this posting the code is essentially untested and likely riddled with bugs!  I'll post again to this PR to let you know when the code is ready for a final review.  In the meantime, feel free to comment on the code and I can address any issues as I go through testing it.

In particular, have a look at the `Spectrograph` parent and children classes.  I've made a lot of changes to this class and have been able to make it where all spectrograph-specific code is held.  Please check the numbers related to the telescope parameters (see `pypit.telescopes`) and the detector parameters in the individual code files (e.g. `pypit.spectrographs.keck_lris.py`).

Beyond that, there are virtually changes everywhere because of how deep settings were embedded in the code.  I've tried to make it so that only the top-level classes use the ParSets.  Everything in `core` and many of the subsidiary functions are only passed individual parameters.

There's also a lot of hopefully innocuous changes where I'm just imposing my style on the code.


pypit/arcimage.py
```diff
@@ -112,4 +94,7 @@ def build_image(self):
         #
         return self.stack
 
+    # TODO: There is not master() method.  Does this mean useframe is
```
profxj: Probably.   It will, of course use the `MasterFrame` master() method.

pypit/arms.py
```diff
@@ -53,10 +60,32 @@ def ARMS(spectrograph, fitstbl, setup_dict):
     basenames = [None]*numsci  # For fluxing at the very end
 
     # Spectrometer class
-    spectro_class = spectro_utils.load_spec_class(spectrograph=spectrograph)
+    if spectrograph is None:
```
profxj: Eventually, this will move into a better place than here.

e.g. the PipeIt class

pypit/arms.py
```diff
@@ -68,154 +97,160 @@ def ARMS(spectrograph, fitstbl, setup_dict):
...
+                if det not in map(int, _par['rdx']['detnum']):
                     msgs.warn("Skipping detector {:d}".format(det))
                     continue
                 else:
                     msgs.warn("Restricting the reduction to detector {:d}".format(det))
             # Setup
             dnum = settings.get_dnum(det)
             msgs.info("Working on detector {:s}".format(dnum))
+            # TODO: Should this be det or dnum; can we deprecate dnum?
             sci_dict[det] = {}
```
profxj: Seeing lines 86-104 on the left be deleted brings
a real warm and fuzzy feeling to my inner being!

pypit/arms.py
```diff
@@ -68,154 +97,160 @@ def ARMS(spectrograph, fitstbl, setup_dict):
...
             sci_image_files = arsort.list_of_files(fitstbl, 'science', sci_ID)
-            # Settings
-            sci_settings = tsettings.copy()
+
             # Instantiate
-            sciI = scienceimage.ScienceImage(file_list=sci_image_files, datasec_img=datasec_img,
-                                             bpm=msbpm, det=det, setup=setup, settings=sci_settings,
-                                             maskslits=maskslits, pixlocn=pixlocn, tslits_dict=tslits_dict,
-                                             tilts=mstilts, fitstbl=fitstbl, scidx=scidx)
+            sciI = scienceimage.ScienceImage(_spectrograph, file_list=sci_image_files,
```
profxj: Looks good.  @jhennawi -- Make note for your new Extraction class.

pypit/arspecobj.py
```diff
@@ -120,8 +120,8 @@ def __repr__(self):
                 self.idx, self.config, self.xobj, self.slitcen, sdet, self.scidx, self.objtype))
 
 
-def init_exp(lordloc, rordloc, shape, maskslits,
-             det, scidx, fitstbl, tracelist, settings, ypos=0.5, **kwargs):
+def init_exp(lordloc, rordloc, shape, maskslits, det, scidx, fitstbl, tracelist, binning=None,
```
profxj: @jhennawi -- Is this the current `SpecObj`?

I've lost track.

pypit/artrace.py
```diff
@@ -207,10 +210,10 @@ def trace_objects_in_slits(slf, det, sciframe, varframe, crmask, doqa=False, **k
     return tracelist
 
 
-def trace_objects_in_slit(det, slitn, tslits_dict, sciframe, skyframe,
-                          varframe, crmask, settings_trace,
-                          trim=2, triml=None, trimr=None, sigmin=2.0, bgreg=None,
-                          maskval=-999999.9, xedge=0.03, standard=False, debug=False):
+def trace_objects_in_slit(det, slitn, tslits_dict, sciframe, skyframe, varframe, crmask,
+                          tracefunc, traceorder, tracefind, nsmooth, trim=2, triml=None,
```
profxj: We might set some of these to defaults.

pypit/biasframe.py
```diff
@@ -8,35 +8,49 @@
...
+# But I don't think you'd ever do that.
+# ----------------------------------------------------------------------
 frametype = 'bias'
 
 # Place these here or elsewhere?
 #  Wherever they be, they need to be defined, described, etc.
 #  These are settings beyond those in the Parent class (ProcessImages)
-additional_default_settings = {frametype: {'useframe': 'none'}}
-
+# additional_default_settings = {frametype: {'useframe': 'none'}}
```
profxj: Because sometimes msbias='overscan', i.e. not an image for a 'str'

pypit/biasframe.py
```diff
@@ -66,57 +80,34 @@ class BiasFrame(processimages.ProcessImages, masterframe.MasterFrame):
...
-        Returns
-        -------
-        stack : ndarray
-
-        """
-        # Wrapper
-        self.stack = self.process(bias_subtract=None, trim=False, overwrite=overwrite)
-        return self.stack.copy()
-    '''
+        # (KBW) Even given above, I don't understand why this is
```
profxj: I concede.  Do as you wish.

pypit/biasframe.py
```diff
@@ -157,63 +148,25 @@ def master(self):
         msframe : ndarray or str or None
 
         """
+        # (KBW) Not sure this is how it should be treated if loaded is
```
profxj: Good question.

pypit/bpmimage.py
```diff
@@ -109,61 +90,22 @@ def build(self):
...
-        else:
-            self.bpm = self.spectro_class.bpm(binning=self.binning, det=self.det,
-                                              shape=self.shape)
-        # Return
-        return self.bpm
+        # Will raise an exception if could not construct the BPM
+        self.bpm_img = self.spectrograph.bpm(shape=self.shape, filename=self.filename,
+                                             det=self.det) if self.msbias is None else \
+                            arprocimg.badpix(self.msbias,
+                                    self.spectrograph.detector[self.det-1]['numamplifiers'],
```
profxj: While it may be a bit risky, I should love to have a 'thisdet' dict or something like that, e.g.

```
self.spectrograph.thisdet = self.spectrograph.detector[self.det-1]
```

Of course, this makes `spectrograph` mutable..

But I've grown tired of passing in self.det-1 (and needing to pass around `det`
just to do this).  Let's discuss.

pypit/calibrations.py
```diff
@@ -82,6 +102,11 @@ def __init__(self, fitstbl, spectro_class=None, save_masters=True, write_qa=True
         # Internals
         self._reset_internals()
 
+    @staticmethod
```
profxj: I was intentionally leaving this list empty for the Parent class.

Sort of to force us to use a Child.

pypit/calibrations.py
```diff
@@ -136,39 +160,46 @@ def get_arc(self):
...
+        if self.msarc is None:  # Otherwise build it
+            msgs.info("Preparing a master {0:s} frame".format(self.arcImage.frametype))
+            self.msarc = self.arcImage.build_image()
+            # Save to Masters
+            if self.save_masters:
+                self.arcImage.save_master(self.msarc, raw_files=self.arcImage.file_list,
+                                          steps=self.arcImage.steps)
+
+        # Save & return
+        self.calib_dict[self.setup]['arc'] = self.msarc
```
profxj: I kind of wanted that `self.shape` in there.  
Or we need another kludge  :)

pypit/calibrations.py
```diff
@@ -245,21 +287,24 @@ def get_datasec_img(self):
...
-            scifile, self.det, self.settings['detector'])
-        self.settings['detector']['naxis0'] = naxis0
-        self.settings['detector']['naxis1'] = naxis1
-        # Datasec image
-        self.datasec_img = arpixels.pix_to_amp(naxis0, naxis1, datasec,
-                                               self.settings['detector']['numamplifiers'])
-        # Return
-        return self.datasec_img, naxis0, naxis1
+        # Generate the spectrograph-specific amplifier ID image
+        return self.spectrograph.get_datasec_img(scifile, self.det)
```
profxj: Love it.

pypit/calibrations.py
```diff
@@ -616,13 +728,14 @@ def __repr__(self):
         return txt
 
 class MultiSlitCalibrations(Calibrations):
+#   (KBW) This could fail because Calibrations does not also accept **kwargs
+#    def __init__(self, fitstbl, steps=None, **kwargs):
+#        Calibrations.__init__(self, fitstbl, **kwargs)
+#   (KBW) I'd prefer this:
+    def __init__(self, fitstbl, spectrograph=None, par=None, wavelength_frame='pixel',
```
profxj: ok

pypit/core/arwave.py
```diff
@@ -185,30 +185,30 @@ def flex_shift(obj_skyspec, arx_skyspec, mxshft=None):
...
-    #
-    msgs.info("Using {:s} file for Sky spectrum".format(skyspec_fil))
-    arx_sky = xspectrum1d.XSpectrum1D.from_file(root+skyspec_fil)
-    #hdu = fits.open(root+'/data/sky_spec/'+skyspec_fil)
-    #archive_wave = hdu[0].data
-    #archive_flux = hdu[1].data
-    #arx_sky = xspectrum1d.XSpectrum1D.from_tuple((archive_wave, archive_flux))
-    # Return
-    return skyspec_fil, arx_sky
+# This has been moved into the spectrographs class
```
profxj: Good call.

pypit/fluxspec.py
```diff
@@ -226,44 +233,54 @@ def generate_sensfunc(self):
...
-        -------
-
-        """
-        # Note the unravel here
-        for sci_obj in arutils.unravel_specobjs(specobjs):
-            if sci_obj is not None:
-                # Do it
-                arflux.apply_sensfunc(sci_obj, self.sensfunc, airmass, exptime, self.settings)
-
+# Does not seem to be used
```
profxj: Might be called by the fluxing script.  Please check.

pypit/masterframe.py
```diff
@@ -104,18 +108,26 @@ def master(self):
...
-        else:
-            # Prevent over-writing the master frame when it is time to save
-            self.settings['masters']['loaded'].append(self.frametype+self.setup)
-            # Hold it
-            self.msframe = msframe
-        # Return
+        self.msframe = msframe
         return msframe.copy()
 
+#        # Load the MasterFrame if it exists and user requested one to load it
```
profxj: How is this being done now?

pypit/scienceimage.py
```diff
@@ -102,13 +96,15 @@ class ScienceImage(processimages.ProcessImages):
 
     """
     # Keep order same as processimages (or else!)
-    def __init__(self, file_list=[], spectrograph=None, settings=None,
-                 tslits_dict=None, tilts=None, det=None, setup=None, datasec_img=None,
-                 bpm=None, maskslits=None, pixlocn=None, objtype='science',
-                 fitstbl=None, scidx=0):
+    def __init__(self, spectrograph, file_list=[], frame_par=None, trace_objects_par=None,
```
profxj: I'd coordinate modifying this object with @jhennawi

pypit/spectrographs/keck_nirspec.py
```diff
@@ -13,20 +13,33 @@
...
         # Get it started
-        spectroclass.Spectrograph.__init__(self)
+        super(KeckNIRSPECSpectrograph, self).__init__()
         self.spectrograph = 'keck_nirspec'
-
+        self.telescope = telescopes.KeckTelescopePar()
+        self.camera = 'NIRSPEC'
+        self.detector = [
+                # Detector 1
+                DetectorPar(dataext=0, dispaxis=0, xgap=0., ygap=0., ysize=1., platescale=0.193,
```
profxj: I love all this.  I might ask that you have each `DetectorPar` key/item pair on a 
separate line.  To improve readability.

pypit/arcomb.py
```diff
@@ -32,16 +32,22 @@ def comb_frames(frames_arr, det, frametype, **kwargs):
...
+def core_comb_frames(frames_arr, printtype=None, frametype='Unknown', saturation=None,
+                     maskvalue=1048577, method='weightmean', satpix='reject', cosmics=None,
+                     n_lohi=[0,0], sig_lohi=[3.,3.], replace='maxnonsat'):
+    """
+    Combine several frames
 
     .. todo::
         - Make better use of np.ma.MaskedArray objects throughout?
         - More testing of replacement code necessary?
+        - Improve docstring...
```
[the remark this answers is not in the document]
kbwestfall: Do you have specific examples?  I've used masked arrays quite a bit and I have had issues now and again.  But those issues have mostly (always?) been because I've assumed some function understands what a masked array is when it doesn't.  When sticking to the MaskedArray class itself and its methods, I've been safe.  Personally, I think the gains from using MaskedArrays is worth the care one has to take when using them (i.e., checking to make sure that they're behaving as you expect/assume).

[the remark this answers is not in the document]
profxj: Related, please see Issue #361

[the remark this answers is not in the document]
profxj: Not sure if this is the right time to deal with this, though.
Should probably be its own PR.

doc/nb/keck_lris_blue_long_400_3400_d560.cfg
```diff
@@ -1,37 +1,27 @@
...
-    sortdir = lris_blue_long_400_3400_d560
-
 [rdx]
-    spectrograph = KECK_LRISb
+    spectrograph = keck_lris_blue
     pipeline = ARMS
+
+[calibrations]
     [[flatfield]]
         frame = /Users/westfall/Work/packages/PYPIT-development-suite/CALIBS/PYPIT_LRISb_pixflat_B400_2x2_15apr2015.fits.gz
```
rcooke-ast: Is there a way to specify this more generally for each user?

pypit/arms.py
```diff
@@ -68,154 +97,160 @@ def ARMS(spectrograph, fitstbl, setup_dict):
...
+            # Arc Image
+            msarc = caliBrate.get_arc()
+            # Bad pixel mask
+            msbpm = caliBrate.get_bpm()
+            # Physical pixel locations on the detector
+            pixlocn = caliBrate.get_pixlocn()
+            # Slit Tracing
+            tslits_dict, maskslits = caliBrate.get_slits()
             if tslits_dict is None: # No slits
+                # TODO: Add a warning?
```
rcooke-ast: Yes indeed - we should add a warning.

pypit/bpmimage.py
rcooke-ast: I see no problem getting an instance of this, for example, at the top of the routine use: ```thisdet = self.spectrograph.detector[self.det-1]``` and then call things like ```thisdet['numamplifiers']```. It would still therefore be immutable. I don't see the advantage of including this into the spectrograph class (or, for that matter, including it in this bpm class)

jhennawi:
Hi @kbwestfall, 

I took a look at what you've been developing. Thanks a lot for what was obviously a massive amount of work on this. It looks really nice. Below is some feedback. 

I went through the iPython notebook that you created demonstrating how this works BiasFrame.ipynb. Note that this does not currently run, presumably because you changed some things while developing. Anyway, I'm using the code that you listed there as way to give you feedback on parset integration. Since it is difficult to comment on GitHub in notebooks, I'm copying those comments here following the relevant lines in your notebook. 

runpar = pypitpar.RunPar(caldir='MF')
rdxpar = pypitpar.ReducePar(spectrograph='shane_kast_blue', masters=None, setup='A_01_aa')

JFH: I  don't like that I have to list these incantations here to get the code to run. Does the code 
automatically write to MF if I don't instantiate the run par?

# Change one default just to make sure it propagates 
biaspar = pypitpar.FrameGroupPar(frametype='bias',
                                 combine=pypitpar.CombineFramesPar(cosmics=25.))

JFH: How does the user know that they need to alter the FrameGroupPar in this way in order to do this with that argument combine= set? I suppose the answer is read through the documentation? This looks very complicated and not that different than what we had before. I don't want to have to figure out a complex hierarchy of settings or read extensive  documentation or thousands of lines of code to change one parameter. I thought the interface we were discussing  was  that there would be a one to one mapping between the parsets and the BaseClass names. The call that I would like to see is thus simply: 

biaspar = pypitpar.BiasFrame(cosmics = 25)

This modifies the parameters from the defaults that BiasFrame usually takes. And if I want to run with 
defaults (which may be instrument specific), then I pass None in for the parset. 


bias_frame = biasframe.BiasFrame(rdxpar['spectrograph'], file_list=kast_blue_bias,
                                 par=biaspar, setup=rdxpar['setup'], root_path=root_path,
                                 mode=rdxpar['masters'])

JFH:  Bloated optional argument lists seem to be the hallmark of PYPIT. This makes it very hard to figure out which arguments are required and which ones are not, and often our codes crash if optional
arguments are not passed.  

This argument list could be dramatically reduced as follows: 

bias_frame = biasframe.BiasFrame(file_list,par=biaspar)

JFH: Let's go over the parameters in your argument list and ask whether they are necessary:
-----
 
 -- spectrograph = this is not formally required since you have a file and so you can get it out of the header in the BiasFrame class. I'm willing to live with it for a homogenous interface to the base classes,  but keep in mind that classes like WaveTilts don't require it so our interface is not all that homogenous. 
Why require it here when you can get it from the header in one microsecond inside BiasFrame. Anwyay, I don't feel that strongly about this. 

-- file_list = I think this should be required rather than optional, and  the user can set it to None if they want to use a master, but that is a matter of taste I suppose. 

-- par = biaspar = I agree with this

-- setup= I don't see why the setup needs to be passed in here at all. I'm sure you will say because we need to write out the master file, but the master file is setup independent. Biasframes don't depend
on spectroscopic setup, they just depend on the binning and or read mode of the detector and all 
of that is in the header. So if you are creating a Master bias from these files, they should have names like" MasterBiasLRIS_Blue1x2-lblue0013.fits" where 1x2 is the detector binning, lblue0013.fits is the first filename in the file list or something, and where the individual detectors are written to this multi-extension fits files with the same dataext as the data itself. There is no need to pass in this setup information because the setup is known from the header. We could debate about the
nomenclature for naming masters, but I really don't get why I have to pass in setup. 

-- root_path = I could imagine that this could be part of the biaspar param list as well, but on the other hand maybe it is better to keep the paths to files as explicit optional arguments as you've done. 

-- mode = I'm not sure what this does as it is not documented. I think it is related to using masters or not. Again, maybe this could be part of the biaspar param list, but maybe better to keep the use master
stuff explicit like file paths. 

It would be very helpful to have a few other notebook examples for other base classes like FlatField, or WaveCalib, etc. If you have time to put those together or update the existing notebooks with parsets, I'm happy to provide feedback. 

Thanks again for the hard work!
Joe

pypit/fluxspec.py
jhennawi: This is definitely used by arflux.py

pypit/scienceimage.py
jhennawi: This class is going to be completely rewritten so I would not spend time modifying it now in its current form.

pypit/arcomb.py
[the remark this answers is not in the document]
jhennawi: Hi @kbwestfall, 

I had many problems with numpy masked arrays in the past. Whether that was due to buggy code, or user error, i.e. the fact that I thought a numpy routine was respecting the mask, whereas it actually was not, is not clear. 

The point is that masking is an important and subtle issue in spectroscopic data reduction. We have several masks and we mask things on multiple occasions and constantly update that mask.  There is no way that this masking can be implicit like with numpy arrays, because it
will result in code that is unreadable and it will cause errors. 

In my opinion we should remove all usage of them from PYPIT and just propagate around an extra mask when needed.

profxj:
It would be a mistake to try to refactor the `MasterFrame` class and functionality in this PR. 
Do not do so.

profxj:
And, while I agree with @jhennawi 's desire to reduce the argument list in some of these classes,
that can also await a new PR.

pypit/bpmimage.py
profxj: Fair point.

pypit/calibrations.py
kbwestfall: This function is actually only ever called by the child at the moment.  I saw this as the default function that a child could overwrite if it needed.  I can move it back to the child, if that makes more sense.

pypit/fluxspec.py
kbwestfall: I don't see it in my current branch.  But I've uncommented this function and removed the leading underscore.  Do we need both `flux_specobjs` and `flux_science`?

pypit/masterframe.py
kbwestfall: The only difference from the uncommented and commented code is the setting line.  I've removed the use of 'loaded'.  I'll have to figure out how to put this back in if 'loaded' is still being used.

pypit/spectrographs/keck_nirspec.py
kbwestfall: Done.

pypit/spectrographs/spectrograph.py
```diff
@@ -0,0 +1,449 @@
...
+        self.bpm_img = None
+
+        # Default time unit
+        self.timeunit = 'mjd'
+
+        # Default extension with the primary header data
+        #   used by arsave.save_2d_images
+        self.primary_hdrext = 0
+
+        # Default sky spectrum
```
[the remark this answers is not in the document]
kbwestfall: Yes.  What's the reason to leave it out of `__init__`?

pypit/arcomb.py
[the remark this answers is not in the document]
kbwestfall: Agreed that pixels can be masked for a number of reasons.  I have a convience `BitMask` class that I found helpful when I was doing this for MaNGA.

Personally, I find MaskedArrays very valuable and have made extensive use of them, but I'm fine with discussing this more.  Let's please move this discussion to #361 or a different, new issue.

doc/nb/keck_lris_blue_long_400_3400_d560.cfg
kbwestfall: Yes, ConfigObj has a way that you can define a path and then use it (like when setting an environmental variable) and use it in other places in the file.  This isn't as powerful as in the normal ConfigParser class, though.  We can talk more about this on a telecon, and I just added #402.

kbwestfall:
Thanks for the comments so far, everyone!

I'm still working through debugging/testing the code.  I broke it in a number of ways that I'm trying to fix.  A few things:
 - Thanks for the deep dive, @jhennawi.  I will work on some more illustrative notebooks, and point you to them when they're ready.
 - On the parameters:
    - I generally write functions where required variables do not have defaults and optional ones do.  These functions should run when no optional arguments are provided.  Admittedly, this can sometimes be hard to stick to if you need a function that has multiple ways of proceeding, but I try to avoid that.  The same thing goes for classes, where I try to use `classmethods` instead of multiple optional arguments in an `__init__` function to provide multiple ways for instantiating an object.
   - I agree that we should make the user-interface with these functions as concise as possible, it's not like I'm adding parameters willy-nilly.  Personally, I'm fine with a long argument list as long as it has an associated docstring (I admit I'm falling behind on that), as long as the arguments make the function more useful.
   - Your example of being able to do `biaspar = pypitpar.BiasFrame(cosmics = 25)` is one way to go, and I agree that it's a nice interface so let's talk through this.  Two pit falls to keep in mind, though: (1) This puts a larger burden on us in terms of making sure that the parameters get appropriately parsed in the the relevant parameter sets and actually makes the parameter sets less useful.  (2) It may push us toward more functions that accept `**kwargs` so that argument lists are much less structured.  I have a pretty strong resistance to this because it makes the code harder to write and document.  matplotlib makes extensive use of `**kwargs`.  The positive is that it makes the code incredibly flexible and, at least in some ways, powerful.  The downside is that the documentation is nigh impenetrable, certainly compared to numpy, where most (all?) of the functions have explicit argument lists.
 - One major change that I made and wee need to keep in mind is the handling of the data and overscan sections.  We should talk about this on a telecon, but the bottom line is that I've written a new parser that takes the string representation of the section and constructs a list of `slice` objects instead of the `int` lists that existed before.  I think this makes some of the relevant code a lot simpler (e.g., see some of the new functions in `arprocimg.py`, like `find_bad_pixels` compared to `badpix`).  Because of that change, we need to be careful that these sections are getting parsed correctly.

Back to debugging ...

jhennawi:
Hi @kbwestfall, 
 
Thanks for the quick reply, and for all the improvements to the code. I respond to your point about the
parset interface below: 

"Your example of being able to do biaspar = pypitpar.BiasFrame(cosmics = 25) is one way to go, and I agree that it's a nice interface so let's talk through this. Two pit falls to keep in mind, though: (1) This puts a larger burden on us in terms of making sure that the parameters get appropriately parsed in the the relevant parameter sets and actually makes the parameter sets less useful."

Let's continue to talk through this. BiasFrame takes a list of parameters, some of which will be ProcessImages parameters, i.e. the parent class,  and some of those could be parameters to the routine itself that are bias specific. 

One option is to have a simple call like: 

biaspar = pypitpar.BiasFrame(cosmics = 25)

Then we have to write a parser which figures out which parsets to change, i.e. the parset for BiasFrame specific parameters, and the parset for ProcessImages. This could be one general
parser routine that takes a list of parsets (in this case the two parsets that BiasFrame depends on), 
and the input argument list, and then modifies the parsets accordingly. I don't think this is that difficult, but I agree it is extra work. I'd rather have that simpler interface, than climb the learning curve of having to understand how to modify the specific parsets. The fact that the parsets names don't match the actual classes makes the current interface less intuitive. 

The second option, would be to simply have something like

biaspar = pypitpar.BiasFrames(use_dark = True)
procpar =pypitpar.ProcessImages(cosmics=25.))

Then the call to BiasFrame would be

bias_frame = biasframe.BiasFrame(file_list=kast_blue_bias, par_bias = biaspar, par_proc = proppar)

Then there is no extra parser to write, and it is clear how the parsets get passed with no 
additional work. 

I am adamant about wanting to avoid the nested hierarchy of parameters that is the current structure of PYPIT. This has caused me countless hours of pain in trying to develop and use the code. I can
never figure out which parameters need to be set, which don't, how to modify them, and where
to figure out what is ingesting what. For that reason, I really think there needs to be a very simple
interface and I think that there needs to be a one to mapping between parset name and class name. 


"(2) It may push us toward more functions that accept **kwargs so that argument lists are much less structured. I have a pretty strong resistance to this because it makes the code harder to write and document. matplotlib makes extensive use of **kwargs. The positive is that it makes the code incredibly flexible and, at least in some ways, powerful. The downside is that the documentation is nigh impenetrable, certainly compared to numpy, where most (all?) of the functions have explicit argument lists."

I'm not advocating using kwargs. I agree with you that this makes code difficult to understand and document. 

Please let me know if you think there are pitfalls with my suggestion above. My main problem with the example in your notebook is that FrameGroupPar is not a PYPIT class name and neither is CombineFramesPar. The mapping to PYPIT classes and PYPIT functionality is thus unclear, and without
reading extensive documentation, I'm not going to understand how to modify parameters. This seems a lot like the current settings model, but with an admittedly much better interface.

Thanks, 
Joe

pypit/spectrographs/spectrograph.py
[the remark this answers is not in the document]
kbwestfall: Done.

kbwestfall:
This PR is ready for your final review.  Sorry this became such a large PR.  If there are specific questions about how this works, feel free to just ask me directly instead of trying to suss it out from the code.

More than the title suggests, the PR has now effectively removed the use of settings from PypeIt.  The pypit files now have a block of lines that set the parameters, but there's not a 1-1 mapping of setting to `ParSet` parameter.  There are no ascii text files that set the defaults; the defaults are all kept internally to the `ParSet` definition.  For some of the evolving documentation of this, check out the branch and do the following:

```
cd doc
make html
open _build/html/pypit_par.html
```

(When you're done looking at the html, you can clean it out by running `make clean` in the `doc/` directory.)

Beyond what's there (not easy to auto-document), I've put in some streamlined ways to set the parameters:

 - I think most of the spectrographs override this default, but the `PypitPar` defaults actually have the sky subtraction, flexure corrections, and flux calibration all turned off (it's easy to change this).  To turn them on using the defaults you can set, e.g.::

        [rdx]
            skysubtract = True
            flexure = True
            fluxcalib = True

 - I've put in place a way that you can change the default image processing parameters for all frames without having to include changes for each frame type.  For a complex example of this, see the `test_pypit_file` function in  `pypit/tests/test_pypitpar.py` and the associated pypit file, `pypit/tests/files/test_pypit_file.pypit`.
  
There are likely a few gotchas that we should keep track of.  The first one that pops to mind is that the `spect` group in the previous pypit file has gone away.  The `number` of frames to group has moved into the definition of each frame group.  E.g., in the new pypit file you would set:

```
[biasframe]
    number = 10
```

But this also means that the specification of a specific file as a certain frame type has gone away.  There are certainly ways I can bring this back, but I think frame typing should be something that's considered a pre-processing step for PypeIt.  I should bring this up as an issue, but just to get us started:  Personally, I'd prefer that we make a clean break between frame typing, and other setup activities, and the call to `run_pypit`.  Given X's `PypitSetup` class and other things that fell out of this PR, the `PYPIT` function is now heavily streamlined.  It's now no longer onerous to have the setup and calibration check code appear in both `run_pypit` and `pypit_setup`.  It's really now just a few lines of code in `PYPIT` between starting the function and quiting because you're only doing a setup or calibration check.  Keck LRIS red currently fails --- well more accurately, I haven't updated the pypit file yet --- in the dev suite because of this frame typing issue...

There may be other functionality that got lost along the way, so please keep your eyes out.  If there's something you notice in your review, let me know and I'll fix it.

Two sticking points I noticed in debugging this PR were
 - the definition and usage of `datasec` and `oscansec` and
 - keeping track of whether or not images were trimmed.

Other than that, all functionality should be the same as it was when using the settings construct.

All tests currently pass.

Failures in the dev suite are:
    - Keck LRIS red: Fails because of frame typing
    - Shane Kast Red: Fails with this traceback:

```
  File "/Library/Frameworks/Python.framework/Versions/3.6/bin/run_pypit", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/westfall/Work/packages/PYPIT/bin/run_pypit", line 17, in <module>
    rtval = run_pypit.main(args)
  File "/Users/westfall/Work/packages/PYPIT/pypit/scripts/run_pypit.py", line 104, in main
    logname=logname)
  File "/Users/westfall/Work/packages/PYPIT/pypit/pypit.py", line 121, in PYPIT
    status = arms.ARMS(fitstbl, setup_dict, par=par, spectrograph=spectrograph)
  File "/Users/westfall/Work/packages/PYPIT/pypit/arms.py", line 223, in ARMS
    specobjs, finalvar, finalsky = sciI.extraction(mswave)
  File "/Users/westfall/Work/packages/PYPIT/pypit/scienceimage.py", line 356, in extraction
    self.obj_model = self.original_optimal(mswave)
  File "/Users/westfall/Work/packages/PYPIT/pypit/scienceimage.py", line 308, in original_optimal
    self.maskslits, self.tslits_dict['slitpix'], doqa=False)
  File "/Users/westfall/Work/packages/PYPIT/pypit/core/arextract.py", line 1753, in obj_profiles
    gdrow = np.where(specobjs[sl][o].boxcar['counts'] > COUNT_LIM)[0]
KeyError: 'counts'
```

jhennawi:
Hi @kbwestfall, 

I love everything that you've done here. My only comment is about this line in the PYPIT files: 

[biasframe]
    number = 10

I'm not sure why we need to specify that 10 bias frames are being in used in the parset list, when there are ten bias frame lines in the pypit file indicating the bias frames that the user wants to have used. There is a minimum bias frame parameter (or at least there used to be), but I don't think this is the number that you have above. 

I find it onerous to specify this frame number twice in the pypit file. I'm not following the logic of why this is now required in the parset. 

Joe

doc/nb/compiled_example.cfg
```diff
@@ -1,53 +1,25 @@
...
-    # Directory relative to calling directory to write qa files.
-    qadir = QA
-    # File for the details of the sorted files.  If None, no output is
-    # created
-    sortdir = lris_blue_long_400_3400_d560
-    # Flag to overwrite any existing output files
-    overwrite = True
-
 # PypIt reduction rules.
 [rdx]
```
profxj: Do we want these .cfg files here (or at all)?

Just checking.

pypit/arcimage.py
```diff
@@ -1,27 +1,18 @@
...
 import numpy as np
-import os
 
 from pypit import msgs
 from pypit import processimages
 from pypit import masterframe
 from pypit.core import arsort
+from pypit.par import pypitpar
 
 from pypit import ardebug as debugger
```
profxj: So nice to see that _debug code disappear!

pypit/arcimage.py
```diff
@@ -60,33 +51,48 @@ class ArcImage(processimages.ProcessImages, masterframe.MasterFrame):
...
+        # MasterFrames: Specifically pass the ProcessImages-constructed
+        # spectrograph even though it really only needs the string name
+        directory_path = None if root_path is None \
+                                else root_path+'_'+self.spectrograph.spectrograph
+        masterframe.MasterFrame.__init__(self, self.frametype, setup,
+                                         directory_path=directory_path, mode=mode)
+
+        # Comments from JFH.  Check if still relevant given changes.
+
+        # TODO, JFH I think the logic below flawed here. The
```
profxj: These can be removed.

pypit/arcimage.py
```diff
@@ -127,4 +120,7 @@ def build_image(self):
         #
         return self.stack
 
+    # TODO: There is no master() method.  Does this mean useframe is
```
profxj: I suppose it must.

pypit/masterframe.py
```diff
@@ -37,44 +34,43 @@ class MasterFrame(object):
     """
     __metaclass__ = ABCMeta
 
-    def __init__(self, frametype, setup, settings):
-        # Start us up
+    def __init__(self, frametype, setup, spectrograph=None, directory_path=None, mode=None):
```
profxj: I still think we should probably have a Master's ParSet,
but that can await a future PR.

pypit/scripts/arcid_plot.py
```diff
@@ -49,10 +49,9 @@ def main(args):
 
     from linetools.utils import loadjson
 
-    from pypit import pyputils
     from pypit import arqa
-
-    msgs = pyputils.get_dummy_logger()
+    from pypit import msgs
+    msgs.reset(verbosity=0)
```
profxj: Can we set the default to highest verbosity here?

It would address Issue #405

profxj:
@jhennawi -- Please give your explicit approval (i.e. click the button).

@rcooke-ast -- Please give a quite code review if you can.  If for no other reason to make yourself familiar with the changes.  Thanks!

jhennawi:
Hi @profxj,

Happy to approve, but I'd like to hear @kbwestfall chime in about the 

[biasframe]
number = 10

point that I raised above. I'd prefer to not to have to specify this as a parameter, given that 10 lines later each biasframe is listed in the pypit file. The simpler we make the pypit file interface the better from my perspective. 

Joe

doc/nb/compiled_example.cfg
kbwestfall: I've left them in just as an example of how to use `PypitPar`.  We can get rid of them because I don't expect many of us will want to use these cfg files, and will use the pypit files instead.

kbwestfall:
@jhennawi, The `number` parameters are used when pypit is matching calibration frames to science frames, and I think it's a minimum number that it must find during that matching.  @profxj and @rcooke-ast will know better what this parameter does.  At the moment this parameter is necessary, but I agree with you that there is likely a way to avoid having to set this.  If we separate the frame typing and matching from the actual reduction of the data, we can remove `number` as a parameter.  Would that mean we would need to have multiple pypit files with calibration and science frames pre-matched so that all pypit would need to do is use all the files provided to it for a given run?  Then you would have multiple runs of pypit, one per pypit file with the appropriate grouping?  I suggest we make this an issue as a possible development path.

pypit/arload.py
```diff
@@ -42,74 +41,34 @@ def load_headers(datlines, settings_spect, settings_argflag):
...
-            # JFH changed to in instead of !=
-            if ((settings_spect['check'][ch] in str(headarr[frhd][kchk]).strip()) == False):
-                print(ch, frhd, kchk)
-                print(settings_spect['check'][ch], str(headarr[frhd][kchk]).strip())
-                msgs.warn("The following file:"+msgs.newline()+datlines[i]+msgs.newline()+"is not taken with the settings.{0:s} detector".format(settings_argflag['run']['spectrograph'])+msgs.newline()+"Remove this file, or specify a different settings file.")
-                msgs.warn("Skipping the file..")
-                skip = True
+        skip = spectrograph.check_headers(headarr)
         if skip:
+            msgs.warn("The following file:"+msgs.newline()+datlines[i]+msgs.newline()+"is not taken with the settings.{0:s} detector".format(
```
rcooke-ast: do we want `settings.blah detector` here? I don't think so, right?

pypit/arload.py
```diff
@@ -216,89 +165,86 @@ def generate_updates(dct, keylst, keys, whddict, headarr):
...
-    if len(ind) == 1:
-        msgs.info("Loaded {0:d} {1:s} frame successfully".format(len(ind), frametype))
-    else:
-        msgs.info("Loaded {0:d} {1:s} frames successfully".format(len(ind), frametype))
-    return frames
+    # Return
+    return fitstbl
+
+
+#def load_frames(fitsdict, ind, det, frametype='<None>', msbias=None, trim=True):
```
rcooke-ast: This can be safely deleted now, right?

pypit/arload.py
```diff
@@ -343,49 +289,49 @@ def load_extraction(name, frametype='<None>', wave=True):
...
-            if settings.argflag['reduce']['masters']['force']:
-                msgs.error("Master calibration file does not exist:"+msgs.newline()+name)
-            else:
-                msgs.warn("Could not read Master calibration file:"+msgs.newline()+name)
-                raise IOError
-        msgs.info("Master {0:s} frame loaded successfully:".format(hdu[0].header['FRAMETYP'])+msgs.newline()+name)
-        head = hdu[0].header
-        data = hdu[exten].data.astype(np.float)
-        return data, head
+#def load_master(name, exten=0, frametype='<None>'):
```
rcooke-ast: this can be likewise deleted...(?)

pypit/arms.py
```diff
@@ -53,10 +66,32 @@ def ARMS(spectrograph, fitstbl, setup_dict):
...
+        # keyword.
+        if par is not None and par['rdx']['spectrograph'] != fitstbl['instrume'][0]:
+            msgs.error('Specified spectrograph does not match instrument in the fits table!')
+        _spectrograph = load_spectrograph(spectrograph=fitstbl['instrume'][0])
+    elif isinstance(spectrograph, basestring):
+        _spectrograph = load_spectrograph(spectrograph=spectrograph)
+    elif isinstance(spectrograph, Spectrograph):
+        _spectrograph = spectrograph
+    else:
+        raise TypeError('Could not instantiate Spectrograph!')
```
rcooke-ast: Is there a  reason why a TypeError is raised, and not a msgs.error?

pypit/arms.py
```diff
@@ -53,10 +66,32 @@ def ARMS(spectrograph, fitstbl, setup_dict):
...
+
+    # Instantiate the parameters
+    _par = _spectrograph.default_pypit_par() if par is None else par
+    if not isinstance(_par, pypitpar.PypitPar):
```
rcooke-ast: I suppose the renaming of PypitPar will change to PypeItPar (probably a future PR, right?)

pypit/arms.py
```diff
@@ -53,10 +66,32 @@ def ARMS(spectrograph, fitstbl, setup_dict):
...
+        raise TypeError('Input parameters must be a PypitPar instance.')
+    required = [ 'rdx', 'calibrations', 'scienceframe', 'objects', 'extract', 'skysubtract',
```
rcooke-ast: A minor comment... I like that the top-level arguments are spelled out, but I think we need to think about changing `rdx` to something a little more transparent.

pypit/arms.py
```diff
@@ -287,80 +342,101 @@ def ARMS(spectrograph, fitstbl, setup_dict):
...
+            helio_dict = dict(refframe='pixel'
+                              if caliBrate.par['wavelengths']['reference'] == 'pixel' 
+                              else caliBrate.par['wavelengths']['frame'],
                               vel_correction=sci_dict['meta']['vel_corr'])
             arsave.save_1d_spectra_fits(all_specobjs, fitstbl[scidx], outfile,
-                                            helio_dict=helio_dict, obs_dict=settings.spect['mosaic'])
+                                        helio_dict=helio_dict, telescope=_spectrograph.telescope)
         elif save_format == 'hdf5':
-            debugger.set_trace()  # NEEDS REFACTORINGj
+            debugger.set_trace()  # NEEDS REFACTORING
```
rcooke-ast: Maybe raise a warning here just before the debugger?

pypit/armsgs.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, log=None, debug=None, verbosity=None, colors=True):
         # Initialize other variables
         # TODO: debug could just be develop=True or False
         self._debug = debug
```
rcooke-ast: I had a feeling that we were removing this functionality? So, I think all references to `self._debug` can be removed.

pypit/bpmimage.py
```diff
@@ -59,48 +62,39 @@ class BPMImage(object):
...
+    frametype = 'bpm'
 
-        self.reduce_badpix = reduce_badpix
-        self.msbias = msbias
+    # Keep order same as processimages (or else!)
+    def __init__(self, spectrograph=None, shape=None, filename=None, det=None, msbias=None,
+                 trim=True):
+
+        # Spectrograph is required
+        if isinstance(spectrograph, basestring):
```
rcooke-ast: I thought we decided that spectrograph would only be an instance of the spectrograph class?

pypit/calibrations.py
```diff
@@ -270,138 +313,164 @@ def get_pixflatnrm(self):
...
+        self.traceSlits = traceslits.TraceSlits(None, self.pixlocn, par=self.par['slits'],
+                                                det=self.det, setup=self.setup,
+                                                directory_path=self.master_dir,
+                                                mode=self.par['masters'], binbpx=self.msbpm)
+
+        # Load via master, as desired
+        if not self.traceSlits.master():
+            # Build the trace image first
+            trace_image_files = arsort.list_of_files(self.fitstbl, 'trace', self.sci_ID)
+            Timage = traceimage.TraceImage(self.spectrograph,
```
rcooke-ast: should this be called `traceImage` instead of `Timage` to keep with our standard?

pypit/spectrographs/keck_nires.py
```diff
@@ -14,18 +14,42 @@
...
 from pypit import msgs
 from pypit import ardebug as debugger
-from. import spectroclass
+from pypit.par.pypitpar import DetectorPar
+from pypit.spectrographs import spectrograph
+from pypit import telescopes
 
-class KeckNIRESpectrograph(spectroclass.Spectrograph):
+# TODO: KeckNIRESSPectrograph instead (i.e. two SS) ?
+class KeckNIRESpectrograph(spectrograph.Spectrograph):
```
rcooke-ast: Or... `KeckNiresSpectrograph`?

pypit/arms.py
kbwestfall: I've been assigned #329, so I'll aim to deal with this in a future PR.  To be honest, I actually fine the Exception types useful as a quick assessment of what's going on.  I don't find `PypitError` as useful, but we should talk about this in a future PR.  Punting on this for now.

pypit/arms.py
kbwestfall: Yep.  Will do this before 0.9.

pypit/arms.py
kbwestfall: I figured out a way around the confusion in my previous implementation before that was confusing this with the python function reduce.  So I think we could change it back to reduce, if that's better.

pypit/arms.py
kbwestfall: I've just commented this out for the time being.  That part of the code is never touched because a few lines above the code says:
```
save_format = 'fits'
if save_format == 'fits':
```

pypit/armsgs.py
kbwestfall: Actually it hasn't been fully removed yet.  Just the bit that was in `pypit.py` that skipped printing the Traceback if the mode wasn't set to `develop`.

pypit/bpmimage.py
kbwestfall: This was my middle ground solution that I didn't change post Tuesday's discussion.  This kind of `if..elif..else` block for the spectrograph is everywhere.  I'm again going to punt on this.  It's easy enough to pull it out in a later PR.

pypit/calibrations.py
kbwestfall: Done.

pypit/spectrographs/keck_nires.py
kbwestfall: Yeah, I agree with you.  We have a somewhat mixed convention at the moment, though.  For example, we have `KeckDEIMOSSpectrograph` and `KeckNIRSPECSpectrograph`, but also `WhtIsisSpectrograph`...  We should discuss a convention and then change these appropriately.  Sorry, I keep doing this, but I'm punting this one to a later PR, as well.

kbwestfall:
Travis is passing, so I'm mergin' it!

@rcooke-ast, I'll raise the pypit format as an issue.  We definitely want this to be easy.  To be clear, there isn't a strict connection between the parameter sets as they are and the config format.  When I first went down that path, I thought there would be a pretty strong connection, but it's more limited now.  So if we want another format, we would just need to write the appropriate parser for it.

However, I think people will need to spend time with the documentation and understand the syntax no matter what we do.  A first step may be to ask if there's a way we can reorganize the parameter set hierarchy to make this easier.  I'd be happy if we reogranize things that effectively removed the nesting in the pypit file, but it's not clear if that makes sense, particularly given the `Calibrations` class.  And I like the current, largely direct connection between the structure of `PypitPar` and what you put into the pypit file to define it.  I'm definitely open to changing things, though.

PR #432, opened 2018-08-06
Extract mpia

jhennawi:
Added xshooter fixes needed to address slit tracing issue in dev suite. Have also now added docs to the flips in the dispaxis. 

Joe

### DIFF
--- a/pypeit/par/pypeitpar.py
+++ b/pypeit/par/pypeitpar.py
@@ -2156,7 +2156,7 @@ class DetectorPar(ParSet):
         options['dispaxis'] = [ 0, 1, -1]
         dtypes['dispaxis'] = int
         descr['dispaxis'] = 'Spectra are dispersed along this axis (0 for row, 1 for column,' \
-                            + 'negative sign means there will be a flip because wavelengths will always be increasing upward)'
+                            'negative sign means there will be a flip because wavelengths will always be increasing upward)'
 
         defaults['xgap'] = 0.0
         dtypes['xgap'] = [int, float]
--- a/pypeit/par/pypeitpar.py
+++ b/pypeit/par/pypeitpar.py
@@ -2155,7 +2155,8 @@ class DetectorPar(ParSet):
         defaults['dispaxis'] = 0
         options['dispaxis'] = [ 0, 1, -1]
         dtypes['dispaxis'] = int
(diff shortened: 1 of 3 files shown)


pypeit/par/pypeitpar.py
```diff
@@ -2155,7 +2155,8 @@ def __init__(self, dataext=None, dispaxis=None, xgap=None, ygap=None, ysize=None
         defaults['dispaxis'] = 0
         options['dispaxis'] = [ 0, 1, -1]
         dtypes['dispaxis'] = int
-        descr['dispaxis'] = 'Spectra are dispersed along this axis (0 for row, 1 for column)'
+        descr['dispaxis'] = 'Spectra are dispersed along this axis (0 for row, 1 for column,' \
```
kbwestfall: I think we'd rather have a new parameter that's specifically for flipping the spectral direction, like `dispflip` or something, because `0 == -0` is True.

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -180,3 +180,114 @@ def bpm(self, shape=None, filename=None, det=None, **null_kwargs):
...
+        # chk_dict is 1-indexed!
+        chk_dict[1] = {}
+        # THIS CHECK IS A MUST! It performs a standard check to make sure the data are 2D.
+        chk_dict[1]['NAXIS'] = 2
+        # Check the CCD name
+        chk_dict[1]['HIERARCH.ESO.DET.CHIP1.NAME'] = 'MIT/LL CCID-20'
+        return chk_dict
+
+    def header_keys(self):
+        head_keys = self.xshooter_header_keys()
```
kbwestfall: Just `return self.xshooter_header_keys()` .

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -180,3 +180,114 @@ def bpm(self, shape=None, filename=None, det=None, **null_kwargs):
...
+
+        Returns:
+            arcparam is modified in place
+
+        """
+        #debugger.set_trace() # THIS NEEDS TO BE DEVELOPED
+        arcparam['lamps'] = ['OH_triplespec']
+
+
+    def bpm(self, shape=None, filename=None, det=None, **null_kwargs):
```
kbwestfall: This is the same as the Parent version so you don't need to redefine it.

PR #438, opened 2018-08-10
Fix flatfield bug

rcooke-ast:
Addresses Issue #418.

In the previous version of the code, the normalized pixel flat contained the slit illumination (spatial) profile. When processing science frames, this normalized pixelflat was then multiplied by the slit profile (So the slit profile was accounted for twice). The fix forces the normalized pixel flat to be truly normalized (with deviations about 1) - it is the flat field frame with both the blaze and the slit illumination profile taken out.

### DIFF
--- a/pypeit/processimages.py
+++ b/pypeit/processimages.py
@@ -399,8 +399,7 @@ class ProcessImages(object):
             msgs.error('No bpm for {0}'.format(self.spectrograph.spectrograph))
 
         # Flat-field the data and return the result
-        self.stack = flat.flatfield(self.stack, self.pixel_flat, self.bpm,
-                                    slitprofile=self.slitprof)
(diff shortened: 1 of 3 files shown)


pypeit/processimages.py
```diff
@@ -400,10 +400,9 @@ def flat_field(self, pixel_flat, bpm, slitprofile=None):
 
         # Flat-field the data and return the result
         self.stack = flat.flatfield(self.stack, self.pixel_flat, self.bpm,
-                                      slitprofile=self.slitprof)
+                                    slitprofile=self.slitprof)
```
jhennawi: Since we always compute a pixel flat and a slit profile, I would probably slightly prefer that slitprofile be a required rather than an optional argument here.

rcooke-ast: Done.

PR #480, opened 2018-09-07
Parsing with spaces

jhennawi:
This very small PR address an issue with the parsing of the .pypeit file which was breaking for the case where the path to the data files has a space in it. In particular: 

/Volumes/GoogleDrive/Team\ Drives/PHYS-GP-Hennawi/PypeIt/PypeIt-development-suite/

Unfortunately Google File Stream defaults to putting Team Drives in a directory with a whitespace. This fix was thus necessary to get the dev suite to work with the new Google Drive dev suite RAW_DATA directory. 

Anyway, I don't see a reason why the pipeline has to crash for paths with a whitespace in them, and my changes to the util.py parser fix this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,7 @@
 - Hot fix for bspline + requirements update
 - Fixed issue with biases being written to disk as untrimmed. 
 - Completely reworked flat fielding algorithm. 
-
+- Fixed some parsing issues with the .pypeit file for cases where there is a whitepsace in the path. 
 
 0.8.1
 -----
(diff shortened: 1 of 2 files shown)


kbwestfall:
Cool.  Glad it worked with that Shane Kast data section.  What I meant by tabulated data section is when the pypeit file looks like this one:

https://github.com/pypeit/PypeIt-development-suite/blob/master/pypeit_files/keck_deimos_830g_m_8500.pypeit

I'm happy if both formats work.

jhennawi:
yeah it handles those no problem.

PR #498, opened 2018-09-14
speedup

rcooke-ast:
After doing some very quick benchmarking and testing, I've increased the speed of the wavelength calibration by a factor of 10 for the test suite and a factor of 20 for the tests. This (hopefully) addresses the PR in the dev suite.


rcooke-ast:
Note: The travis failure is due to the flexure and nothing else. I haven't checked the dev suite yet (as I don't seem to have the latest), but all of the wavelength solutions being tested in travis still work.

PR #503, opened 2018-09-21
Turn off flexure for Keck NIRES

kbwestfall:
Sorry that this is coming long after you asked me to look at this.

This turns off the flexure correction for Keck NIRES in the default pypeit parameters.  As it is, you effectively can't set any parameter to `None` unless it is also `None` by default.  We can change this, but it may have pretty significant follow-on effects for the general implementation of the parameter sets.  To turn the flexure off, you just have to set the entire ParSet to None.

This needs to be tested to make sure the behavior is as expected, but I hope this fixes the problem.

### DIFF
--- a/pypeit/bpmimage.py
+++ b/pypeit/bpmimage.py
@@ -112,7 +112,7 @@ class BPMImage(object):
         else:
             _datasec = datasec
             if self.spectrograph is None and _datasec is None:
-                _datasec = [[slice(None),slice(None)]]
+                _datasec = [(slice(None),slice(None))]
                 _numamplifiers = 1
             if _datasec is None:
                 # Get the data sections from the spectrograph definition
--- a/pypeit/check_requirements.py
+++ b/pypeit/check_requirements.py
@@ -17,6 +17,6 @@ for requirement in install_requires:
         raise ImportError("Package: {:s} not installed!".format(pkg))
     else:
         if pkg_resources.parse_version(pv) < pkg_resources.parse_version(version):
-            print("Version of package {:s} = {:s}".format(pkg, pv))
-            raise ImportError("You need version >= {:s}".format(version))
+            raise ImportError('Your version of {0} is incompatible with PypeIt.  '.format(pkg)
+                                + 'Please update to version >= {0}'.format(version))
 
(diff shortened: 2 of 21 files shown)


kbwestfall:
This now includes a number of fixes to the tests, as well.  Most of these are just changing relevant paths to work with the dev suite.  Non-trivial changes are:

`test_flatfield.py`
 - removed test_step_by_step; we need tests of the steps in flat.fit_flat
 - changed value in assertion; check that this is valid

`test_flexure.py`
 - fixed but shift is substantially different than it used to be

There are still a number of failures in the test-suite that need to be fixed, and I'm not comfortable doing them myself.  They're in:  `test_flux.py`, `test_fluxspec.py`, and `test_wavecalib.py`.  Can someone(s) volunteer to fix these?  The first two have to do with errors in specobjs, and the latter has to do with the arc line parameters.

kbwestfall:
Setting a parameter to None in a `ParSet` is (almost) always like telling it to use the default.  The only exception is when the default value of the parameter is actually None.  Turning this functionality off would basically break ParSets.  The way I envisioned turning off different modules that require a given parameter set was to not even define the parameter set; e.g., par['flexure'] = None, not par['flexure']['method'] = None.  If you want the ability to set parameters to None, the easiest thing to implement is to have their default be None or to allow 'None' (i.e. a string) be one of the allowed values.  We can talk about this more on Tuesday.

jhennawi:
Hi @kbwestfall, 

Okay I think I see your point. For the flexure parset it is particularly confusing, since the docs say: 

"Method used to correct for flexure. Use None for no correction." Whereas actually following instructions and setting the method parameter to None will actually end up setting this to the default. Thus the user thinks they are not correcting whereas they are. 

My suggestion is that we fix this by allowing 'nocorr' or something along those lines. 'None' has the disadvantage that a user may try to set it to None in code, which then does not achieve the desired result. If we have to set the entire parset to None, well how do you do that in a .pypeit file using cfg? 

I can imagine a case where a user wants to turn off flexure compensation for one reduction in a pypeit file, but who does not want or know how to modify this stuff in code. 

Joe

kbwestfall:
@jhennawi, you make a couple of good points.
 - Whatever we decide, I will improve the docs to make this less confusing
 - I had originally intended flexure corrections to effectively always be off, meaning that the user would have to turn them on using the pypeit file.  Then we introduced the spectrograph defaults, which basically always turns the flexure correction on.  I think this means that the user effectively can't turn off the flexure correction if it's on by default for the specific spectrograph.  I agree that this should be changed.  I can add a "method" that will skip the correction.

jhennawi:
Hi @kbwestfall , 

Rather than have a method that skips the correction, why don't we just come up with a convenient sounding string that indicates no correction and document that.  Then the parsets will work as they are. It seems too much trouble to make a separate method for this. 

Joe

PR #508, opened 2018-09-23
slit boundary tweaks

jhennawi:
This PR implements changes to the flat field routine to allow the slit/order boundaries to be tweaked. The TraceSlits class does a nice job of finding slits, but the boundaries are too liberal. This is typical when one is tracing a sawtooth filtered image. The flat field routines create a slit/order illumination profile, which can then be used to determine a more meaningful set of slit boundaries. This dramatically improves the quality of the reductions. 

This was not a trivial thing to implement. The actual tweaks are easy, but the point is that the flat field code runs downstream in calibrations from the codes that generate slits and tilts. As such, the previous slits and tilts need to be updated by the flat routine, as do the master files etc. I've implemented this in a sensible way, but it required that I touch many different parts of the calibration routines. 

There are some other things in here related to X-shoote development. They probably don't belong in this PR, but X-shooter is still under development so this does not break anything.

I also need to pull in PR #503 because I needed to check that NIRES was working properly and that PR was a fix to NIRES. So this PR should be merged after 503 clears.


pypeit/par/pypeitpar.py
```diff
@@ -431,6 +431,11 @@ def __init__(self, frame=None, illumflatten=None, method=None): #, params=None,
         dtypes['illumflatten'] = bool
         descr['illumflatten'] = 'Use the flat field to determine the illumination profile of each slit.'
 
+        defaults['tweak_slits'] = True
+        dtypes['tweak_slits'] = bool
+        descr['tweak_slits'] = 'Use the illumination flat field to tweak the slit edges. illumflatten must be set to true for this to work'
```
kbwestfall: Add this to the validate method.  I.e., check that illumflatten is true when tweak_slits is true.

pypeit/par/pypeitpar.py
```diff
@@ -404,7 +404,7 @@ class FlatFieldPar(ParSet):
     For a table with the current keywords, defaults, and descriptions,
     see :ref:`pypeitpar`.
     """
-    def __init__(self, frame=None, illumflatten=None, method=None): #, params=None, twodpca=None):
+    def __init__(self, frame=None, illumflatten=None, tweak_slits = None, method=None): #, params=None, twodpca=None):
```
kbwestfall: Sorry that I'm nit-picking:  In general, please separate arguments of a function with a comma and a space, and don't include spaces between a keyword argument and its default value.  I.e. do this:

```
def foo(arg1, arg2, kwarg1=0., kwarg2=1.)
```

not what you tend to do, which is:

```
def foo(arg1,arg2,kwarg1 = 0.,kwarg2 = 1.)
```

pypeit/traceslits.py
```diff
@@ -149,6 +149,8 @@ def __init__(self, mstrace, pixlocn, par=None, det=None, setup=None, master_dir=
         self.lordpix = None
         self.rordpix = None
         self.slitpix = None
+        self.lcen_tweak = None   # Place holder for tweaked slit boundaries from flat fielding routine.
```
kbwestfall: Is it useful to keep both the tweaked values and the originals?

pypeit/core/arc.py
```diff
@@ -79,8 +78,33 @@ def setup_param(spectro_class, msarc_shape, fitstbl, arc_idx,
...
+
+    for islit in range(nslits):
+        msgs.info("Extracting an approximate arc spectrum at the centre of slit {:d}".format(islit + 1))
+        # Create a mask for the pixels that will contribue to the arc
+        spat_img = np.outer(np.ones(nspec), np.arange(nspat))  # spatial position everywhere along image
+        trace_img = np.outer(trace[:,islit], np.ones(nspat))  # left slit boundary replicated spatially
+        arcmask = (slitpix > 0) & inmask & (spat_img > (trace_img - box_rad)) & (spat_img < (trace_img + box_rad))
+        this_mean, this_med, this_sig = sigma_clipped_stats(arcimg, mask=~arcmask, sigma=3.0, axis=1)
+        arc_spec[:,islit] = this_med.data
+        if ~np.any(arc_spec[:,islit]):
```
kbwestfall: Use `not`:
```
>>> t = True
>>> ~t
-2
```

pypeit/core/flat.py
```diff
@@ -109,27 +125,51 @@ def fit_flat(flat, mstilts, thismask, slit_left, slit_righ, inmask = None,spec_s
...
-    nspec = flat.shape[0]
-    nspat = flat.shape[1]
-    piximg = mstilts * (nspec-1)
-    # Compute the approximate number of pixels sampling each spatial pixel for this slit
-    npercol = np.fmax(np.floor(np.sum(thismask)/nspec),1.0)
+    shape = flat.shape
+    if shape != tilts_dict['tilts'].shape:
+        msgs.error('Something is very wrong. Tilt image shape does not match flat field image shape')
+
+    nspec = shape[0]
```
kbwestfall: `nspec, nspat = shape`

pypeit/core/flat.py
```diff
@@ -109,27 +125,51 @@ def fit_flat(flat, mstilts, thismask, slit_left, slit_righ, inmask = None,spec_s
...
 
+    # Fit the spectral direction of the blaze. We do this in the log
     log_flat = np.log(np.fmax(flat, 1.0))
     inmask_log = ((flat > 1.0) & inmask)
     log_ivar = inmask_log.astype(float)/0.5**2 # set errors to just be 0.5 in the log
 
-    # Flat field pixels for fitting spectral direction
-    fit_spec = thismask & inmask & (edgmask == False)
+    # Flat field pixels for fitting spectral direction. Restrict to original slit pixels
+    fit_spec = thismask_in & inmask & (edgmask_in == False)
```
kbwestfall: `np.invert(edgmask_in)`

pypeit/core/flat.py
```diff
@@ -212,88 +238,136 @@ def fit_flat(flat, mstilts, thismask, slit_left, slit_righ, inmask = None,spec_s
...
+                                                               inmask = imed, nord=4,upper=5.0, lower=5.0,fullbkpt = fullbkpt)
 
+    # Evaluate and save
+    illumflat = np.ones_like(flat)
+    illumflat[thismask], _ = spat_set.value(ximg[thismask])
+    norm_spec_spat = np.ones_like(flat)
+    norm_spec_spat[thismask] = flat[thismask]/np.fmax(spec_model[thismask], 1.0)/np.fmax(illumflat[thismask],0.01)
+
+    if tweak_slits:
+        tweak_left = False
```
kbwestfall: Pull this tweaking algorithm into a function.  How robust is this algorithm?  I.e., do you have a good sense of when/if it will break?

pypeit/core/flat.py
```diff
@@ -28,8 +29,10 @@
 import scipy
 
 
-def fit_flat(flat, mstilts, thismask, slit_left, slit_righ, inmask = None,spec_samp_fine = 1.2,  spec_samp_coarse = 50.0,
-             spat_samp = 5.0, spat_illum_thresh = 0.01, npoly = None, trim_edg = (3.0,3.0),debug = False):
+
+def fit_flat(flat, tilts_dict, thismask_in, slit_left_in, slit_righ_in, inmask = None,spec_samp_fine = 1.2, spec_samp_coarse = 50.0,
```
kbwestfall: Can you break this function into smaller chunks?  I suggest pulling out the tweaking algorithm into a different function.  You could also pull the debug plots into separate qa plotting functions.

pypeit/core/extract.py
```diff
@@ -1171,7 +1171,7 @@ def parse_hand_dict(hand_extract_dict):
 
 def objfind(image, thismask, slit_left, slit_righ, inmask = None, FWHM = 3.0,
             hand_extract_dict = None, std_trace = None, ncoeff = 5, nperslit = 10,  bg_smth = 5.0, pkwdth = 3.0,
-            maskwidth = 3.0, sig_thresh = 10.0, peak_thresh = 0.0, abs_thresh = 0.0, trim_edg = (3,3), objmask_nthresh = 2.0,
+            maskwidth = 3.0, sig_thresh = 5.0, peak_thresh = 0.0, abs_thresh = 0.0, trim_edg = (3,3), objmask_nthresh = 2.0,
```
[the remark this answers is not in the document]
jhennawi: So I'm just lagging behind on putting this stuff in the parsets because I've been too busy trying to repair all the things in the code that are broken. All of these hard wired parameters are going to go into the parset.

pypeit/par/pypeitpar.py
jhennawi: Added to the validate method.

pypeit/par/pypeitpar.py
jhennawi: Fixed. Thanks!

pypeit/core/arc.py
jhennawi: Fixed.

pypeit/core/flat.py
jhennawi: Fixed.

pypeit/core/flat.py
jhennawi: I've moved this algorithm into a function as you requested.

pypeit/core/flat.py
jhennawi: I broke our the slit tweaking. I'm leaving the debug/QA stuff where it is because this may end up being subprocess plot in interactive mode. If that is the case, it needs to be defined in the function as with the other interactive subprocess plots. If we dump this out to QA and don't plot to the screen then it could go into a separate function later.

jhennawi:
Hi @kbwestfall, 

I've addressed all your requests in the code. Your main concern is below: 

"I imagine it will come with testing, but my main question is about robustness. Do you have a sense of when/if the tweaking algorithm will break? My spidey sense is worried about the code being slow and fragile given some of the hard-coding and brute-force stepping involved. I'm happy to be proven wrong. And if more testing is needed, I'm happy to just merge it now and put it through its paces."

This code is extremely robust and does not fail anywhere in the test suite. I have tested it extensively on every single spectrograph. It is neither buggy nor slow. That is more than I can say for the rest of the pypeit core functionality (setup, arclines, tilts, slits) which are all crashing or failing in multiple places in the dev suite.  You can try the flat fielding code in the dev suit with debug = True in fit_flat, or I can show you how it works in our next telecon. 

Things are not really hardwired anywhere. Eventually the parameters will go into the parsets, but I have not gotten around to that yet for lack of time. In practice the defaults work fine for just every spectrograph I've tested on. I code things in pixel units, and decide on parameters relative to things like the slit width or image sizes, such that everything is basically self similar, and one set of parameters governs the entire procedure.

jhennawi:
Sounds good @kbwestfall, please merge #503 and then merge this.

PR #511, opened 2018-09-25
Xshooter ema

EmAstro:
This is a small PR.
It mainly deals with a bug in procimg.py when defining the gain image and adds some xshooter specific parameters.


pypeit/core/procimg.py
```diff
@@ -456,7 +456,12 @@ def gain_frame(datasec_img, namp, gain_list):
     msgs.warn("Should probably be measuring the gain across the amplifier boundary")
 
     # Loop on amplifiers
-    gain_img = np.zeros_like(datasec_img)
+    # ToDo
+    # EMA --> I think there was a bug here.
+    # the previous command was: gain_img = np.zeros_like(datasec_img)
+    # but it returns an array of integers, so if the gain was <0.6 it
+    # was rounded to 0 causing a lot of problems.
+    gain_img = np.zeros_like(datasec_img,dtype=float)
```
jhennawi: This is a massive bug @profxj! We have been rounding the gain to an integer all along!!!

PR #515, opened 2018-09-27
Adding GNIRS class

feigewang:
Adding gnirs class.  I add one dispaxis (=2) option in the pypeitpar.py to flip the image.


pypeit/spectrographs/spectrograph.py
```diff
@@ -143,6 +143,9 @@ def load_raw_frame(self, raw_file, det=None):
             img = img.T
         elif self.detector[_det - 1]['dispaxis'] == -1:
             img = np.flip(img.T, axis=0)
+        elif self.detector[_det - 1]['dispaxis'] == 2:
```
kbwestfall: My original suggestion was that we add a keyword to DetectorPar, something like `dispflip`.  Then the code would look like this:

```
if self.detector[_det-1]['dispaxis'] == 1:
    # Transpose so that wavelength is along rows
    img = img.T
if self.detector[_det-1]['dispflip']:
    # Flip so that pixel number correlates with wavelength
    img = np.flip(img, axis=0)
```

pypeit/telescopes.py
```diff
@@ -12,13 +12,19 @@
...
+                                               elevation=4160.0)
+
 class KeckTelescopePar(TelescopePar):
     def __init__(self):
         super(KeckTelescopePar, self).__init__(name='KECK',
                                                longitude=155.47833,
                                                latitude=19.82833,
                                                elevation=4160.0,
-                                               fratio=15,
                                                diameter=10)
```
kbwestfall: Please put `fratio=15` back.

pypeit/par/pypeitpar.py
```diff
@@ -2334,7 +2334,7 @@ def __init__(self, dataext=None, dispaxis=None, xgap=None, ygap=None, ysize=None
 
         # TODO: Should this be detector-specific, or camera-specific?
         defaults['dispaxis'] = 0
-        options['dispaxis'] = [ 0, 1, -1]
+        options['dispaxis'] = [ 0, 1, -1, 2]
```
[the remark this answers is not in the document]
feigewang: Done. Thanks!

pypeit/spectrographs/spectrograph.py
feigewang: I have put back the deleted line in KeckTelescopePar

jhennawi:
I've added in the dispflip and documented it. It appears to work fine. Merging.

pypeit/spectrographs/spectrograph.py
jhennawi: I've added this now Kyle. I think we are good.

pypeit/telescopes.py
jhennawi: Done.

PR #521, opened 2018-09-29
PypeIt Setup lives

kbwestfall:
This PR is a major step toward PypeIt setup being functional again.  There are a few major changes:
   - `fitstbl` is now a PypeIt object called `PypeItMetaData`.  The purpose for making it a new object was to streamline the frame typing and matching, collecting most of the relevant code into a class with some core code in `framematch`.  I basically picked through and refactored some of `core.load` and `core.fsort` functions into new modules.
   - I've painstakingly gone through all the spectrograph classes to update the necessary header keywords and frame matching criteria so that PypeIt setup is successful for the instruments in the dev suite.  I'm doing this PR now despite not quite having all the tests working in `test_setups.py`.  This is one of the more important things for you look at because I don't want to have to fix more conflicts when other branches get merged.
   - I'm now typing the frames using bits using the `BitMask` class I pulled from the MaNGA DAP.  I've also incorporated this into `ScienceImage`.

There are also a lot of minor style and doc edits.

I wanted to put this on your radar with the PR and start having you review the code.  But there is still some work to do before it gets merged:
   - I have one more test to fix.  It's Friday and I'm done for the day...
   - I'm worried that there is a memory leak somewhere.  I need to do some profiling, which is made more difficult by the fact that the dev suite has ballooned.  In tests yesterday, I was running out of memory on my 16 GB laptop when running the dev suite.  Even in `master`, the first DEIMOS reduction (1200 line grating) was requiring 10GB of RAM.  Does that seem crazy to anyone else?

Please start reviewing the code, but I won't want to merge it until I've fixed/understood the issues above.


pypeit/scienceimage.py
```diff
@@ -4,27 +4,50 @@
...
+class ScienceImageBitMask(BitMask):
+    """
+    Define a bitmask used to set the reasons why each pixel in a science
+    image was masked.
+    """
+    def __init__(self):
+        # TODO:
+        #   - Can IVAR0 and IVAR_NAN be consolidated into a single bit?
+        #   - Is EXTRACT ever set?
+        mask = {       'BPM': 'Component of the instrument-specific bad pixel mask',
```
profxj: Whoa, I like this.

pypeit/scienceimage.py
```diff
@@ -147,7 +173,9 @@ def __init__(self, spectrograph, file_list, det = None, objtype='science', scidx
         self.skymask = None
         self.objmask = None
         self.outmask = None
+        # TODO: (KBW) I'd like to change the bitmask nomenclature
```
profxj: Please do.

pypeit/scienceimage.py
```diff
@@ -737,227 +789,52 @@ def __repr__(self):
...
-                                          self.sciframe.shape, self.maskslits, self.det, self.scidx, self.fitstbl,
-                                          self.tracelist,
-                                          binning=self.binning, objtype=self.objtype)
-        # Step
-        self.steps.append(inspect.stack()[0][3])
-        # Return
-        return self.specobjs
-'''
\ No newline at end of file
+#def unpack_bitmask(bitmask):
```
profxj: Remove this

jhennawi:
Hi @kbwestfall, 

@EmAstro put in some very minor changes to deal with X-shooter, which I just merged in.  Now there appear to be conflicts with your branch. Can you take a look. 

Joe

kbwestfall:
@jhennawi, @EmAstro , I was able to fix the conflicts with master, but yes please let me know if this wasn't done right.  I'll wait to merge until you approve.  The conflicts had to do with the parameter used and some of the header keywords.  I should have used all the parameters that are currently in master, replacing the old ones.  The `binning` keyword is now handled in the specific VLT spectrograph class now in `validate_metadata`.  There are a couple of other such methods for other spectrographs that need to do similar things.

pypeit/core/framematch.py
```diff
@@ -0,0 +1,451 @@
...
+    if len(exprng) != 2:
+        raise ValueError('exprng must have two elements.')
+    indx = np.ones(len(exptime), dtype=bool) if exprng[0] is None \
+                else exptime > exprng[0]
+    if exprng[1] is not None:
+        indx &= (exptime < exprng[1])
+    return indx
+    
+
+def group_AB_frames(file_list, targets, coords, max_nod_sep=2):
```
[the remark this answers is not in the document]
kbwestfall: Don't get too impressed.  This is largely a copy of the function that was in `fsort`, and I haven't tested that it actually works!

pypeit/core/wave.py
```diff
@@ -486,55 +486,55 @@ def geomotion_correct(specObjs, maskslits, fitstbl, scidx, time, longitude, lati
...
-                if not hasattr(specobj, attr):
-                    continue
-                if 'WAVE' in getattr(specobj, attr).keys():
-                    msgs.info('Applying {0} correction to '.format(refframe)
-                              + '{0} extraction for object:'.format(attr)
-                              + msgs.newline() + "{0}".format(str(specobj)))
-                    getattr(specobj, attr)['WAVE'] = getattr(specobj, attr)['WAVE'] * vel_corr
-    # Return
-    return vel, vel_corr  # Mainly for debugging
+#def geomotion_correct_oldbuggyversion(specobjs, maskslits, fitstbl, scidx, time, longitude, latitude, elevation,
```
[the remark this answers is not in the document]
kbwestfall: Done, but there are a lot of other functions that I think @jhennawi has made obsolete.  I'll leave him to delete the rest of them.

pypeit/metadata.py
```diff
@@ -0,0 +1,841 @@
...
+        required_for_ABBA = ['ra', 'dec'] # and for standards?
+        added_by_fsort = ['frametype', 'framebit']
+        if any([ c not in data.keys() for c in required_columns]):
+            msgs.warn('Columns are missing.')
+
+        # Number of files to read
+        numfiles = len(file_list)
+
+        # Loop on files
+        for i in range(numfiles):
```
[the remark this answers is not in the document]
kbwestfall: yep, you're right.  I only ever used `i` to select the file.  Fixed.

pypeit/metadata.py
```diff
@@ -0,0 +1,841 @@
...
+#            utc = self.get_utc(headarr)
+#            data['utc'].append('None' if utc is None else utc)
+#            if utc is None:
+#                msgs.warn('UTC is not listed as a header keyword in file:' + msgs.newline()
+#                          + file_list[i])
+
+            # TODO: Read binning-dependent detector properties here? (maybe read speed too)
+    
+            # Now get the rest of the keywords
+            for k in data.keys():
```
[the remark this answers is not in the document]
kbwestfall: Yep, good point.  Fixed.

pypeit/metadata.py
```diff
@@ -0,0 +1,841 @@
...
+    nfile : int, optional
+      Number of files to mimic
+    spectrograph : str, optional
+      Name of spectrograph to mimic
+    notype : bool (optional)
+      If True, do not add image type info to the fitstbl
+
+    Returns
+    -------
+    fitstbl : Table
```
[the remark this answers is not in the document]
kbwestfall: `PypeItMetaData`.  Fixed.

pypeit/pypeit.py
```diff
@@ -11,68 +11,76 @@
...
-
-    Inherited Attributes
-    --------------------
     """
     __metaclass__ = ABCMeta
 
     def __init__(self, spectrograph, setups_path=None, verbosity=2,
                  overwrite=True, logname=None, show=False):
 
+        # Spectrometer class
```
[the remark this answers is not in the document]
kbwestfall: Good point.  I just moved these tests into `load_spectrograph`.  I.e., you can now pass `load_spectrograph` a `Spectrograph` instance, that it will simply return, or a string.  I'm not crazy about the idea that `load_spectrograph` can return `None` because it means that the relevant class is not well defined.  I understand why this is allowed for the master frames, but it could lead to problems for users.

pypeit/scienceimage.py
kbwestfall: Done.  @jhennawi please make sure you agree with this change.

kbwestfall:
Tests pass now, including those requiring the dev suite.  I still need to work out the memory issues and continue testing by running the dev suite.

There are some subtleties with the frame typing that I need to think through.  I'm not sure of the strategy that we should go for.  I think I'd rather have `pypeit_setup` be virtually bullet proof when using the defaults, even if that means the frame typing is poor.  I think generally people are going to want to edit their pypeit files and/or use SPIT.  So think we should aim to make the defaults as general as possible just to make sure that pypeit_setup succeeds, even if that means users will virtually always need to edit their pypeit file before doing the reduction.  We can discuss this in a couple of weeks.

pypeit/ginga.py
```diff
@@ -54,60 +51,88 @@ def connect_to_ginga(host='localhost', port=9000, raise_err=False):
...
-    bitmask: ndarray (2D)
-      bitmask produced by PypeIt extraction illustrating which pixels were masked and why
-
-    exten: int, optional
-      extension of image in fits file. Only passed in if inp is a file
-
-    Returns
-    -------
-
+def show_image(inp, chname='Image', waveimg=None, bitmask=None, mask=None, exten=0, cuts=None,
```
jhennawi: I don't like that the user has to supply this bitmask object as an argument into the call. There is no reason for this extra argument, why not just directly import bitmask. These show_image routines are at the moment only every going to use a single bitmask, which is that created by the scienceimage class. If we decided to add additional bit masks in the future then we can make this change. 

Also, have you checked all the dependencies on show_image to see that they are not now going to break because of this change? See for example the show method in science image, or the show_twod.py script. 

We should create tests for these. But I would try to grep show_image for the moment, because it is used by many routines.  

Joe

pypeit/ginga.py
```diff
@@ -54,60 +51,88 @@ def connect_to_ginga(host='localhost', port=9000, raise_err=False):
...
+            image in other channels to it.
+
+    Returns:
+        ginga.util.grc.RemoteClient, ginga.util.grc._channel_proxy: The
+        ginga remote client and the channel with the displayed image.
+
+    Raises:
+        ValueError:
+            Raised if `cuts` is provided and does not have two elements
+            or if bitmask is provided but mask is not.
```
jhennawi: Thanks for updating the docs!!

pypeit/scienceimage.py
jhennawi: I agree, looks great.

jhennawi:
Hi @kbwestfall, 

Let me add that I agree with your points on frame-typing. In general the task of automating the matching of frames to science data is not really something you can fully automate. The point is only the user knows which calibrations they should use depending on what happened at the telescope, etc. If in the future we convince people at Keck to take data in a more standard way, this might change. But for the moment, I'd work under the assumption that users will edit the pypeit files. 

Joe

jhennawi:
Hi Kyle, 

I'm branched from your branch, and I'm not able to get the tests to all run. Namely, test_spectrographs fails with the following error: 

I think the issue is with your dev suite files. It looks like you are somehow substituting underscores for colons in the file "XSHOO.2016-08-02T08:45:49.494.fits.gz which does exist. I'm not really sure why you would do such a thing, but perhaps this was generated in an automated way somehow. 

Joe

_____________________________________________ test_vltxshooternir ______________________________________________

    @dev_suite_required
    def test_vltxshooternir():
        s = spectrographs.vlt_xshooter.VLTXShooterNIRSpectrograph()
        example_file = os.path.join(os.environ['PYPEIT_DEV'], 'RAW_DATA', 'VLT_XSHOOTER',
                                    'NIR', 'XSHOO.2016-08-02T08_45_49.494.fits.gz')
>       assert os.path.isfile(example_file), 'Could not find example file for VLT Xshooter NIR read.'
E       AssertionError: Could not find example file for VLT Xshooter NIR read.
E       assert False
E        +  where False = <function isfile at 0x10c3ff950>('/Volumes/GoogleDrive/Team Drives/PHYS-GP-Hennawi/PypeIt/PypeIt-development-suite/RAW_DATA/VLT_XSHOOTER/NIR/XSHOO.2016-08-02T08_45_49.494.fits.gz')
E        +    where <function isfile at 0x10c3ff950> = <module 'posixpath' from '/Users/joe/anaconda3/lib/python3.6/posixpath.py'>.isfile
E        +      where <module 'posixpath' from '/Users/joe/anaconda3/lib/python3.6/posixpath.py'> = os.path

pypeit/tests/test_spectrographs.py:127: AssertionError

jhennawi:
Also, did you run the dev suite script all the way through. I thought that dev suite was running Keck_NIRES. Anyway, when I try to run NIRES it fails because of this line: 

# Use the ARMS pipeline
par['rdx']['pipeline'] = 'ARMS'

This is an easy fix, since you removed the pipeline parameter. But anyway I'd suggest trying to run the dev suite all the way through. I think it dies on a couple of setups that are crashing because of slits or arclines, for which there are all issues. 

Joe

jhennawi:
Hi Kyle, 

I'm getting more failures in the dev suite. Specifically everything fails. The failure is below. It looks like
you forgot to import msgs in core/paths.py. This is an easy fix, but I seem to be missing something because I don't see how you could have run on the dev suite. 

  File "/Users/joe/anaconda3/bin//run_pypeit", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/joe/python/PypeIt/bin/run_pypeit", line 17, in <module>
    rtval = run_pypeit.main(args)
  File "/Users/joe/python/PypeIt/pypeit/scripts/run_pypeit.py", line 97, in main
    pypeIt.init_setup(args.pypeit_file, redux_dir, calibration_check=True)
  File "/Users/joe/python/PypeIt/pypeit/pypeit.py", line 523, in init_setup
    overwrite=self.overwrite, redux_path=self.par['rdx']['redux_path'])
  File "/Users/joe/python/PypeIt/pypeit/core/paths.py", line 37, in make_dirs
    msgs.info("Creating Science directory")
NameError: name 'msgs' is not defined

jhennawi:
I fixed the messages bug and now there is another crash: 

  File "/Users/joe/anaconda3/bin//run_pypeit", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/joe/python/PypeIt/bin/run_pypeit", line 17, in <module>
    rtval = run_pypeit.main(args)
  File "/Users/joe/python/PypeIt/pypeit/scripts/run_pypeit.py", line 98, in main
    pypeIt.reduce_all(reuse_masters=args.use_masters)
  File "/Users/joe/python/PypeIt/pypeit/pypeit.py", line 229, in reduce_all
    sci_dict = self.reduce_exposure(sci_ID, reuse_masters=reuse_masters)
  File "/Users/joe/python/PypeIt/pypeit/pypeit.py", line 348, in reduce_exposure
    sciimg, sciivar, skymodel, objmodel, ivarmodel, outmask, sobjs, vel_corr = self._extract_one()
  File "/Users/joe/python/PypeIt/pypeit/pypeit.py", line 838, in _extract_one
    = wave.geomotion_correct(sobjs, maskslits, self.fitstbl, scidx,
NameError: name 'scidx' is not defined

jhennawi:
Hi @kbwestfall, 

Sorry for the confusion, I just realized that you didn't get to run on the dev suite yet. Anyway, you will encounter a bunch of crashes. I've put what I think are some of the fixes in spec_headers_test. I got to a point where I didn't know how to proceed, because your entries in the fitstable for the standard appear to be wrong, namely the exposure time is somehow zero. So for that reason the fluxing is going to fail. There is also a problem with std_header as the fluxing codes requires it. This stuff is all a bit of a mess, so feel free to clean it up using whatever cards are in your fitstable. 

While I would like to have branched from your branch, it appears it is not ready yet so I'm going to push forward from master and hopefully the merge will not be too bad. That said, I tried to run the NIRES and X-shooter dev suites through and they are all failing all over the place due to your changes. Things like the pypeline variable are easy enough to fix, but there is a bunch of binning and frame matching and other stuff which is now broken. 

While I realize that the echelle modes are not in the dev suite yet, and not official pypeit pipelines yet, the point is @EmAstro and @feigewang have spent a lot of time trying to get these codes working. If your changes break all that, it adds a bunch of work for us to figure out how to get things working again. 

So please add to the list of tasks before you merge that NIRES, X-shooter, and GNIRS need to not fail because of the changes you've made. Consult with @feigewang and @EmAstro, who can try help you with which files to run on, but they are in the dev suite. 

Thanks, 
Joe

pypeit/ginga.py
kbwestfall: Good point about the other scripts.  I'll fix this.

About passing bitmask and mask, you can't import scienceimage into ginga and ginga into scienceimage.  It creates a circular import.  You solved this by putting the import into the function but this is bad coding practice.  Imports go at the top.  So you can't only pass the mask data without also passing the bitmask object that interprets it.  There are a few ways around this, if we're really concerned about minimizing the number of arguments.

jhennawi: I guess we can make the bitmask class a separate module to solve this?

EmAstro:
Thanks for this massive amount of work @kbwestfall!
Is the echelle mode a renaming of ARMED or there are substantial changes? If you make Echelle to run, @feigewang and I can help you on fixing up the NIRES, X-shooter, and GNIRS spectrographs

kbwestfall:
@EmAstro, the idea is that we need to convert ARMED into a `PypeIt` subclass like `MultiSlit` but called `Echelle` (we can change the name if you want).  At the moment, this branch (`spec_headers`) will fail anytime you use the spectrographs where the `pypeline` method returns `Echelle`.  So, yes, we need to fix this.  I could try to do this myself, but I think you would do a better job of it, and I'm not going to be able to get to this until next week.  It would actually make sense to me if you would plan to put in some PRs that implement this in master that I can then merge with this `spec_headers` branch.  Does that make sense?

jhennawi:
Hi Guys,

For the moment just change the pypeline variable to MultiSlit and Xshooter
and NIRES will run as a multislit pipeline.

Kyle keep in mind that these instruments need to run up to the point they
were running up to In master so as to not break our development efforts.
NIRES runs all the way through and Xshooter crashes in wavelengths.

Joe

profxj:
closing this PR to reopen on the `gmos` branch

PR #528, opened 2018-10-11
Bitmask Hotfix

kbwestfall:
Very minor change to fix the change in the ScienceImage bitmask naming convention I introduced.

### DIFF
--- a/pypeit/pypeit.py
+++ b/pypeit/pypeit.py
@@ -859,7 +859,7 @@ class MultiSlit(PypeIt):
             # Set to sciivar. Could create a model but what is the point?
             ivarmodel = np.copy(sciivar)
             # Set to inmask in case on objects were found
-            outmask = sciI.bitmask
+            outmask = sciI.mask
             # empty specobjs object from object finding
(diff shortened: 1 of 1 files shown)


pypeit/pypeit.py
```diff
@@ -859,7 +859,7 @@ def _extract_one(self, std=False):
             # Set to sciivar. Could create a model but what is the point?
             ivarmodel = np.copy(sciivar)
             # Set to inmask in case on objects were found
-            outmask = sciI.bitmask
+            outmask = sciI.mask
```
jhennawi: Is sciI.mask a bitmask in the sense that it has integer bits set?

kbwestfall: Yeah, I changed "mask" to mean the integer value of the mask, and bitmask to the object that deciphers the mask values.

We forgot to talk about this today...

We can change this if you want in a different PR, but this change is needed now for master to pass the dev suite reductions.

jhennawi:
Looks great @kbwestfall thanks for fixing this!

PR #549, opened 2018-11-15
Adjustments to par['calibrations']['slits']  for XSHOOTER NIR

EmAstro:
This is the smallest PR ever (only 6 characters changed) to tweak xshooter NIR par for order finding. In this way we loose some precision on the last order, but we manage to get the first one that was previously missed.

xshooter_nirslits


feigewang:
Looks great!

PR #550, opened 2018-11-16
Bad pixel mask for DEIMOS + LACosmics adjustments

freddavies:
Fleshes out the DEIMOS bad pixel mask after hunting for bad columns on all detectors across a few different exposures.

Also starts to adjust LACosmics parameters for different instruments. Particularly, setting `objlim` to lower values greatly improves the cosmic ray rejection for DEIMOS and LRIS red with no adverse effects. The latter is quite sensitive to the binning on the detector, however, so some additional functionality was added to use different parameter values depending on the binning.

I also made the global default parameters modestly more aggressive, which seems to work well for various instruments on the dev suite. But, it's probably worth a closer look. See Figure 4 in http://adsabs.harvard.edu/abs/2001PASP..113.1420V -- the previous default `objlim=5` is more tuned towards heavily undersampled data (a la HST imaging), but I don't know how exactly that translates to spectra.

### DIFF
--- a/pypeit/par/pypeitpar.py
+++ b/pypeit/par/pypeitpar.py
@@ -280,7 +280,7 @@ class ProcessImagesPar(ParSet):
         dtypes['sigfrac'] = [int, float]
         descr['sigfrac'] = 'Fraction for the lower clipping threshold in LA cosmics routine.'
 
-        defaults['objlim'] = 2.5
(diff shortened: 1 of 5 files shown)


freddavies:
Cancelled pending some adjustments to how binning is passed around.

PR #630, opened 2019-02-26
Hot fixes for binning and 2d images not being saved

jhennawi:
This PR provides a hot fix to two massive bugs: 

- Fixed a bug that was introduced when the binning was switched to the PypeIt convention.
- Fixed a bug whereby 2d images were not being saved if no objects were detected.
- Revamped the naming convention of output files to have the original filename in it.

Tests all passed on the current version of Cooked (v1.0). I'm running the dev-suite now but do not anticipate any problems. 

Joe

screen shot 2019-02-26 at 00 51 27


CHANGES.rst
```diff
@@ -1,6 +1,8 @@
 0.9.3dev
 --------
-
+- Fixed a bug that was introduced when the binning was switched to the PypeIt convention.
+- Fixed a bug whereby 2d images were not being saved if no objects were detected.
```
profxj: Are you sure this was a `bug`?

pypeit/core/parse.py
```diff
@@ -786,7 +786,9 @@ def sec2slice(subarray, one_indexed=False, include_end=False, require_dim=None,
     ## JFH This is a kludge to get this to work becuase I cannot understand the nonsense below. The binning
     ## convention is now flipped relative to before and the code below appends to lists, so this is the
     ## easiest fix
-    _binning = _binning[::-1]
+    #from IPython import embed
```
profxj: remove

pypeit/scienceimage.py
```diff
@@ -6,7 +6,7 @@
 from pypeit import processimages
 from pypeit import utils
 from pypeit import ginga
-from pypeit.core import coadd2d
+from pypeit.core import coadd2d, pixels
```
profxj: Is `pixels` used?

pypeit/scripts/show_2dspec.py
```diff
@@ -8,17 +8,25 @@
...
 
 import argparse
 from astropy.table import Table
 from pypeit import ginga
 from pypeit.spectrographs import util
 from pypeit.processimages import ProcessImagesBitMask as bitmask
 from pypeit.core import pixels
+import os
+import numpy as np
+import IPython
```
profxj: remove `IPython`

pypeit/spectrographs/spectrograph.py
```diff
@@ -278,20 +278,25 @@ def get_datasec_img(self, filename, det=1, force=True):
             # Get the image shape
             raw_naxis = self.get_raw_image_shape(filename, det=det)
 
-            binning = self.get_meta_value(filename, 'binning')
-#            binning = self.parse_binning(filename)
+            binning_pypeit = self.get_meta_value(filename, 'binning')
 
             data_sections, one_indexed, include_end, transpose \
                     = self.get_image_section(filename, det, section='datasec')
+            if transpose:
```
profxj: Please add a comment on what `binning_raw` is.
Same with `binning_pypeit`.

profxj: note that it had been my intent to have binning *always* be (spec,spat)
anywhere in the code.  That is, if it is written as (spat,spec) to the header
or whatever, it is *immediately* flipped so that any/every variable will have it 
in that manner.  Can we maintain this?

CHANGES.rst
jhennawi: Yes it was a bug since those images were meant to be saved all along, and also the code uses the existence of 2d images to decide whether a frame has been reduced.

pypeit/scienceimage.py
jhennawi: no, removed. it was briefly used here.

pypeit/scripts/show_2dspec.py
jhennawi: Note the script has an option to embed after running in case the user wants to inspect some property of the images. IPython is thus a dependency.

pypeit/spectrographs/spectrograph.py
jhennawi: Docs added. Take a look at what I wrote there. @kbwestfall and I thought through this quite carefully. The preferred solution is to totally nuke processimages and start from scratch since nearly all of the code it wraps is incoherent garbage. The current stopgap is this fix, which was done thoughtfully to not interfere with sec2slice does, which works in the fits convention.

jhennawi:
Dev suite passed.

pypeit/spectrographs/spectrograph.py
kbwestfall: I don't really agree with the sentiment of Joe's third sentence, but let me explain the thinking on binning.  The issue is primarily that we want to be able to have the data/overscan sections defined either directly in `spectrograph.detector` or read from the fits headers.  To keep things consistent, it's important then to have, e.g., `spectrograph.detector[0]['datasec']` be ordered as it would be read via a FITS header.  Depending on the setup of the given spectrograph, `spectrograph.get_image_section` then provides the desired section and, based on the spectral axis defined the instrument, whether or not that section should be transposed to match the way that `spectrograph.load_raw_frame` orients the data array.  An alternative approach to what we've already done is to force `get_image_section` to always return (spec,spat).  This isn't something Joe and I talked about yesterday, but it may mean we don't need to use/return `transpose`.  The only thing I insisted on was that `parse.sec2slice` should never transpose the section with respect to the provided binning or vice versa.  The order of the section should be identical to the order of binning when passed to that function.

kbwestfall:
Are you running the dev suite with Kast fluxing?  The final call if I run `./pypeit_test kast` fails for me because it can't find the correct spec1d file.

On a slightly different note, the `flux_spec` script should not know whether or not it's being run by a unit test.  The unit test should appropriately call `flux_spec.main` to run the script as if a user ran it.

pypeit/scripts/show_2dspec.py
profxj: Ok.  Add to the install notes and 
.travis and requirements.txt.

pypeit/spectrographs/spectrograph.py
profxj: Am a bit confused, but I'm convinced you two came to a good solution.

jhennawi:
I fixed the fluxing issue @kbwestfall. Update the the dev-suite.

pypeit/scripts/show_2dspec.py
jhennawi: Sounds good, can you instruct me on how to do that?

jhennawi:
I doubt you've ever seen this much green!

screen shot 2019-02-26 at 18 01 50

pypeit/scripts/show_2dspec.py
profxj: Add a link to iPython in the `installing.rst` file
Like the others.

Add a line in the `requirements.txt` file. Like the others
I.e.  copy+paste

profxj:
I've seen that kind of green.  Am looking for the 
final line showing all passed.  Thanks

jhennawi:
All didn't pass because your buggy flexure code fails;) Also there was a bug in fluxing because of the new filenames which I've sensed fixed. It will take another day to run the dev suite fully through to get you that status message. Perhaps we should reduce the number of DEIMOS setups we are testing. 

Anyway, I'm fine waiting but anyone running LRIS on binned data is going to get garbage results until then.

jhennawi:
Oh and by the way, the reason why this bug even exists and needed to fixed was that someone didn't bother to address the VIS_1x2 failure in the dev suite in the past;)

jhennawi:
screen shot 2019-02-27 at 11 41 14
screen shot 2019-02-27 at 11 51 09

PR #638, opened 2019-03-04
MasterFrame and other IO issues

kbwestfall:
Primary changes are a significant refactoring of MasterFrame regarding its I/O routines.

I've also started working through a number of path definition issues.  Paths are now defined by the `PypeIt` class and should be kept fixed through a run of pypeit.  I don't think I've fixed all the path names, but I'm working in that direction, and given that the scripts will need access to these definitions, we probably want to pull this up to a higher level.

List of changes:

- Use matplotlib 'agg' backend to the top-level `__init__.py` to allow
  for running the code under a screen; may need a better approach.
- Numerous doc and style changes
- Add `master_type` to `MasterFrame` (and derived classes), which is
  used to set the name of the master frame output file.
- Significant edits to `MasterFrame` to streamline IO for derived
  classes.  Lead to significant changes to `Calibrations`.
- Main paths now set in `PypeIt`.
- Allow `connect_to_ginga` to start up the ginga viewer.
- Add a pytest `skipif` that checks if the Cooked directory exists in
  the dev-suite.  Use this to run the tests that only need the raw image
  data or don't need the dev-suite at all.
- Move wavelength calibration save/load out of `pypeit.wavecalib` into
  `pypeit.core.wavecal.waveio.py`
- Rename default directory for calibration masters to `Masters` and
  removed inclusion of spectrograph name.

Testing is ongoing, but please start reviewing the code.


pypeit/pypeit.py
```diff
@@ -295,33 +330,48 @@ def reduce_exposure(self, frames, bg_frames=[], std_outfile=None):
...
+            dict: The dictionary containing the primary outputs of
+            extraction.
+
         """
+
+        # TODO:
+        # - bg_frames should be None by default
+        # - change doc string to reflect that more than one frame can be
+        #   provided
+
```
jhennawi: the codes require that rather than None it be an empty list. we can debate this, but trying to make it None will break a bunch of stuff.

pypeit/pypeit.py
```diff
@@ -295,33 +330,48 @@ def reduce_exposure(self, frames, bg_frames=[], std_outfile=None):
...
         # if show is set, clear the ginga channels at the start of each new sci_ID
         if self.show:
+            # TODO: Put this in a try/except block?
             ginga.clear_all()
 
-        # Save the frame
+        # Save the frames
+        # TODO: Should these be set to self?  They're not used elsewhere
+        # in the class or in run pypeit.
         self.frames = frames
```
jhennawi: probably should not be set to self.

pypeit/pypeit.py
```diff
@@ -295,33 +330,48 @@ def reduce_exposure(self, frames, bg_frames=[], std_outfile=None):
...
         self.bg_frames = bg_frames
         # Is this an IR reduction?
+        # TODO: Why specific to IR?
```
jhennawi: ir_redux is synonomous with using a frame to bg subtract. We could name this something less specific though.

pypeit/pypeit.py
```diff
@@ -295,33 +330,48 @@ def reduce_exposure(self, frames, bg_frames=[], std_outfile=None):
...
-        # JFH Why does this need to be ordered?
+        # TODO: JFH Why does this need to be ordered?
         sci_dict = OrderedDict()  # This needs to be ordered
         sci_dict['meta'] = {}
         sci_dict['meta']['vel_corr'] = 0.
         sci_dict['meta']['ir_redux'] = self.ir_redux
 
         # Print status message
         msgs_string = 'Reducing target {:s}'.format(self.fitstbl['target'][self.frames[0]]) + msgs.newline()
+        # TODO: Print these when the frames are actually combined,
```
jhennawi: Yes I agree that would probably make for cleaner code.

pypeit/scripts/chk_tilts.py
```diff
@@ -36,6 +36,7 @@ def main(pargs):
 
     import subprocess
 
+    # TODO: THIS IS OUT OF DATE
```
jhennawi: I agree.

pypeit/scripts/flux_spec.py
```diff
@@ -103,6 +103,7 @@ def main(args, unit_test=False):
     par = pypeitpar.PypeItPar.from_cfg_lines(cfg_lines=spectrograph_def_par.to_config(),
                                              merge_with=config_lines)
 
+    # TODO: Remove this.  Put this in the unit test itself.
```
jhennawi: I agree never liked this here.

pypeit/traceslits.py
```diff
@@ -152,60 +151,60 @@ def __init__(self, mstrace, spectrograph, par, binning=None,
...
-        # dict
-        slf.tc_dict = ts_dict['tc_dict']
-
-        # Load the pixel objects?
-        if load_pix_obj:
-            slf._make_pixel_arrays()
-
-        # Return
-        return slf
+#    # TODO I think this routine should be deprecated.
```
jhennawi: It is not used anymore. Loading the entire class from a master file is not a good idea. Anyway the master format changed so it won't work anymore.

pypeit/traceslits.py
```diff
@@ -1017,141 +1059,185 @@ def __repr__(self):
         txt += '>'
         return txt
 
-
-    def save_master(self, tslits_dict, mstrace=None, outfile=None, overwrite=True):
+    def save(self, outfile=None, overwrite=True, traceImage=None):
```
jhennawi: So I previously changed the interface to these master frames to take an argument. For save this is the master to be written, for load it is the file to read from. There are important  reasons for that. 

At times I need to be able to write master files to disk using the save_master method without running the entire class. This is the case when I create pseudo masters as are the case for the 2d-coadds.  You could fix that by for example adding an optional argument where the master is specified instead of using the attributed. Otherwise I have to instantiate the class and set class attributes to the thing I want to write which is pretty tedious. Another solution would be for this save method to just wrap some core method that takes arguments.  

Here is an example from the 2d coadd codes: 


    # Write out the psuedo master files to disk
    master_key_dict = stack_dict['master_key_dict']
    waveImage = waveimage.WaveImage(None, None, None, None, None,
                                      master_key=master_key_dict['arc'], master_dir=master_dir)
    waveImage.save_master(waveimg_psuedo)
    traceSlits = traceslits.TraceSlits(None, None, None, master_key=master_key_dict['trace'], master_dir=master_dir)
    traceSlits.save_master(tslits_dict_psuedo)

I think you can give me this functionality back (which you've removed) by just adding optional arguments. 

The scripts often need to load these masters and that is why those utility functions exist. It looks like you changed those into static methods which I agree is a better approach. 

Anyway I think these are easy fixes. Ideal you would have caught this conflict with 2d coadds via a test, but the 2d coadds are not fully implemented yet.

pypeit/waveimage.py
```diff
@@ -134,9 +147,43 @@ def __repr__(self):
         txt = '<{:s}: >'.format(self.__class__.__name__)
         return txt
```
jhennawi: Need a way to pass an optional argument here for pseudo masters. See my comment before.

pypeit/waveimage.py
```diff
@@ -134,9 +147,43 @@ def __repr__(self):
...
+        Returns:
+            tuple: Returns an `numpy.ndarray`_ with the wavelength
+            image.  Also returns the primary header, if requested.
+        """
+        return super(WaveImage, self).load('WAVE', ifile=ifile, return_header=return_header)
 
+    
 
 
+# TODO: Provide a staticmethod
```
jhennawi: I agree a static method is a more elegant approach.

pypeit/wavetilts.py
```diff
@@ -294,108 +330,153 @@ def run(self, maskslits=None, doqa=True, debug=False, show=False):
...
 
-        self.tilts_dict = {'tilts':self.final_tilts, 'coeffs':self.coeffs, 'slitcen': self.slitcen, 'func2d':self.par['func2d'],
-                           'nslit': self.nslits, 'spat_order': self.spat_order, 'spec_order': self.spec_order}
+        self.tilts_dict = {'tilts':self.final_tilts, 'coeffs':self.coeffs, 'slitcen':self.slitcen,
+                           'func2d':self.par['func2d'], 'nslit':self.nslits,
+                           'spat_order':self.spat_order, 'spec_order':self.spec_order}
         return self.tilts_dict, maskslits
 
-    def load_master(self, filename, exten=0, force=False):
+    def save(self, outfile=None, overwrite=True):
```
jhennawi: optional arguments required if I want to save something not created by the class, i.e. tilts_dict = None or something.

pypeit/core/flat.py
```diff
@@ -1045,7 +1045,7 @@ def slit_profile_pca(mstrace, tilts, msblaze, extrap_slit, slit_profiles,
     return slit_profiles, mstracenrm, extrap_blz
 
 
-
+# TODO: Is this ever called?
```
profxj: probably not.  Feel free to deprecate.

pypeit/ginga.py
```diff
@@ -45,10 +53,17 @@ def connect_to_ginga(host='localhost', port=9000, raise_err=False):
     try:
         tmp = ginga.get_current_workspace()
     except:
+        if allow_new:
```
profxj: good one.

pypeit/masterframe.py
```diff
@@ -1,240 +1,221 @@
...
-    master_dir = os.path.join(redux_path, tmppar['caldir'])
-    # Spectrograph
-    if spectrograph is not None:
-        master_dir += '_'+spectrograph.spectrograph
-    # Return
-    return master_dir
-
+        raw_files = {}
+        prefix = 'F'
+        for k, v in hdr.items():
```
profxj: Having a hard time parsing this.  Please add a comment on what
is being done.

pypeit/par/pypeitpar.py
```diff
@@ -1800,7 +1800,7 @@ def __init__(self, caldir=None, setup=None, trim=None, badpix=None, biasframe=No
 
         # Fill out parameter specifications.  Only the values that are
         # *not* None (i.e., the ones that are defined) need to be set
-        defaults['caldir'] = 'MF'
+        defaults['caldir'] = 'Masters'
```
profxj: I suspect this will break our make `Cooked` script, which is fine.
Just a head's up.

pypeit/processimages.py
```diff
@@ -808,6 +804,7 @@ def show(self, attr='stack', idx=None, display='ginga'):
         # Show
         viewer, ch = ginga.show_image(img)
 
+    # TODO: Is this ever used?
```
profxj: probably by @jhennawi 's stacking code.

pypeit/pypeit.py
```diff
@@ -112,26 +121,28 @@ def __init__(self, pypeit_file, verbosity=2, overwrite=True, reuse_masters=False
...
-        paths.make_dirs(self.spectrograph.spectrograph, self.par['calibrations']['caldir'],
-                        self.par['rdx']['scidir'], self.par['rdx']['qadir'],
-                        overwrite=self.overwrite, redux_path=self.par['rdx']['redux_path'])
+
+        # Report paths
+        msgs.info('Setting reduction path to {0}'.format(self.par['rdx']['redux_path']))
+        msgs.info('Master calibration data output to: {0}'.format(self.calibrations_path))
+        msgs.info('Science data output to: {0}'.format(self.science_path))
+        msgs.info('Quality assessment plots output to: {0}'.format(self.qa_path))
+        # TODO: Is anything written to the qa dir or only to qa/PNGs?
```
profxj: yes, an HTML file is written in QA/

pypeit/scripts/arcid_plot.py
```diff
@@ -44,6 +43,8 @@ def main(args):
 
     from linetools.utils import loadjson
 
+    # TODO: This must be an out-dated script that is never used.
```
profxj: yup

pypeit/scripts/flux_spec.py
profxj: I couldn't figure out how to do so.

pypeit/scripts/run_pypeit.py
```diff
@@ -60,7 +61,9 @@ def main(args):
     import sys
     import traceback
 
-    from pypeit import check_requirements
+    # This is done in the package level __init__; don't need to do it
```
profxj: agreed.  remove.

pypeit/tests/test_metadata.py
```diff
@@ -1,15 +1,57 @@
...
 from pypeit.par.util import parse_pypeit_file
 from pypeit.pypeitsetup import PypeItSetup
-from pypeit.tests.tstutils import dev_suite_required
+from pypeit.tests.tstutils import dev_suite_required, data_path
 from pypeit.metadata import PypeItMetaData
 from pypeit.spectrographs.util import load_spectrograph
-
+from pypeit.scripts import setup
+
+def test_read_combid():
```
profxj: Nice!

pypeit/traceslits.py
```diff
@@ -1,57 +1,59 @@
...
-import os
-from subprocess import Popen
 
 from scipy import ndimage
 
-from importlib import reload
-
 from astropy.io import fits
 
+# TODO: I'd rather we not use 'as' ...
```
profxj: ok, but how does one deal with multiple instances of `utils`, `io`, etc.?

pypeit/core/flat.py
```diff
@@ -1045,7 +1045,7 @@ def slit_profile_pca(mstrace, tilts, msblaze, extrap_slit, slit_profiles,
     return slit_profiles, mstracenrm, extrap_blz
```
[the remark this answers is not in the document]
kbwestfall: Okay, commenting it out for now.

pypeit/core/wavecal/kdtree_generator.py
```diff
@@ -9,6 +9,9 @@
 it's only purpose is to generate a KD Tree with the desired patterns.
 """
 
+# TODO: Why not use scipy?: 
+# https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.html
+
```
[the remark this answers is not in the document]
kbwestfall: Thanks for pointing this out!  I've added a note to the file, but we might deprecate the file at some point.

pypeit/pypeit.py
```diff
@@ -144,13 +155,46 @@ def __init__(self, pypeit_file, verbosity=2, overwrite=True, reuse_masters=False
         self.sciI = None
         self.obstime = None
```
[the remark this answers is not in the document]
kbwestfall: Thanks!

pypeit/pypeit.py
kbwestfall: I dislike empty lists as a default.  Leaving it for now and leaving the comment.

pypeit/pypeit.py
kbwestfall: Okay, not longer saving `frames` or `bg_frames` to `self`.

pypeit/pypeit.py
kbwestfall: That's what I thought.  Leaving it for now, but we should rename it at some point.

pypeit/pypeit.py
kbwestfall: Sounds good.  Leaving this as a TODO for now.

pypeit/scripts/chk_tilts.py
kbwestfall: Okay, I've added a note at the top of the script (which propagated to readthedocs), but I'm leaving the script in case someone eventually updates it.

pypeit/scripts/flux_spec.py
kbwestfall: I'm going to leave this as a comment and I can try to deal with it in another PR.

pypeit/traceslits.py
kbwestfall: I think this might have actually been your original comment and I just commented out the function.

pypeit/waveimage.py
kbwestfall: I had missed this one.  Added now.

pypeit/__init__.py
```diff
@@ -37,6 +37,14 @@ def signal_handler(signalnum, handler):
...
 # Ignore all warnings given by python
+# TODO: I'd rather we not do this.  Is there a way we can redirect
+# warnings to pypeit.msgs ?
 warnings.resetwarnings()
 warnings.simplefilter('ignore')
 
+# TODO: Need some way of selectively doing this.  Once you import
+# pypeit, this affects the behavior of pyplot for *anything* else you
+# plot in a given session.
+from matplotlib import pyplot
```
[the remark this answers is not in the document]
kbwestfall: Yeah, and it meant I couldn't plot things within pdb either.  So I'm going to comment this out in the push to develop and only use it when I'm running the dev suite.  I'm not sure what else to do except to add arguments to all the QA functions that allow you to set the backend to `agg`, which is not ideal.

pypeit/arcimage.py
```diff
@@ -56,23 +63,56 @@ def __init__(self, spectrograph, files=None, det=1, par=None, master_key=None,
...
+
+        Args:
+            ifile (:obj:`str`, optional):
+                Name of the master frame file.  Defaults to
+                :attr:`file_path`.
+            return_header (:obj:`bool`, optional):
+                Return the header
+
+        Returns:
+            tuple: Returns three `numpy.ndarray`_ objects with the raw
```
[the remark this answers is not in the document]
kbwestfall: Good catch!  Had just copied over the doc.  Fixed.

pypeit/calibrations.py
```diff
@@ -80,36 +83,41 @@ class Calibrations(object):
...
     """
     __metaclass__ = ABCMeta
 
-    def __init__(self, fitstbl, par, spectrograph, redux_path=None, reuse_masters=False, save_masters=True,
-                 write_qa=True, show=False):
+    def __init__(self, fitstbl, par, spectrograph, caldir=None, qadir=None, reuse_masters=False,
+                 show=False):
 
-        # Check the type of the provided fits table
+        # Check the types
```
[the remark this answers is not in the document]
kbwestfall: I prefer doing this too, but "pythonic" convention is usually that it's better to ask for forgiveness than permission.  E.g.:

https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html

But that leads to try-except blocks everywhere, which I find distracting.

pypeit/calibrations.py
```diff
@@ -215,33 +221,35 @@ def get_arc(self):
...
-                                          par=self.par['arcframe'], master_key=self.arc_master_key,
-                                          master_dir=self.master_dir, reuse_masters=self.reuse_masters)
+                                          par=self.par['arcframe'],
+                                          master_key=self.master_key_dict['arc'],
+                                          master_dir=self.master_dir,
+                                          reuse_masters=self.reuse_masters)
 
         # Load the MasterFrame (if it exists and is desired)?
-        self.msarc, _ = self.arcImage.master(prev_build=prev_build)
+        self.msarc = self.arcImage.load()
```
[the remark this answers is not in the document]
kbwestfall: Yeah, but I'll double check the derived classes that overwrite the base class method to make sure this is still adhered to.

pypeit/calibrations.py
```diff
@@ -317,122 +321,130 @@ def get_bpm(self):
...
+        # allowing for arcs of different shape becuase of X-shooter etc.
         self.shape = procimg.trim_frame(dsec_img, dsec_img < 1).shape
 
         # Build it
-        self.msbpm = self.spectrograph.bpm(shape=self.shape, filename=sci_image_files[0], det=self.det)
+        self.msbpm = self.spectrograph.bpm(shape=self.shape, filename=sci_image_file, det=self.det)
+
         # Record it
-        self.calib_dict[self.bpm_master_key]['bpm'] = self.msbpm
+        self.calib_dict[self.master_key_dict['bpm']]['bpm'] = self.msbpm
```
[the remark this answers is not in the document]
kbwestfall: Wasn't as easy as that, but I've added a new `_update_cache` method that does this.

pypeit/core/flat.py
kbwestfall: Commented out for now.

pypeit/ginga.py
kbwestfall: Thanks!  Was worried there was some reason not to do this.  Should we make the `allow_new=True` by default?

pypeit/masterframe.py
kbwestfall: Done.  Let me know if it's still unclear.

pypeit/par/pypeitpar.py
kbwestfall: Yeah, I need to remake Cooked and run the cooked tests to make sure that I haven't missed somewhere `MF` is hard-coded.

pypeit/processimages.py
kbwestfall: I don't see it anywhere.  Will comment it for now and see if that breaks anything.

pypeit/pypeit.py
kbwestfall: Roger that.

pypeit/scripts/arcid_plot.py
kbwestfall: Okay, made a note that gets propagated to the docs, but leaving the script in case someone wants to update it.

pypeit/scripts/flux_spec.py
kbwestfall: Okay.  I'm punting on this one for now.

pypeit/scripts/run_pypeit.py
kbwestfall: Done.

pypeit/traceslits.py
kbwestfall: It's actually not used in this file, so I'm just going to remove it.

I would generally just do: `import linetools.utils` in cases where there are multiple `utils` modules to load.  But I realize this makes the subsequent calls more onerous.

pypeit/traceslits.py
kbwestfall: Here's the fix I've implemented:
```
waveImage = waveimage.WaveImage(None, None, None, None, None,
                                master_key=master_key_dict['arc'], master_dir=master_dir)
waveImage.save(mswave=waveimg_psuedo)

traceSlits = traceslits.TraceSlits(None, None, master_key=master_key_dict['trace'],
                                   master_dir=master_dir)
traceSlits.save(tslits_dict=tslits_dict_psuedo)
```

This should only be a stop-gap until we come up with a better solution.

pypeit/waveimage.py
kbwestfall: Addressed.

pypeit/wavetilts.py
kbwestfall: Only done for `WaveImage` and `TraceSlits` as a temporary stop-gap.  Not implemented generally.

pypeit/tests/test_metadata.py
kbwestfall: Thanks.  Ran into problems reading one of the pypeit files in the dev suite because it had these defined.  So I added a test to make sure I caught this failure mode.

pypeit/traceslits.py
jhennawi: Thanks @kbwestfall, this looks great. I'm happy to discuss better solutions, but I'm afraid these pseudo masters, although not very elegant, are something that we have to live with.

pypeit/wavetilts.py
jhennawi: Okay that is fine.

jhennawi:
Looks great @kbwestfall, I will approve when screenshots of dev suite and tests are posted.

pypeit/wavetilts.py
kbwestfall: Okay, good.  Thanks.

pypeit/traceslits.py
```diff
@@ -1017,141 +1056,187 @@ def __repr__(self):
...
+        msgs.info('Saving master frame to {0}'.format(_outfile))
+
+        # Build the header
+        hdr = fits.Header()
+        #   - Set the master frame type
+        hdr['FRAMETYP'] = (self.master_type, 'PypeIt: Master calibration frame type')
+        #   - List the completed steps
+        hdr['STEPS'] = (','.join(self.steps), 'Completed reduction steps')
+        #   - Provide the file names
+        if traceImage is not None:
```
jhennawi: Please remove these dependencies on the traceImage class. As per our discussion these save routines either: 

1) Need to wrap core routines that save the master file, which take simple arguments
 
OR 

2) Stay in the class, but then allow the things that are being saved to be passed in as optional arguments. Here those things are simply mstrace and tslits_dict. I do not want to see save methods that expect the class as an argument. This is anyway redundant since you are in the class. 

The easiest way out here is to simply make mstrace the optional argument, and if the trace image.files is None, don't bother to write that stuff about the files to the header. That stuff makes no sense for pseudo masters anyway.

kbwestfall:
Dev suite is passing.  Still waiting on DEIMOS, but here's the rest of them.
[dev_suite_status.pdf](https://github.com/pypeit/PypeIt/files/2957214/dev_suite_status.pdf).

Once DEIMOS is done, I'll post that result (ZD-7500 will likely fail), remake cooked, fix any tests that fail because of the path changes, and then post the clean bill.

kbwestfall:
test_deimos

DEIMOS finally finished.  The failure is in 600ZD_M_7500.

I broke some of the tests, so I'll fix those, compile cooked, run the cooked tests and post the result hopefully this morning or early afternoon.

jhennawi:
Hi @kbwestfall, 

Nice work! Approve, please just get the tests working and the new version of Cooked.

kbwestfall:
mfio_tests

Tests pass and cooked uploaded to the team drive as Cooked_pypeit_dev_mfio.tar.gz.

I'm working on a test that will check both run_pypeit and reuse_masters.

kbwestfall:
Okay, added a test in `test_scripts.py` that tests a few ways of running `run_pypeit` for Kast blue.  All tests pass, so I think this merge is finally ready to go.  @jhennawi merge when you're satisfied, or let me know if you run into other problems.

PR #644, opened 2019-03-14
Hotfix for saving and reusing masters in memory

kbwestfall:
The `calib_dict` object wasn't being setup correctly in `Calibrations`.


kbwestfall:
Tests pass and Dev Suite Kast passes.

PR #656, opened 2019-03-22
Minor code tweaks from our March run

jhennawi:
This branch has very minor code tweaks made to improve functionality of the develop branch during our last observing run. The most significant change is a minor change to the background frame matching logic in pypeit.py to allow for frames to serve as a background frame for multiple science frames. I also made some effort to prevent the pypeit_setup from crashing if files are empty. Instead these files are just ignored, although I still think this needs work to work in all cases. 

dev suite is running now overnight. Will post test results and generate cooked tomorrow.


pypeit/core/coadd.py
```diff
@@ -1294,67 +1294,234 @@ def order_phot_scale(spectra, phot_scale_dicts, nsig=3.0, niter=5, debug=False):
 
     return collate(spectra_list_new)
 
-def order_median_scale(spectra, nsig=3.0, niter=5, overlapfrac=0.03, num_min_pixels=50, SN_MIN_MEDSCALE=5.0, debug=False):
+def order_median_scale(wave, wave_mask, fluxes_in, ivar_in, sigrej=3.0, niter=5, min_overlap_pix=21, min_overlap_frac=0.03,
```
kbwestfall: Let's make lines of code no longer than 100 characters, and doc strings no wider than 72 characters.

pypeit/core/coadd.py
```diff
@@ -1294,67 +1294,234 @@ def order_phot_scale(spectra, phot_scale_dicts, nsig=3.0, niter=5, debug=False):
...
+                       max_rescale_percent=50.0, sn_min=1.0, debug=False):
     '''
     Scale different orders using the median of overlap regions. It starts from the reddest order, i.e. scale H to K,
       and then scale J to H+K, etc.
     Parameters:
```
kbwestfall: Args:

pypeit/core/coadd.py
```diff
@@ -1294,67 +1294,234 @@ def order_phot_scale(spectra, phot_scale_dicts, nsig=3.0, niter=5, debug=False):
...
-      spectra: XSpectrum1D spectra
-      nsig: float
-        sigma used for sigma_clipping median
-      niter: int
+
+    wave: float ndarray (nspec,)
```
kbwestfall: Indent and properly format.
```
    r"""
    Args:
        wave (numpy.ndarray):
            Floating point array with common wavelength grid for all the
            orders.  Shape is :math:`(N_{\rm spec},)`.
    ...
    """
```

pypeit/core/coadd.py
```diff
@@ -1294,67 +1294,234 @@ def order_phot_scale(spectra, phot_scale_dicts, nsig=3.0, niter=5, debug=False):
...
-                allok_iord = sigs[iord, :] > 0
-                allok_iordm1 = sigs[iord-1, :] > 0
                 plt.figure(figsize=(12, 6))
-                plt.plot(wave[allok], spectra.data['flux'][iord, allok], '-',lw=10,color='0.7', label='Scale region')
-                plt.plot(wave[allok_iord], spectra.data['flux'][iord,allok_iord], 'r-', label='reference spectrum')
-                plt.plot(wave[allok_iordm1], fluxes_raw[iord - 1,allok_iordm1], 'k-', label='raw spectrum')
-                plt.plot(spectra.data['wave'][iord - 1, allok_iordm1], spectra.data['flux'][iord - 1, allok_iordm1], 'b-',
-                         label='scaled spectrum')
-                mny, medy, stdy = stats.sigma_clipped_stats(fluxes[iord, allok], sigma=nsig, iters=niter)
+                plt.plot(wave[allok_iord_ref], fluxes_out[iord_ref, allok_iord_ref], '-',lw=10,color='0.7', label='Scale region')
```
kbwestfall: comma separation

pypeit/core/coadd.py
```diff
@@ -1294,67 +1294,234 @@ def order_phot_scale(spectra, phot_scale_dicts, nsig=3.0, niter=5, debug=False):
...
+                      'Consider decreasing min_overlap_frac = {:5.3f}'.format(min_overlap_frac))
+            fluxes_out = fluxes_in[iord-1,:]
+            sigs_out = sigs_in[iord-1,:]
+
+
+
+def merge_order(spectra, wave_grid, extract='OPT', orderscale='median', niter=5, sigrej_final=3., SN_MIN_MEDSCALE = 5.0,
+                overlapfrac = 0.01, num_min_pixels=10,phot_scale_dicts=None, qafile=None, outfile=None, debug=False):
+    """
+        routines for merging orders of echelle spectra.
```
kbwestfall: First line of the docstring is aligned with the quotes:
```
    """
    Combine spectra from multiple echelle orders.

    [[Here's a longer description of the algorithm.]]]

    Args:
        spectra (XSpectrum1D):
            Spectra to combine.
        ...
    """
```

pypeit/core/coadd.py
```diff
@@ -1294,67 +1294,234 @@ def order_phot_scale(spectra, phot_scale_dicts, nsig=3.0, niter=5, debug=False):
...
+        wave_grid (numpy array): The final wavelength grid for the output. Note that the wave_grid should be
+          consistent with your spectra, if not please rebin it first.
+        extract (str): 'OPT' or 'BOX'
+        orderscale (str): which method you want use to scale different orders: median, photometry or None.
+        niter (int): number of iteration for rejections in the overlapped part
+        sigrej_final (float): sigma for rejections in the overlapped part
+        SN_MIN_MEDSCALE (float): minimum SNR for scaling different orders
+        overlapfrac (float): minimum overlap fraction for scaling different orders.
+        phot_scale_dicts (dict): A dict contains photometric information of each order. An example is given below.
+          phot_scale_dicts = {'0': {'filter': None, 'mag': None, 'mag_type': None, 'masks': None},
```
kbwestfall: Add two colons so that sphinx will interpret this as a code block:
```
        phot_scale_dicts (dict):
            Photometric information for each order.  Format is::

                phot_scale_dicts = {'0': {'filter': None, 'mag': None, 'mag_type': None, 'masks': None},
                ...
```

pypeit/core/coadd.py
```diff
@@ -1294,67 +1294,234 @@ def order_phot_scale(spectra, phot_scale_dicts, nsig=3.0, niter=5, debug=False):
...
+
+    ## Scaling different orders
+    if orderscale == 'photometry':
+        # Only tested on NIRES.
+        if phot_scale_dicts is not None:
+            spectra = order_phot_scale(spectra, phot_scale_dicts, debug=debug)
+        else:
+            msgs.warn('No photometric information is provided. Will use median scale.')
+            orderscale = 'median'
+    elif orderscale == 'median':
```
kbwestfall: If `orderscale` gets changed from photometry to median, this `elif` won't catch it.  You can change it to `if`, and then replace the `else` with `if orderscale not in ['photometry', 'median']: `

pypeit/core/coadd.py
```diff
@@ -1390,179 +1557,135 @@ def ech_coadd(files,objids=None,extract='OPT',flux=True,giantcoadd=False,ordersc
...
-                if len(qafile.split('.')) == 1:
-                    msgs.info("No fomat given for the qafile, save to PDF format.")
-                    qafile_iord = qafile + '.pdf'
+            # Coadding
+            spec1d_final = coadd_spectra(spectra, wave_grid_method=wave_grid_method, niter=niter,
+                                   scale_method=scale_method, do_offset=do_offset, sigrej_final=sigrej_final,
+                                   do_var_corr=do_var_corr, qafile=qafile, outfile=outfile,
+                                   do_cr=do_cr, debug=debug, **kwargs)
+            return spec1d_final
+        else:
```
kbwestfall: There's no need for this `else`.  The function will return if it entered the `if` block.

pypeit/core/coadd.py
```diff
@@ -1390,179 +1557,135 @@ def ech_coadd(files,objids=None,extract='OPT',flux=True,giantcoadd=False,ordersc
...
-            #spec1d_final = coadd_spectra(spectra_coadd_rebin, wave_grid_method=wave_grid_method, niter=niter,
-            #                                  scale_method=scale_method, do_offset=do_offset, sigrej_final=sigrej_final,
-            #                                  do_var_corr=do_var_corr, qafile=qafile, outfile=outfile,
-            #                                  do_cr=do_cr, debug=debug, **kwargs)
+            spec1d_final = merge_order(spectra_coadd_rebin, wave_grid, extract=extract, orderscale=orderscale,
+                                       niter=niter, sigrej_final=sigrej_final, SN_MIN_MEDSCALE=SN_MIN_MEDSCALE,
+                                       overlapfrac=overlapfrac, num_min_pixels=num_min_pixels,
+                                       phot_scale_dicts=phot_scale_dicts, qafile=qafile, outfile=outfile, debug=debug)
             return spec1d_final
         else:
```
kbwestfall: Same comment about the uneccessary `else`.

pypeit/core/coadd.py
```diff
@@ -1390,179 +1557,135 @@ def ech_coadd(files,objids=None,extract='OPT',flux=True,giantcoadd=False,ordersc
...
-            ### Do NOT remove this part althoug it is deprecated.
-            # we may need back to using this pieces of code after fixing the coadd_spectra problem on first order.
-            #kwargs['echelle'] = True
-            #kwargs['wave_grid_min'] = np.min(wave_grid)
-            #kwargs['wave_grid_max'] = np.max(wave_grid)
-            #spec1d_final = coadd_spectra(spectra_coadd_rebin, wave_grid_method=wave_grid_method, niter=niter,
-            #                                  scale_method=scale_method, do_offset=do_offset, sigrej_final=sigrej_final,
-            #                                  do_var_corr=do_var_corr, qafile=qafile, outfile=outfile,
-            #                                  do_cr=do_cr, debug=debug, **kwargs)
+            spec1d_final = merge_order(spectra_coadd_rebin, wave_grid, extract=extract, orderscale=orderscale,
```
kbwestfall: If you're just going to return the object, there's no need to assign it to something.  I.e., just `return merge_order(...)`.

pypeit/core/coadd.py
```diff
@@ -1390,179 +1557,135 @@ def ech_coadd(files,objids=None,extract='OPT',flux=True,giantcoadd=False,ordersc
...
             if outfile is not None:
                 for iord in range(len(spectra_list)):
-                    outfile_iord = outfile.replace('.fits','_ORDER{:04d}.fits'.format(iord))
-                    msgs.info('Saving the final calibrated spectrum of order {:d} as {:s}'.format(iord,outfile))
+                    outfile_iord = outfile.replace('.fits', '_ORDER{:04d}.fits'.format(iord))
+                    msgs.info('Saving the final calibrated spectrum of order {:d} as {:s}'.format(iord, outfile))
                     spectra_list[iord].write_to_fits(outfile_iord)
             return spectra_list
 
+    elif nfile==1:
```
kbwestfall: I would separate the cases of `nfiles == 1` and `nfiles > 1` into separate functions.

pypeit/core/coadd.py
```diff
@@ -1390,179 +1557,135 @@ def ech_coadd(files,objids=None,extract='OPT',flux=True,giantcoadd=False,ordersc
...
+        msgs.info('Only find one spectrum. Thus only order merging will be performed.')
+        spectra = load.ech_load_spec(files, objid=objids, order=None, extract=extract, flux=flux)
+
+        # Get wave_grid
+        norder = spectra.nspec
+        flux, sig, wave = unpack_spec(spectra, all_wave=True)
+        dloglam = np.median(np.log10(wave[0,1:])-np.log10(wave[0,:-1]))
```
kbwestfall: `dloglam = np.median(np.diff(np.log10(wave[0])))`

pypeit/spectrographs/spectrograph.py
```diff
@@ -860,6 +861,36 @@ def slit2order(self, slit):
...
+    @property
+    def loglam_minmax(self):
+        pass
+
+    def wavegrid(self, binning=None, midpoint=False,samp_fact=1.0):
+        """
+        Routine to generate a fixed wavelength grid in log_10 lambda. Mostly used by echelle spectrographs
+
+        Args:
+            binning:
```
kbwestfall: Docs.

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -729,23 +718,13 @@ def slitmask(self, tslits_dict, pad=None):
     @property
     def dloglam(self):
```
kbwestfall: Should we make this a user-defined parameter in a `ParSet` so that this default can be overridden?

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -729,23 +718,13 @@ def slitmask(self, tslits_dict, pad=None):
...
-
-        # Define the grid for VLT-XSHOOTER NIR
-        logmin, logmax = self.loglam_minmax
-        loglam_grid = utils.wavegrid(logmin, logmax, self.dloglam)
-        if midpoint:
-            loglam_grid = loglam_grid + self.dloglam/2.0
-
-        return np.power(10.0,loglam_grid)
-
+        return np.log10(5000.0), np.log10(11000)
```
kbwestfall: Same comment here about allowing this to be a parameter.

pypeit/utils.py
```diff
@@ -28,16 +29,18 @@ def wavegrid(wave_min, wave_max, dwave, osamp=1.0):
            Maximum wavelength. Can be in linear or log.
         dwave: float
            Delta wavelength interval
-        osamp: float
-           Oversampling factor
+        samp_fact: float
```
kbwestfall: ```samp_fact (float, optional):``` or better yet ```samp_fact (:obj:`float`, optional):```

pypeit/metadata.py
```diff
@@ -321,11 +321,22 @@ def _build(self, files, strict=True, usrdata=None):
                 data[meta_key].append(value)
             msgs.info('Added metadata for {0}'.format(os.path.split(ifile)[1]))
 
-#        # Additional bits and pieces
-#        self._add_bkg_pairs(data, 'empty')
+        # JFH Changed the below to now crash if some files have None in their MJD. This is the desired behavior
```
kbwestfall: "not" or "now"?

kbwestfall: PS - I worry a bit about things like this because these warning messages could easily fly by a user unnoticed.  We should think about if we're logging warnings in a way that a user can easily parse.

pypeit/core/coadd.py
jhennawi: Docs are still under construction.

pypeit/spectrographs/vlt_xshooter.py
jhennawi: It is a fixed format spectrograph, so no. This number is basically a constant and only developers should be considering changing it.

pypeit/metadata.py
jhennawi: Sorry it was supposed to be not. Anyway I think the code should crash when something is seriously wrong. I don't think a corrupt file should crash setup. I think if there is a corrupt file, it should be written to the .sorted file with some kind of warning or info. 

The issue I had was the following. I was running PypeIt while observing. The instrument controller would create an empty file to hold the current exposure and then update it with the data at readout. This meant that when I copied the data over I always had one empty file. This crashed setup because of the line above. The only work around was to constantly delete this one empty file. People running PypeIt while observing is something we want to aspire for, so I think having setup constantly crash if there is one file that is either empty or corrupt is a bad idea.

pypeit/spectrographs/vlt_xshooter.py
jhennawi: Seem my point above. These are constants for the spectrograph that only developers should touch.

pypeit/core/coadd.py
jhennawi: I'm fine with that but I don't know of a way to count characters on a line. My development IDE PyCharm does not do this for me.

jhennawi:
Screen Shot 2019-03-22 at 20 26 46

jhennawi:
Dev suite passed!
Screen Shot 2019-03-23 at 00 24 44

pypeit/core/coadd.py
kbwestfall: It seems like there should be a way to set a limit according to this:

https://stackoverflow.com/questions/17319422/how-do-i-set-the-maximum-line-length-in-pycharm

at least for the 100 character length.  May not be possible to do this and the doc string limitation.

pypeit/spectrographs/vlt_xshooter.py
kbwestfall: Got it.

pypeit/metadata.py
kbwestfall: So this try except block doesn't really have anything to do with the success of the `time.Time` call, right?  Shouldn't this check get put further up in the function?  And should we remove the incomplete/corrupt files from the internal metadata table?

pypeit/core/coadd.py
kbwestfall: I was hoping we can address this before we merge.

pypeit/metadata.py
jhennawi: My suggestion would be to never crash in pypeit_setup. There is really no reason to crash there, and any crash was just bad coding. We are classifying files and assigning meta data. If pypeit_setup comes to something it cannot classify, then it should simply put that in a separate setup "Unknown" or something and leave the file in the .sorted list with a bunch of None values for the meta data.   This is enough warning for the user who can then figure out what went wrong. 

The try: except: block that I added was a stopgap to prevent the crash. I'm not sure why that line is there, it is just validating that the times are okay but not returning anything. It is exactly this "validation" which is in my opinion the flawed approach, because the code crashes when it fails to validate which is (in my opinion) not the desired behavior. 

Right now I'd like to leave this try:except in there to prevent crashes, but I'd be much happier if we made an Issue on not having pypeit_setup ever crash and instead just putting unclassifiable files in their own bin with None meta data.

pypeit/core/coadd.py
jhennawi: @feigewang can you comment. 

Anyway this stuff is under construction. Nobody is using it except developers.

feigewang: Changed according to Kyle's comments. Thanks!

PR #783, opened 2019-10-31
Hotfix for bug that was introduced by PR#775 related to the bpm. 

jhennawi:
This fixes the bug that was introduced into the bpm by PR #775 that causes all instruments except  WHT_ISIS to crash.

### DIFF
--- a/pypeit/calibrations.py
+++ b/pypeit/calibrations.py
@@ -414,7 +414,7 @@ class Calibrations(object):
         # Save & return
         self._update_cache('bias', 'bias', self.msbias)
 
-        # Commenting this out for now until it gets fixed. I'm not convinced this should be here. 
+        # Commenting this out for now until it gets fixed. I'm not convinced this should be here.
 
         # If we need to make a bad pixel mask using the bias frames, do it now
         #if self.par['badpix']:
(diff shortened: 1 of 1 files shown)


jhennawi:
Hi @profxj and @rcooke-ast, this PR is a stop-gap to get the develop branch working again. We were lax about merging PR #775 in without running tests. This broke the entire dev suite. I've simply commented out the offending lines of code in calibrations to get things working again. I'll leave it to @profxj and @rcooke-ast to discuss the path forward. A few of my thoughts are below.

We currently require a science frame to create the bpm. This has the disadvantage that one cannot simply run the calibrations script without a science frame. That said, I don't think pypeit will run without a science frame. For this reason, I don't see any obvious reason to add this bpm functionality based on bias frames at the present moment. While I'd like PypeIt to run on just lists of calibration files, the fitstable and various other things do not support this. You have to have a science image. 

@rcooke-ast I think this should work straightforwardly for WHT_ISIS as well. I am not sure why you need to do something different and make a bpm from a bias frame. This PR will break WHT_ISIS because your overloading of the bpm function is not consistent with its calling sequence. I suggest that you fix this, and make WHT_ISIS part of the dev suite as soon as possible to guarantee that it will continue to work as PypeIt is developed.

PR #801, opened 2019-12-28
Fixing a bug.

marijana777:
Previous version: After coadding spectra by using combspec in coadd1d.py, wavelengths were unexpectedly shifted.
New version: Wavelengths are not artificially shifted (as it should be).
Reason for the bug: Precision in weights_flat in compute_stack was not enough high.
Modification: Weights_flat set to float64 type (instead of float32).
Attached: An example of a coadded spectrum with shifted wavelengths (flux vs wavelength) before the changes in the code. In this example two identical spectra (in blue) were coadded. The coadded spectrum is shown in black. After the changes in the code, the blue and black line overlap, as expected.  
example

### DIFF
--- a/pypeit/core/coadd1d.py
+++ b/pypeit/core/coadd1d.py
@@ -1273,7 +1273,6 @@ def compute_stack(wave_grid, waves, fluxes, ivars, masks, weights):
     fluxes_flat = fluxes[ubermask].flatten()
     ivars_flat = ivars[ubermask].flatten()
     vars_flat = utils.inverse(ivars_flat)
-    #weights_flat = np.float64(weights[ubermask].flatten())
     weights_flat = weights[ubermask].flatten()
 
     # Counts how many pixels in each wavelength bin
@@ -2078,6 +2077,7 @@ def combspec(waves, fluxes, ivars, masks, sn_smooth_npix,
               spectrum on wave_stack wavelength grid. True=Good.
     '''
 
+    # We cast to float64 because of a bug in np.histogram
     waves = np.float64(waves)
     fluxes = np.float64(fluxes)
     ivars = np.float64(ivars)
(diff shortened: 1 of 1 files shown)


marijana777:
Hi @jhennawi . Thank you very much for your comments. 

To respond the first comment, my understanding is that there is a bug in Python as described in https://github.com/numpy/numpy/issues/10319 . I checked the calculated values by the code in multiple places. For the example I was using, np.histogram was supposed to return an array with all values equal 2., but instead for some elements in the array the code returned 2.00026. After changing to the 64 bit precision, the code returned all values equal 2., as expected. The part of the code is weights_total, wave_edges = np.histogram(waves_flat,bins=wave_grid,density=False,weights=weights_flat) in compute_stack. 

To respond the second comment, my input fluxes were float32. Function sn_weights defines weights = np.zeros_like(flux_stack), so in my example weights had the same type, or float32. 
The issue could be also solved if input fluxes (and maybe waves and ivars) are set to float64 in the beginning of the combspec function, as I did in the next commit (pushed ~ one minute ago).

pypeit/core/coadd1d.py
```diff
@@ -2077,6 +2078,9 @@ def combspec(waves, fluxes, ivars, masks, sn_smooth_npix,
               spectrum on wave_stack wavelength grid. True=Good.
     '''
 
+    waves = np.float64(waves)
```
jhennawi: Okay this looks good. Maybe add a comment here that we cast to float64 because of a bug in np.histogram. Then I will approve. Thanks for tracking this down!!

marijana777:
Thank you @jhennawi . Just added the comment.

jhennawi:
The reason this branch is failing tests is this line: 

188 of core/wave.py

subpix_grid = np.linspace(max_corr-3., max_corr+3., 7.)

which needs to be changed to: 

subpix_grid = np.linspace(max_corr-3., max_corr+3., 7)

I cannot make the change since you are forked @marijana777. Can you please make this change, push it, and then you should pass Travis so we can merge. 

Many Thanks!
Joe Hennawi

kbwestfall:
Sorry, @marijana777, I didn't realize I had closed this!  Certainly wasn't my intent.  Merging now.

PR #820, opened 2020-01-25
Hotfix HDU

rcooke-ast:
The function `spectrograph.get_rawimage()` previously returned an HDUlist, even though the rest of the arguments of this function were detector/HDU specific. Furthermore, the returned HDUlist is (from what I can tell) not used anywhere in the code. I therefore propose that a list of the headers should be returned, as this is useful (and less memory intensive) to pass around than a full HDU list.

Following on from this point, I think the original header info (or fitstbl) should be passed around in case something is needed from the header during the reduction, and a `processedImage` doesn't seem to carry this information from the raw files. This is needed for the KCWI reduction. We could use the HDUlist, but it's a question of memory... Any objections?

Running on the devsuite now to make sure I haven't broken anything... 🤞


rcooke-ast:
--- PYPEIT DEVELOPMENT SUITE PASSED 41/41 TESTS (Masters ignored) ---

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -282,7 +282,7 @@ def get_rawimage(self, raw_file, det):
         # Need the exposure time
         exptime = hdu[self.meta['exptime']['ext']].header[self.meta['exptime']['card']]
         # Return, transposing array back to orient the overscan properly
```
jhennawi: I'm confused. GMOS images are multi-extension fits, but you appear to be only passing back extension 0, and extension 1, whereas before the entire hdu was passed back. Not sure this matters, but it is confusing.

pypeit/spectrographs/keck_deimos.py
```diff
@@ -488,7 +488,7 @@ def get_rawimage(self, raw_file, det):
 
         # Return
         exptime = hdu[self.meta['exptime']['ext']].header[self.meta['exptime']['card']]
```
jhennawi: Why is this different than gemini_gmos, where you passed back extension 0 and 1.

pypeit/spectrographs/spectrograph.py
```diff
@@ -473,7 +473,7 @@ def get_rawimage(self, raw_file, det):
                 oscansec_img = pix_img.copy()
 
         # Return
-        return raw_img, hdu, exptime, rawdatasec_img, oscansec_img
+        return raw_img, headarr, exptime, rawdatasec_img, oscansec_img
```
jhennawi: Now it is a headarr. Shouldn't this be the same for all instruments that have multiple extensions?

jhennawi:
Hi @rcooke-ast regarding your point about passing the headers around, or the fitstable. I'm opposed to the latter because it adds a dependency on the fitstable, which then means all users would have to run pypeit_setup succesfully before running the base classes. When you are developing or testing or scripting, you don't want to have to do that.  Currently the only base class that depends on fitstable is calibrations and pypeit, and there is no way around it for those two classes. The rest of the classes operate on files or specific objects, so creating this additional dependency is unnecessary and undesirable.  

Can you give us an example of a particular piece of information that you need to reduce KCWI data that you don't have? My inclination would be to put only the specific information you need in the processedimage object. This is how we deal with binning and other such attributes. The other thing we are discussing is adding run-time instrument specific information in spectrograph. My thinking is that you want to put things there, and not pass headers around.

pypeit/spectrographs/gemini_gmos.py
rcooke-ast: You're right - we should just use the function `get_headarr()` for all. I'll fix that now for all instruments.

rcooke-ast:
@kbwestfall - That's a valid point. The way I view it is that all of the other arguments returned are based on either one detector and/or one extension of the fits file. I can't see why we would need all of the other HDUs loaded into memory (especially since we would have none of their datasec/oscansec details, so we'd have to call `get_rawimage()` again, anyway). The only useful bit of information is that sometimes a single data extension benefits from knowing the header from different HDUs (e.g. sometimes relevant information is split over head0 and head1).

@jhennawi - In hindsight, I agree with your thoughts on not passing around fitstable. In principle, we can also not pass all of the headers around (I agree that's a little overkill), but maybe have another definition in `spectrograph.py` that lists key information that should be saved into the master frames (originally header information from the raw frames). I think, in principle, we can get around this, provided that all of the master frames list the filenames of the raw frames that were used in the combination. I've just found a minor bug where this doesn't happen, which is included in my future KCWI PR (basically, head0 is written to a master calibration file which contains the input raw frames, but if you run pypeit without the -m flag, the master frames generated have `None` for the `head0` parameter). So, there's a simple fix, and we can simply get rid returning `hdu/headarr`.

Regarding KCWI - the main issue is the wavelength calibration. It can be performed with FeAr or ThAr, and we can't hardcode the lamps in the `par['calibrations']['wavelengths']['lamps']` parameter. So, just before the wavelength fitting, I need to grab the lamps from the original header. Otherwise, the user needs to figure out which lamps are on themselves, and set this in the `.pypeit` file.

So, as far as I see it, we have two options:
(1) pass around all headers (i.e. return `headarr` from `get_rawimage`)
(2) return neither `hdu` nor `headarr`, but make sure all master calibrations contain the original raw filenames.

I'm leaning towards option (2)... if agreed, I'll make the appropriate changes to this PR.

kbwestfall:
Hi @rcooke-ast, all,

A few more thoughts/comments:
  - I'm fine with the change from `HDUList` to `list` returns.  I'd like us to get to a point where we are opening files as few times as possible, limited by what we think is reasonable to hold in memory.  But this seems like a low-priority goal that no one is really pursuing...
 - `binning` is a good example of something that we pull from the headers in the initial setup with `fitstbl` but becomes something that we pass around separate from the `fitstbl` object.  You could follow that approach with the specific lamp that's used in the calibrations.  The process is non-trivial and I don't remember the details, but you would add the parameter to the list of additional metadata items.  (I would do this after merging with the `deimos_merge3` branch because this was moved to its own module, `pypeit.core.meta`.)  Then you can add a lamp keyword (or something) to the relevant calibration methods in `Calibrations` and pass it the value from the `fitstbl`.  I'd prefer that we do that instead having to read in the data from the raw file again.  I.e., I'd rather parse all the header data we need from the files at the beginning of the pypeit run so that the code will know if it can keep going or not early on.
 - Given the above, we probably want to wait to merge this until after the PR #817 gets merged.  But eventually, you also need to post the results of the tests.
 - There are a couple of places where you edited the docstrings that don't follow the right syntax.  We all need to learn how to do this right.  For a given docstring, you have to choose to use either Numpy or Google style, you can't mix them.  The relevant resources are in the `sphinx.readme` file at the top level of the repo.

rcooke-ast:
@kbwestfall - good points. I think the best thing to do is close this PR until we talk about this in more detail.

PR #825, opened 2020-01-31
Identify script

rcooke-ast:
Launch an identify window outside of PypeIt runtime.

(just found `git cherry-pick`... wow...)

This doesn't change anything in the main code, and merges easily with the `deimos_merge3` branch, so I don't see an issue in bumping this ahead of the deimos PR.

### DIFF
--- a/bin/pypeit_identify
+++ b/bin/pypeit_identify
@@ -1,0 +1,11 @@
+#!/usr/bin/env python
+
+"""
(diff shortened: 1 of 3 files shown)


rcooke-ast:
Sorry - I see now that `deimos_merge3` has changed some things from `identify.py`... I'll wait  :-)

PR #928, opened 2020-05-05
KCWI move Alignments to DataContainer

rcooke-ast:
As discussed in PR #920, the Alignments frame has now moved to a DataContainer (I think!). Was pretty straightforward, tbh.


pypeit/alignframe.py
```diff
@@ -49,295 +121,176 @@ class Alignment():
...
-            checksum (:obj:`bool`, optional):
-                Passed to `astropy.io.fits.HDUList.writeto`_ to add
-                the DATASUM and CHECKSUM keywords fits header(s).
-            float_dtype (:obj:`str`, optional):
-                Convert floating-point data to this data type before
-                writing.  Default is 32-bit precision.
-        """
+                    msgs.error("Alignment profiling failed to generate traces")
+                align_traces[:, bar, sl] = align_prof[sls][bar].TRACE_SPAT
+        return align_traces
```
kbwestfall, line 230: May not be an issue, and I guess this goes back to the use of `extract.objfind` to get the traces, but the nested `for` loop here feels inefficient.  Is there a way to optimize this better?  We can do that in a different PR.

pypeit/calibrations.py
```diff
@@ -301,82 +301,48 @@ def get_align(self):
...
 
         Returns:
-            ndarray or str: :attr:`align`
+            :class:`pypeit.alignframe.Alignments`:
 
         """
         # Check for existing data
         if not self._chk_objs(['msbpm', 'slits']):
-            msgs.error("Don't have all the objects")
+            msgs.warn('Must have the bpm and slits to make the alignments!  Skipping and may crash down the line')
```
kbwestfall, line 312: If this is true (i.e, you need these things to proceed) I think this should crash instead of just throwing a warning.

rcooke-ast: I feel like it's better to throw an error at the place where the problem occurs, but I was just following suit with `get_flats`, `get_tilts`, and `get_wv_calib` in `calibrations.py`. I've made the change in this instance, but should we also change the other methods as well?

pypeit/alignframe.py
rcooke-ast: This part is super fast compared to the object tracing, so I'm not too concerned. I think the problem is because the object tracing is stored in a dictionary. I'm not really sure how to convert this into a single for loop without changing the `extract.objfind`...

pypeit/calibrations.py
kbwestfall: Sorry, didn't realize we were doing this elsewhere.  @profxj what do you suggest?

rcooke-ast:
Thanks @profxj - tests and docs added, and also added to frametypes. I've also added a script to plot the alignments to double check they're working as they should 😊

PR #947, opened 2020-06-05
KCWI reductions

rcooke-ast:
This PR brings KCWI reductions to completion.

The main improvement here includes:

* The integration of `IFUreduce` into `Reduce`
* Including the relative spectral illumination profile in the generation of the flat images
* A change to the `FlatImages` data container (now stores illumflat and pixelflat separately, and also includes the relative spectral illumination profile).


rcooke-ast:
OK - KCWI is now ready for review. Travis tests pass, and I'm going to start a run on the dev suite. Will report back on Thursday if it succeeds 🤞

pypeit/flatfield.py
```diff
@@ -203,10 +295,32 @@ def show(self, slits=None, wcs_match=True):
...
+        else:
+            # Show everything that's available (anything that is None will not be displayed)
+            image_list = zip([self.pixelflat_norm, illumflat_pixel, self.pixelflat_raw, self.pixelflat_model,
+                              self.pixelflat_spec_illum, illumflat_illum, self.illumflat_raw],
+                             ['pixelflat_norm', 'pixelflat_spat_illum', 'pixelflat_raw', 'pixelflat_model',
+                              'pixelflat_spec_illum', 'illumflat_spat_illum', 'illumflat_raw'],
+                             [(0.9, 1.1), (0.9, 1.1), None, None,
+                              (0.8, 1.2), (0.9, 1.1), None])
+        # Display frames
+        show_flats(image_list, wcs_match=wcs_match, slits=slits)
```
profxj, line 322: To what extent has this been tested?  GUIs are often neglected.  :)

pypeit/flatfield.py
```diff
@@ -1011,8 +1166,184 @@ def spatial_fit(self, norm_spec, spat_coo, median_slit_width, spat_gpm, gpm, deb
...
+        (2) Using this rough scale, perform a joint b-spline fit to all slits. This
+            step ensures that a single functional form is used in step 3 to fit all
+            slits. It also ensures that the model covers the min and max wavelength
+            range of all slits.
+        (3) Calculate the relative scale of each slit, using the joint model
+            calculated in step (2).
+
+        Parameters
+        ----------
+        debug : bool
```
profxj, line 1184: add `gpm`

pypeit/flatfield.py
```diff
@@ -1011,8 +1166,184 @@ def spatial_fit(self, norm_spec, spat_coo, median_slit_width, spat_gpm, gpm, deb
...
+        slitmask = self.slits.slit_img(initial=True, flexure=flex)
+        tilts = self.wavetilts.fit2tiltimg(slitmask, flexure=flex)
+        waveimg = wavecalib.build_waveimg(self.spectrograph, tilts, self.slits, self.wv_calib, spat_flexure=flex)
+        msgs.info('Performing a joint fit to the flat-field response')
+        # Grab some parameters
+        trim = self.flatpar['slit_trim']
+        spec_samp_fine = self.flatpar['spec_samp_coarse']
+        rawflat = self.rawflatimg.image.copy() / self.msillumflat.copy()
+        # Grab the BPM and the slit images
+        gpm = np.ones_like(rawflat, dtype=bool) if self.rawflatimg.bpm is None else (
```
profxj, line 1204: does this over-write the input `gpm`?

pypeit/reduce.py
```diff
@@ -247,6 +246,8 @@ def extract(self, global_sky, sobjs_obj):
                 Sky estimate
             sobjs_obj (:class:`pypeit.specobjs.SpecObjs`):
                 List of SpecObj that have been found and traced
+            scaleImg (:obj:`numpy.ndarray`, float):
```
profxj, line 249: `_`

pypeit/reduce.py
```diff
@@ -609,6 +653,61 @@ def local_skysub_extract(self, global_sky, sobjs,
...
+            master_dir = self.caliBrate.master_dir
+            master_key = self.caliBrate.fitstbl.master_key(0, det=self.det) + "_" + sciName
+
+            regfile = masterframe.construct_file_name(buildimage.SkyRegions,
+                                                      master_key=master_key,
+                                                      master_dir=master_dir)
+            # Check if a file exists
+            if os.path.exists(regfile):
+                msgs.info("Loading SkyRegions file for: {0:s} --".format(sciName) + msgs.newline() + regfile)
+                skymask_init = fits.getdata(regfile).astype(np.bool)
```
profxj, line 692: Shouldn't the read just mimic how we read other `PypeItImages` ?

pypeit/reduce.py
```diff
@@ -1138,7 +1237,7 @@ def local_skysub_extract(self, global_sky, sobjs,
         return self.skymodel, self.objmodel, self.ivarmodel, self.outmask, self.sobjs
 
 
-class IFUReduce(Reduce):
+class IFUReduce(MultiSlitReduce, Reduce):
```
profxj, line 1240: Didn't know you can do that.  Do you need to?
ie.  isn't just `MultiSlitReduce` enough?

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -115,13 +118,12 @@ def config_specific_par(self, scifile, inp_par=None):
...
 
         # Templates
-        print(self.get_meta_value(headarr, 'dispname'))
         if self.get_meta_value(headarr, 'dispname') == 'BH2':
             par['calibrations']['wavelengths']['method'] = 'full_template'  # 'full_template'
             par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_kcwi_BH2_4200.fits'
             par['calibrations']['wavelengths']['lamps'] = ['FeI', 'ArI', 'ArII']
-            par['calibrations']['wavelengths']['n_first'] = 2
-            par['calibrations']['wavelengths']['n_final'] = 6
+            # par['calibrations']['wavelengths']['n_first'] = 4
```
profxj, line 125: do you mean for these to be commented out?

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -180,7 +182,12 @@ def default_pypeit_par(self):
         """
         par = pypeitpar.PypeItPar()
         par['rdx']['spectrograph'] = 'keck_kcwi'
-        # Set wave tilts order
+
+        # Subtract the detector pattern from all frames
+        #par['baseprocess']['use_pattern'] = True
```
profxj, line 187: remove?

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -441,9 +453,50 @@ def get_rawimage(self, raw_file, det):
...
                 oscansec_img = pix_img.copy()
 
+        # Calculate the pattern frequency
+        hdu = self.calc_pattern_freq(raw_img, rawdatasec_img, oscansec_img, hdu)
+
         # Return
         return detpar, raw_img, hdu, exptime, rawdatasec_img, oscansec_img
 
+    def calc_pattern_freq(self, frame, rawdatasec_img, oscansec_img, hdu):
+        # TODO :: PATTERN FREQUENCY ALGORITHM HAS NOT BEEN TESTED WHEN BINNING != 1x1
```
profxj, line 463: doc string

pypeit/calibrations.py
```diff
@@ -504,33 +505,57 @@ def get_flats(self):
...
-            self.flatimages = flatField.run(show=self.show)
-
-            # Save to Masters
-            self.flatimages.to_master_file(masterframe_filename)
-            # Save slits too, in case they were tweaked
-            self.slits.to_master_file()
+        pixelflatImages, illumflatImages = None, None
+        # Check if the image files are the same
+        pix_is_illum = Counter(illum_image_files) == Counter(pixflat_image_files)
+        if len(pixflat_image_files) > 0:
```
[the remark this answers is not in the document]
rcooke-ast: Not the KCWI dev suite, but the dev suite touches this, as I found a fail. Once the dev suite passes, this should be OK.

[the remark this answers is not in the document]
rcooke-ast: Sorry, just re-read the question - the KCWI dev suite has separate `illumflat` and `pixelflat` files. So, yes, the dev suite has a specific check for different file types.

pypeit/core/procimg.py
```diff
@@ -388,6 +396,210 @@ def subtract_overscan(rawframe, datasec_img, oscansec_img,
...
+
+
+def pattern_frequency(frame, axis=1):
+    """
+    Using the supplied 2D array, calculate the pattern frequency
+    along the specified axis.
+
+    Args:
+        frame (:obj:`numpy.ndarray`):
+            2D array to measure the pattern frequency
```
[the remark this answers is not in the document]
rcooke-ast: Done!

pypeit/core/procimg.py
```diff
@@ -388,6 +396,210 @@ def subtract_overscan(rawframe, datasec_img, oscansec_img,
...
+    amp = np.fft.rfft(arr, axis=1)
+    idx = (np.arange(arr.shape[0]), np.argmax(np.abs(amp), axis=1))
+
+    # Construct the variables of the sinusoidal waveform
+    amps = (np.abs(amp))[idx] * (2.0 / arr.shape[1])
+    phss = np.arctan2(amp.imag, amp.real)[idx]
+    frqs = idx[1]
+
+    # This does a reasonable job, but needs to be much better - a fit or CC would be best to get subpixel sampling.
+    # posn = np.linspace(0.0, 1.0, arr.shape[1])[np.newaxis, :]
```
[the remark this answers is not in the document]
rcooke-ast: Removed

pypeit/flatfield.py
```diff
@@ -110,21 +129,77 @@ def _bundle(self):
...
+                    dkey = 'bspline-{}'.format(self.spat_id[ss])
+                    # Save
+                    d.append(copy.deepcopy({dkey: spat_bspl}))
             else: # Add to header of the primary image
                 d[0][key] = self[key]
         # Return
         return d
 
-    def fit2illumflat(self, slits, initial=False, flexure_shift=None):
+    # def merge_with(self, flatimgs):
```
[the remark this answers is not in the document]
rcooke-ast: Removed

pypeit/flatfield.py
rcooke-ast: I've tested all combinations of this, and tested from the script. So, I think this is OK.

pypeit/core/procimg.py
```diff
@@ -388,6 +396,210 @@ def subtract_overscan(rawframe, datasec_img, oscansec_img,
...
+
+        # Use the above to as initial guess parameters in chi-squared minimisation
+        cosfunc = lambda xarr, *p: p[0] * np.cos(2.0 * np.pi * p[1] * xarr + p[2])
+        xdata, step = np.linspace(0.0, 1.0, overscan.shape[1], retstep=True)
+        xdata_all = (np.arange(osd_slice[1].start, osd_slice[1].stop) - os_slice[1].start) * step
+        model_pattern = np.zeros_like(oscandata)
+        val = np.zeros(overscan.shape[0])
+        # Get the best estimate of the amplitude
+        for ii in range(overscan.shape[0]):
+            popt, pcov = curve_fit(cosfunc, xdata, overscan[ii, :], p0=[amps[ii], use_fr, phss[ii]],
```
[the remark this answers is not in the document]
rcooke-ast: well-spotted - I've caught the errors now

pypeit/core/procimg.py
```diff
@@ -332,8 +333,15 @@ def subtract_overscan(rawframe, datasec_img, oscansec_img,
             Frame from which to subtract overscan
         numamplifiers (int):
             Number of amplifiers for this detector.
-        datasec_img (np.ndarray):
-        oscansec_img (np.ndarray):
+        datasec_img (:obj:`numpy.ndarray`):
```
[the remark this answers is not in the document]
rcooke-ast: Yes, I'm going to presume you are correct - I've now copied the rest of the code:
`numpy.ndarray_`

pypeit/flatfield.py
rcooke-ast: Done!

pypeit/flatfield.py
rcooke-ast: Well-caught - I've updated this!

pypeit/reduce.py
rcooke-ast: Fixed

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: I was going to ping you on this @profxj - I can't figure out why, but `baseprocess` can't be used to update this argument for all frames, so I have to put this into all frames manually. What is the preferred approach, using `baseprocess` or explicitly list each frametype?

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Yes - the default values work fine - removed

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Done!

pypeit/reduce.py
rcooke-ast: Yep, both are needed (and order is important). When writing this, I read about multiple inheritance and method resolution order, and decided this order was needed and both classes needed to be specified. I've forgotten why...

pypeit/reduce.py
rcooke-ast: Yeah... that was a total hack by me... I've now implemented this as a `PypeItImage` - thanks for the suggestion!

pypeit/core/skysub.py
```diff
@@ -113,9 +115,15 @@ def global_skysub(image, ivar, tilts, thismask, slit_left, slit_righ, inmask=Non
 
     # Init
     (nspec, nspat) = image.shape
-    piximg = tilts * (nspec-1)
+    if use_wave:
```
jhennawi, line 118: Please remove this argument and this code. You can simply deal with this in the calling code by creating a psuedo tilt image, i.e. 

tilt_psuedo = (waveimg - wavimg.min())/(waveimg.max()-waveimg.min()) 

There is a good reason why we use an image in pixel coordinates as the x-coordinate for sky-subtraction, because then the break-point spacing bsp is in pixels.  I'm not sure if you dealt with this in your code or just got lucky somehow. But I would avoid doing this fit in any other coordinates besides tilts*(nspec-1).

pypeit/flatfield.py
```diff
@@ -48,19 +50,33 @@ class FlatImages(datamodel.DataContainer):
...
-                         desc='Mirrors SlitTraceSet mask for the Flat-specific flags'),
-        'spat_bsplines': dict(otype=np.ndarray, atype=bspline.bspline,
-                              desc='B-spline models for Illumination flat'),
-        'spat_id': dict(otype=np.ndarray, atype=np.integer, desc='Slit spat_id '),
+        'spat_id': dict(otype=np.ndarray, atype=np.integer, desc='Slit spat_id'),
+        'procflat': dict(otype=np.ndarray, atype=np.floating, desc='testing'),
+        'pixelflat': dict(otype=np.ndarray, atype=np.floating, desc='testing'),
+        'flat_model': dict(otype=np.ndarray, atype=np.floating, desc='testing'),
+        'bpmflats': dict(otype=np.ndarray, atype=np.integer, desc='testing'),
+        'spat_bsplines': dict(otype=np.ndarray, atype=bspline.bspline, desc = 'testing'),
```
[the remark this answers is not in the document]
rcooke-ast: It seems that class properties need to be included in the data model. Not super clean, but I couldn't find a better way around this. Perhaps @kbwestfall has an alternative solution?

pypeit/core/skysub.py
rcooke-ast: Great suggestion - I've implemented this!

pypeit/calibrations.py
```diff
@@ -501,52 +502,67 @@ def get_flats(self):
...
+            # Initialise the pixel flat
+            illumFlatField = flatfield.FlatField(illum_flat, self.spectrograph,
+                                                 self.par['flatfield'], self.slits, self.wavetilts,
+                                                 self.wv_calib, spat_illum_only=True)
+            # Generate
+            illumflatImages = illumFlatField.run(show=self.show)
+
+        # Merge the illum flat with the pixel flat
+        if pixelflatImages is not None:
+            msgs.info("Merging illumflat parameters into FlatImages")
```
jhennawi, line 546: Can you add a comment making it more clear what this merge is doing. In other words, if the illumflatimages is None, I presume it takes the the illuminatino function from the pixflatimages, but if not, it uses the illumflatimages. 

I also don't follow why you are printing out a message in this case but not in the else.

pypeit/calibrations.py
```diff
@@ -501,52 +502,67 @@ def get_flats(self):
...
+            self.flatimages = flatfield.merge(pixelflatImages, illumflatImages)
         else:
-            self.flatimages = flatfield.FlatImages(None, None, None, None)
+            # No pixel flat, but there might be an illumflat
```
jhennawi, line 549: This is not transparent. This will use the pixelflat generated from the illumination flats as the pixel flat. Since you set spat_illum_only=True, I presume that is an image full of ones, but it is unclear. Please add a comment explaining what is going on here and print out some info to be consistent with the above print statement.  Also, does this fail in the case where I want to use a user defined flat field file via self.flatimages above, but then want to use the illumination function generated from twilight flats? This is actually the case with LRIS, and I think that your code does not owrk.

pypeit/core/skysub.py
```diff
@@ -116,6 +116,9 @@ def global_skysub(image, ivar, tilts, thismask, slit_left, slit_righ, inmask=Non
     piximg = tilts * (nspec-1)
     if inmask is None:
         inmask = (ivar > 0.0) & thismask & np.isfinite(image) & np.isfinite(ivar)
+    else:
```
jhennawi, line 119: The docs require that the input be of type bool. Rather than casting, I would prefer that you just fault if this is not a boolean mask. If necessary you can cast outside of this routine. All the codes in PypeIt use boolean arrays as masks, and we are not providing this flexibility to use 1s or 0s in arrays elsewhere.

pypeit/flatfield.py
```diff
@@ -110,21 +130,59 @@ def _bundle(self):
...
+                for ss, spat_bspl in enumerate(self[key]):
+                    # Naming
+                    spat_bspl.hdu_prefix = 'ILLUMFLAT_SPAT_ID-{}_'.format(self.spat_id[ss])
+                    dkey = 'bspline-{}'.format(self.spat_id[ss])
+                    # Save
+                    d.append(copy.deepcopy({dkey: spat_bspl}))
             else: # Add to header of the primary image
                 d[0][key] = self[key]
         # Return
         return d
```
jhennawi, line 155: I am not sure I'm following while these have been converted from attributes to properties? Can you elaborate on the logic behind that change?

pypeit/flatfield.py
```diff
@@ -110,21 +130,59 @@ def _bundle(self):
...
+            return np.zeros(self.procflat.shape, dtype=np.int)
+
+    @property
+    def spat_bsplines(self):
+        if self.illumflat_spat_bsplines is not None:
+            return self.illumflat_spat_bsplines
+        else:
+            return self.pixelflat_spat_bsplines
+
+    def fit2illumflat(self, slits, frametype='illum', initial=False, flexure_shift=None):
```
jhennawi, line 185: This frametype argument that you are adding is not documented. I cannot see any reason for it. Can you please explain why this code needs to know the frametype? I thought we agreed you would do illum/pixel specific things outside of this class?

pypeit/flatfield.py
```diff
@@ -150,7 +215,7 @@ def fit2illumflat(self, slits, initial=False, flexure_shift=None):
                                                       slitid_img=_slitid_img,
                                                       flexure_shift=flexure_shift)
```
jhennawi, line 217: Let's chat I'm not following any of this.

pypeit/flatfield.py
jhennawi: Great. I would run pypeit_chk_flats on a few setups after dev suite is done. Hard to test GUIs of course.

pypeit/flatfield.py
```diff
@@ -384,11 +518,13 @@ def show(self, wcs_match=True):
...
-        show_flats(self.mspixelflat, self.msillumflat, self.rawflatimg.image, self.flat_model,
-                   wcs_match=wcs_match, slits=self.slits)
+        # Prepare the images to show, their names and their cuts
+        image_list = zip([self.mspixelflat, self.msillumflat, self.rawflatimg.image, self.flat_model],
+                         ['pixelflat', 'spat_illum', 'raw', 'model', 'spec_illum'],
+                         [(0.9, 1.1), (0.9, 1.1), None, None, (0.8, 1.2)])
+        show_flats(image_list, wcs_match=wcs_match, slits=self.slits)
 
-    def fit(self, debug=False):
+    def fit(self, illumflat=False, debug=False):
```
jhennawi, line 527: The default behavior of this routine should be to generate both pixel and illumination flats. There should be an option to short circuit that called spat_illum_only or something like that. This is how your codes work above, but you have switched the convention here and set the default to be not creating the illumflat. I think you should use the same spat_illum_only here, and the default should spat_illum_only=False. The docs should be changed to state that if True, only the illumination flat will be generated and the code will exit.

pypeit/flatfield.py
```diff
@@ -551,14 +690,14 @@ def fit(self, debug=False):
         norm_spec = np.ones_like(rawflat)
         norm_spec_spat = np.ones_like(rawflat)
         twod_model = np.ones_like(rawflat)
```
jhennawi, line 692: I'm not following why you need to change how the bpms were being handled.

pypeit/flatfield.py
```diff
@@ -939,18 +1080,25 @@ def fit(self, debug=False):
             # TODO: Add some code here to treat the edges and places where fits
             #  go bad?
 
+        # No need to continue if we're just doing the spatial illumination
```
jhennawi, line 1083: Please change this to be illum_only or something as I mentioned above.

pypeit/flatfield.py
```diff
@@ -1011,8 +1159,185 @@ def spatial_fit(self, norm_spec, spat_coo, median_slit_width, spat_gpm, gpm, deb
         return exit_status, spat_coo_data, spat_flat_data, spat_bspl, spat_gpm_fit, \
                spat_flat_fit, spat_flat_data_raw
 
+    def spectral_illumination(self, gpm, debug=False):
+        """
```
jhennawi, line 1163: Given your usage, it looks like gpm should be an optional input.

pypeit/par/pypeitpar.py
```diff
@@ -254,6 +259,11 @@ def __init__(self, trim=None, apply_gain=None, orient=None,
         dtypes['use_illumflat'] = bool
         descr['use_illumflat'] = 'Use the illumination flat to correct for the illumination profile of each slit.'
 
+        defaults['use_specillum'] = False
+        dtypes['use_specillum'] = bool
+        descr['use_specillum'] = 'Use the relative spectral illumination profiles to correct the spectral' \
```
jhennawi, line 264: Add a note indicating this is used mainly (exclusively) for IFUs.

pypeit/reduce.py
```diff
@@ -238,15 +237,17 @@ def parse_manual_dict(self, manual_dict, neg=False):
...
         """
         Main method to extract spectra from the ScienceImage
 
         Args:
-            global_sky (ndarray):
+            global_sky (:obj:`numpy.ndarray`_):
                 Sky estimate
             sobjs_obj (:class:`pypeit.specobjs.SpecObjs`):
                 List of SpecObj that have been found and traced
+            scaleImg (:obj:`numpy.ndarray`_, float):
```
jhennawi, line 249: scaleimg is not used in this routine. Please remove this parameter.

pypeit/core/skysub.py
```diff
@@ -1108,16 +1111,18 @@ def ech_local_skysub_extract(sciimg, sciivar, fullmask, tilts, waveimg, global_s
...
         would select the first 10%, the inner 30%, and the final 20% of all slits
 
     Parameters
     ----------
-    text : str
+    skyreg : str
         The sky region definition.
+    nslits : int
+        Number of slits on the detector
     resolution: int, optional
```
jhennawi, line 1126: I cannot follow your documentation. I don't understand what resolution is or why it is 1000. I think specifying percentages is bad news. Why not use absolute pixel coordinates at some reference (y, i.e. spectral) location, since that is what the user sees in the GUI. Then you can compute the percentages internally, but the user is pointing to a specific pixel. Perhaps I'm not following your procedure.

pypeit/reduce.py
```diff
@@ -354,9 +356,17 @@ def run(self, basename=None, ra=None, dec=None, obstime=None,
                               show=self.reduce_show & (not self.std_redux),
                               manual_extract_dict=manual_extract_dict)
 
+        # Check if the user wants to overwrite the skymask with a pre-defined sky regions file
+        skymask_init, usersky = self.load_skyregions(skymask_init)
+
+        # Define the trim regions for different pypelines
```
jhennawi, line 362: This is not allowed, since it violates the inheritance structure that we set up for this class. extract is general, and is supposed to handle all cases. If you need to change that and do something IFU specifici, make a child that overloads it. I think what you want to achieve here with edge trimming can be trivially accomodated via a parset which would be the better approach.

pypeit/reduce.py
```diff
@@ -354,9 +356,17 @@ def run(self, basename=None, ra=None, dec=None, obstime=None,
...
+        if self.spectrograph.pypeline == "IFU":
+            trim_edg = (0, 0)
+        else:
+            trim_edg = (3, 3)
+
         # Global sky subtract
-        self.initial_sky = \
-            self.global_skysub(skymask=skymask_init).copy()
+        self.initial_sky = self.global_skysub(skymask=skymask_init, trim_edg=trim_edg).copy()
```
jhennawi, line 369: Remove the trim_edge, this can be handled via parsets.

pypeit/reduce.py
```diff
@@ -528,35 +535,73 @@ def global_skysub(self, scaleImg=None, skymask=None, update_crmask=True, trim_ed
...
-        # Select the edges to use: Selects the edges tweaked by the
-        # illumination profile if they're present; otherwise, it
-        # selects the original edges from EdgeTraceSet. To always
-        # select the latter, use the method with `original=True`.
-        # TODO -- JXP removed these from KCWI4..
-        #original = True if ((trim_edg[0] == 0) and (trim_edg[1] == 0)) else False
-        #left, right = self.slits.select_edges(original=original)
-        #left, right = self.slits.select_edges()
-
         if self.par['reduce']['skysub']['joint_fit']:
```
jhennawi, line 538: As I think we are unlikely to use joint_fit in any other context except for IFUs, my suggestion is that you simply overload global skysub rather than adding this if then. Perhaps @profxj would like to comment. But we created this inheritance structure for exactly this reason.

pypeit/reduce.py
```diff
@@ -528,35 +535,73 @@ def global_skysub(self, scaleImg=None, skymask=None, update_crmask=True, trim_ed
...
-                = skysub.global_skysub(self.sciImg.image/scalefact, self.sciImg.ivar, wavenorm,
+                = skysub.global_skysub(self.sciImg.image, self.sciImg.ivar, tilt_wave,
                                        thismask, self.slits_left, self.slits_right, inmask=inmask,
                                        sigrej=sigrej, trim_edg=trim_edg,
                                        bsp=self.par['reduce']['skysub']['bspline_spacing'],
                                        no_poly=self.par['reduce']['skysub']['no_poly'],
-                                       pos_mask=(not self.ir_redux), show_fit=True)#show_fit)
-            # Apply the scaling factor to the sky image
-            self.global_sky *= scaleImg
+                                       pos_mask=(not self.ir_redux), use_wave=True, show_fit=show_fit)
```
jhennawi, line 552: use_wave is no longer an option to global_skysub, so technically this should crash? Have you tested this recently?

pypeit/reduce.py
```diff
@@ -528,35 +535,73 @@ def global_skysub(self, scaleImg=None, skymask=None, update_crmask=True, trim_ed
...
+            msgs.info("Correcting science frame for relative spectral illumination")
+            scaleFact = scaleImg + (scaleImg == 0)
+            sciImg, varImg = flat.flatfield(self.sciImg.image.copy(), scaleFact, self.sciImg.fullmask,
+                                            varframe=utils.inverse(self.sciImg.ivar.copy()))
+            self.sciImg.image = sciImg.copy()
+            self.sciImg.ivar = utils.inverse(varImg)
+            # Repeat the sky subtraction
+            msgs.info("Repeating global sky subtraction")
+            self.global_sky[thismask] \
+                = skysub.global_skysub(self.sciImg.image, self.sciImg.ivar, tilt_wave,
```
jhennawi, line 599: Same comment about use_wave.

pypeit/reduce.py
```diff
@@ -609,6 +654,62 @@ def local_skysub_extract(self, global_sky, sobjs,
 
         return None, None, None, None, None
 
+    def load_skyregions(self, skymask_init):
```
jhennawi, line 657: I am not following why there are two ways to load in these sky regions. I am also not convinced that we want this to be a master file, but we should discuss that.

pypeit/spectrographs/keck_kcwi.py
jhennawi: Does the baseprocess documentation here help? https://pypeit.readthedocs.io/en/latest/pypeit_par.html 

I've used this for e.g. use_illumflat and other things, so I'm not sure why it would fail for use_pattern.

pypeit/reduce.py
rcooke-ast: I agree. I've handled this with a parset now.

pypeit/reduce.py
rcooke-ast: Done!

pypeit/reduce.py
rcooke-ast: Some IFUs will not want to joint fit, so I would still need the if/then statement. There might be some cases (e.g. multislit which contains a narrow slit with very limited sky regions) that one might wish to perform a joint fit (at least for that slit). I agree otherwise, this is a pretty specific need... I've left it in for now unless there is resistance.

jhennawi: I am not happy with it, because I don't agree that we would ever do this in multislit. We certainly have never had the need to. The whole point of local_skysub_extract is to deal with these narrow slits and model the sky and object simultaneously. So your usage is compeltely IFU specific and should be a child not an if/then.

rcooke-ast: OK - I'll overload

rcooke-ast:
All tests pass and the dev suite (`develop`) passes. With one exception:

```
Running pypeit_coadd_1dspec on gemini_gmos/GS_HAM_R400_860COADD1D  --- FAILED
No matching objects for SPAT1508-SLIT1789-DET01.  Odds are you input the wrong OBJID
```

pypeit/spectrographs/keck_kcwi.py
jhennawi: I think baseprocess onlyh works with pypeit files, sorry. @kbwestfall is the expert. In code, I think I've been setting it by hand using that function that X created.

profxj:
Unit tests pass [pulled from Slack]

================ 229 passed, 573 warnings in 451.42s (0:07:31) ================= (edited)

rcooke-ast:
Hi @jhennawi - All tests pass except for the one in the dev suite, as mentioned above. This appears to have also been seen in previous PRs (e.g. [950](https://github.com/pypeit/PypeIt/pull/950#issuecomment-644997149)), so may not be introduced by the code here?

jhennawi:
Sorry to be a stickler guys, but the rule has always been that you post the screenshot of the terminal output for the tests after you run by pypeit_test develop. Not: 

pypeit_test reduce
pypeit_test afterburn
or "I ran the tests and they all passed"

For all of the above, you could have fixed one test, but that fix might have broke another. That is why we have the script, the development doc, and this wonderful (although somewhat tedious) procedure.

jhennawi:
Also, don't worry about the test that is failing @rcooke-ast, it is a known issue (not actually an issue, just a tedious test to fix because we have to identify the object pixel by hand).

rcooke-ast:
OK - I'll re-run the tests overnight - sorry wan't aware of the screenshot idea. Also, I haven't done the "afterburn" thing, so I'll check that too, and send a screenshot. I agree - that's a good way to make sure the code doesn't get broken along the way with multiple development.

jhennawi:
Hi @rcooke-ast, don't run reduce or afterburn. Those are for before you run the final test that gets the screenshot. pypeit_test develop will run everything you need to get the PR approved, i.e. it does the reductions (which is what reduce does), and then it runs the things on those reductions (which is what afterburn does).

rcooke-ast:
image

rcooke-ast:
Dev suite tests (1 fail, as mentioned above):
image

PR #980, opened 2020-07-25
KCWI - making cubes...

rcooke-ast:
This PR contains a script that converts a series of spec2D input files into datacubes for IFU data. Most likely, some rearrangement of the functions will be necessary, but I'm unsure of where the best place to put them would be, so happy to receive comments about that.

I've also included a new output for the spec2D files, which is needed to make sure the spec2D files all contain the same relative spectral scaling.

All tests and dev suite tests are passing.


pypeit/reduce.py
```diff
@@ -1195,128 +1198,163 @@ def find_objects_pypeline(self, image, std_trace=None,
...
+        self.scaleimg *= scaleFact
+        sciImg, varImg = flat.flatfield(self.sciImg.image.copy(), scaleFact, self.sciImg.fullmask,
+                                        varframe=utils.inverse(self.sciImg.ivar.copy()))
+        self.sciImg.image = sciImg.copy()
+        self.sciImg.ivar = utils.inverse(varImg)
+        return
 
-        Returns:
-            numpy.ndarray: image of the the global sky model
+    def illum_profile_spatial(self, skymask=None, trim_edg=(0, 0)):
```
profxj, line 1221: need a doc string

pypeit/reduce.py
```diff
@@ -1195,128 +1198,163 @@ def find_objects_pypeline(self, image, std_trace=None,
...
+            plt.imshow(spatScaleImg, vmin=0.99, vmax=1.01)
+            plt.show()
+            plt.subplot(133)
+            plt.plot(xplt[0::2], coeff_fit[::2, 2], 'rx')
+            plt.plot(xplt[1::2], coeff_fit[1::2, 2], 'bx')
+            plt.show()
+        # Apply the relative scale correction
+        self.apply_relative_scale(spatScaleImg)
+
+    def illum_profile_spectral(self, global_sky, skymask=None):
```
profxj, line 1309: add a doc string

pypeit/reduce.py
```diff
@@ -1195,128 +1198,163 @@ def find_objects_pypeline(self, image, std_trace=None,
...
+        trim = self.par['calibrations']['flatfield']['slit_trim']
+        gpm = (self.sciImg.fullmask == 0)
+        scaleImg = flatfield.illum_profile_spectral(self.sciImg.image.copy(), self.waveimg, self.slits,
+                                                    model=global_sky, gpmask=gpm, skymask=skymask, trim=trim,
+                                                    flexure=self.spat_flexure_shift)
+        # Now apply the correction to the science frame
+        self.apply_relative_scale(scaleImg)
+
+    def joint_skysub(self, skymask=None, update_crmask=True, trim_edg=(0,0),
```
profxj, line 1318: same

pypeit/reduce.py
```diff
@@ -1332,6 +1370,80 @@ def global_skysub(self, skymask=None, update_crmask=True, trim_edg=(0,0),
...
+    def global_skysub(self, skymask=None, update_crmask=True, trim_edg=(0,0),
+                      show_fit=False, show=False, show_objs=False):
+        """
+        Perform global sky subtraction. This IFU-specific routine ensures that the
+        edges of the slits are not trimmed, and performs a spatial and spectral
+        correction using the sky spectrum, if requested. See Reduce.global_skysub()
+        for parameter definitions.
+        """
+        # Generate a global sky sub for all slits separately
+        global_sky_sep = Reduce.global_skysub(self, skymask=skymask, update_crmask=update_crmask, trim_edg=trim_edg,
```
profxj, line 1384: We have generally done a `super` call, although I admit I don't know the difference.

pypeit/reduce.py
```diff
@@ -1332,6 +1370,80 @@ def global_skysub(self, skymask=None, update_crmask=True, trim_edg=(0,0),
...
+                                              show_fit=show_fit, show=show, show_objs=show_objs)
+        # If the joint fit or spec/spat sensitivity corrections are not being performed, return the separate slits sky
+        if not self.par['reduce']['skysub']['joint_fit'] and \
+                not self.par['scienceframe']['process']['use_specillum'] and \
+                not self.par['scienceframe']['process']['use_illumflat']:
+            return global_sky_sep
+
+        # Do the spatial scaling first
```
profxj, line 1392: remove?

pypeit/scripts/coadd_datacube.py
```diff
@@ -0,0 +1,452 @@
...
+                        action='store_true')
+    parser.add_argument('--det', default=1, type=int, help="Detector")
+    parser.add_argument('-o', '--overwrite', default=False, action='store_true',
+                        help='Overwrite any existing files/directories')
+
+    return parser.parse_args() if options is None else parser.parse_args(options)
+
+
+def dar_fitfunc(radec, coord_ra, coord_dec, datfit, wave, obstime, location, pressure, temperature, rel_humidity):
+    (diff_ra, diff_dec) = radec
```
profxj, line 48: short doc string, please

pypeit/scripts/coadd_datacube.py
```diff
@@ -0,0 +1,452 @@
...
+
+
+def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, rel_humidity,
+                   wave_ref=None, numgrid = 10):
+    """Apply a differental atmospheric refraction correction to the input ra/dec.
+    This implementation is based on ERFA, which is called through astropy
+
+    Parameters
+    ----------
+    wave_arr (ndarray):
```
profxj, line 65: this is a mix of Google and numpy doc style

profxj: I think you can make it Google simply by changing `Parameters` to `Args:`

Oh, and you'd have to change the Returns approach..

pypeit/scripts/coadd_datacube.py
```diff
@@ -0,0 +1,452 @@
...
+    dec_diff = spl_dec(wave_arr) - spl_dec(wave_ref)
+
+    return ra_diff, dec_diff
+
+
+def generate_masterWCS(crval, cdelt, equinox=2000.0):
+    """ Generate a WCS that will cover all input spec2D files
+
+    Parameters
+    ----------
```
profxj, line 130: Google+numpy

pypeit/scripts/coadd_datacube.py
```diff
@@ -0,0 +1,452 @@
...
+    w.wcs.crval = crval  # RA, DEC, and wavelength zeropoints
+    w.wcs.crpix = [0, 0, 0]  # RA, DEC, and wavelength reference pixels
+    #w.wcs.cd = np.array([[cdval[0], 0.0, 0.0], [0.0, cdval[1], 0.0], [0.0, 0.0, cdval[2]]])
+    w.wcs.cdelt = cdelt
+    w.wcs.lonpole = 180.0  # Native longitude of the Celestial pole
+    w.wcs.latpole = 0.0  # Native latitude of the Celestial pole
+    return w
+
+
+def twoD_Gaussian(tup, amplitude, xo, yo, sigma_x, sigma_y, theta, offset):
```
profxj, line 162: Should this go in `utils` or the coming `fitting.py` module?

pypeit/scripts/coadd_datacube.py
```diff
@@ -0,0 +1,452 @@
...
+    xo = float(xo)
+    yo = float(yo)
+    a = (np.cos(theta)**2)/(2*sigma_x**2) + (np.sin(theta)**2)/(2*sigma_y**2)
+    b = -(np.sin(2*theta))/(4*sigma_x**2) + (np.sin(2*theta))/(4*sigma_y**2)
+    c = (np.sin(theta)**2)/(2*sigma_x**2) + (np.cos(theta)**2)/(2*sigma_y**2)
+    g = offset + amplitude*np.exp( - (a*((x-xo)**2) + 2*b*(x-xo)*(y-yo) + c*((y-yo)**2)))
+    return g.ravel()
+
+
+def calculate_offset(image, im_ref, nfit=3):
```
profxj, line 180: Add to `flexure.py` ?

pypeit/scripts/coadd_datacube.py
```diff
@@ -0,0 +1,452 @@
...
+    """Calculate the x,y offset between two images
+
+    Parameters
+    ----------
```
profxj, line 184: mix of Google and numpy

pypeit/spec2dobj.py
```diff
@@ -61,6 +61,7 @@ class Spec2DObj(datamodel.DataContainer):
         'objmodel': dict(otype=np.ndarray, atype=np.floating, desc='2D object model image'),
         'ivarmodel': dict(otype=np.ndarray, atype=np.floating, desc='2D ivar model image'),
         'tilts': dict(otype=np.ndarray, atype=np.floating, desc='2D tilts image'),
+        'scaleimg': dict(otype=np.ndarray, atype=np.floating, desc='2D multiplicative scale image that has been applied to the science image'),
```
profxj, line 64: If we are modifying the datamodel, which I presume is desired/necessary, please increment the `version` value.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -38,6 +41,8 @@ def __init__(self):
         self.grating = None
         self.optical_model = None
         self.detector_map = None
+        self.slicescale = 0.00037718  # Degrees per 'large slicer' slice
```
profxj, line 44: Can these be read from the Header?  And/Or should they be set dynamically inside `config_par`?

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -218,7 +214,7 @@ def default_pypeit_par(self):
         #par['calibrations']['flatfield']['tweak_slits'] = False  # Do not tweak the slit edges (we want to use the full slit)
         par['calibrations']['flatfield']['tweak_slits_thresh'] = 0.0  # Make sure the full slit is used (i.e. when the illumination fraction is > 0.5)
         par['calibrations']['flatfield']['tweak_slits_maxfrac'] = 0.0  # Make sure the full slit is used (i.e. no padding)
-        par['calibrations']['flatfield']['slit_trim'] = 0  # Make sure the full slit is used (i.e. no padding)
+        par['calibrations']['flatfield']['slit_trim'] = 3  # Make sure the full slit is used (i.e. no padding)
```
profxj, line 217: Should you update the comment?

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -607,51 +617,251 @@ def bpm(self, filename, det, shape=None, msbias=None):
...
-        hdr['LATPOLE'] = (0.0, 'Celestial latitude of native pole')
-        hdr['HISTORY'] = "  {0:s} {1:s}".format(kgeom.progid, systime(0, kgeom.timestamp))
-        hdr['HISTORY'] = "  {0:s} {1:s}".format(pre, systime(0))
-        return hdr
+        if 'Mask' in hdr['BNASNAM']:
+            return True
+        else:
+            return False
+
+    @staticmethod
```
profxj, line 640: Are you adding this because it changes from frame to frame?
Put another way, why not use the `meta` approach?
Note that X-Shooter has frames with multiple binnings.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -607,51 +617,251 @@ def bpm(self, filename, det, shape=None, msbias=None):
...
+            The header of the raw frame.
+
+        Returns
+        -------
+        list : x,y pixel binning
+        """
+        binning = hdr['BINNING']
+        return [int(ibin) for ibin in binning.split(',')]  # [binspectral, binspatial]
+
+    def get_scales(self, hdr):
```
profxj, line 656: Consider having this part of the `meta`.  And it seems you
can remove the hard-coded values at the head of the Class.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -607,51 +617,251 @@ def bpm(self, filename, det, shape=None, msbias=None):
...
+        pxscl = self.platescale * self.get_binning(hdr)[1]
+        slscl = self.slicescale
+        ifunum = hdr['IFUNUM']
+        if ifunum == 2:
+            slscl = self.slicescale/2.0
+        elif ifunum == 3:
+            slscl = self.slicescale/4.0
+        return pxscl, slscl
+
+    def get_wcs(self, hdr, slits, wave0, dwv):
```
profxj, line 678: Is this fully KCWI specific?  Or can it be a method in `spectrographs.py`?

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -607,51 +617,251 @@ def bpm(self, filename, det, shape=None, msbias=None):
...
+        w.wcs.crval = [ra, dec, wave0]  # RA, DEC, and wavelength zeropoints
+        w.wcs.crpix = [crpix1, crpix2, crpix3]  # RA, DEC, and wavelength reference pixels
+        w.wcs.cd = np.array([[cd11, cd12, 0.0], [cd21, cd22, 0.0], [0.0, 0.0, dwv]])
+        w.wcs.lonpole = 180.0  # Native longitude of the Celestial pole
+        w.wcs.latpole = 0.0  # Native latitude of the Celestial pole
+
+        return w
+
+    def get_radec_image(self, alignments, slits, wcs, flexure=None):
+        """Get the WCS for a frame
```
profxj, line 784: I think the doc purpose is out of date.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -607,51 +617,251 @@ def bpm(self, filename, det, shape=None, msbias=None):
...
+        w.wcs.ctype = ["RA---TAN", "DEC--TAN", "AWAV"]
+        w.wcs.crval = [ra, dec, wave0]  # RA, DEC, and wavelength zeropoints
+        w.wcs.crpix = [crpix1, crpix2, crpix3]  # RA, DEC, and wavelength reference pixels
+        w.wcs.cd = np.array([[cd11, cd12, 0.0], [cd21, cd22, 0.0], [0.0, 0.0, dwv]])
+        w.wcs.lonpole = 180.0  # Native longitude of the Celestial pole
+        w.wcs.latpole = 0.0  # Native latitude of the Celestial pole
+
+        return w
+
+    def get_radec_image(self, alignments, slits, wcs, flexure=None):
```
profxj, line 783: Can we put this in `slittrace.py` ?
I see nothing KCWI specific here.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -607,51 +617,251 @@ def bpm(self, filename, det, shape=None, msbias=None):
...
+            minmax[:, 0] = np.min(evalpos)
+            minmax[:, 1] = np.max(evalpos)
+            slitID = np.ones(evalpos.size) * slit_idx - 12.0
+            world_ra, world_dec, _ = wcs.wcs_pix2world(slitID, evalpos, onslit_init[0], 0)
+            # Set the RA first and DEC next
+            raimg[onslit] = world_ra.copy()
+            decimg[onslit] = world_dec.copy()
+        return raimg, decimg, minmax
+
+    def get_dar_params(self, hdr):
```
profxj, line 828: I should prefer we slurp these in as part of `meta`.  Can that be done?

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: This was lazy of me... should have realised that was a better approach - fixed throughout!

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: I'm not really sure what to do with this... because, it should probably be part of the DetectorContainer (`platescale` is there already). I could add another parameter to the DetectorContainer (`slicescale`?).

pypeit/spec2dobj.py
profxj: I actually hadn't thought about this.
@kbwestfall -- what do you think?

kbwestfall: No, I would just increment the version specifically for the datamodel in question.  I don't see a significant reason to tie the datamodel to the pypeit version.  Actually, the main reason they're different, I guess, is so that we can have fine tuned control on the viable version of the datamodel.  I.e., we want old `spec2dobj` files to fault when they're read with the new code, but not the `arcimage` masters, etc.

rcooke-ast: OK - I'll change this to `1.0.1` when I push me next set of commits. That makes sense to me in hindsight.

pypeit/reduce.py
```diff
@@ -1195,128 +1198,188 @@ def find_objects_pypeline(self, image, std_trace=None,
...
+        model_ivar = self.sciImg.ivar.copy()
+        for nn in range(numiter):
+            msgs.info("Performing iterative joint sky subtraction - ITERATION {0:d}/{1:d}".format(nn+1, numiter))
             self.global_sky[thismask] \
-                = skysub.global_skysub(self.sciImg.image, self.sciImg.ivar, tilt_wave,
+                = skysub.global_skysub(self.sciImg.image, model_ivar, tilt_wave,
                                        thismask, self.slits_left, self.slits_right, inmask=inmask,
                                        sigrej=sigrej, trim_edg=trim_edg,
-                                       bsp=self.par['reduce']['skysub']['bspline_spacing'],
+                                       bsp=self.par['reduce']['skysub']['bspline_spacing'] / np.sqrt(nslits),
```
jhennawi, line 1375: I think you are adding this factor here because you are better sampled given the number of slits. I'm not sure this is the correct thing to do and I worry it results in overfitting in addition to slowing things down. Even for very long slits, we usually set the bspline spacing to like 0.7 pixels or so. Unless you are super agressively spectrally sampled, I'm not sure this is doing anything, and it can lead to artifacts. Anyway, you can easily inspect the fits and see if this helps with the debug option.

pypeit/scripts/coadd_datacube.py
```diff
@@ -0,0 +1,412 @@
...
+    coord_min = [np.min(all_ra), np.min(all_dec), np.min(all_wave)]
+    coord_dlt = [dspat, dspat, dwv]
+    masterwcs = generate_masterWCS(coord_min, coord_dlt)
+
+    # Generate the output binning
+    if combine:
+        numra = int((np.max(all_ra)-np.min(all_ra)) * cosdec / dspat)
+        numdec = int((np.max(all_dec)-np.min(all_dec))/dspat)
+        numwav = int((np.max(all_wave)-np.min(all_wave))/dwv)
+        xbins = np.arange(1+numra)-0.5
```
jhennawi, line 371: Can we name zbins spec_bins or something to make this more clear?

pypeit/scripts/coadd_datacube.py
```diff
@@ -0,0 +1,412 @@
...
+    # Make the cube
+    msgs.info("Generating datacube")
+    # TODO :: Need to include variance weights and make a variance cube
+    if combine:
+        pix_coord = masterwcs.wcs_world2pix(all_ra, all_dec, all_wave * 1.0E-10, 0)
+        hdr = masterwcs.to_header()
+    else:
+        pix_coord = wcs.wcs_world2pix(np.vstack((all_ra, all_dec, all_wave*1.0E-10)).T, 0)
+        hdr = wcs.to_header()
+    datacube, edges = np.histogramdd(pix_coord, bins=(xbins, ybins, zbins), weights=all_sci*all_ivar)
```
jhennawi, line 390: You are computing an inverse variance weighted datacube here. While that is formally optimal, In my experience co-adding/extracting images with noisy individual pixel weights is rarely a good idea.  For co-adding images, I've found it is usually preferable to work with a single weight per exposure, or a smoothly varying weight function (in the spectral direction), the describes the S/N ratio as a function of wavelength.  It is always better to use a model for the inverse variance, which you don't have here, since you don't have a model of your objects and have only a model of the sky. So I think ivar must be that computed from the image which is noisy. This will also be biased in the low-count regime and I think produce coadded cubes with excess noise.  I think it might be useful to compute a "vanilla" coadded cube without these noisy weights, i.e. using individual exposure weights, or using weights that are smoother than the ivar image.

jhennawi: Hi @rcooke-ast the reason I adopt a smoothly varying S/N ratio as a function of wavelength is to deal exactly with this non-overlapping issue, i.e. if one were to somehow combine two setups with different wavelength coverage. The problem is you need to extract an object to get a smoothly varying S/N ratio. 

The problem with what you are doing in my opinion is that suppose that the transparency changes but the sky stays the same, which can often be the case. Then you are going to weight to exposures the same, whereas you should be waiting the one with higher transparency more. 

Implementing correct S/N ratio weights, and also smoothly varying wavelength weights requires an estimate of the S/N for the same reference object on each image. This sounds challenging before you have the full cube. One approach could be to do a preliminary stack, perhaps uniform weigted, or with approximate single exposure weights, and then create the cube, use object finding to locate a bright object in the preliminary cube, estimate S/N for that object in each exposure, and then recompute the cube. That is in my opinion the "optimal" way to do it.

rcooke-ast:
Tests pass

image

and so does the dev suite

image

The one failure is `NIRES_QL`, I did a `./build_nires_masters` and this fixed the failure (`pypeit_test develop -i keck_nires`):

image

PR #991, opened 2020-08-10
Data type bug fix in `show_2dspec`

kbwestfall:
This fixes a bug that crept in via the change in the data types for the spec2d files when displaying them in ginga.


kbwestfall:
Tests pass
```
======================= 235 passed in 747.99s (0:12:27) ========================
```

PR #996, opened 2020-08-11
Tests cleanup files written to disk

finagle29:
Running `pytest` creates a lot of files that then have to be manually added to the `.gitignore` or reverted to clean up the git repository. This PR makes each test function cleanup after itself (with the exception of tests that create intermediate files required by subsequent functions in the same test module and the `spec1d_b27-J1217p3905_KASTb_2015May20T045733.560.fits` problem).

Let me know if any of the tests' output files are in fact necessary for determining the correctness of PypeIt after test runs and I can whitelist them from this PR.


finagle29:
`============================ 235 passed, 748 warnings in 644.31s (0:10:44) ===========================`

PR #1010, opened 2020-08-25
Add P200/TripleSpec

feigewang:
As titled, P200/TripleSpec is now in PypeIt. dev suite passed.

Screen Shot 2020-08-25 at 2 06 32 PM


pypeit/core/arc.py
```diff
@@ -906,7 +906,9 @@ def detect_lines(censpec, sigdetect=5.0, fwhm=4.0, fit_frac_fwhm=1.25, input_thr
     xrng = np.arange(detns.size, dtype=np.float)
 
     if cont_subtract:
-        cont_now, cont_mask = iter_continuum(detns, inmask=None if bpm is None else np.invert(bpm),
+        # FW: for trimmed orders (i.e. set spec_max or spec_min in spectrograph file), you will have lot zeros.
+        # so I mask those zeros if bpm is not given.
+        cont_now, cont_mask = iter_continuum(detns, inmask= (detns!=0) if bpm is None else np.invert(bpm),
```
profxj, line 911: Seems reasonable to me, but @jhennawi will need to comment on whether this precipitates running the full DevSuite.

pypeit/spectrographs/p200_tspec.py
```diff
@@ -0,0 +1,273 @@
...
+            platescale      = 0.37,
+            darkcurr        = 0.085,
+            saturation      = 28000,
+            nonlinear       = 0.9,
+            mincounts       = -1e10,
+            numamplifiers   = 1,
+            gain            = np.atleast_1d(3.8),
+            ronoise         = np.atleast_1d(3.5),
+            datasec         = np.atleast_1d('[:,:]'),
+            oscansec        = np.atleast_1d('[:,:]')  # Is this a hack??
```
profxj, line 83: You can remove the hack comment.

pypeit/spectrographs/p200_tspec.py
```diff
@@ -0,0 +1,273 @@
...
+        detector = detector_container.DetectorContainer(**detector_dict)
+        return detector
+
+    @property
+    def pypeline(self):
+        return 'Echelle'
+
+    def default_pypeit_par(self):
+        """
+        Set default parameters for Shane Kast Blue reductions.
```
profxj, line 94: Modify `Shane Kast Blue`

pypeit/spectrographs/p200_tspec.py
```diff
@@ -0,0 +1,273 @@
...
+        par['rdx']['spectrograph'] = 'p200_tspec'
+        # Wavelengths
+        # 1D wavelength solution
+        par['calibrations']['wavelengths']['rms_threshold'] = 0.3
+        par['calibrations']['wavelengths']['sigdetect']=5.0
+        par['calibrations']['wavelengths']['fwhm']= 5.0
+        par['calibrations']['wavelengths']['n_final']= [3,4,4,4,4]
+        par['calibrations']['wavelengths']['lamps'] = ['OH_NIRES']
+        par['calibrations']['wavelengths']['method'] = 'reidentify'
+        #par['calibrations']['wavelengths']['method'] = 'holy-grail'
```
profxj, line 106: remove

pypeit/spectrographs/p200_tspec.py
```diff
@@ -0,0 +1,273 @@
...
+        -------
+        bpix : ndarray
+          0 = ok; 1 = Mask
+
+        """
+        msgs.info("Custom bad pixel mask for TSPEC")
+        bpm_img = self.empty_bpm(filename, det, shape=shape)
+
+        # Fill in bad pixels if a master bias frame is provided
+        if msbias is not None:
```
profxj, line 224: Are you sure you want this?  Are there even `bias` frames ever taken?

pypeit/spectrographs/p200_tspec.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    Child to handle Keck/NIRES specific code
+    """
+    ndet = 1
+
+    def __init__(self):
+        # Get it started
+        super(P200TSPECSpectrograph, self).__init__()
+        self.spectrograph = 'p200_tspec'
+        self.telescope = telescopes.KeckTelescopePar()
```
finagle29, line 26: I don't know if `self.telescope` is used for much, but #990 added a `P200TelescopePar` for what it's worth

feigewang: Good catch. Thanks for that.

pypeit/core/arc.py
jhennawi: I would prefer it if you just used the bpm option here in the routine that calls detect_lines. This is a bug that in principle impacts many spectrographs and as such, I think that necessitates a more transparent implementation. I think we never had a problem because it just set the continuum to zero, but we need to be a bit more careful here.

pypeit/core/wavecal/templates.py
```diff
@@ -810,6 +813,35 @@ def main(flg):
...
+    if flg & (2**34):
+        reid_path = os.path.join(resource_filename('pypeit', 'data'), 'arc_lines', 'reid_arxiv')
+        iroot = 'p200_triplespec_MasterWaveCalib.fits'
+        iout = 'p200_triplespec.fits'
+        # Load
+        old_file = os.path.join(reid_path, iroot)
+        par = fits.open(old_file)
+        pyp_spec = par[0].header['PYP_SPEC']
+        spectrograph  = load_spectrograph(pyp_spec)
+        orders = spectrograph.orders
```
jhennawi, line 826: Not sure why you don't just add the orders to the file that you are reading in or hard wire them here. It seems that this is technically safer than grabbing them from spectrograph which can change. Anyway, it does not matter since it is fixed format. I can live with this.

pypeit/spectrographs/p200_tspec.py
```diff
@@ -0,0 +1,269 @@
...
+            datasec         = np.atleast_1d('[:,:]'),
+            oscansec        = np.atleast_1d('[:,:]')
+            )
+        detector = detector_container.DetectorContainer(**detector_dict)
+        return detector
+
+    @property
+    def pypeline(self):
+        return 'Echelle'
+
```
jhennawi, line 91: If static, i.e. if this does not depend on self, then we have been making them @classmethod, see other spectrograph file examples.

pypeit/spectrographs/p200_tspec.py
```diff
@@ -0,0 +1,269 @@
...
+            binning (optional):
+
+        Returns:
+            np.ndarray:
+
+        """
+        norders = order_vec.size
+        return np.full(norders, 0.15)
+
+    @property
```
jhennawi, line 259: We got rid of these @propertys. See e.g. gemini_gnirs spectrograph file.

jhennawi (changes requested):
Minor change requested. I think running the whole dev suite is not necessary, but the shane_kast_blue and the fixed format echelle's should be run. You should also run the unit tests. I can help you with running the tests.

pypeit/spectrographs/p200_tspec.py
feigewang: What do you mean we got rid of these properties. I can see this property in GNIRS and NIRES

feigewang:
@jhennawi It would be great if you can help to run the test for other spectrograph. I am going to run the unit tests.

pypeit/core/arc.py
feigewang: @jhennawi  Ok. I just reversed it. It seems this is no longer a problem if I do not use the holy grail method for wavelength. Since I have already archived the wavelength solution. I suppose we do not need to change the inmask anymore.

feigewang:
unit test passed.

Screen Shot 2020-08-25 at 3 42 30 PM

feigewang:
> Minor change requested. I think running the whole dev suite is not necessary, but the shane_kast_blue and the fixed format echelle's should be run. You should also run the unit tests. I can help you with running the tests.

Okay. I was able to run the test on my desktop. Shane_kast_blue and gemini_gnirs fully passed, the quick look for NIRES failed because I did not set NIRES_MASTER variable on my computer. unit tests also passed. So @jhennawi please go ahead to merge once you approve the changes.

jhennawi (approved):

profxj:
merging

PR #1023, opened 2020-09-07
Hot fix for IR redux bug when only negative objects are identified. #1021

jhennawi:
This one line PR fixes a bug in IR redux. Already tested it on NIRES. Other tests should not be necessary.


profxj (approved):
Hot fixes normally should go to `master`, but `develop` is 
just about to go to `master`.

PR #1026, opened 2020-09-09
Fix installation from sdist

finagle29:
This PR includes the ``pypeit/data`` dir in sdists, which are used to build/install PypeIt from pip on non macOS py3.7 platforms.


kbwestfall (approved):
Looks good.  Thanks!

profxj (approved):
Thanks!

PR #1027, opened 2020-09-10
Flexure and ref frame applied to wavelength image

rcooke-ast:
This PR uses the `global_skysub` sky image to calculate the flexure of an entire slit. This avoids an object-by-object flexure correction, and makes sure the saved wavelength image is close to the extracted object wavelengths.


pypeit/par/pypeitpar.py
```diff
@@ -686,13 +686,9 @@ def __init__(self, spec_method=None, spec_maxshift=None, spectrum=None):
...
         # Fill out parameter specifications.  Only the values that are
         # *not* None (i.e., the ones that are defined) need to be set
-        defaults['spec_method'] = 'skip'
-        options['spec_method'] = FlexurePar.valid_methods()
-        dtypes['spec_method'] = str
-        descr['spec_method'] = 'Method used to correct for flexure. Use skip for no correction.  If ' \
-                          'slitcen is used, the flexure correction is performed before the ' \
-                          'extraction of objects (not recommended).  ' \
-                          'Options are: None, {0}'.format(', '.join(options['spec_method']))
+        defaults['perform'] = False
```
profxj, line 689: Can we call this `spec_perform`? To be clear it is for spectral flexure.

pypeit/specobj.py
```diff
@@ -467,29 +467,6 @@ def apply_flux_calib(self, wave_sens, sensfunc, exptime, telluric=None, extinct_
             self[attr+'_FLAM_SIG'] = flam_sig
             self[attr+'_FLAM_IVAR'] = flam_var
 
-
```
profxj, line 470: I'd prefer we keep this around.  We may find better performance for some instruments on an object-by-object flexure correction.

profxj (changes requested):
Thanks for this!  My main request is that we keep the `spec_flexure_obj` 
approach as well.  

And I'll even argue that is the better option for most cases
when an object is extracted (if there is none in the slit, then this new option 
is the only choice).  The main issue I can think of is atmospheric dispersion.
This can make the trace of an object notably different from the slit profile.
Or at least measurably different.  :)

profxj:
One other bit:  Are we still recording `FLEX_SHIFT` to the `SpecObj`?
Our users demand it!  :)

Same goes for the helio-correction.

pypeit/specobj.py
```diff
@@ -467,29 +467,6 @@ def apply_flux_calib(self, wave_sens, sensfunc, exptime, telluric=None, extinct_
...
-        """
-        # Apply
-        for attr in ['BOX', 'OPT']:
-            if self[attr+'_WAVE'] is not None:
-                msgs.info('Applying {0} correction to '.format(refframe)
-                          + '{0} extraction for object:'.format(attr)
-                          + msgs.newline() + "{0}".format(str(self.NAME)))
-                self[attr+'_WAVE'] *= vel_corr
-                # Record
-                self['VEL_TYPE'] = refframe
```
profxj, line 490: Confirm these are being set, i.e. propagated to the header.

profxj: Same with `FLEX_SHIFT` which was set in the `flexure_interp()` method which may not be called any longer.

profxj (commented):
two more comments.

pypeit/core/flexure.py
```diff
@@ -302,134 +304,124 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
+        slit_spat = slits.spat_id[islit]
+        mask = (slitmask == slit_spat) & gpm
 
-        flex_list.append(flex_dict.copy())
+        # Fill in the boxcar extraction tags
+        box_denom = moment1d(waveimg * mask > 0.0, traces[:, islit], 2, row=spec_vec)[0]
+        wghts = (box_denom + (box_denom == 0.0))
+        sky_flux = moment1d(skyimg * mask, traces[:, islit], 2, row=spec_vec)[0] / wghts
+        # Denom is computed in case the trace goes off the edge of the image
+        sky_wave = moment1d(waveimg * mask, traces[:, islit], 2, row=spec_vec)[0] / wghts
```
jhennawi, line 397: Let's chat. I don't see why you are extracting a sky spectrum here. As I said in Slack, you have a bspline model for the sky as a function of the spectral pixel position, so you can just evaluate that model on a grid of spectral points, and this can be what you cross-correlate with the archival sky. There is no reason to extract a sky spectrum from the skyimg model.

pypeit/par/pypeitpar.py
jhennawi: Agreed.

jhennawi (changes requested):
Nice work on this @rcooke-ast. Let's discuss your algorithm. You already have a b-spline fit to the global sky so there is no reason to boxcar extract it. This just wastes computation time, whereas you could just evaluate the bspline model. Then you just need to evaluate that same bspline model at the shifted location instead of interpolating. Let's discuss a more efficient implementation.

rcooke-ast (commented):
Comments addressed. Note, I haven't stored the "global" shifts to the specobjs yet... can't think of a sensible/clean way to do this...

rcooke-ast:
OK - I came up with a way to store the total (global) shifts in the meta of the specobjs, and I've confirmed that it's storing these correctly.

rcooke-ast:
All tests pass... running dev suite now...
image

rcooke-ast:
Dev suite passes:
image

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
                 corr_cen=corr.size/2, smooth=smooth_sig_pix, success=success)
 
 
-def spec_flexure_obj(specobjs, slitord, bpm, method, sky_file, mxshft=None):
-    """Correct wavelengths for flexure, object by object
+def spec_flexure_slit(slits, slitord, slitmask, bpm, sky_file, method="boxcar", specobjs=None, slitspecs=None,
+                      waveimg=None, mxshft=None, slitShift=None):
```
jhennawi, line 309: Please add some docs here explaining the boxcar vs slitcen functionality, as the information below is not enough to understand what the difference is.

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
-          'boxcar' -- Recommneded
-          'slitpix' --
+        method (:obj:`str`, optional)
+          'boxcar' -- Recommended for object extractions
+          'slitcen' -- Recommended when no objects are being extracted
         sky_file (str):
             Sky file
+        specobjs (:class:`pypeit.specobjs.Specobjs`_, optional):
+            Spectral extractions
+        slitspecs (list, optional):
```
jhennawi, line 328: line tools dependencies should not be in this routine. This may be left over from the past, but we agreed to get rid of these.

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
+        slitShift (list, optional):
+            A list of spectral shifts (in pixels) that have been globally applied
+            to each slit and the waveimg. This list should have the same size as
+            the number of slits.
 
     Returns:
-        list:  list of dicts containing flexure results
-            Aligned with specobjs
-            Filled with a basically empty dict if the slit is skipped or there is no object
+        tuple: Two element tuple containing (1) the input waveimg (`numpy.ndarray`_),
```
jhennawi, line 343: Add that if waveimg is not passed in then the first variable is None.

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
+               results of each slit. The latter is filled with a basically empty dict if
+               the slit is skipped
     """
     sv_fdict = None
     msgs.work("Consider doing 2 passes in flexure as in LowRedux")
+    waveimg_flex_corr = None if waveimg is None else waveimg.copy()
+
+    # Determine the method
+    slit_cen = False
+    if specobjs is None: slit_cen = True
```
jhennawi, line 354: Replace with the more concise: 

slit_cen = True if specobjs is None else False

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
                 corr_cen=corr.size/2, smooth=smooth_sig_pix, success=success)
```
jhennawi, line 307: Can we change this bpm to slit_bpm or something like that. Usually bpm in PypeIt is reserver for the actual bpm image.

pypeit/core/flexure.py
jhennawi: Also add to the documentation that this slit specs contain sky spectra extracted for each slit.

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
+        slitspecs (list, optional):
+            A list of linetools.xspectrum1d, one for each slit
+        waveimg (`numpy.ndarray`_, optional):
+            A 2D array providing the wavelength (in A) of each pixel of an image.
+            If none, the spectral flexure correction will not be applied to the
+            wavelength image. If provided, the correction will be applied and
+            returned.
         mxshft (int, optional):
             Passed to flex_shift()
+        slitShift (list, optional):
```
jhennawi, line 337: I'm not sure I understand the reason for this parameter, or what it is used for. Can you elaborate?

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
-                    fdict = copy.deepcopy(sv_fdict)
+            fdict = spec_flex_shift(slitspecs[islit], sky_spectrum, sky_lines, mxshft=mxshft)
+            # Failed?
+            if fdict is not None:
+                # Update dict
+                for key in ['polyfit', 'shift', 'subpix', 'corr', 'corr_cen', 'smooth', 'arx_spec']:
+                    flex_dict[key].append(fdict[key])
+                # Interpolate
+                sky_wave_new = specobj.SpecObj.flexure_interp(fdict['shift'], sky_wave)
+                flex_dict['sky_spec'].append(
```
jhennawi, line 398: If anything this should be operating on specobjs objects and not a line tools object. It seems silly to me that we are calling a method in specobj and then appending stuff to an Xspectrum1d. Only one object is needed here.

pypeit/specobj.py
```diff
@@ -350,39 +350,67 @@ def copy(self):
...
+        Returns:
+            `numpy.ndarray`_: Wavelength scale corrected for spectral flexure
+
+        """
+        npix = wave.size
+        x = np.linspace(0., 1., npix)
+        f = interpolate.interp1d(x, wave, bounds_error=False, fill_value="extrapolate")
+        twave = f(x + shift / (npix - 1))
+        # If this shift is to be applied onto another wave scale, reinterpolate
+        if wavein is not None:
```
jhennawi, line 374: I thought we agreed to use the tilts image for this and not to use interp1d as you are doing it here.

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
+                    xspectrum1d.XSpectrum1D.from_tuple((sky_wave_new, sky_flux)))
+                # Now correct the waveimg
+                if waveimg_flex_corr is not None:
```
jhennawi, line 401: Waveimg should not be corrected via a method in Specobjs, see my comment below. I also don't agree with how you are interpolating the waveimg.

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
+
+                # Interpolate
+                new_sky = sobj.apply_spectral_flexure(fdict)
+                if slitShift is not None:
+                    sobj.update_flex_shift(slitShift[islit])
+                # Update dict
+                for key in ['polyfit', 'shift', 'subpix', 'corr', 'corr_cen', 'smooth', 'arx_spec']:
+                    flex_dict[key].append(fdict[key])
+                flex_dict['sky_spec'].append(new_sky)
+
```
jhennawi, line 451: It looks like all the code below is defunct, since it just throws an error. @profxj do you want to comment.

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
-    # Loop over slits, and then over objects here
+    # Loop over slits, and then over objects
     for islit in range(nslits):
-        i_slitord = slitord[islit]
-        msgs.info("Working on flexure in slit (if an object was detected): {:d}".format(islit))
+        msgs.info("Working on spectral flexure of slit: {:d}".format(islit))
 
-        indx = specobjs.slitorder_indices(i_slitord)
-        this_specobjs = specobjs[indx]
         # Reset
```
jhennawi, line 374: Can we add a attribute to this dict indicating the method that was used, i.e. slits vs object?

pypeit/core/flexure.py
```diff
@@ -303,147 +305,226 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
...
-    return flex_list
-
-
-def spec_flexure_qa(specobjs, slitords, bpm, basename, det, flex_list,
-               slit_cen=False, out_dir=None):
+    return waveimg_flex_corr, flex_list
+
+
+def spec_flexure_corrQA(ax, this_flex_dict, cntr, name):
+    # Fit
```
jhennawi, line 482: Can we indicate via the QA somehow which method was used. In other words, how will the QA look if slit vs object methods are used?

pypeit/reduce.py
```diff
@@ -184,6 +186,7 @@ def __init__(self, sciImg, spectrograph, par, caliBrate,
         # SpecObjs object
         self.sobjs_obj = None  # Only object finding but no extraction
         self.sobjs = None  # Final extracted object list with trace corrections applied
+        self.slitShift = np.zeros(self.slits.nslits)  # Global spectral flexure slit shifts (in pixels) that are applied to all slits.
```
jhennawi, line 189: I think camelcase is reserved for object names in PEP.

pypeit/reduce.py
```diff
@@ -390,13 +392,23 @@ def run(self, basename=None, ra=None, dec=None, obstime=None,
...
                 self.sobjs.make_neg_pos() if return_negative else self.sobjs.purge_neg()
 
         else:  # No objects, pass back what we have
+            # Apply a global flexure correction to each slit
+            # provided it's not a standard star
+            if self.par['flexure']['spec_method'] != 'skip' and not self.std_redux:
+                self.spec_flexure_correct(basename+"_global", mode='slit', update_waveimg=True)
             #Could have negative objects but no positive objects so purge them
             if self.ir_redux:
                 self.sobjs_obj.make_neg_pos() if return_negative else self.sobjs_obj.purge_neg()
```
jhennawi, line 414: The line

 if self.par['flexure']['spec_method'] != 'skip' and not self.std_redux:
                 self.spec_flexure_correct(basename+"_global", mode='slit', update_waveimg=True)

Should be moved out of this if self.nobj> 0 else sequence, since it needs to be done either way. As it is coded here it is redundantly replicating the same line of code.

pypeit/reduce.py
```diff
@@ -412,13 +424,17 @@ def run(self, basename=None, ra=None, dec=None, obstime=None,
         # Finish up
         if self.sobjs.nobj == 0:
             msgs.warn('No objects to extract!')
-        else:
+        elif self.par['flexure']['spec_method'] not in ['skip', 'slit_cen']:
```
jhennawi, line 427: I can imagine that someone would want to do both slit method and the object methods, so you should make that a possibility. In particular running the object flexure code after your slit flexure runs would be a good way of checking that whether the object corrections are even necessary.

pypeit/reduce.py
```diff
@@ -412,13 +424,17 @@ def run(self, basename=None, ra=None, dec=None, obstime=None,
...
+            # Apply a refined estimate of the flexure to objects, and then apply reference frame correction to objects
             # TODO -- Should we move these to redux.run()?
             # Flexure correction if this is not a standard star
-            if not self.std_redux:
-                self.spec_flexure_correct(self.sobjs, basename)
-            # Heliocentric
-            self.helio_correct(self.sobjs, ra, dec, obstime)
+            if (not self.std_redux):
```
jhennawi, line 431: Why the added paranthesis here, I think: 

if not self.std_redux 

works just fine.

pypeit/reduce.py
```diff
@@ -412,13 +424,17 @@ def run(self, basename=None, ra=None, dec=None, obstime=None,
...
+                self.spec_flexure_correct(basename, mode='object', specObjs=self.sobjs, update_waveimg=False)
```
jhennawi, line 432: Can we make basename an attribute of self and not pass it here?

pypeit/reduce.py
```diff
@@ -412,13 +424,17 @@ def run(self, basename=None, ra=None, dec=None, obstime=None,
...
+            # Apply a reference frame correction to each object
+            self.refframe_correct(ra, dec, obstime, specObjs=self.sobjs, update_waveimg=False)
+
```
jhennawi, line 435: Can you just consolidate these two calls to reframe_correct into a single call to reframe_correct, that simply decides what to do based on the self.par['flexure']['spec_method']. That would simplify the logic in self.run and 
then we would not need these update_waveimg=True, False toggles, i.e. you can just do the right thing based on the self.par['flexure']['spec_method']

On that note, shouldn't the there be a separate parameter for reframe correction, i.e. we are currently assuming the user wants both objects extracted and the 2-d waveimg hello centric corrected.

pypeit/reduce.py
```diff
@@ -645,60 +660,115 @@ def load_skyregions(self, skymask_init):
...
+            # Initialise
+            slitSpecs = None
+            gd_slits = np.logical_not(self.reduce_bpm)
+            wvimg = self.waveimg.copy() if update_waveimg else None
+            # Perform some checks
+            if mode == "object" and specObjs is None:
+                msgs.warn("No spectral extractions provided for flexure, using slit center instead")
+                mode = "slit"
+            elif mode not in ["object", "slit"]:
+                msgs.warn("mode must be 'object' or 'slit'. Assuming 'slit'.")
```
jhennawi, line 686: I don't think this should be a warning, but rather it should fault if an unrecognized parameter value is input.

pypeit/reduce.py
```diff
@@ -645,60 +660,115 @@ def load_skyregions(self, skymask_init):
...
+                # TODO :: Need to think about spatial flexure - is the appropriate spatial flexure already included in trace_spat via left/right slits?
+                trace_spat = 0.5 * (self.slits_left + self.slits_right)#/(self.slits.nspat-1)
+                trace_spec = np.arange(self.slits.nspec)#/(self.slits.nspec-1)
+                slitSpecs = []
+                for ss in range(self.slits.nslits):
+                    if not gd_slits[ss]:
+                        slitSpecs.append(None)
+                        continue
+                    slit_spat = self.slits.spat_id[ss]
+                    mask = (self.slitmask == slit_spat)
```
jhennawi, line 699: Call this "thismask" to be consistent with other usage in this module?

pypeit/reduce.py
```diff
@@ -645,60 +660,115 @@ def load_skyregions(self, skymask_init):
...
+                    box_denom = moment1d(self.waveimg * mask > 0.0, trace_spat[:, ss], 2, row=trace_spec)[0]
+                    wghts = (box_denom + (box_denom == 0.0))
+                    slit_sky = moment1d(self.global_sky * mask, trace_spat[:, ss], 2, row=trace_spec)[0] / wghts
+                    # Denom is computed in case the trace goes off the edge of the image
+                    slit_wave = moment1d(self.waveimg * mask, trace_spat[:, ss], 2, row=trace_spec)[0] / wghts
+                    slitSpecs.append(xspectrum1d.XSpectrum1D.from_tuple((slit_wave, slit_sky)))
```
jhennawi, line 705: I really dislike adding an XSpectr1d dependency here to this routine.

pypeit/reduce.py
```diff
@@ -645,60 +660,115 @@ def load_skyregions(self, skymask_init):
...
+
+            # Measure flexure
+            waveimg_new, flex_list = flexure.spec_flexure_slit(self.slits, self.slits.slitord_id, self.slitmask,
+                                                               self.reduce_bpm, self.par['flexure']['spectrum'],
+                                                               method=self.par['flexure']['spec_method'],
+                                                               mxshft=self.par['flexure']['spec_maxshift'],
+                                                               specobjs=specObjs, waveimg=wvimg, slitspecs=slitSpecs,
+                                                               slitShift=self.slitShift)
```
jhennawi, line 713: if update_waveimg would be clearer logic to me. I realize they are equivalent, but that would make the code easier to follow.

pypeit/reduce.py
```diff
@@ -645,60 +660,115 @@ def load_skyregions(self, skymask_init):
...
+            if waveimg_new is not None:
+                self.waveimg = waveimg_new.copy()
+                # Store the slit shifts that were applied to each slit
+                # These corrections are later needed so the headers of the specobjs contains the total spectral flexure
+                self.slitShift = np.zeros(self.slits.nslits)
```
jhennawi, line 718: It seems problematic to me that slits that got skipped have a value of 0.0. Maybe they should be -999 or nan or something.

pypeit/specobj.py
```diff
@@ -350,39 +350,67 @@ def copy(self):
         # Return
         return copy.deepcopy(self)
 
-    def flexure_interp(self, sky_wave, fdict):
+    @staticmethod
```
jhennawi, line 353: Given that this is now being used to interpolate wavelength images, I think this actually should just be a utility method somewhere in core/wave.py since in that context it makes little sense to import Specobs and use a static method on specobjs.

pypeit/specobj.py
```diff
@@ -350,39 +350,67 @@ def copy(self):
...
-        x = np.linspace(0., 1., cut_sky.npix)
-        f = interpolate.interp1d(x, cut_sky.wavelength.value, bounds_error=False, fill_value="extrapolate")
-        twave = f(x + fdict['shift'] / (cut_sky.npix - 1)) * units.AA
-        new_sky = xspectrum1d.XSpectrum1D.from_tuple((twave, cut_sky.flux))
-        # Save
-        self.FLEX_SHIFT = fdict['shift']
+        twave = self.flexure_interp(fdict['shift'], fdict['sky_spec'].wavelength.value) * units.AA
+        new_sky = xspectrum1d.XSpectrum1D.from_tuple((twave, fdict['sky_spec'].flux))
+        # Save - since flexure may have been applied/calculated twice, this needs to be additive
+        self.update_flex_shift(fdict['shift'])
```
jhennawi, line 401: One have you created a single line method below? This method is not used anywhere else as far as I can tell and awyway, it is just a single line of code.

pypeit/specobj.py
```diff
@@ -350,39 +350,67 @@ def copy(self):
...
         # Return
         return new_sky
 
+    def update_flex_shift(self, shift):
+        """Store the total spectral flexure shift in pixels
+
```
jhennawi, line 407: I'm not following what we are summing here. Perhaps this is for when we have done both slit and object based flexure shifts, but perhaps we should be storing those shifts individually then as well.

jhennawi (changes requested):
Hi @rcooke-ast, 

Thanks for the hard work on this PR. I can see that it is not trivial undertaking. 

I have many minor comments that I think would clean up the code a bit. The only major comment I have is that the wavelength image interpolation using the flexure should be done via the tilts as we discussed previously, analogous to how spatial flexure is being treated. Currently you are still doing interp1d, which I don't agree with. Shall we discuss over Slack. 

Thanks!
Joe

rcooke-ast (commented):
Thanks @jhennawi for the review. Comments addressed.

pypeit/specobj.py
rcooke-ast: That's right... this is the total shift. If you prefer, we can store a `local` and a `global` spectral flexure?

pypeit/specobj.py
rcooke-ast: It is also used in `reduce.py`... So I thought it made sense to keep it in specobj...

pypeit/reduce.py
jhennawi: Yeah, you are right. I agree. The logic is slightly tricky here, but I don't think it can be simplified.

pypeit/reduce.py
jhennawi: Okay good.

pypeit/reduce.py
jhennawi: Okay I think I agree. For some reason I thought a user might want the wavelength image to be uncorrected but the objects corrected, but I doubt that is a desired usage case.

pypeit/reduce.py
```diff
@@ -645,60 +662,119 @@ def load_skyregions(self, skymask_init):
...
-            # QA
-            flexure.spec_flexure_qa(sobjs, self.slits.slitord_id, self.reduce_bpm, basename, self.det, flex_list,
-                                    out_dir=os.path.join(self.par['rdx']['redux_path'], 'QA'))
+            # Initialise
+            slitSpecs = None
+            gd_slits = np.logical_not(self.reduce_bpm)
+            # Perform some checks
+            if mode == "local" and specObjs is None:
+                msgs.warn("No spectral extractions provided for flexure, using slit center instead")
+                mode = "global"
```
jhennawi, line 684: I think this should crash if the mode is different than specified.

pypeit/reduce.py
```diff
@@ -645,60 +662,119 @@ def load_skyregions(self, skymask_init):
...
+                    mask = (self.slitmask == slit_spat)
+                    box_denom = moment1d(self.waveimg * mask > 0.0, trace_spat[:, ss], 2, row=trace_spec)[0]
+                    wghts = (box_denom + (box_denom == 0.0))
+                    slit_sky = moment1d(self.global_sky * mask, trace_spat[:, ss], 2, row=trace_spec)[0] / wghts
+                    # Denom is computed in case the trace goes off the edge of the image
+                    slit_wave = moment1d(self.waveimg * mask, trace_spat[:, ss], 2, row=trace_spec)[0] / wghts
+                    slitSpecs.append(xspectrum1d.XSpectrum1D.from_tuple((slit_wave, slit_sky)))
+
+            # Measure flexure
+            flex_list = flexure.spec_flexure_slit(self.slits, self.slits.slitord_id, self.reduce_bpm,
```
jhennawi, line 709: I cannot follow the logic here.  If mode == 'local', then what is this line of code doing? I think the code would be much easier to read if it does one thing if mode == 'local' and another thing if mode == 'global'.

pypeit/reduce.py
```diff
@@ -645,60 +662,119 @@ def load_skyregions(self, skymask_init):
...
+                for islit in range(self.slits.nslits):
+                    if (not gd_slits[islit]) or len(flex_list[islit]['shift']) == 0:
+                        continue
+                    self.slitshift[islit] = flex_list[islit]['shift'][0]
+                # Apply the new wavelength solution
+                msgs.info("Regenerating wavelength image")
+                self.waveimg = self.wv_calib.build_waveimg(self.tilts, self.slits,
+                                                           spat_flexure=self.spat_flexure_shift,
+                                                           spec_flexure=self.slitshift)
+
```
jhennawi, line 727: So if mode == 'local' where is the flexure being applied? Is this done internally to the specobjs? I'm struggling with the logical flow here.

pypeit/specobj.py
jhennawi: Yeah, I think storing local, global, and the sum of the flexure is a good thing to do.

pypeit/specobj.py
jhennawi: Okay.

pypeit/wavecalib.py
```diff
@@ -179,10 +179,25 @@ def build_waveimg(self, tilts, slits, spat_flexure=None):
...
         Returns:
             `numpy.ndarray`_: The wavelength image.
         """
+        # Check spectral flexure type
+        if spec_flexure is None: spec_flex = np.zeros(slits.nslits)
+        elif type(spec_flexure) is float: spec_flex = spec_flexure*np.ones(slits.nslits)
+        elif type(spec_flexure) is np.ndarray:
+            spec_flex = spec_flexure.copy()
+            assert(spec_flexure.size == slits.nslits)
+        spec_flex /= (slits.nspec - 1)
```
jhennawi, line 199: This variable should probably be named to tilt_flex to make this more clear.

pypeit/wavecalib.py
```diff
@@ -211,11 +226,11 @@ def build_waveimg(self, tilts, slits, spat_flexure=None):
                 # # TODO: Put this in `SlitTraceSet`?
                 # evaluate solution --
                 image[thismask] = self.wv_fit2d.eval(
-                    tilts[thismask], x2=np.full_like(tilts[thismask], slits.ech_order[islit]))
+                    tilts[thismask] + spec_flex[islit], x2=np.full_like(tilts[thismask], slits.ech_order[islit]))
```
jhennawi, line 229: Did you check the sign here, it can be rather confusing?

pypeit/wavecalib.py
```diff
@@ -179,10 +179,25 @@ def build_waveimg(self, tilts, slits, spat_flexure=None):
...
+                Spectral flexure correction in pixels. If a float,
+                the same spectral flexure correction will be applied
+                to all slits. If a numpy array, the length of the
+                array should be the same as the number of slits. The
+                value of each element is the spectral shift in pixels
+                to be applied to each slit.
 
         Returns:
             `numpy.ndarray`_: The wavelength image.
         """
```
jhennawi, line 192: Why are we not similarly type checking the spatial flexure? The code is not asymmetric.

pypeit/wavecalib.py
```diff
@@ -179,10 +179,25 @@ def build_waveimg(self, tilts, slits, spat_flexure=None):
                 Image holding tilts
             slits (:class:`pypeit.slittrace.SlitTraceSet`):
             spat_flexure (float, optional):
```
jhennawi, line 181: Why are spatial and spectra flexure treated differently in the input arguments and documentation? It should be symmetric.

pypeit/wavecalib.py
```diff
@@ -179,10 +179,25 @@ def build_waveimg(self, tilts, slits, spat_flexure=None):
...
+                value of each element is the spectral shift in pixels
+                to be applied to each slit.
 
         Returns:
             `numpy.ndarray`_: The wavelength image.
         """
+        # Check spectral flexure type
+        if spec_flexure is None: spec_flex = np.zeros(slits.nslits)
+        elif type(spec_flexure) is float: spec_flex = spec_flexure*np.ones(slits.nslits)
+        elif type(spec_flexure) is np.ndarray:
```
jhennawi, line 196: I usually use isinstance for this type of thing. Not sure which is correct.

jhennawi (changes requested):
We are converging. Thanks for the changes @rcooke-ast

pypeit/reduce.py
rcooke-ast: While I appreciate the concern, I don't think this needs to be addressed with this PR - it's a separate issue that needs more thought. That said, it's a fairly localised dependency (it's only really needed in `flexure.spec_flex_shift`). I've added a TODO.

rcooke-ast (commented):
Comments addressed, @jhennawi - thanks!

rcooke-ast:
Tests passing:
image
image

profxj (approved):
all set

profxj:
merging.

PR #1036, opened 2020-09-23
Datacube generation with more optimal weights

rcooke-ast:
Following on from comments by @jhennawi in PR #980, this PR uses a more optimal weighting scheme, based on the wavelength dependent relative S/N ratio of the observations. The script has been checked against KCWI reductions, and the results are visually improved.

The code is also cleaned up a little bit. These changes do not modify any of the reduction code, so I won't post any test results.


profxj (changes requested):
Only a few things for me.  Is there a test of this in the Dev Suite?
I forget.

rcooke-ast (commented):
Thanks @profxj - comments addressed! Note, in this update I've also made changes to coadd2d. The "cooked" unit tests touch the coadd3d code, but not the dev suite. Hope that's not an issue...

profxj (approved):
Giving my thumbs up

pypeit/core/datacube.py
```diff
@@ -0,0 +1,338 @@
...
+    spl_dec = interp1d(wave_grid, dec_grid, kind='cubic')
+
+    # Evaluate the differentials at the input wave_arr
+    ra_diff = spl_ra(wave_arr) - spl_ra(wave_ref)
+    dec_diff = spl_dec(wave_arr) - spl_dec(wave_ref)
+
+    return ra_diff, dec_diff
+
+
+def generate_whiteLightImgs(all_ra, all_dec, all_wave, all_sci, all_wghts, all_idx,
```
jhennawi, line 121: Camel case I think reserved for classes under PEP.

pypeit/io.py
```diff
@@ -478,6 +482,58 @@ def dict_to_hdu(d, name=None, hdr=None, force_to_bintbl=False):
     return fits.BinTableHDU.from_columns(cols, header=_hdr, name=name)
 
 
+def read_spec2d_file(ifile, filetype='coadd2d'):
+    """
+    Read a PypeIt file of type "filetype", akin to a standard PypeIt file.
```
jhennawi, line 487: Please add docs, I'm not following what this does or why it is needed.

pypeit/scripts/coadd_datacube.py
```diff
@@ -387,25 +220,31 @@ def coadd_cube(files, det=1, overwrite=False):
         spec_bins = np.arange(1+numwav)-0.5
     else:
         # TODO :: This is KCWI specific - probably should put this in the spectrograph file, or just delete it.
+        msgs.warn("This routine is Keck/KCWI specific")
         slitlength = int(np.round(np.median(slits.get_slitlengths(initial=True, median=True))))
         xbins = np.arange(1 + 24) - 12.0 - 0.5
```
jhennawi, line 225: Why are these hardwired?

pypeit/core/datacube.py
```diff
@@ -0,0 +1,338 @@
...
+        # Extract the spectrum
+        pix_coord = whitelightWCS.wcs_world2pix(np.vstack((all_ra[ww], all_dec[ww], all_wave[ww] * 1.0E-10)).T, 0)
+        spec, edges = np.histogramdd(pix_coord, bins=bins, weights=all_sci[ww])
+        norm, edges = np.histogramdd(pix_coord, bins=bins)
+        ivar, edges = np.histogramdd(pix_coord, bins=bins, weights=all_ivar[ww])
+        nrmSpec = (norm > 0) / (norm + (norm == 0))
+        spec = (spec*nrmSpec)[0, 0, :]
+        all_spec[:, ff] = spec.copy()
+        all_snr[:, ff] = spec*np.sqrt(ivar[0, 0, :])
+
```
jhennawi, line 311: @rcooke-ast take a look at core.coadd.sn_weights. I think that does something very similar to what you are doing here. That code is very well tested, and ideally we would not be doing this in multiple ways.

jhennawi (changes requested):
Hi @rcooke-ast, just one significant comment on how you are determining the weights, as you appear to have written your own code for this, whereas we of course have been doing this for a while in 1d coadds.

pypeit/core/datacube.py
rcooke-ast: Let's talk about this on Slack

rcooke-ast (commented):
@jhennawi comments addressed

rcooke-ast:
For the record:
image

pypeit/core/coadd.py
```diff
@@ -741,22 +757,39 @@ def sn_weights(waves, fluxes, ivars, masks, sn_smooth_npix, const_weights=False,
...
+            ref_spec = np.argmax(sn2)
+            if verbose:
+                msgs.info(
+                    "The reference spectrum (ref_spec={0:d}) has a typical S/N = {1:.3f}".format(ref_spec, sn2[ref_spec]))
+
+        elif (ref_spec < 0) or (ref_spec >= nstack):
+            msgs.error("Provided reference spectrum (ref_spec={0:d}) is not within range 0 <= ref_spec < {1:d}".format(ref_spec, nstack))
+    elif (ref_spec >= 0) and (ref_spec < nstack):
+        if verbose: msgs.info("Computing relative weights")
+        relative_weights = True
```
jhennawi, line 783: I think you should require relative weights to just be true if the user wants relative weights. That simplifies the logic here, and you can get rid of this last elif.

pypeit/core/coadd.py
```diff
@@ -741,22 +757,39 @@ def sn_weights(waves, fluxes, ivars, masks, sn_smooth_npix, const_weights=False,
...
+    # TODO: Should ivar weights be deprecated??
     if ivar_weights:
         if verbose:
             msgs.info("Using ivar weights for merging orders")
         weights = np.zeros_like(flux_stack) # Should this be zeros_like?
         for iexp in range(nstack):
-            sn_med1 = utils.fast_running_median(ivar_stack[mask_stack[:, iexp],iexp], sn_smooth_npix)
+            sn_med1 = utils.fast_running_median(ivar_stack[mask_stack[:, iexp], iexp], sn_smooth_npix)
             # TODO Change to scipy.interpolate?
             sn_med2 = scipy.interpolate.interp1d(spec_vec[mask_stack[:, iexp]], sn_med1, kind='cubic',
```
jhennawi, line 794: Can you also add your fix to the fill_value here?

pypeit/core/coadd.py
```diff
@@ -766,19 +799,36 @@ def sn_weights(waves, fluxes, ivars, masks, sn_smooth_npix, const_weights=False,
             weights[:, iexp] = sn_conv
     else:
         weights = np.zeros_like(flux_stack)
+        refscale = (sn_val[:, ref_spec] > 0) / (sn_val[:, ref_spec] + (sn_val[:, ref_spec] == 0))
```
jhennawi, line 802: refscale is only needed if relative_weights == True and so it should be computed in the if relative_weights if block above, not here.

pypeit/core/coadd.py
```diff
@@ -766,19 +799,36 @@ def sn_weights(waves, fluxes, ivars, masks, sn_smooth_npix, const_weights=False,
...
         for iexp in range(nstack):
+            if relative_weights:
```
jhennawi, line 804: Please change the control flow here to be 

if ivar_weights: 

elif relative_weights:

else: 
old code which was the default. 

Otherwise the control flow is impossible to follow with all of these if relative weights statements inserted. If this means that you have to do the loop over exposures twice so be it.

pypeit/core/coadd.py
```diff
@@ -766,19 +799,36 @@ def sn_weights(waves, fluxes, ivars, masks, sn_smooth_npix, const_weights=False,
...
+                if iexp == ref_spec:
+                    # Reference spectrum gets a weight of 1
+                    weights[:, iexp] = 1.0
+                    continue
+                else:
+                    # Compute the relative (S/N)^2 and update the mask
+                    sn2[iexp] /= sn2[ref_spec]
+                    mask_stack[:, iexp] *= (mask_stack[:, ref_spec]) | (sn_val[:, ref_spec] != 0)
+                    sn_val[:, iexp] *= refscale
```
jhennawi, line 813: You are dividing a noisy pixel based estimate of the S/N ratio by another noisy (although less noisy since it is the reference spectrum) pixel based estimate of the S/N ratio. I fail to see why dividing the ratio of two noisy quantities would ever produce better results than the quantities themselves.

pypeit/core/coadd.py
```diff
@@ -687,27 +687,42 @@ def interp_spec(wave_new, waves, fluxes, ivars, masks):
...
+            Inverse variance noise vectors for the spectra
+        masks : bool ndarray, shape = (nspec, nexp)
+            Mask for stack of spectra. True=Good, False=Bad.
+        sn_smooth_npix : float
             Number of pixels used for determining smoothly varying S/N ratio weights.
+        const_weights : bool
+            Use a constant weights for each spectrum?
+        ivar_weights : bool
+            Use inverse variance weighted scheme?
+        relative_weights : bool
```
jhennawi, line 715: Maybe make it clear here that relative_weights needs to be True to do relative weighting, but that ref_spec is optional.

pypeit/core/coadd.py
```diff
@@ -687,27 +687,42 @@ def interp_spec(wave_new, waves, fluxes, ivars, masks):
...
+            Perform relative weights? See ref_spec for further details
+        ref_spec : int
+            Index of spectrum (i.e. 0 <= ref_spec < nstack) to be used as a reference.
```
jhennawi, line 718: Clarify that if ref_spec == -1, then the code automatically sets the reference spectrum to be the highest S/N ratio spectrum. Also make it clear this is only used if the relative weights option is True.

pypeit/core/datacube.py
```diff
@@ -0,0 +1,344 @@
...
+        # Calculate the S/N in a given spectral bin
+        flux_stack[:, ff] = spec[0, 0, :] * np.sqrt(normspec)  # Note: sqrt(nrmspec), is because we want the S/N in a _single_ pixel (i.e. not spectral bin)
+        ivar_stack[:, ff] = ivar_spec
+
+    mask_stack = (flux_stack != 0.0) & (ivar_stack != 0.0)
+    # Obtain a wavelength of each pixel
+    wcs_res = whitelightWCS.wcs_pix2world(np.vstack((np.zeros(numwav), np.zeros(numwav), np.arange(numwav))).T, 0)
+    wave_spec = wcs_res[:, 2] * 1.0E10
+    # Identify the reference spectrum (the one with the highest typical S/N ratio)
+    medsnr = np.median(flux_stack*np.sqrt(ivar_stack), axis=0)
```
jhennawi, line 324: This seems redundant to me since the sn_weights code already does it if you don't set ref_spec.

jhennawi (approved):
Thanks @rcooke-ast this looks great.

rcooke-ast (commented):
Forgot to post comments!

rcooke-ast:
Merging

PR #1045, opened 2020-10-05
Telluric Initialization Improvements

finagle29:
This PR does 3 things:
- improves initial guess in `core.coadd.solve_poly_ratio` by first performing a linear fit and using that as the initial guess iff the absolute value of the fit's r value is > 0.8
- adds a parameter `use_huber` to TellFitPar to allow users to disable Huber loss rescaling in the initialization of the object model (n.b. that Huber loss is always used in the fitting of the telluric model)
- adds the currently installed package version to the `check_requirements` error message to aid in debugging


pypeit/core/coadd.py
```diff
@@ -349,6 +349,7 @@ def poly_ratio_fitfunc_chi2(theta, flux_ref, thismask, arg_dict):
     wave_max = arg_dict['wave_max']
     func = arg_dict['func']
     model = arg_dict['model']
+    use_huber = arg_dict['use_huber']
```
jhennawi, line 352: I would prefer that you removed the option to not use the huber loss function, unless you can demonstrate that this option is actually needed.

pypeit/core/coadd.py
```diff
@@ -375,7 +376,8 @@ def poly_ratio_fitfunc_chi2(theta, flux_ref, thismask, arg_dict):
     robust_scale = 2.0
     huber_vec = scipy.special.huber(robust_scale*chi_std, chi_vec)
     loss_function = np.sum(np.square(huber_vec*mask_both))
-    #chi2 = np.sum(np.square(chi_vec))
+    if not use_huber:
```
jhennawi, line 379: Same goes here, eliminate this please.

pypeit/core/coadd.py
```diff
@@ -476,7 +478,7 @@ def median_filt_spec(flux, ivar, mask, med_width):
 def solve_poly_ratio(wave, flux, ivar, flux_ref, ivar_ref, norder, mask = None, mask_ref = None,
```
jhennawi, line 478: Please remove use_huber option unless it is really needed.

pypeit/core/coadd.py
```diff
@@ -549,29 +551,40 @@ def solve_poly_ratio(wave, flux, ivar, flux_ref, ivar_ref, norder, mask = None,
...
-    wave_min = wave.min()
-    wave_max = wave.max()
 
     # Now compute median filtered versions of the spectra which we will actually operate on for the fitting. Note
     # that rejection will however work on the non-filtered spectra.
     med_width = (2.0*np.ceil(median_frac/2.0*nspec) + 1).astype(int)
     flux_med, ivar_med = median_filt_spec(flux, ivar, mask, med_width)
     flux_ref_med, ivar_ref_med = median_filt_spec(flux_ref, ivar_ref, mask_ref, med_width)
 
+    lr = scipy.stats.linregress(wave[mask], scale_fun(flux_ref_med[mask]))
```
jhennawi, line 576: Please add a comment to describe what you are doing here. Also, these lines of code fail if norder = 1, which I think was actually never a use case (since we would use robust_median_ratio for that), but you should maybe put a line at the beginning causing the code to fault if border =1, and also add that to the docs. Maybe also clarify please in the docs that norder=2 is a linear fit.

pypeit/core/coadd.py
```diff
@@ -549,29 +551,40 @@ def solve_poly_ratio(wave, flux, ivar, flux_ref, ivar_ref, norder, mask = None,
...
+    if np.abs(lr.rvalue) > 0.8:
+        leg_slope = lr.slope * (wave_max - wave_min) / 2.0
+        leg_int = lr.intercept + lr.slope * (wave_max + wave_min)/2.0
+        guess = np.append([leg_int, leg_slope], np.zeros(norder-2))
```
jhennawi, line 580: So why cut on the lr.rvalue? If all this is doing is a linear fit, shouldn't we be using the errors here. stats.linregress does not take errors, but there are other linear fitting codes that do.

pypeit/core/coadd.py
```diff
@@ -549,29 +551,40 @@ def solve_poly_ratio(wave, flux, ivar, flux_ref, ivar_ref, norder, mask = None,
...
+
     arg_dict = dict(flux = flux, ivar = ivar, mask = mask,
                     flux_med = flux_med, ivar_med = ivar_med,
                     flux_ref_med = flux_ref_med, ivar_ref_med = ivar_ref_med,
                     ivar_ref = ivar_ref, wave = wave, wave_min = wave_min,
-                    wave_max = wave_max, func = func, model=model, norder = norder, guess = guess, debug=debug)
+                    wave_max = wave_max, func = func, model=model, norder = norder, guess = guess,
+                    use_huber = use_huber, debug=debug)
```
jhennawi, line 587: Remove use_huber please.

pypeit/core/telluric.py
```diff
@@ -885,7 +885,8 @@ def init_star_model(obj_params, iord, wave, flux, ivar, mask, tellmodel):
     # set the flux_ref to be the data here, i.e. flux
     scale, fit_tuple, flux_scale, ivar_scale, outmask = coadd.solve_poly_ratio(
         wave, flam_model, flam_model_ivar, flux, ivar, obj_params['polyorder_vec'][iord],
-        mask=flam_model_mask, mask_ref=mask, func=obj_params['func'], model=obj_params['model'])
+        mask=flam_model_mask, mask_ref=mask, func=obj_params['func'], model=obj_params['model'],
+        use_huber=obj_params['use_huber'])
```
jhennawi, line 889: Again, remove use_huber.

pypeit/core/telluric.py
```diff
@@ -1000,7 +1001,8 @@ def init_poly_model(obj_params, iord, wave, flux, ivar, mask, tellmodel):
     # set the flux_ref to be the data here, i.e. flux
     scale, fit_tuple, flux_scale, ivar_scale, outmask = coadd.solve_poly_ratio(
         wave, tellmodel, tellmodel_ivar, flux, ivar, obj_params['polyorder_vec'][iord],
-        mask=tellmodel_mask, mask_ref=mask, func=obj_params['func'], model=obj_params['model'], scale_max=1e5)
+        mask=tellmodel_mask, mask_ref=mask, func=obj_params['func'], model=obj_params['model'],
+        scale_max=1e5, use_huber=obj_params['use_huber'])
```
jhennawi, line 1005: again no use_huber.

pypeit/core/telluric.py
```diff
@@ -1457,7 +1459,7 @@ def qso_telluric(spec1dfile, telgridfile, pca_file, z_qso, telloutfile, outfile,
 def star_telluric(spec1dfile, telgridfile, telloutfile, outfile, star_type=None, star_mag=None, star_ra=None, star_dec=None,
                   func='legendre', model='exp', polyorder=5, mask_abs_lines=True, delta_coeff_bounds=(-20.0, 20.0),
                   minmax_coeff_bounds=(-5.0, 5.0), only_orders=None, sn_clip=30.0, tol=1e-3, popsize=30, recombination=0.7, polish=True,
-                  disp=False, debug_init=False, debug=False, show=False):
+                  use_huber=True, disp=False, debug_init=False, debug=False, show=False):
```
jhennawi, line 1462: another use_huber.

pypeit/core/telluric.py
```diff
@@ -1543,10 +1545,12 @@ def star_telluric(spec1dfile, telgridfile, telloutfile, outfile, star_type=None,
...
 
-def poly_telluric(spec1dfile, telgridfile, telloutfile, outfile, z_obj=0.0, func='legendre', model='exp', polyorder=3,
-                  fit_wv_min_max=None, mask_lyman_a=True, delta_coeff_bounds=(-20.0, 20.0),
-                  minmax_coeff_bounds=(-5.0, 5.0), only_orders=None, sn_clip=30.0, tol=1e-3, popsize=30, maxiter=3,
-                  recombination=0.7, polish=True, disp=False, debug_init=False, debug=False, show=False):
+def poly_telluric(spec1dfile, telgridfile, telloutfile, outfile, z_obj=0.0, func='legendre',
+                  model='exp', polyorder=3, fit_wv_min_max=None, mask_lyman_a=True,
+                  delta_coeff_bounds=(-20.0, 20.0), minmax_coeff_bounds=(-5.0, 5.0),
+                  only_orders=None, sn_clip=30.0, tol=1e-3, popsize=30, maxiter=3,
+                  recombination=0.7, polish=True, use_huber=True, disp=False, debug_init=False,
```
jhennawi, line 1552: Remove use_huber please.

pypeit/core/telluric.py
```diff
@@ -1574,7 +1578,7 @@ def poly_telluric(spec1dfile, telgridfile, telloutfile, outfile, z_obj=0.0, func
                       polyorder_vec=polyorder_vec, exptime=meta_spec['core']['EXPTIME'],
                       func=func, model=model, sigrej=3.0,
                       output_meta_keys=('airmass', 'polyorder_vec', 'exptime', 'func'),
```
jhennawi, line 1580: another use_huber.

pypeit/par/pypeitpar.py
```diff
@@ -1684,7 +1684,7 @@ def __init__(self, objmodel=None, redshift=None, delta_redshift=None, pca_file=N
                  bounds_norm=None, tell_norm_thresh=None, only_orders=None, pca_lower=None, pca_upper=None,
                  star_type=None, star_mag=None, star_ra=None, star_dec=None, mask_abs_lines=None,
                  func=None, model=None, polyorder=None, fit_wv_min_max=None, mask_lyman_a=None,
-                 delta_coeff_bounds=None, minmax_coeff_bounds=None, tell_grid=None):
+                 delta_coeff_bounds=None, minmax_coeff_bounds=None, use_huber=None, tell_grid=None):
```
jhennawi, line 1687: again.

pypeit/par/pypeitpar.py
```diff
@@ -1790,6 +1790,10 @@ def __init__(self, objmodel=None, redshift=None, delta_redshift=None, pca_file=N
         dtypes['minmax_coeff_bounds'] = list
         descr['minmax_coeff_bounds'] = "Paramters setting the polynomial coefficient bounds for telluric optimization."
 
+        defaults['use_huber'] = True
```
jhennawi, line 1793: I think we should get rid of use_huber.

pypeit/par/pypeitpar.py
```diff
@@ -1812,11 +1816,11 @@ def __init__(self, objmodel=None, redshift=None, delta_redshift=None, pca_file=N
...
         k = numpy.array([*cfg.keys()])
-        parkeys = ['objmodel','redshift', 'delta_redshift', 'pca_file', 'npca', 'bal_wv_min_max', 'bounds_norm',
-                   'tell_norm_thresh', 'only_orders', 'pca_lower', 'pca_upper',
-                   'star_type','star_mag','star_ra','star_dec','mask_abs_lines',
-                   'func','model','polyorder','fit_wv_min_max','mask_lyman_a',
-                   'delta_coeff_bounds','minmax_coeff_bounds','tell_grid']
+        parkeys = ['objmodel','redshift', 'delta_redshift', 'pca_file', 'npca', 'bal_wv_min_max',
+                   'bounds_norm', 'tell_norm_thresh', 'only_orders', 'pca_lower', 'pca_upper',
+                   'star_type', 'star_mag', 'star_ra', 'star_dec', 'mask_abs_lines', 'func',
+                   'model', 'polyorder', 'fit_wv_min_max', 'mask_lyman_a', 'delta_coeff_bounds',
```
jhennawi, line 1822: also here.

pypeit/scripts/tellfit.py
```diff
@@ -175,6 +175,7 @@ def main(args):
                                          mask_abs_lines=par['tellfit']['mask_abs_lines'],
                                          delta_coeff_bounds=par['tellfit']['delta_coeff_bounds'],
                                          minmax_coeff_bounds=par['tellfit']['minmax_coeff_bounds'],
+                                         use_huber=par['tellfit']['use_huber'],
```
jhennawi, line 178: and here and below.

pypeit/scripts/tellfit.py
```diff
@@ -186,6 +187,7 @@ def main(args):
                                          delta_coeff_bounds=par['tellfit']['delta_coeff_bounds'],
                                          minmax_coeff_bounds=par['tellfit']['minmax_coeff_bounds'],
                                          only_orders=par['tellfit']['only_orders'],
+                                         use_huber=par['tellfit']['use_huber'],
```
jhennawi, line 190: use_huber

jhennawi (changes requested):
Thanks for the hard work here @finagle29. Unless you can convince me that the the use_huber=False option is needed, I'd prefer that you remove this. I cannot think of applications where we would want to do this but you can try to convince me otherwise. 

Otherwise, I just have a minor comments about your fix which you could look at. 

Regarding tests and dev suite, we really need to run all the afterburner scripts that touch solve_poly_ratio, but it may be easier to just re-run the entire dev suite if you are setup to do that. If you have problems running it let me know and I can help.  I've been meaning to take the new parallel version for a spin. 

Thanks!
Joe

pypeit/core/coadd.py
finagle29: removed option to not use Huber loss

pypeit/core/coadd.py
finagle29: rm'd

pypeit/core/coadd.py
finagle29: rm'd

pypeit/core/coadd.py
finagle29: rm'd

pypeit/core/telluric.py
finagle29: rm'd

pypeit/core/telluric.py
finagle29: rm'd

pypeit/core/telluric.py
finagle29: rm'd

pypeit/core/telluric.py
finagle29: rm'd

pypeit/core/telluric.py
finagle29: rm'd

pypeit/par/pypeitpar.py
finagle29: rm'd

pypeit/par/pypeitpar.py
finagle29: rm'd

pypeit/par/pypeitpar.py
finagle29: rm'd

pypeit/scripts/tellfit.py
finagle29: rm'd

pypeit/scripts/tellfit.py
finagle29: rm'd

pypeit/core/coadd.py
finagle29: Nice catch!
It looks like in `core.coadd.scale_spec` https://github.com/pypeit/PypeIt/blob/develop/pypeit/core/coadd.py#L1275 we do use norder=1. I have a couple ideas on how to address this.

1. make `norder` intuitive, i.e. `norder=1` is a linear fit
   a. keep `scale_spec` as is (changing behavior)
   b. adjust scale spec (keep its behavior) and add a guard for `norder=1` before my linear guess code
2. keep `norder` as is, add a guard for `norder=1` before my linear guess code.

pypeit/core/coadd.py
finagle29: I'll look into cutting on something like reduced chi^2

finagle29: Okay, I've replaced stats.linregress with np.polyfit, which takes errors, and handle error propagation from the spectrum rescaling appropriately. I found that using a quadratic fit for the guess generally improved many of the cases where a linear guess was worse than a constant guess. I've also moved to cutting on reduced chi^2.

pypeit/core/coadd.py
jhennawi: Let's chat on Slack.

finagle29:
unit tests passing! `================ 242 passed, 723 warnings in 1215.12s (0:20:15) ================`
dev suit running

finagle29:
unit tests and dev suite passing
`================ 250 passed, 732 warnings in 1319.89s (0:21:59) ================`
`--- PYPEIT DEVELOPMENT SUITE PASSED 89/89 TESTS (Masters ignored) ---`

pypeit/core/coadd.py
```diff
@@ -365,16 +365,18 @@ def poly_ratio_fitfunc_chi2(theta, flux_ref, thismask, arg_dict):
...
+    #    chi_vec, np.invert(mask_both), cenfunc='median', stdfunc=utils.nan_mad_std, maxiters=5, sigma=2.0)
+    chi_std = np.std(chi_vec)
     # The Huber loss function smoothly interpolates between being chi^2/2 for standard chi^2 rejection and
     # a linear function of residual in the outlying tails for large residuals. This transition occurs at the
     # value of the first argument, which we have set to be 2.0*chi_std, which is 2-sigma given the modified
     # errors described above from Schlegel's code.
     robust_scale = 2.0
     huber_vec = scipy.special.huber(robust_scale*chi_std, chi_vec)
-    loss_function = np.sum(np.square(huber_vec*mask_both))
+    loss_function = np.sum(huber_vec*mask_both)
```
jhennawi, line 379: Wow good catch, I can't believe I was screwing this up with the np.square.

pypeit/core/coadd.py
```diff
@@ -549,24 +552,46 @@ def solve_poly_ratio(wave, flux, ivar, flux_ref, ivar_ref, norder, mask = None,
     # Determine an initial guess
     ratio = robust_median_ratio(flux, ivar, flux_ref, ivar_ref, mask=mask, mask_ref=mask_ref,
                                 ref_percentile=ref_percentile, max_factor=scale_max)
+    wave_min = wave.min()
+    wave_max = wave.max()
+
```
jhennawi, line 557: Take a look at the changes made in my branch mosfire_oct20. Please test on your problem.

pypeit/core/coadd.py
```diff
@@ -549,24 +552,46 @@ def solve_poly_ratio(wave, flux, ivar, flux_ref, ivar_ref, norder, mask = None,
...
     if 'poly' in model:
         guess = np.append(ratio, np.zeros(norder-1))
+        scale_fun = lambda x: x
```
jhennawi, line 560: See my branch.

pypeit/core/coadd.py
```diff
@@ -549,24 +552,46 @@ def solve_poly_ratio(wave, flux, ivar, flux_ref, ivar_ref, norder, mask = None,
...
     # that rejection will however work on the non-filtered spectra.
     med_width = (2.0*np.ceil(median_frac/2.0*nspec) + 1).astype(int)
     flux_med, ivar_med = median_filt_spec(flux, ivar, mask, med_width)
     flux_ref_med, ivar_ref_med = median_filt_spec(flux_ref, ivar_ref, mask_ref, med_width)
 
+    # Here we compute a linear regression to the scaled flux in order to get a better guess.
+    if norder > 1:
+            # we need scale_mask to mask negative points in the polyfit
+            scale_mask = mask_fun(flux_ref_med)
+
```
jhennawi, line 587: Doing this with robust_fit now, take a look at my branch.

pypeit/core/telluric.py
```diff
@@ -399,7 +399,7 @@ def tellfit_chi2(theta, flux, thismask, arg_dict):
         chi_vec = totalmask * (flux - tell_model*obj_model) * np.sqrt(flux_ivar)
         robust_scale = 2.0
         huber_vec = scipy.special.huber(robust_scale, chi_vec)
-        loss_function = np.sum(np.square(huber_vec * totalmask))
+        loss_function = np.sum(huber_vec * totalmask)
```
jhennawi, line 402: Again great catch!!

pypeit/core/coadd.py
finagle29: I think mosfire_oct20 works great to fix the problems I was having. One weird thing, though, is when I test with real data the fit looks great, but I still get the same error message from `scipy.optimize.minimize`. However, when I look at nearby cuts through parameter space, the loss function looks smooth. I think this may truly be a numerical precision problem then, and we might want to switch to using a different optimization method in `poly_ratio_fitfunc`, like Powell e.g.

pypeit/core/coadd.py
finagle29: your solution in mosfire_oct20 looks good

finagle29:
closing in favor of #1071

PR #1047, opened 2020-10-12
Merges develop into master (1.2.0 tag)

kbwestfall:
As titled.

**Reminder**: Please only approve and let me do the merge.

@profxj , can you check that I merged the requirements.txt file correctly?  `pytest` isn't really a requirement to run the code anyway, but it might be worth figuring out what is causing Travis to fail.


profxj (approved):
Approving.

as for `pytest`, I tried a few things and had no luck.
I fear I have no clue and am hopeful it will be fine
again in a week or two.

deborape (approved):
Approved!

PR #1124, opened 2021-01-12
Dbsp angle case fix

eteq:
I don't really understand what might be different about the input files, but when I tried running PypeIt on a palomar DBSP dataset I have, this change was necessary to get it to run correctly. (it seems that the "D" in "deg" is sometimes capitalized and sometimes not...)


pypeit/spectrographs/p200_dbsp.py
```diff
@@ -95,7 +95,7 @@ def compound_meta(self, headarr: List[fits.Header], meta_key: str):
             return Time(headarr[0]['UTSHUT']).mjd
         elif meta_key == 'dispangle':
             try:
-                return Angle(headarr[0]['ANGLE']).deg
+                return Angle(headarr[0]['ANGLE'].replace('Deg', 'deg')).deg
```
finagle29, line 98: I think that using `headarr[0]['ANGLE'].lower()` might make the code more robust (for example if the DBSP controller writes `"20 Deg 37 Min"` in the `ANGLE` field).

finagle29 (changes requested):
Thanks for finding this bug and contributing a fix for it!

pypeit/spectrographs/p200_dbsp.py
jhennawi: I agree, can someone make that change please, then I'll approve.

finagle29 (approved):

pypeit/spectrographs/p200_dbsp.py
finagle29: done. waiting for CI

jhennawi (approved):
Approved.

pypeit/spectrographs/p200_dbsp.py
eteq: Oops, missed this thread until now, but thanks for taking care of it, @finagle29 !

PR #1128, opened 2021-01-18
Documentation updates

chloe-mt-cheng:
Hi there!  I have suggested some updates to the documentation, both based on our discussion of the Coadd1D noise pattern in the lris channel in the PypeIt Slack (edits to coadd1d.rst), as well as some suggestions that I would have found useful when running PypeIt the first time (edits to cookbook.rst and reduction_tips.rst).  I am unfamiliar with the reStructuredText syntax, so I apologize in advance if any of the markup is incorrect, please feel free to edit.  Also, if any of the information is incorrect, please edit as necessary.  Thanks again for all of your help!


jhennawi (approved):
Many thanks for the help with this!!

PR #1132, opened 2021-01-22
Updated docs to deprecate PySide2.

jhennawi:
I updated the docs to be less agnostic about the choice of PySide2 vs PyQt5. PySide2 instructions are still in there, but my goal was to make it less likely for users to accidentally install both.


jhennawi:
Closing, as changes were put in PR # 1130.

PR #1138, opened 2021-02-08
Make par_outfile a real command line parameter

eteq:
Looks to me like this was just a copy-and-paste error since this really shouldn't ever be a boolean.


jhennawi (approved):
Good catch.

jhennawi (commented):
Good catch, but you need to pull in the updated release into this branch, before we can marge.

jhennawi (changes requested):
Pull in the updated release branch then it will be approved.

kbwestfall (approved):
Looks good!

jhennawi (approved):
Thanks!

PR #1140, opened 2021-02-09
Prevent crash when no slit edges are found

kbwestfall:
A short PR that prevents crashes when no slit edges are found.  This addresses an issue when DEIMOS has detectors with no slits, which leads to an unavoidable crash when attempting to reduce all detectors.  This is a quick-and-dirty fix in the sense that the code now treats these detectors in a similar way to long-slit data by placing "slit edges" at the edges of the detector and then carries on.  At best, this leads to wasted compute time; at worst, it might mean the code crashes in later stages because there is no flux.

Running unit tests now, and then I'll start the dev-suite.

(This also updates `develop` to match the recent `release` hotfix.)


jhennawi:
I am very much against adding in bogus slits, since this is going to break everything downstream. I think edgetrace and the following reduction logic simply need to be fixed to deal with this. 

I don't see this as a particularly urgent issue, since in principle, a workaround already exists, which is that the deimos user simply excludes the detectors with no slits in their pypeit file. 

If you want to go forward with this bogus slit strategy, then I suggest you mask the bogus slits on this detector such that they don't get reduced, and then also add an issue as a reminder to fix this flawed approach in the future.

jhennawi (changes requested):
See my comment.

kbwestfall:
Thanks for reviewing the PR, @jhennawi .  A couple of comments for you:

- There's a subtlety in that I have a vague memory that bounding the detector with slit edges may be needed for some long-slit data.  (Maybe LRIS?  Does anyone else also have a memory of this?)  So the difficulty is that in some cases you want the code to add "bogus" edges because the slit edges actually fall off the detector, and in some cases the detector has no slits, so you'd like it to skip them and move on to the next detector.  I'd be interested to hear any ways of automatically deciding which to do (maybe put a threshold on the flux in the image?).

- Yes, there is a work around, but it's labor intensive when you're trying to run 100s of masks, like we're doing for the ADAPs.  This is meant to ease the hand-holding limitations, at the expense of wasted compute time.  We could and should do better in handling failure modes and reporting them to the user so they know to address them, so this is only a stop-gap solution.

I've tried to address your concern, which I agree with, by adding in a catch that is triggered by a parameter.  By default, the behavior is still as it was in the original posting of the PR.  I.e., it will add "bogus" slit edges and keep going.  At least in the data that Marla passed on, I don't think this causes the code to fault, even though there is no actual flux.  The reason I left this is because of the ambiguity of what to do with long-slits that have edges that fall off the detector.

The added parameter (`sync_bound`) prevents this behavior and basically allows the tracing code to be unsuccessful if no edges are found.  For this, `pypeit.edgetrace.EdgeTraceSet.sync` now returns a boolean that indicates if the tracing was "successful" or not.  For now, this is only False if no edges are found when `sync` is called.  The success flag is then propagated back to `Calibrations` and to the loop over detectors in `PypeIt.reduce_exposure`.  If the calibrations are unsuccessful (only true if the tracing fails, but could also be extended to other calibration steps), the code issues a warning and continues on to the next detector without running `PypeIt.reduce_one`.

Tests pass, and I've restarted the DevSuite.

jhennawi:
Hi @kbwestfall, 

Maybe it is better to chat at the next telecon. But I'm again not enamored with this bogus slits fix. The only reason it is not causing the code to fault is that the bogus slit is getting masked, but I don't think this is a good way to code and it will probably cause other problems that we are not aware of. The fake slit edge when slits fall of the detector is a very different situation and it is very easy to tell which regime you are in. For example, in one case edge trace finds something but cannot sync the edges, whereas in the other edge trace finds nothing. 

It is clear what the right thing to do here is, and I realize you are trying to fix it quickly. On the other hand, I'm not appreciating where the fault is occurring, and why this simply cannot be fixed in the logic in pypeit.py. Namely, I don't agree that crashing is unavoidable when some detectors don't have slits, pypeit.py should just be skipping those detectors based on whether or not edgetrace found slits on that detector. 

Joe

pypeit/calibrations.py
```diff
@@ -744,13 +749,17 @@ def get_tilts(self):
 
         return self.wavetilts
 
-    def run_the_steps(self):
+    def run_the_steps(self, quiet=False):
```
profxj: document what `quiet` does

profxj (approved):
all good

pypeit/par/pypeitpar.py
```diff
@@ -2773,6 +2774,19 @@ def __init__(self, filt_iter=None, sobel_mode=None, edge_thresh=None, follow_spa
                                 '`center_mode` for these edges and place them at the edge of ' \
                                 'the detector (with the relevant shape).'
 
+        defaults['sync_bound'] = True
```
jhennawi: I think the default should be False here as per our discussion. 

Also, I think sync_bound is not the best name for this parameter. If I understand correctly, the *only* thing this governs is whether or not a fake left and right slit edge are added at the detector boundaries when no slits are found, is that correct? I would call it maybe "slits_full_detector" or something like that, and I should add that if this parameter is set edgetrace probably should not even try to find slits and should just return the edges consistent with the detector.

pypeit/pypeit.py
```diff
@@ -467,6 +467,11 @@ def reduce_exposure(self, frames, bg_frames=None, std_outfile=None):
             # These need to be separate to accomodate COADD2D
             self.caliBrate.set_config(frames[0], self.det, self.par['calibrations'])
             self.caliBrate.run_the_steps()
+            if not self.caliBrate.success:
```
jhennawi, line 470: Nice!

jhennawi (changes requested):
Thanks @kbwestfall, a minor change request for the sync_bounds parameter.

pypeit/calibrations.py
kbwestfall: Good catch.  This was left in by mistake.  Removed.

pypeit/pypeit.py
kbwestfall: Thanks!

pypeit/par/pypeitpar.py
kbwestfall: Changed the default to False, and I changed the name of the parameter to `bound_detector`.  I also added some text to the slit-tracing troubleshooting page (doc/slit_tracing.rst).  I'm running tests now and will restart the dev-suite.  I did get a failure in a previous version that I need to look into.  Please look through and provide any edits on the doc text that I added.  Here's what it looks like when it's formatted:
tmpdocs

kbwestfall:
Sorry I dropped this for a while.  Tests still pass:

`======================= 256 passed in 1135.60s (0:18:55) =======================`  

Just a heads up that I had to deal with a few errors in the dev-suite.  There are a couple Gemini GMOS setups where I have added the `bound_detector=True` flag to the cached pypeit files.  I've also set this to be the default for the MMT MMIRS, MMT BlueChannel, and NOT ALFOSC spectrographs (this was already true for Shane Kast, as well).  I think everything should pass now, but I have to restart it.  I'll keep you posted, but hopefully this will be ready to merge by the end of the week.

kbwestfall:
At long last:
```
--- PYPEIT DEVELOPMENT SUITE PASSED 96/96 TESTS (Masters ignored) ---
Testing Started at 2021-03-02T17:18:29.135778
Testing Completed at 2021-03-04T21:37:13.761070
Total Time: 2 days, 4:18:44.625292
```

pypeit/par/pypeitpar.py
jhennawi: Hi Kyle, 

This looks amazing. I would for 2) add an explanation of when this behaviour is required, i.e. "for instruments for which the entire detector is illuiminated so that the slit edges project to be off the detector, or for situations where the detector is windowed (Kast) in such a way that the slit edges are not covered by the windowed detector."

For the warning, make it more clear that the "Users should be sure to add a standard star to the PypeIt file (and have it classified as standard), in which case PypeIt will use the standard star trace as a crutch for object tracing." 

Joe

jhennawi (approved):
Hi @kbwestfall, 

Great job and great docs. I'm approving, but please see my last comment on the docs, and also please do run the dev suite before merging.

kbwestfall:
Thanks for the doc improvements @jhennawi !  I added them in.  I did not edit the code since the dev-suite successfully ran (see the post this morning).  I'll merge as soon as the CI tests pass.

PR #1144, opened 2021-02-20
deBug on Mosfire_oct20 with GMOS data

Tang-SL:
The y axis size of datasec of GMOS data could either be 4224 (Red grism) or 2112 (Blue grism), with the original version, one may get an error such as:
`ValueError: could not broadcast input array from shape (656,1152) into shape (1356,1152)`
This is a kludge solution, but at least solve the above error.


jhennawi (changes requested):
Are these numbers in header cards? Usually they are listed as datsaec and oscansec, etc. It would be best to not hardwire these numbers, if we don't have to. I realize that the previous code (@profxj) has hardwired numbers, but we would prefer not to do that.

Tang-SL (commented):
For 'autoid.py', I added the igood definition that @jhennawi mentioned in the slack, this solves the "not enough useful IDs" issue in wavelength calibration. Sometimes the arc lines are detected, while rejected later because the original wave_soln_arxiv[:,iarxiv] are mostly zeros. The "igood" limits the good solution only.

For the get_rawimage function, I replaced the manually set numbers with the nodpix. When binning is applied, the datasec in header changes, but the nodpix does not, so xbin is needed to bin the nodpix. Nodpix is then used to find the position of row 1 and row 2. GMOS data have 48 unuseful rows, the whole datasec is made up of: unuseful rows + nod position 1 + node position 2 + node position 3. For example, the nodpix of Hamamatsu detector is 1392, thus the whole datasec along the spatial direction is 1392*3 + 48 = 4224 pixels. The number 48 is not written in the header, thus has to be written manually.

ps. I am not sure whether the xbin is along spatial axis or wavelength axis, in my case, both are 2, so always passes tests.

jhennawi (approved):
Thanks, approved!

PR #1162, opened 2021-03-26
pypeit_collate_1d and HISTORY entries

badpandabear:
This is the first pull request for the pypeit_collate_1d tool and for adding HISTORY entries for PypeIt output files.  
Features in this version:
Grouping spectra from spec1d files by ra/dec or by pixel and coadding
Filtering by slit bitmask (requires spec2d file)
Writing metadata to output directory
Creating a an archive directory with files and metadata for KOA
Adds RA/DEC of objects to RA_OBJ and DEC_OBJ in coadded output.
Accepts [coadd1d] parameters in its .collate1d file, or in a separate .coadd1d file with the same prefix as the .collate1d file.

Items planned for future:
fluxing
removing dependence on spec2d files
refactoring code out of scripts/pypeit_collate_1d.py so that they can appear in api docs.
writing signal to noise information to output
writing a QA PDF
adding weather information (although GUIDFWHM has been added to metadata)
Supporting HISTORY for more types of processing/with more  parameters


doc/collate1d.rst
```diff
@@ -0,0 +1,226 @@
+====================
+Collating 1D Spectra
+====================
+
```
jhennawi, line 4: Nice docs!

pypeit/history.py
```diff
@@ -0,0 +1,144 @@
+"""
```
jhennawi, line 1: Very nice!!

pypeit/pypeit.py
```diff
@@ -338,6 +348,7 @@ def reduce_all(self):
             # Loop on unique comb_id
             u_combid = np.unique(self.fitstbl['comb_id'][grp_science])
             for j, comb_id in enumerate(u_combid):
+                #import pdb; pdb.set_trace()
```
jhennawi: Can you remove this. Have you tried 

from IPython import embed; embed()

A much more fun way to debug in my opinion;)

pypeit/spectrographs/util.py
```diff
@@ -38,6 +42,21 @@ def load_spectrograph(spec):
     if spec in classes.keys():
         return classes[spec]()
 
+    # Check if we were given a file, and if so try to read the spectrograph type from its header
+    if os.path.isfile(spec):
```
jhennawi, line 46: Nice, I like this option.

pypeit/scripts/collate_1d.py
```diff
@@ -0,0 +1,823 @@
...
+#https://stackoverflow.com/questions/3853722/python-argparse-how-to-insert-newline-in-the-help-text
+class SmartFormatter(argparse.HelpFormatter):
+
+    def _split_lines(self, text, width):
+        if text.startswith('R|'):
+            return text[2:].splitlines()
+        # this is the RawTextHelpFormatter._split_lines
+        return argparse.HelpFormatter._split_lines(self, text, width)
+
+class KOAArchiveDir():
```
jhennawi: Do we really want to make the name KOA specific? We support non-Keck instruments as well, which this should also work with. We can adopt the KOA framework, but maybe name it something different.

pypeit/scripts/collate_1d.py
```diff
@@ -0,0 +1,823 @@
...
+            distance via an Astropy Angle.
+
+        Returns (bool): True if the SpecObj matches this group,
+            False otherwise.
+        """
+
+        if not self._config_key_match(spec1d_header):
+            return False
+
+        if self.match_type == 'ra/dec':
```
jhennawi, line 386: The default unit should be in arcseconds as we discussed. Maybe I'm missing how it is coded.

pypeit/scripts/collate_1d.py
```diff
@@ -0,0 +1,823 @@
...
+        is compatible with the ones in this SourceObject.
+
+        Args:
+        spec_obj (:obj:`pypeit.specobj.SpecObj`): 
+            The SpecObj to compare with this SourceObject.
+
+        spec1d_header (:obj:`astropy.io.fits.Header`):
+            The header from the spec1d that dontains the SpecObj.
+            
+        tolerance (float OR :obj:`astropy.coordinates.Angle`):
```
jhennawi: My preference for this argument is that it would not be an astropy.coordinates.Angle object but rather either be a float, or a astropy quantity with units of angle, i.e. u.arcsec or u.deg. But I guess that depends on whether users will be calling this function directly. My point is that there are slightly more lines of code required to instantiate an astropy angle object, as compared to just passing in the units of the matching tolerance.

pypeit/scripts/collate_1d.py
```diff
@@ -0,0 +1,823 @@
...
+#
+# See top-level LICENSE file for Copyright information
+#
+# -*- coding: utf-8 -*-
+"""
+This script collates multiple 1d spectra in multiple files by object, 
+runs flux calibration/coadding on them, and produces files suitable
+for KOA archiving.
+"""
+
```
jhennawi, line 11: So far we have had all our scripts import methods and classes from core or other directories, rather than in this module where the module in scripts also has the methods and classes it needs. I actually prefer your approach, but it depends on whether these methods are used elsewhere.

pypeit/scripts/collate_1d.py
```diff
@@ -0,0 +1,823 @@
...
+        spec2d_file = os.path.join(path, filename.replace('spec1d', 'spec2d', 1))
+
+        if not os.path.exists(spec2d_file):
+            msgs.error(f'Could not find matching spec2d file for {spec1d_file}')
+
+        spec2d_files.append(spec2d_file)
+
+    return spec2d_files
+
+def is_float(s):
```
jhennawi: Should this be in pypeit/utils.py?

jhennawi (changes requested):
Thanks for this great PR @badpandabear. My only substantive comment is how collate1d interfaces with pypeit_flux_setup, which does some similar operations. I'm wondering whether we can use some of this nice code that you have developed to improve pypeit_flux_setup and thus have less code duplicating operations.

pypeit/coadd1d.py
```diff
@@ -110,12 +112,18 @@ def to_file(self, ofile, primary_hdr=None, **kwargs):
...
+            subheader = spectrograph.subheader_for_spec(self.head0, self.head0,
+                                                        extra_header_cards = ['RA_OBJ', 'DEC_OBJ'])
         else:
             subheader = {}
         # Add em in
         for key in subheader:
             primary_hdr[key] = subheader[key]
+
+        # Add history
+        if history is not None:
```
profxj, line 124: very convenient!

pypeit/pypeit.py
badpandabear: FIxed. Sorry about that, I have a bad habit of putting these in code and forgetting to remove them. Thanks for finding this.

pypeit/scripts/collate_1d.py
badpandabear: This code is getting the RA/DEC from the SpecObj which is in decimal degrees.  The tolerance does default to arcseconds and is handled in main() at around line 777.

pypeit/scripts/collate_1d.py
badpandabear: I changed it to take a floating point value and an optional "unit" argument that defaults to arcseconds.

pypeit/scripts/collate_1d.py
badpandabear: I was planning to split up pypeit_collate_1d.py in the next pull release. Do you want me to do that in this one?

pypeit/scripts/collate_1d.py
badpandabear: I agree, and changed it to ArchiveDir.

pypeit/scripts/collate_1d.py
badpandabear: Done

pypeit/scripts/collate_1d.py
profxj: I am fine waiting for the next PR

pypeit/scripts/collate_1d.py
```diff
@@ -0,0 +1,817 @@
...
+
+    for spec1d_file in spec1d_files:
+        sobjs = specobjs.SpecObjs.from_fitsfile(spec1d_file)
+
+        for sobj in sobjs.specobjs:
+
+            if sobj.MASKDEF_ID in exclude_map:
+                msgs.info(f'Excluding {sobj.MASKDEF_ID} in {spec1d_file} because of flags {exclude_map[sobj.MASKDEF_ID]}')
+                continue
+            if sobj.OPT_COUNTS is None:
```
profxj: What if the spectrum was only extracted with BOXCAR?
Can we allow for that?

pypeit/scripts/collate_1d.py
```diff
@@ -0,0 +1,817 @@
...
+    coAdd1d = coadd1d.CoAdd1D.get_instance(source.spec1d_file_list,
+                                           [x.NAME for x in source.spec_obj_list],
+                                           spectrograph=spectrograph, par=par['coadd1d'])
+
+    # Run
+    coAdd1d.run()
+    # Save to file
+    coAdd1d.save(source.coaddfile)
+
+def read_file_list(lines, block):
```
profxj, line 507: Probably should go in `io.py`  (eventually)

pypeit/scripts/collate_1d.py
```diff
@@ -0,0 +1,817 @@
...
+
+    # Strip out the coadd list
+    is_config, coadd_lines = read_file_list(lines, 'coadd1d')
+
+    cfg_lines = list(lines[is_config])
+
+    return cfg_lines
+
+
+def find_spec2d_from_spec1d(spec1d_files):
```
profxj, line 606: This is fine, but let me suggest that for the next PR that we
add `spec2d_filename` to the `SpecObj` data model.

Will help us maintain provenance.

pypeit/tests/test_history.py
```diff
@@ -0,0 +1,218 @@
+"""
```
profxj, line 1: Thanks for the unit tests!

profxj (approved):
Great stuff.  Put in a number of comments
but am approving as none of them were 
especially onerous.

pypeit/scripts/collate_1d.py
badpandabear: I added BOX_COUNTS to this check, so that only objects with no counts are excluded. With this change BOXCAR only can be done by changing ex_value in the coadd1d parameters. But it's still not possible to mix BOX and OPT in the same run of collate_pypeit_1d. Would we ever want to do that?

pypeit/scripts/collate_1d.py
jhennawi: Fine with me.

pypeit/scripts/collate_1d.py
profxj: good question.  I'll say no for now. :)

jhennawi:
@badpandabear I'm ready to approve here after develop is merged in and tests are posted. I'd suggest that you probably wait for mosfire_oct20 though, since that is next in the merge order.

badpandabear:
Screenshot_unit_test_deimos_collate_1d
Screenshot_devsuite_deimos_collate_1d
Unit tests and devsuite complete

jhennawi (approved):

profxj:
merging

PR #1185, opened 2021-05-14
Blackbody standards and fixes for KCWI headers

rcooke-ast:
This PR contains a few old, minor edits to make KCWI data work with data taken at the telescope (those files don't contain pressure/humidity/temperature, and use a different keyword for the exposure time).

I also include a bunch of Blackbody standard stars... This is probably overkill to have them as individual files, but I'm open to hearing suggestions of how best to include this into PypeIt.


pypeit/data/standards/blackbody/make_spec.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+from astropy import table
+import astropy.constants as cons
+import astropy.units as u
+
+
+def blackbody_func(a, teff, outname):
```
profxj: add brief doc string, thx

profxj (approved):
All good.  One minor nit from me.

badpandabear (approved):

pypeit/data/standards/blackbody/make_spec.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Args:
+        a (float):
+            flux normalisation factor
+        teff (float):
+            Effective temperature of the blackbody
+        outname (string):
+            The filename to write the spectrum to.
+
+    """
```
jhennawi: Why are you writing these files out to disk and archiving them if you can generate the spectrum on the fly?

jhennawi (changes requested):
Hi @rcooke-ast, 

Thanks for this PR. Two comments: 

1) Can you please provide some information on why this blackbody standards are useful and why we need them in PypeIt, compared to other standard stars like X-shooter and the stuff in calspec?

2) Why are you archiving .fits files for these stars? It appears that you are just generating the standard star spectrum on the fly from a table of magnitudes and temperatures, so why not just generate the standard spectrum on the fly? Perhaps I've missed something. 

Thanks,
Joe

pypeit/data/standards/blackbody/make_spec.py
rcooke-ast: Done - now generating these during runtime.

pypeit/data/standards/blackbody/make_spec.py
rcooke-ast: Done.

jhennawi (approved):
Thanks @rcooke-ast. I'm approving but please run unit tests (but not dev suite tests), before merging.

rcooke-ast:
Tests passed:
image

PR #1189, opened 2021-05-24
update MK extinction data

rcooke-ast:
Update for Mauna Kea extinction data: [Buton et al. 2013](https://ui.adsabs.harvard.edu/abs/2013A%26A...549A...8B/abstract)


jhennawi (changes requested):
Thanks for updating this @rcooke-ast. Can you add a note to the header about telluric lines are being treated here. In other words, are they being interpolated over (which is what I gather) from inspecting the numbers. 

Thanks!

rcooke-ast:
Thanks @jhennawi - A skim read suggests they have a high resolution telluric spectrum, and convolve this with a Gaussian to a resolution of R~1000. So, telluric absorption is included, but at a somewhat low resolution. Seems reasonable to me...

feigewang (commented):
Thanks @rcooke-ast for updating this. Could you please add one comment in the header to describe the uncertainty of the extinction curve. It seems to be 0.001-0.04 mag from 0.3-1.0 um by looking at their table.

jhennawi (approved):
Thanks @rcooke-ast!

rcooke-ast:
Thanks @feigewang - good idea - I've included this now.

feigewang (approved):
Thanks

PR #1192, opened 2021-05-25
Fix local_skysub_extract object grouping logic

finagle29:
Occasionally, PypeIt finds and traces objects, but skips them during extraction. This bug can be traced to the object grouping logic in `local_skysub_extract`, where objects with overlapping extraction masks should be extracted in the same group. In the old code, the blue and green objects would be grouped together, but in the next iteration of the while loop, the code would start grouping again with the red object, skipping over the orange object.
maskwidth_skipped_trace

This PR fixes that by doing the grouping correctly first, and then iterating over the groups to extract them. The grouping is done by considering the objects as vertices in a graph, with an edge between them if their extraction masks overlap, e.g.
the graph for the above has vertices [blue, orange, green, red] and edges between (blue, green) and (green, orange). Then the correct groups for extraction are the connected components of the graph, e.g. [blue, green, orange] and [red] for the above case.


pypeit/core/skysub.py
```diff
@@ -675,29 +675,75 @@ def local_skysub_extract(sciimg, sciivar, tilts, waveimg, global_sky, rn2_img,
     # TODO Can this be simply replaced with spat_img above (but not spat_pix since that could have holes)
     spatial_img = thismask * ximg * (np.outer(xsize, np.ones(nspat)))
```
jhennawi, line 667: Can you add some more comments on what the philosophy behind the graph search and depth-first search, along with what this code is trying to achieve. The previous code was admittedly lacking such documentation, but as many (including yours truly) are not that familiar with graphs.

pypeit/core/skysub.py
```diff
@@ -675,29 +675,75 @@ def local_skysub_extract(sciimg, sciivar, tilts, waveimg, global_sky, rn2_img,
...
-        group = np.array([], dtype=np.int)
-        group = np.append(group, i1)
-        # The default value of maskwidth = 3.0 * FWHM = 7.05 * sigma in objfind with a log(S/N) correction for bright objects
-        min_spat1 = np.maximum(sobjs[i1].TRACE_SPAT - sobjs[i1].maskwidth - 1, slit_left)
-        max_spat1 = np.minimum(sobjs[i1].TRACE_SPAT + sobjs[i1].maskwidth + 1, slit_righ)
-        for i2 in range(i1 + 1, nobj):
-            left_edge = sobjs[i2].TRACE_SPAT - sobjs[i2].maskwidth - 1
-            righ_edge = sobjs[i2].TRACE_SPAT + sobjs[i2].maskwidth + 1
-            touch = (left_edge < max_spat1) & (sobjs[i2].TRACE_SPAT > slit_left) & (righ_edge > min_spat1)
+    # First, we find all groups of objects to local skysubtract together
```
jhennawi, line 668: I suggest that we make from line  679 to 731 a method in the specobjs class, which returns the object groups, i.e. 

groups = sobjs.get_extraction_groups()

Or something like that. The reason why I think it should be a method in specobjs is that it only has to do with the specobjs properties, i.e. the TRACE_SPAT, nobjs, and maskwidth. 

The only slight subtlety is when model_full_slit is set above. This argument is used for narrow slit echelle reductions where the goal is to do simply perform local sky-subtraction on the whole slit grouping all objects together. This is being hacked in above at present by just making the maskwidth large such that everything touches. I can live with that remaining as is for now, but it is admittedly not very elegant.

jhennawi (changes requested):
Thanks for catching this bug and for implementing this elegant fix @finagle29. 

I have one substantive comment, about code rearrangement, and a request for some more documentation info. 

I would have to dig in to the code completely  follow the logic here, but this is mainly because I don't fully understand how  graphs work. 

Since this is altering the guts of the extraction and could have unforeseen consequences, my suggestion is that we run the full dev suite tests and then I can inspect some outputs for various spectrographs to make sure that all looks kosher. 

Another thing that you could look into is whether the old code and the new code produce the same extraction min_spat1 and max_spat1 for various setups in the dev suite for the non-degenerate cases (i.e. where the bug is not present). 

Thanks!
Joe

pypeit/core/skysub.py
finagle29: I put some comments in `SpecObjs.get_extraction_groups` and `utils.DFS`, can you take a look and let me know if that looks sufficient.

pypeit/core/skysub.py
finagle29: The main bit of this code now lives in `SpecObjs.get_extraction_groups`, and I put the DFS routine into `utils` since it could be more widely useful than just here.

For `model_full_slit`, we would want one group with all of the objects, and then the `min_spat` and `max_spat` later just being the slit edges, right?

jhennawi: Yeah that is right, one group with all the objects and  min_spat1  and max_spat1 being the left and right slit edges. I'm a bit confused about what happens if the slit edges run off the detector. 

groups = sobjs.get_extraction_groups(model_full_slit=model_full_slit)

Would be cleaner than the way it is implemented currently. We will need to test all of this though. 

I also realised that we are updating the mask widths but never using them anywhere so I commented that out. Please do a pull.

pypeit/core/skysub.py
```diff
@@ -675,29 +675,23 @@ def local_skysub_extract(sciimg, sciivar, tilts, waveimg, global_sky, rn2_img,
...
-                min_spat1 = np.maximum(np.minimum(left_edge, min_spat1), slit_left)
-                group = np.append(group, i2)
+        left_edges = np.array([sobjs[i].TRACE_SPAT - sobjs[i].maskwidth - 1 for i in group])
+        righ_edges = np.array([sobjs[i].TRACE_SPAT + sobjs[i].maskwidth + 1 for i in group])
+
+        min_spat1 = np.maximum(np.amin(left_edges, axis=0), slit_left)
+        max_spat1 = np.minimum(np.amax(righ_edges, axis=0), slit_righ)
+
         # Create the local mask which defines the pixels that will be updated by local sky subtraction
         min_spat_img = np.outer(min_spat1, np.ones(nspat))
```
jhennawi: While we are here, I noticed that this could be done more efficiently with broadcasting, which means we would not have to multiply.

finagle29: Good catch, I've replaced these (and a couple other instances) with the appropriate slices to use broadcasting. The 2 remaining `np.outer` calls in this file require more digging to convert to use broadcasting, so I'd like to punt on those.

finagle29:
Unit tests passing locally
`=============== 278 passed, 4654 warnings in 3235.95s (0:53:55) ================`
If this looks good to you @jhennawi , then could you @profxj start running the dev suite?

debora-pe (approved):
All looks good to me!

jhennawi (changes requested):
Thanks @finagle29. We are ready to run tests, and then we can merge.

jhennawi:
Screenshot 2021-06-01 at 08 33 55

The one failure is associated with pypeit_multislit_flexure which was prematurely pushed to the dev suite, so we will disregard that.

jhennawi:
Visual inspections look good. Great work @finagle29. Let me run the unit tests since I have the full dev suite.

jhennawi:
Screenshot 2021-06-02 at 15 24 39

jhennawi:
Tests passed. Approving and merging. Great work @finagle29!

jhennawi (approved):
Approved!

PR #1198, opened 2021-06-11
Merges develop into release (1.4.1 tag)

kbwestfall:
As titled. New tag has already been pushed. Anyone can approve and merge.

In prep for this, I updated the docs and fixed a minor issue in `pypeit/test/tstutils.py`.  **Everyone should check that the telluric grids are in the correct directory**: `pypeit/data/telluric/atm_grids`.  The grid files should **not** (also) be one directory up.


badpandabear (approved):

profxj (approved):
approved!

PR #1210, opened 2021-06-17
Updates to the relative spectral illumination

rcooke-ast:
Several minor updates to calculating the relative spectral illumination of slits. See Slack kcwi channel for the results.


pypeit/core/meta.py
```diff
@@ -140,10 +140,10 @@ def define_additional_meta(nlamps=20):
                        'object': dict(dtype=str, comment='Alternative object name (cf. target)'),
                        'obstime': dict(dtype=str, comment='Observation time'),
                        'oscansec': dict(dtype=str, comment='Overscan section (windowing)'),
-                       'pressure': dict(dtype=float, comment='Pressure at obstime'),
+                       'pressure': dict(dtype=units.quantity.Quantity, comment='Pressure at obstime'),
```
jhennawi: I try to avoid using astropy Quantities in production code because they make numpy operations 10-100 times slower. An alternative here would be a float and you specify the units in the comments. As long as you are not multiplying this into an array and making that a quantity though it probably okay as is.

pypeit/flatfield.py
```diff
@@ -1355,10 +1233,9 @@ def show_flats(image_list, wcs_match=True, slits=None):
             clear = False
 
 
-def illum_profile_spectral(rawimg, waveimg, slits, model=None, gpmask=None, skymask=None, trim=3, flexure=None):
+def illum_profile_spectral(rawimg, waveimg, slits, ref_idx=0, smooth_npix=None, model=None, gpmask=None, skymask=None, trim=3, flexure=None):
     """
-    Generate a rough estimate of the relative spectral scaling of slits
-    using a low order polynomial. This routine is for slit-based IFUs.
+    Determine the relative spectral illumination of all slits.
```
jhennawi, line 1238: Can you please make it clear that this is only used for image slicing IFUs as in the previous documentation.

pypeit/par/pypeitpar.py
```diff
@@ -497,7 +497,7 @@ def __init__(self, method=None, pixelflat_file=None, spec_samp_fine=None,
                  spec_samp_coarse=None, spat_samp=None, tweak_slits=None, tweak_slits_thresh=None,
                  tweak_slits_maxfrac=None, rej_sticky=None, slit_trim=None, slit_illum_pad=None,
                  illum_iter=None, illum_rej=None, twod_fit_npoly=None, saturated_slits=None,
-                 slit_illum_relative=None):
+                 slit_illum_relative=None, ref_idx=None):
```
jhennawi: How about renaming ref_idx to  slit_illum_ref_idx to make it more clear that this parameter is associated with that option.

pypeit/par/pypeitpar.py
```diff
@@ -614,6 +614,11 @@ def __init__(self, method=None, pixelflat_file=None, spec_samp_fine=None,
                                    'extracted from the slit; \'continue\' - ignore the ' \
                                    'flat-field correction, but continue with the reduction.'
 
+        defaults['ref_idx'] = 0
+        dtypes['ref_idx'] = int
+        descr['ref_idx'] = 'The index of a reference slit (0-indexed) used for estimating ' \
```
jhennawi: Please specify that this parameter is only needed if slit_illum_relative is True or being performed.

pypeit/reduce.py
```diff
@@ -1362,6 +1362,18 @@ def apply_relative_scale(self, scaleImg):
         # Check that scaleimg is set to the correct shape
         if self.scaleimg.size == 1:
             self.scaleimg = np.ones_like(self.sciImg.image)
+        debug = False
```
jhennawi: Make debug an argument to apply_relative_scale?

pypeit/scripts/coadd_datacube.py
```diff
@@ -143,10 +148,17 @@ def coadd_cube(files, parset, overwrite=False):
         rel_humidity = spec.get_meta_value([spec2DObj.head0], 'humidity')
         coord = SkyCoord(raval, decval, unit=(units.deg, units.deg))
         location = spec.location  # TODO :: spec.location should probably end up in the TelescopePar (spec.telescope.location)
-        ra_corr, dec_corr = dc_utils.dar_correction(waveimg[onslit_gpm], coord, obstime, location,
-                                                    pressure, temperature, rel_humidity, wave_ref=wave_ref)
-        raimg[onslit_gpm] += ra_corr
-        decimg[onslit_gpm] += dec_corr
+        if pressure.value == 0.0:
```
jhennawi: Again, beware of astropy quantities, they are really bad news in terms of performance. Fine as long as you don't accidentally convert an array into a Quantity.

jhennawi (changes requested):
Minor comments @rcooke-ast. The main concern is using astropy.quantities in meta values. I'm not necessarily opposed, but quantities are very bad news in terms of computation and make numpy operations 10-100 times slower, so just make sure you are not multiplying them into images.

pypeit/reduce.py
rcooke-ast: Debug is not needed here - it was just a check to make sure the right operation is being applied, and it is - deleted.

pypeit/par/pypeitpar.py
rcooke-ast: Done!

pypeit/par/pypeitpar.py
rcooke-ast: Done!

pypeit/flatfield.py
rcooke-ast: Hmmm... Prof X and I discussed that this function might be used for correcting the vignetting in DEIMOS. So, I removed this from the docstring. I can take it out, provided this gets removed if it's used for DEIMOS. Given this, let me know what you prefer.

pypeit/core/meta.py
rcooke-ast: OK - this is only used for the DAR correction, so I'm happy to return this to float.

pypeit/scripts/coadd_datacube.py
rcooke-ast: See earlier comment - this is now reverted back to float.

rcooke-ast:
Thanks @jhennawi! All comments addressed except one (about the docstring). Let me know if this is OK, or if you prefer that I change it.

debora-pe (approved):
@rcooke-ast I took a look at this. I must say that I am not super familiar with this part of the code, but it looks good to me. It would be good to see the result of the tests.

rcooke-ast:
Thanks @deborape  - much appreciated! I think I'm the only one who uses this code, so I suspect no one is very familiar with it!!

rcooke-ast:
Tests pass (updated after fixing coadd3d scripts - see commits below):
image

pypeit/core/meta.py
jhennawi: I don't feel strongly about it as long as you are not accidentally casting an array to a quantity. That said, if you don't use the quantity, please specify the units in the comment.

pypeit/flatfield.py
jhennawi: Maybe just write "currently only used for IFUs" and then we can update it.

jhennawi (approved):
@rcooke-ast see my suggestion re the doc-string, but no biggie. Thanks!

PR #1219, opened 2021-07-02
Enable codecov

tepickering:
This PR makes a few minor CI workflow tweaks:

- adds python 3.9 to weekly cron tests
- changes `master` to `release` in list of branches that can trigger workflow when changes are pushed
- enables coverage calculation within the CI tests and uploads results to codecov.io

The existing `tox` configuration already had everything set up to calculate coverage and generate reports.

### DIFF
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python: [3.8]
+        python: [3.8, 3.9]
         toxenv: [test-alldeps, test-numpydev, test-linetoolsdev, test-gingadev, test-astropydev, conda]
     steps:
     - name: Check out repository
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python: [3.8, 3.9]
-        toxenv: [test, test-alldeps, test-linetoolsdev, test-gingadev, test-astropydev, conda]
+        toxenv: [test, test-alldeps-cov, test-linetoolsdev, test-gingadev, test-astropydev, conda]
     steps:
     - name: Check out repository
       uses: actions/checkout@v2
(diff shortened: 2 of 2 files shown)


tepickering:
oops, a test check is "hung" because i renamed it. shall i add it back?

profxj (approved):
Thx!

kbwestfall (approved):

jhennawi (changes requested):
Thanks!

jhennawi (approved):

tepickering:
ope, looks like there's a "c is less efficient" test failure...

kbwestfall:
Hi @tepickering.  A couple of requests:

 - The test failure you're getting is usually fixed if you just re-run the tests, but this test is becoming a problem.  Can you edit the test assertion in `test_bspline.py::test_cholesky_band_versions` to be `assert ctime < 2*pytime, 'C is less efficient!'`?  That should keep this failure from happening, and still alert us if something more dramatic happens.
 - The "Tox env 3.8-test-alldeps" check is stalled.  Is that because you renamed it?  Is it possible to fix or remove this?  Is it something you can do from your end, or do we need @profxj to override something to allow for the merge to continue?

tepickering:
ok, i'll go fix that assertion and maybe take a deeper peek why it sometimes complains. e.g. is the python really python or using optimized library?

the stalled check is due to a renamed test. i added the `-cov` to an existing test. @projxj has manually cleared those before, but i could also just add it back at a slight efficiency cost (running same test case twice).

kbwestfall:
> ok, i'll go fix that assertion and maybe take a deeper peek why it sometimes complains. e.g. is the python really python or using optimized library?

I'm not sure what's happening, and the behavior is annoyingly inconsistent.  I'm wondering if it has to do with the compilation and initial linking of the c library object.

> the stalled check is due to a renamed test. i added the `-cov` to an existing test. @projxj has manually cleared those before, but i could also just add it back at a slight efficiency cost (running same test case twice).

In that case, I think we should have @profxj manually override to clear and merge it.

tepickering:
yeah, compilation is my strong suspicion, but haven't looked at what's actually happening. at any rate, timing tests should always use multiple runs, especially when the execution times are small. if we want to keep those tests around, it might be worth refactoring them to use the `timeit` module: https://docs.python.org/3/library/timeit.html.

kbwestfall:
Good idea on using timeit.  I can implement this in an upcoming PR.  @profxj , I think you should go-ahead with the merge.

PR #1229, opened 2021-07-14
Updates to pypeit_collate_1d for archiving

badpandabear:
This PR contains changes to pypeit_collating_1d for creating a draft archive file for KOA. The changes include:

1. Added spec1d extraction text files and .pypeit files to the archive directory when using --archive_dir
2. Added the spec1d filename, spec1d text file name, and pypeit file name to the metadata.
3. Added s2n and wave_rms to the metadata.
4. Add KOA related metadata fields that were already in the spec2d file to the spec1d file for DEIMOS data.
5. Split up files in the archive_dir into subdirectories by year and month. This is to avoid creating a directory with too many files.
6. Only create by_id_meta.dat and by_object_meta.dat when using --archive_dir. I made this change because I thought the metadata was getting too archive specific and not useful when just collating.
7. Create a new collate collate_report.dat file with similar information to by_object_meta.dat but with less archive_dir specific information. This is created regardless of whether --archive_dir is specified and is intended to give information on what collation was performed.


badpandabear:
Unit tests:
`========================== 285 passed, 442 warnings in 805.66s (0:13:25) =============`

debora-pe (changes requested):
Hi @badpandabear Thank you for this work. It's great. I have some comments, suggestions. And I found a problem but actually I don't know if it's because I did the main reduction in a different branch that this one. 

Here are the comments.

-- In the `collate_report.dat` and `by_object_meta.dat` can you call `ra` and `dec` --> `objra` and `objdec` instead? I feel that `ra` and `dec` can be confused with the value in `by_id_meta.dat`.

-- The files are split into subdirectories by year and month, which looks nice. But many times we will collate spectra that are taken during different months and even years. how would this work for these cases?

-- I noticed that there is a problem with `guidfwhm`, `progpi`, `semester` and `progid`. In `collate_report.dat` and `by_object_meta.dat` they are all `None` while in `by_id_meta.dat` some spectra have those correct and some are `None`. I have checked the header of those files and this info should be present, both in the raw file and in the spec2D. They are not present in the spec1D files though. Maybe this is my problem because I run the main reduction not in this branch?

-- I have a suggestion for the History of the collated spec1D. As it is now I think is a bit hard to read. the suggestion is this:

```
HISTORY 2021-07-14T20:15 PypeIt Coadded 20 objects from 10 spec1d files         
HISTORY From spec1d_DE.20160106.49294-071715B1_DEIMOS_20160106T134128.694.fits
HISTORY SPAT0153-SLIT0162-DET01                           
HISTORY SPAT0141-SLIT0147-DET05
HISTORY From spec1d_DE.20160106.46528-071715B1_DEIMOS_20160106T125521.821.fits
HISTORY SPAT0152-SLIT0162-DET01                          
HISTORY SPAT0137-SLIT0147-DET05 
HISTORY From spec1d_DE.20160106.45260-071715B1_DEIMOS_20160106T123414.592.fits
HISTORY SPAT0152-SLIT0162-DET01                         
HISTORY SPAT0137-SLIT0147-DET05                         
(...)     
```   

I felt that the `File 0` and `from file 0` writing can be hard to follow if you are trying to quickly look at it. Would this change be possible?

-- When I look at the tables for the KOA archive I always get confuse about what is what and I realized why. The name of the table are not really intuitive. So one table (`by_id_meta.dat`) is a list of the reduced spec1D (BTW I think we can add also a column with the spec2D filename) and their link to the KOA ID. So I guess we can call that table something like "main_reduction" or something like that.
The other table (`by_object_meta.dat`) is a list of all the collated spec1D so we can call that table something like "coadded_spec1Ds". These are just suggestions to give you an idea, but I think it's important to give them names that are more intuitive.

doc/collate1d.rst
```diff
@@ -190,37 +190,140 @@ Archiving
...
++-----------------+-----------------------------------------------------------+
+| wav_rms         | The RMS in pixels of the wavelength solution.             |
++-----------------+-----------------------------------------------------------+
+| source_id       | The KOAID of the original source image.                   |
++-----------------+-----------------------------------------------------------+
+| spec1d_filename | The name of the spec1d file containing the spectrum.      |
++-----------------+-----------------------------------------------------------+
+| dispname        | The grating used for the source image.                    |
++-----------------+-----------------------------------------------------------+
+| decker          | The slitmask used for the source image.                   |
```
bpholden: I have a nomenclature complaint. I would prefer that the slit mask name be called mask name or slit mask name and not decker. A decker is something weirdly specific and DEIMOS does not have one.

bpholden (commented):
Overall this seems great. I just do not like the column name Decker. KOA uses aperture which is more general. Slit mask name is what the FITS header keyword is. I would prefer either.

badpandabear:
@deborape I've started working on these changes. For the metadata file names, how does "reduced_files_meta.dat" and "coadded_files_meta.dat" sound?

The problem you had with guidfwhm, progpi, semester and progid was probably caused by using a different version to reduce the files. I added those values to the spec1d header in this PR.  If you can point me to the data that had this problem I can double check.

Also regarding your question about where the coadded files go, currently I use the MJD from the header. It looks like the coadding code uses the MJD from the first spec1d file.   Would a different organization (like semester and progid) be better?

bpholden:
> Also regarding your question about where the coadded files go, currently I use the MJD from the header. It looks like the coadding code uses the MJD from the first spec1d file. Would a different organization (like semester and progid) be better?

@badpandabear I would suggest semester and progid, yes. That will match the intent of the individuals that designed the masks at least.

astronomeara:
Semester and progid also helps on the KOA statistics end

debora-pe:
Hi @badpandabear 

> @deborape I've started working on these changes. For the metadata file names, how does "reduced_files_meta.dat" and "coadded_files_meta.dat" sound?

Yes, those file names sound good!

> The problem you had with guidfwhm, progpi, semester and progid was probably caused by using a different version to reduce the files. I added those values to the spec1d header in this PR. If you can point me to the data that had this problem I can double check.

Yes, you are right. I re-run the reduction in this branch and `progpi`, `semester` and `progid` are now visible in the tables. However,  `guidfwhm` is still `None`. Is it possible that this is due to the fact that `subheader_for_spec` in  `pypeit/spectrographs/keck_deimos.py` does not include GUIDFWHM in the `koa_header_cards`? I tested it for 2 different datasets and for both `guidfwhm` is always `None`.

> Also regarding your question about where the coadded files go, currently I use the MJD from the header. It looks like the coadding code uses the MJD from the first spec1d file. Would a different organization (like semester and progid) be better?

Yes, as the others said, an organization by semester and progid would be better. Also, though, there will still be cases where we combine spectra from different semester+progids, so it would good if we can save that info somewhere. So you could use semester+progid of the first spec1D file for the subdirectories, but maybe save the list of different semester+progids in the History?


I have a few additional comments.
When I was running `pypeit_collate1D`, at first I did not run it in the same directory where the .pypeit file is, so it crushed but only after all the collated 1D files were created. 

- It would be better if a check for the .pypeit file is done at the beginning of the script (still only when `--archive_dir` is used).

- For the spec1D.txt the check does not need to be done at the beginning of the script (where it is done now it's fine), but I am wondering if we should not have the script crush, but instead record that missing file somewhere.
 
- Also, we could give the option to input where to find the .pypeit file, with the default being the current directory. It could be a `--pypeit_file` parset where the user points to the .pypeit file.

- And I would also give the option the save all the collated 1D files in a specific directory, which it could be done by an `--outdir` parset. The default could still be the current directory.

- A very minor thing. I found a typo in `history.py` line 82: "Callibration frames" --> "Calibration frames"

badpandabear:
Hi @deborape , thanks for the feedback.

You are absolutely correct about the `guidfwhm`, that's something I just missed when I was adding the other KOA header keywords.

Something I should clarify, is that currently the script doesn't look for the .pypeit file in the current directory.
It looks in the parent directory of the spec1d files. My thought that was when collating from multiple reductions, the spec1d files would be specified as something like `"*/Science/spec1d*.fits"`, and as long as the .pypeit files were in the directory above `Science`, archiving would still work.  I'll add this to the documentation and also add the `--pypeit_file` parameter you recommended.

debora-pe:
Hi @badpandabear.
The changes that you made are really great! Thank you so much for doing that.

I have only a few minor things.
- please in the description of the script, explicitly say which is the default `--outdir`.
- please call the first column of reduced_files_meta.dat `koaid`
- I believe that for the koaid you are using the raw file name. Sometime the raw files are compressed and ends in `fits.gz` but the koaid would be without `.gz`. So can you have the koaid to be without `.gz` if present (also in coadded_files_meta).
- I would put the spec2d files in reduced_files_meta.dat in a new column to match the koaid.
- So, a suggestion for the first 4 columns of reduced_files_meta.dat is: `koaid`, `spec1d file`, `spec1d info`, `spec2d file`

That's everything. 
Great work.

debora-pe (approved):
Great work @badpandabear! Thank you!

bpholden (approved):

badpandabear:
Dev suite passed except for known issues with shane_kast_red:
--- PYPEIT DEVELOPMENT SUITE FAILED 2/128 TESTS (Masters ignored) ---
Failed tests:
    shane_kast_red/600_7500_d57 pypeit (without masters)
    shane_kast_red/1200_5000_d57 pypeit (without masters)
Skipped tests:
Testing Started at 2021-08-09T10:34:57.593043
Testing Completed at 2021-08-09T13:29:03.132094
Total Time: 2:54:05.539051

PR #1242, opened 2021-08-09
Show full version on readthedocs

kbwestfall:
This fixes the version displayed on readthedocs.  @profxj, this simply displays the full version on readthedocs for the "release" version.  I didn't realize we didn't have this before, but I would suggest turning on the readthedocs for the tagged version of the code.  That is, you should be able to set the docs for the '1.4.2' as visible.  In the future, this means that people will see the docs associated with the current `release` version in the `release` version of readthedocs, or they can use the drop-down menu at the bottom left to select the tagged version they have installed.


profxj (approved):
thx!

rcooke-ast (approved):
Approved

PR #1246, opened 2021-08-12
Add CentOS to CI

tepickering:
This PR adds a section of CI tests that are run under CentOS. GitHub Actions does not support CentOS directly, but it does support the ability to run tests within custom containers. The containers used here are the stock ones for CentOS 7 and 8 that are available from Docker Hub. 

The version pinning for `bottleneck` was removed since it doesn't appear to be strictly necessary and was causing acceptable pre-built packages to be ignored. 

The `fetch-depth` parameter is now explicitly set to 1 to make it clear only the latest commit is cloned. This matches the default and shouldn't change anything, but guards against the default changing in the future.


jhennawi:
Thanks for doing this!

jhennawi (approved):

profxj (approved):
approving

PR #1253, opened 2021-08-24
Astrometric corrections for KCWI

rcooke-ast:
This update includes a few KCWI updates:
* allows the contbars frame to be used (or not) for the astrometric correction
* adds KCWI documentation


pypeit/io.py
```diff
@@ -34,6 +34,22 @@
 import pypeit
 import time
 
+
+def get_time_string(codetime):
+    if codetime < 60.0:
```
profxj: add a brief doc string

pypeit/slittrace.py
```diff
@@ -441,26 +451,75 @@ def get_radec_image(self, wcs, initial=True, flexure=None, trace_cen=None):
...
+            tilt_spl = RegularGridInterpolator((np.arange(tilts.shape[0]), np.arange(tilts.shape[1])), tilts*(nspec-1), method='linear')
 
         # Initialise the output
         raimg = np.zeros((self.nspec, self.nspat))
         decimg = np.zeros((self.nspec, self.nspat))
         minmax = np.zeros((self.nslits, 2))
-
         # Get the slit information
         slitid_img_init = self.slit_img(pad=0, initial=initial, flexure=flexure)
+        # Debugging when astrometric=True
```
profxj: remove?

profxj (approved):
Solid.  Only one nit from me.

pypeit/slittrace.py
```diff
@@ -15,6 +15,8 @@
 from astropy.coordinates import SkyCoord, Angle
 from astropy import units
 from astropy.stats import sigma_clipped_stats
+from skimage import transform as tf
```
debora-pe: `skyimage` is not a package that is included in the Pypeit requirements. That should be done.

pypeit/slittrace.py
```diff
@@ -413,24 +415,32 @@ def get_slitlengths(self, initial=False, median=False):
...
+        tilts : `numpy.ndarray`
+            Spectral tilts.
+        locations : `numpy.ndarray`_, list
+            locations along the slit of the alignment traces. Must
+            be a 1D array of the same length as alignments.traces.shape[1]
         maxslitlen : int
             This is the slit length in pixels, and it should be the same
             value that was passed to get_wcs() to generate the WCS that
             is passed into this function as an argument.
+        initial : bool
```
debora-pe: This should be `astrometric`?

doc/skysub.rst
```diff
@@ -48,3 +53,74 @@ lines on faint galaxy continua), set
...
+and drag from the gray to the black regions (i.e. you must click
+and drag within this small window for it to work).
+
+Alternatively, you can click the "Enter regions" button, which
+will request input from the command line. You should now enter
+the regions in the same format as above for the "user_regions".
+
+If you're happy with the sky regions, press the
+"Continue (and save changes)" button. If you do not wish to save
+the sky regions, press the "Continue (and don't save changes)".
```
debora-pe: button

debora-pe (approved):
Looks good to me! Only minor comments.
The most important one is about the `skyimage` package.

Please, also post the results of the tests.

pypeit/slittrace.py
tepickering: note that the `skimage` module is provided by the `scikit-image` package. and, yes, `scikit-image` needs to be added to the requirements in `setup.cfg`.

doc/skysub.rst
rcooke-ast: Done!

pypeit/io.py
rcooke-ast: Thanks - I should have caught that...

pypeit/slittrace.py
rcooke-ast: Quite right - good catch!!

pypeit/slittrace.py
rcooke-ast: done (and I've also deleted the debugging comments below, too...)

pypeit/slittrace.py
rcooke-ast: Given that `scikit-image` is only used when creating datacubes (and only when an astrometric correction is requested) then I'm tempted to not add this to the list of "must haves" for PypeIt. I've come up with an alternative solution (to catch an import error, and added some info to the installation and KCWI docs) let me know what you think!

rcooke-ast (commented):
Comments addressed - thanks all!

pypeit/core/datacube.py
```diff
@@ -599,10 +597,12 @@ def coadd_cube(files, parset, overwrite=False):
     wave_ref = None
     whitelight_img = None  # This is the whitelight image based on all input spec2d frames
     weights = np.ones(numfiles)  # Weights to use when combining cubes
+    locations = parset['calibrations']['alignment']['locations']
     for ff, fil in enumerate(files):
         # Load it up
         spec2DObj = spec2dobj.Spec2DObj.from_file(fil, det)
         detector = spec2DObj.detector
+        flexure = None  #spec2DObj.sci_spat_flexure
```
jhennawi, line 605: Can you get rid of this line and set flexure to be None in the keyword argument online 625?

pypeit/io.py
```diff
@@ -34,6 +32,33 @@
 import pypeit
 import time
 
+
```
jhennawi, line 35: Does this rather belong in pypeit.uitls?

pypeit/pypeit.py
```diff
@@ -921,19 +921,7 @@ def print_end_time(self):
...
-        elif codetime/60.0 < 60.0:
-            mns = int(codetime/60.0)
-            scs = codetime - 60.0*mns
-            msgs.info('Execution time: {0:d}m {1:.2f}s'.format(mns, scs))
-        else:
-            hrs = int(codetime/3600.0)
-            mns = int(60.0*(codetime/3600.0 - hrs))
-            scs = codetime - 60.0*mns - 3600.0*hrs
-            msgs.info('Execution time: {0:d}h {1:d}m {2:.2f}s'.format(hrs, mns, scs))
+        msgs.info(io.get_time_string(time.time()-self.tstart))
```
jhennawi: Thanks for cleaning this up.

pypeit/slittrace.py
```diff
@@ -413,24 +417,32 @@ def get_slitlengths(self, initial=False, median=False):
             slitlen = np.median(slitlen, axis=1)
         return slitlen
 
-    def get_radec_image(self, wcs, initial=True, flexure=None, trace_cen=None):
+    def get_radec_image(self, wcs, alignments, tilts, locations,
+                        astrometric=True, initial=True, flexure=None):
```
jhennawi, line 421: Can you please add a comment to the doc that this method is currently only used for IFUs?

jhennawi (changes requested):
Very minor comments. Thanks for this PR.

jhennawi (approved):

pypeit/core/datacube.py
rcooke-ast: There are multiple instances of flexure in this code, so I think having a single place to change this makes more sense... just in case flexure is supported with future IFUs.

pypeit/io.py
rcooke-ast: Sure - I'll put it there!

pypeit/pypeit.py
rcooke-ast: np

pypeit/slittrace.py
rcooke-ast: Done!

rcooke-ast (commented):
@jhennawi comments addressed - thanks!

rcooke-ast:
Tests pass... merging
image

PR #1259, opened 2021-09-06
KCWI BL grating support

rcooke-ast:
* added BL grating support to KCWI
* added features and updated the docs for the identify GUI (including how to incorporate a wavelength solution into the full_template archive).
* added a `skip_boxcar` parameter.

There is a matching PR for the dev suite. Here is a screenshot of the fit:

image


doc/construct_template.rst
```diff
@@ -51,7 +51,31 @@ Once all of the slits and master wavelength calibration files are
 selected, use the function :func:`pypeit.core.wavecal.templates.build_template` to turn the list of slits,
 wavelength ranges, and master wavecalib files into a final template.
 
-The input looks like the following:
+For longevity, it is a good idea to obtain the PypeIt Development Suite and combine your solution
```
profxj, line 54: Thanks for this!

Is there a link to this file from `new_spectrograph.rst`?
If not, kindly make one.

pypeit/core/wavecal/templates.py
```diff
@@ -764,6 +764,13 @@ def main(flg):
         slits = [1026, 1021]
         lcut = [4350.0, 8000.0]
         build_template([wfile1, wfile2], slits, lcut, binspec, outroot, lowredux=False, normalize=True)
+        # FeAr BL
```
profxj: this is ok, but I'm trying to move all of these into separate modules
in the `spectrographs` folder.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -130,6 +130,10 @@ def config_specific_par(self, scifile, inp_par=None):
             par['calibrations']['wavelengths']['method'] = 'full_template'
             par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_kcwi_BM.fits'
             par['calibrations']['wavelengths']['lamps'] = ['FeI', 'ArI', 'ArII']
+        elif self.get_meta_value(headarr, 'dispname') == 'BL':
```
profxj, line 148: looks like there might be some code replication in this
`if/else` block.  Reduce if possible.

profxj (approved):
Just a few minor things from me.

Thanks!

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Done!

doc/construct_template.rst
rcooke-ast: It was not - Just added

pypeit/core/wavecal/templates.py
rcooke-ast: Sorry about this... If I should be doing something else, just let me know (I've written the docs such that this method is valid - sorry if I'm out of date!).

rcooke-ast:
Tests pass...
image

debora-pe (approved):
I went through the changes and everything looks good to me! Nice work.
It would be nice to have a KCWI BL dataset in the Dev Suite to test.

Also, would be easy for you to run the Dev Suite? Even if they are minor, you made changes to reduce and pypeitpar and it would be good to check that everything is still OK.

Anyway I'm approving.

rcooke-ast:
Thanks @deborape - I don't currently have any KCWI BL datasets myself (I'm implementing this as a favour for other users). That said, all of the same algorithms are touched by the KCWI BH2 dataset in the dev suite, so i think this is OK.

I'll run the dev suite now, and report back the results - thanks again!

pypeit/core/wavecal/templates.py
profxj: this is fine.  But, yeah, see this an example for 
the future:
https://github.com/pypeit/PypeIt/blob/release/pypeit/core/wavecal/spectrographs/templ_keck_deimos.py

pypeit/reduce.py
```diff
@@ -276,19 +276,20 @@ def extract(self, global_sky, sobjs_obj):
             self.sobjs = self.sobjs_obj.copy()
             # Purge out the negative objects if this was a near-IR reduction unless negative objects are requested
```
jhennawi: I don't understand the control flow here and the addition of this skip_optimal parameter. Can you please elaborate. The the user is skipping optimal, and also skipping boxcar, then it is unclear to me what this method is doing. Basically it is just filling things with zeros. If that is the case, this should just be done outside of the extract, i.e. you are not extracting anything so why is the extract method being called in the control flow? In other words, skip_boxcar is not an appropriate name for the parameter. It should be "skip_extraction" and this if then statement should go somewhere in reduce.py before extract is called to prevent extraction and simply fill everything with zeros.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -245,8 +245,9 @@ def default_pypeit_par(cls):
         par['scienceframe']['process']['use_biasimage'] = False
         par['scienceframe']['process']['use_darkimage'] = False
 
-        # Don't do optimal extraction for 3D data.
-        par['reduce']['extraction']['skip_optimal'] = True  # Because extraction occurs before the DAR correction, don't to optimal - boxcar will also be rubbish
+        # Don't do 1D extraction for 3D data - it's meaningless because the DAR correction must be performed on the 3D data.
+        par['reduce']['extraction']['skip_boxcar'] = True  # Because extraction occurs before the DAR correction, don't to boxcar
```
jhennawi: See my comment above. I think you want to change this to skip_extraction

jhennawi (changes requested):
Thanks for this PR @rcooke-ast. The improvements to the identify script and additional documentation are very much appreciated. Please see my comments about skip_boxcar. That is my only significant feedback. I don't think it belongs in the extract method and I think you should call this something else. Going into the extract method at all when your plan is not to extract anything is at best an opaque control flow. Maybe I've missed what you are aiming for here though. 

Also, since this PR modifies reduce, the entire dev suite needs to be run before this can be approved.

pypeit/reduce.py
```diff
@@ -472,9 +472,19 @@ def run_extraction(self, global_sky, sobjs_obj, skymask, ra=None, dec=None, obst
             if self.par['flexure']['spec_method'] != 'skip' and not self.std_redux:
                 self.spec_flexure_correct(mode='global')
 
-            # Extract + Return
-            self.skymodel, self.objmodel, self.ivarmodel, self.outmask, self.sobjs \
-                = self.extract(self.global_sky, self.sobjs_obj)
+            if self.par['reduce']['extraction']['skip_extraction']:
```
jhennawi: I think this can be simplified to have line 469 state: 

469 if self.nobj > 0 and not self.par['reduce']['extraction']['skip_extraction']:

490 else: 

You are doing the same setups twice in 476-483 that are being done in the else. 

Ideally we would not have entered run_extraction at all, which was my initial intent, however I can see here that this routine is creating a bunch of outputs like tilts, waving etc. that are needed.  I'm pretty sure this is massively inefficient, since those images were probably already created to create the global sky. That is why this class needs to be refactored, which is on my list of things to do.

rcooke-ast:
All tests pass...
image
image

profxj:
--- PYPEIT DEVELOPMENT SUITE PASSED 128/128 TESTS (Masters ignored) ---
Testing Started at 2021-10-03T12:55:27.003549
Testing Completed at 2021-10-03T18:12:56.018397
Total Time: 5:17:29.014848

jhennawi (approved):

PR #1260, opened 2021-09-06
Enabling more than 26 setups

jhennawi:
This PR extends the setup dictionary to allow more setups than the number of letters in the alphabet. The first 26 setups will be A, B, C etc. and then after this there are 26^2 = 676 setups corresponding to AA, AB etc. So the total number allowed is now 702. As this is a single line change, tests should not be required. For what its worth test_coremeta passes, although a bunch of tests are failing because my Cooked is not up to date (which is why I dislike Cooked!).


profxj:
I'd be amazed if this doesn't break a bunch of tests.

Please run em.  Thx.

jhennawi:
Screenshot 2021-09-08 at 22 06 46

Unit tests pass. The only failure is also a failure on develop due to the recent headers PR (I think).

profxj (approved):
approved

rcooke-ast (approved):
Approved - nice and simple 👍

PR #1280, opened 2021-10-27
MOSFIRE maskdef ingestion

debora-pe:
As title this ingests MOSFIRE slitmask information to PypeIt.

As for DEIMOS, MOSFIRE now:
- uses slitmask info in the slit edge tracing
- associates RA, Dec and Object name to each extracted object
- extracts undetected objects using the predicted position from slitmask info.

The dither offeset recorded in the header is used as default `slitmask_offset`, but the user has the possibility to provide the `maskdef_id` of a slit with a bright object that can trace the offset.

This also includes:
- improvements in the frame typing
- documentation
- unit tests


debora-pe:
Multi-slit
multi-slit

long2pos
long2pos

LONGSLIT-3x0.7
longslit

joshwalawender:
@deborape I've looked through this.  I'm not intimately familiar with PypeIt code, so I can only review the conceptual aspects for MOSFIRE.

For file typing, it looks reasonable.  I wish we didn't have to use the OBJECT header keyword as that is free text set by the user or script which executed the observation and it is not forced to reflect the state of the instrument, but given that there is no other way I can think of to uniquely identify the lamps off flats, I think we have to use it.

One other thought on typing is that for historical reasons, the dome flat lamp state is recorded in multiple keywords in the MOSFIRE headers: FLATSPEC, FPOWER, FLAMP1, and FLAMP2.  The FPOWER (power level 0-20 where 0=high and 20=low), FLAMP1, and FLAMP2 (on or off) values reflect the state of the current lamp control system and are pulled from that service. FLATSPEC, on the other hand, is a MOSFIRE keyword which does not control anything and is simply set by the scripts which we use to control the dome flat lamps.  Those scripts control FPOWER, FLAMP1, and FLAMP2 directly.  Thus, while FLATSPEC is a nice summary, it is possible for it to be mismatched to reality (this should be exceedingly rare).  Thus there may be an advantage to using FLAMP1 and FLAMP2 instead of FLATSPEC.

doc/dev/mosfireframes.rst
```diff
@@ -0,0 +1,119 @@
...
+The general procedure used to assign frames a given type is described
+here: :ref:`frame_types`.
+
+MOSFIRE frame typing
+--------------------
+
+The primary typing of MOSFIRE frames is performed by
+:func:`pypeit.spectrographs.keck_mosfire.KeckMOSFIRESpectrograph.check_frame_type`.
+This function checks the values of various header keywords against a
+set of criteria used to classify the frame type. The headers cards
```
profxj: headers -> header

doc/dev/radec_object.rst
```diff
@@ -61,32 +65,44 @@ Using the coordinates of the slit center available from the slitmask design and
...
-  This parameter is optional (default value is **bright_maskdef_id=None**) and is ignored
-  if **slitmask_offset** is provided.
-- **use_alignbox**: flag to use stars in alignment boxes to compute the slitmask offset.
-  If this is set to True PypeIt will NOT compute the offset using **nsig_thrshd** or
-  **bright_maskdef_id**. The default is **use_alignbox=False**.
+
+- **bright_maskdef_id**: ``maskdef_id`` (corresponding to ``dSlitId`` and ``Slit_Number`` in the DEIMOS
+  and MOSFIRE slitmask design, respectively) of a slit containing a bright object that will be used
+  to compute the slitmask offset. This parameter is optional (default value is **bright_maskdef_id=None**)
+  and is ignored if **slitmask_offset** is provided. However, this parameter is highly recommended
```
profxj, line 89: Since I think of it here, should we change the `type` allowed for the offset that gave Ebeling's crash be `int` or `float`?  I think either will work fine with the code.  Just a thought.

doc/include/datamodel_slittrace.rst
```diff
@@ -1,10 +1,10 @@
-Version: 1.1.1
+Version: 1.1.4
 
 =====================  =================  ==========  ===========================================================================================================================
 Obj Key                Obj Type           Array Type  Description                                                                                                                                                                         
 =====================  =================  ==========  ===========================================================================================================================
 ``spat_id``            ndarray            int         Slit ID number from SPAT measured at half way point.
-``maskdef_id``         ndarray            int         Slit ID number from slitmask design (implemented only for :doc:`deimos`).
+``maskdef_id``         ndarray            int         Slit ID number from slitmask design (implemented only for :doc:`deimos` and :doc:`mosfire`).
```
profxj, line 7: Eventually we'll just want to point to a file on slitmasks that list all supported spectrographs.  But that time isn't here (yet).  :)

doc/mosfire.rst
```diff
@@ -0,0 +1,140 @@
...
+except when the *LONGSLIT* mask is used.
+
+``PypeIt`` also assigns to each extracted 1D spectrum the corresponding RA, Dec and object name
+information from the slit-mask design, and forces the extraction of undetected object at the location
+expected from the slit-mask design. This functionality is also the default for MOSFIRE and can be
+controlled through the flags **assign_obj** and **extract_missing_objs** in :ref:`pypeit_par:SlitMaskPar Keywords`.
+See `Additional Reading`_ .
+
+By default, ``PypeIt`` uses the dither offset recorded in the header of the science frames to find the
+objects in the slits. However, a better procedure would be to let ``PypeIt`` computes the offset using
```
profxj: computes -> compute

doc/mosfire.rst
```diff
@@ -0,0 +1,140 @@
...
+a bright object in one of the slits in the slit-mask. To do so, the uses should provide in the :doc:`pypeit_file`
+the ``maskdef_id`` (corresponding to "Slit_Number" in the MOSFIRE slit-mask design) of the slit containing the
+bright object. Here is an example::
+
+    [reduce]
+       [[slitmask]]
+            use_dither_offset = False
+            bright_maskdef_id = 1
```
profxj, line 71: Is this numbering in MOSFIRE IDs or PypeIt slit_ids?  
I presume the former (and the text suggests as much).

Maybe just add a sentence that this SLIT_ID will differ
from that given by PypeIt.

doc/mosfire.rst
```diff
@@ -0,0 +1,140 @@
...
+design information contained as meta data in the MOSFIRE observations, and assign to each extracted 1D spectrum
+the corresponding RA, Dec and object name information from the slit-mask design. However, forces the extraction
+of undetected object is not performed by default.
+Note that the RA and Dec recorded in the MOSFIRE observations of `long2pos` masks are mock numbers, therefore
+the users should not use them.
+
+LONGSLIT
+********
+Similarly to `long2pos`, `LONGSLIT` masks with a single slit that is shorter than the length of the
+detector will have random slits that fill the mask. By default, ``PypeIt`` is capable of excluding
```
profxj, line 112: And part of the issue is that those extra slits can cause the code to crash, right?  Maybe add a little bit of why we are excluding them.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -127,6 +131,63 @@ def default_pypeit_par(cls):
...
+                pix_start, pix_end = find_longslit_pos(scifile)
+                # exclude the random slits outside the longslit from slit tracing
+                par['calibrations']['slitedges']['exclude_regions'] = ['1:0:{}'.format(pix_start),
+                                                                       '1:{}:2040'.format(pix_end)]
+                par['calibrations']['slitedges']['det_buffer'] = 0
+                # artificially add left and right edges
+                par['calibrations']['slitedges']['bound_detector'] = True
+
+        # Turn on the use of mask design
+        if 'LONGSLIT' not in self.get_meta_value(headarr, 'decker'):
```
profxj: should this simply be an `else` off the above?

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -176,22 +240,38 @@ def compound_meta(self, headarr, meta_key):
...
-                    # TODO: This should be changed to except on a specific error.
-                    FLATSPEC = int(headarr[0].get('FLATSPEC'))
-                    PWSTATA7 = int(headarr[0].get('PWSTATA7'))
-                    PWSTATA8 = int(headarr[0].get('PWSTATA8'))
-                    if FLATSPEC == 0 and PWSTATA7 == 0 and PWSTATA8 == 0:
+            # KOAIMTYP non good frame typing
+            # if headarr[0].get('KOAIMTYP', None) is not None:
+            #     return headarr[0].get('KOAIMTYP')
+            # else:
+            try:
```
profxj: Youch!  Pretty painful.  But I don't see a way to simplify..

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -287,15 +369,17 @@ def lamps(self, fitstbl, status):
...
             # Check if all are off
-            return np.all(np.array([fitstbl[k] == 0 for k in fitstbl.keys() if 'lampstat' in k]),
-                          axis=0)
-        if status == 'arcs':
-            # Check if any arc lamps are on
-            arc_lamp_stat = [ 'lampstat{0:02d}'.format(i) for i in range(1,6) ]
-            return np.any(np.array([ fitstbl[k] == 1 for k in fitstbl.keys()
-                                            if k in arc_lamp_stat]), axis=0)
+            return fitstbl['lampstat01'] == 'off'
+            # return np.all(np.array([fitstbl[k] == 0 for k in fitstbl.keys() if 'lampstat' in k]),
```
profxj: remove?

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -432,3 +516,233 @@ def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table,
...
+                       'Slitmask design matching not possible. Turn parameter `use_maskdesign` off')
+
+        targ_dist_center = np.array(ssl['Target_to_center_of_slit_distance'], dtype=float)
+
+        slit_centers = np.array(ssl['Slit_length'], dtype=float) / 2.
+        # Some # adjustment for long2pos
+        if 'long2pos' in self.get_meta_value(filename, 'decker'):
+            slit_centers_wrong = np.array(ssl['Slit_length'], dtype=float) / 2.
+            # correct slit length
+            ssl['Slit_length'] = '22.970', '7.010', '22.970'
```
profxj: Are these written to the header too?  If so, maybe pull
from there in case they change some day.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -432,3 +516,233 @@ def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table,
...
+        """
+         Provides the slit edges positions predicted by the slitmask design using
+         the mask coordinates already converted from mm to pixels by the method
+        `mask_to_pixel_coordinates`.
+
+        If not already instantiated, the :attr:`slitmask`, :attr:`amap`,
+        and :attr:`bmap` attributes are instantiated.  If so, a file must be provided.
+
+        Args:
+        ccdnum (:obj:`int`):
```
profxj: the variables may need to be indented for sphinx to work right.  not sure.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -432,3 +516,233 @@ def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table,
...
+            Detector number
+        filename (:obj:`str`, optional):
+            The filename to use to (re)instantiate the :attr:`slitmask` and :attr:`grating`.
+            Default is None, i.e., to use previously instantiated attributes.
+        debug (:obj:`bool`, optional):
+            Run in debug mode
+        Returns:
+            Three `numpy.ndarray`_ and a :class:`~pypeit.spectrographs.slitmask.SlitMask`.
```
profxj: usually we write --   `tuple: ...`

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -432,3 +516,233 @@ def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table,
...
+        # Re-initiate slitmask and amap and bmap
+        if filename is not None:
+            # Reset the slitmask
+            self.get_slitmask(filename)
+
+        if self.slitmask is None:
+            msgs.error('Unable to read slitmask design info. Provide a file.')
+
+        # Hard-Coded for MOSFIRE
+        platescale = 0.1798
```
profxj: is this in the detector object?  If so, can we pull it from there?

profxj: or follow what you do below (read from header)

pypeit/tests/test_frametype.py
```diff
@@ -67,3 +67,55 @@ def test_deimos():
...
         shutil.rmtree(output_path)
 
+
+@dev_suite_required
+def test_mosfire():
+    # Raw DEIMOS directory
+    raw_dir = os.path.join(os.getenv('PYPEIT_DEV'), 'RAW_DATA', 'keck_mosfire')
+
+    # Get the list of setup directories
+    setups = [os.path.join('/Users/dpelliccia/PypeIt/PypeIt-development-suite/RAW_DATA/keck_mosfire', isetup)
```
profxj: use raw_dir, not `deplliccia`

pypeit/tests/test_maskdesign_matching.py
```diff
@@ -15,128 +15,167 @@
 # to traced slits
 
 # Load flats files
-def deimos_flat_files():
-    return [os.path.join(os.getenv('PYPEIT_DEV'), 'RAW_DATA', 'keck_deimos', '1200G_Cooper', ifile)
+def flat_files(instr='keck_deimos'):
+    if instr == 'keck_deimos':
+        return [os.path.join(os.getenv('PYPEIT_DEV'), 'RAW_DATA', 'keck_deimos', '1200G_Cooper', ifile)
                 for ifile in ['d0115_0023.fits.gz', 'd0115_0024.fits.gz', 'd0115_0025.fits.gz']]
+    if instr == 'keck_mosfire':
```
profxj: `elif`

pypeit/tests/test_maskdesign_matching.py
```diff
@@ -15,128 +15,167 @@
...
+
+        slits = edges.get_slits()
+        # Check that the `maskdef_id` assigned to the first and last slits is correct
+        if name == 'keck_deimos':
+            assert slits.maskdef_id[0] == 1098247, 'DEIMOS maskdef_id not found or wrong'
+            assert slits.maskdef_id[-1] == 1098226, 'DEIMOS maskdef_id not found or wrong'
+            # `maskdef_id` is the slit_id ("dSlitId") from the DEIMOS slit-mask design. If the matching is done
+            # correctly these are the slit_id values that the first and last slits should have. These values are
+            # coming from a reduction run with DEEP2 IDL-based pipeline.
+        if name == 'keck_mosfire':
```
profxj: `elif`

pypeit/tests/test_slitmask.py
```diff
@@ -19,145 +20,185 @@ def data_path(filename):
...
-def deimos_flat_files():
-    return [os.path.join(os.getenv('PYPEIT_DEV'), 'RAW_DATA', 'keck_deimos', '830G_M_8500', ifile)
-            for ifile in ['DE.20100913.57161.fits.gz', 'DE.20100913.57006.fits.gz']]
 
+# Load flats files
+def flat_files(instr='keck_deimos'):
+    if instr == 'keck_deimos':
+        return [os.path.join(os.getenv('PYPEIT_DEV'), 'RAW_DATA', 'keck_deimos', '830G_M_8500', ifile)
+                for ifile in ['DE.20100913.57161.fits.gz', 'DE.20100913.57006.fits.gz']]
+    if instr == 'keck_mosfire':
```
profxj: `elif`

profxj (approved):
Great stuff!  I don't think any of my requests warrant
my clicking on `Request changes`, so I am approving.

doc/dev/mosfireframes.rst
debora-pe: Done!

doc/dev/radec_object.rst
debora-pe: Done!

doc/mosfire.rst
debora-pe: Done

doc/mosfire.rst
debora-pe: Done

doc/mosfire.rst
debora-pe: Done

pypeit/spectrographs/keck_mosfire.py
debora-pe: Yes, done.

pypeit/spectrographs/keck_mosfire.py
debora-pe: No, unfortunately these are not written in the header.  I added a method that provides these numbers, so in case we need to make some modifications in the future, it will be easier.

pypeit/spectrographs/keck_mosfire.py
debora-pe: Yes! Thanks. Done

pypeit/spectrographs/keck_mosfire.py
debora-pe: Done.

pypeit/spectrographs/keck_mosfire.py
debora-pe: Done.

pypeit/tests/test_frametype.py
debora-pe: Thanks! Done.

pypeit/tests/test_maskdesign_matching.py
debora-pe: Done

pypeit/tests/test_maskdesign_matching.py
debora-pe: Done

pypeit/tests/test_slitmask.py
debora-pe: Done

pypeit/edgetrace.py
```diff
@@ -596,9 +598,9 @@ def empty_design_table(rows=None):
                     # table.Column(name='SLITRFOC', dtype=float, length=length,
                     #              description='Right edge of the slit in mm at the focal plane'),
                     table.Column(name='SLITLOPT', dtype=float, length=length,
-                                description='Left edge of the slit in pixel predicted by optical model'),
+                                description='Left edge of the slit in pixel predicted by slitmask design'),
                     table.Column(name='SLITROPT', dtype=float, length=length,
-                                description='Right edge of the slit in pixel predicted optical model'),
+                                description='Right edge of the slit in pixel predicted slitmask design'),
```
kbwestfall: Is it worth changing the names of these columns?  Maybe ``SLIT?DES``?  I don't remember what these were supposed to hold.  Is this the prediction before or after the matching is performed?  I.e., if a user wants to check the quality of the registration between the measured trace edges and the expected position given the slitmask design, can they look at the distribution of, e.g., ``TRACELPIX - SLITLOPT``?

pypeit/edgetrace.py
```diff
@@ -1059,6 +1069,29 @@ def initial_trace(self, bpm=None):
...
+    def _parse_exclude_regions(self):
+        """
+        Parse the `exclude_regions` parset.
+
+        Returns:
+            :obj:`tuple`: Returns two arrays with the starting
+            and ending pixels of the regions to exclude in this detector
+        """
+        if self.par['exclude_regions'] is None:
+            return
```
kbwestfall: A few recommendations:

- Given you effectively have the same `if` statement at line 987 before calling `_parse_exclude_regions`, I would have this raise an error (i.e., use `msgs.error`).  If you'd rather not, I would use `return None, None` instead. 
- Also, you don't need the `else` block below this.  I.e., remove the `else` line and change the indentation of the rest of the code.
- I would explicitly set the datatype of the arrays to `int` and do the conversion from `str` to `int` in the `for` loop.  Something like: `dets[i], reg_start[i], reg_end[i] = [int(s) for s in region.split(':')]`.  That way you don't need to do the `int` conversion in line 990.

pypeit/edgetrace.py
```diff
@@ -519,6 +519,7 @@ def _init_internals(self):
         self.coeff_t = None             # Coefficients of the x-correlation between LEFT edges predicted
                                         # by the optical model and traced on the image.
         self.maskfile = None            # File used to slurp in slit-mask design
+        self.slitmask = None            # SlitMask instance that hold info on slitmask design
```
kbwestfall, line 522: Nice.  Moving slitmask to this class makes more sense than having it in `spectrograph`.

pypeit/pypeit.py
```diff
@@ -582,9 +582,16 @@ def reduce_exposure(self, frames, bg_frames=None, std_outfile=None):
             # get object positions from slitmask design and slitmask offsets for all the detectors
             spat_flexure = np.array([ss.spat_flexure for ss in sciImg_list])
             platescale = np.array([ss.detector.platescale for ss in sciImg_list])
+            # get the dither offset if available
+            if self.par['reduce']['slitmask']['use_dither_offset']:
+                dither = self.spectrograph.parse_dither_pattern(
+                    ['{}/{}'.format(self.fitstbl['directory'][frames[0]], self.fitstbl['filename'][frames[0]])])
```
kbwestfall: Anytime we're constructing file paths, I think we should be using `os.path.join`.  I would do that here, or you can use the `self.fitstbl.frame_paths` function.

pypeit/slittrace.py
```diff
@@ -1438,8 +1447,8 @@ def average_maskdef_offset(calib_slits, platescale):
                       'Especially for dithered observations!')
             msgs.warn('To provide a value set `slitmask_offset` in `SlitMaskPar`')
         else:
-            # define which are the blue and red detectors. This is hard coded for DEIMOS but no other
-            # instrument should ever get to this point.
+            # define which are the blue and red detectors. This is hard coded for DEIMOS. However it works
+            # perfectly fine also with MOSFIRE (although the nomenclature "red and blue dets" does not apply).
```
kbwestfall: Maybe just flag this as an issue for a later PR, but I worry this will lead to confusion.  I.e., even though the code works fine for MOSFIRE, the code reads as DEIMOS specific and will likely confuse someone who tries to read/alter the code in the long term.  It would be good to think of a way to abstract this.

On abstracting this, here are a few thoughts/questions:  Are the blue and red detectors treated completely independently?  Could you pull out the code starting at line 1457 into a core function that accepts an array of detectors and returns `calib_dets` for those detectors?  Then you could have a `Spectrograph` function that returns the relevant detector numbers.  You'd have to call that function twice for DEIMOS and once for MOSFIRE.

Also, unrelated to the MOSFIRE-specific changes, the control flow in this function is awkward.  FWIW, it seems like you should be able to have two if statements that return if either is true, and then have everything in the currently buried `else` block execute if both of them aren't.

pypeit/spectrographs/keck_mosfire.py
kbwestfall: Do we know what will cause these lines to throw errors?  Is it because the specific header keys may not exist?  In general, I think `try` blocks should be as small as possible because the larger they get, the harder they are to debug.  I would put small `try` blocks around accessing each header keyword, as annoying as that is.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -432,3 +518,305 @@ def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table,
...
+                msgs.info('{0:^5d}{1:^14d} {2:^9.3f} {3:^12.3f}    {4:^16.2f} {5:^14.2f}'.format(num,
+                            self.slitmask.slitid[sortindx][i], self.slitmask.onsky[:,2][sortindx][i],
+                            self.slitmask.onsky[:,3][sortindx][i], top_edges[sortindx][i], bot_edges[sortindx][i]))
+                num += 1
+            msgs.info('*' * 92)
+
+        return top_edges, bot_edges, sortindx, self.slitmask
+
+
+def _CSUnumslits():
```
kbwestfall: I think all of the functions below should be methods of the MOSFIRE spectrograph class; you can make them `staticmethod`s, as necessary.

kbwestfall (changes requested):
Excellent PR!  Attention to the docs and unit tests is much appreciated!  I have a few mostly minor comments and requests.

pypeit/edgetrace.py
debora-pe: Yes, these are the expected positions before the x-correlation. I changed their names (I was already thinking about doing it) and I have explicitly say in the description that they are before x-correlation

pypeit/edgetrace.py
debora-pe: Thank you for the suggestions. I implemented them!

pypeit/pypeit.py
debora-pe: Thank you! I used `self.fitstbl.frame_paths`

pypeit/slittrace.py
debora-pe: Thank you for pushing on this. I have been lazy about changing it. I followed your suggestions, let me know if you think it looks good.

pypeit/spectrographs/keck_mosfire.py
debora-pe: I actually don't know why there is a `try` block. It was already here. I didn't see any case for which the specific header keys don't exist, but I'll investigate more.

pypeit/spectrographs/keck_mosfire.py
debora-pe: Thank you so much!!! I never understood what `staticmethod` does before now. I moved everything to MOSFIRE spectrograph class.

pypeit/slittrace.py
```diff
@@ -1401,93 +1418,102 @@ def get_maskdef_objpos_offset_alldets(sobjs, calib_slits, spat_flexure, platesca
...
+    if not blue_and_red:
+        _, median_off, _ = sigma_clipped_stats(slitmask_offsets, sigma=2.)
+        for cs in calib_slits:
+            # assign median to each det
+            cs.maskdef_offset = median_off
+        msgs.info('Average Slitmask offset: {:.2f} pixels ({:.2f} arcsec).'.format(median_off, median_off * platescale))
+
+        return calib_slits
+
+    else:
```
kbwestfall: I don't think you need this `else`.

pypeit/slittrace.py
```diff
@@ -1401,93 +1418,102 @@ def get_maskdef_objpos_offset_alldets(sobjs, calib_slits, spat_flexure, platesca
...
-                if calib_slits[i].det in blue_dets:
-                    calib_slits[i].maskdef_offset = median_off_blue
-                if calib_slits[i].det in red_dets:
-                    calib_slits[i].maskdef_offset = median_off_red
+            # if median was not computed for blue dets, use the remaining (red) dets to compute the median for all dets
+            _, median_off, _ = sigma_clipped_stats(slitmask_offsets, sigma=2.)
+            for cs in calib_slits:
+                cs.maskdef_offset = median_off
+            msgs.info('Average Slitmask offset for all the detectors: '
+                      '{:.2f} pixels ({:.2f} arcsec).'.format(median_off, median_off * platescale))
```
kbwestfall: I think you can combine this block and the identical block at line 1474.  I think starting with line 1472, you could do:
```
    indx_b = np.where(np.in1d(calib_dets, spectrograph_dets[0]))[0]
    if not blue_and_red or slitmask_offsets[indx_b].size == 0:
        _, median_off, _ = sigma_clipped_stats(slitmask_offsets, sigma=2.)
        for cs in calib_slits:
            # assign median to each det
            cs.maskdef_offset = median_off
        msgs.info('Average Slitmask offset: {:.2f} pixels ({:.2f} arcsec).'.format(median_off, median_off * platescale))
        return calib_slits

    # compute median if these dets have values of slitmask_offsets
    _, median_off, _ = sigma_clipped_stats(slitmask_offsets[indx_b], sigma=2.)
    for cs in calib_slits[indx_b]:
        ...
```

pypeit/spectrographs/spectrograph.py
```diff
@@ -469,6 +469,16 @@ def bpm(self, filename, det, shape=None, msbias=None):
...
 
+    def list_detectors(self):
+        """
+        List the detectors of this spectrograph, e.g., array([[1, 2, 3, 4], [5, 6, 7, 8]])
+        They are separated if they are split into blue and red detectors
+
+        This method is not defined for all spectrographs.
+
+        """
+        return None, None
```
kbwestfall: Nit-picking: I would do `return None, False` here, just to make sure that the last object is always a boolean.

kbwestfall (changes requested):
I had a quick look at the new updates, and I had a couple of more comments on the `slittrace` changes.

jhennawi (approved):
Excellent work @deborape. Thanks for this very valuable PR. MOSFIRE users will be very grateful. Please run the dev suite and tests before you merge.

pypeit/spectrographs/spectrograph.py
debora-pe: Done

pypeit/slittrace.py
debora-pe: Done. Thank you!

pypeit/slittrace.py
debora-pe: Removed!

pypeit/spectrographs/keck_mosfire.py
debora-pe: I removed the try block.

kbwestfall (approved):
Thanks!

debora-pe:
Dev Suite passes
```
--- PYPEIT DEVELOPMENT SUITE PASSED 135/135 TESTS (Masters ignored) ---
Testing Started at 2021-11-16T16:27:36.603520
Testing Completed at 2021-11-16T23:01:43.530707
Total Time: 6:34:06.927187
```

Unit tests pass except one
`====== 1 failed, 302 passed in 1550.08s (0:25:50) ========`
which I think it's related to the bug that should be fixed in PR #1294 
`FAILED test_coadd.py::test_coadd_datacube - TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'`
I can wait to merge after that PR clears.

debora-pe:
All good now.

Dev Suite
```
--- PYPEIT DEVELOPMENT SUITE PASSED 135/135 TESTS (Masters ignored) ---
Testing Started at 2021-11-18T10:32:26.105575
Testing Completed at 2021-11-18T15:23:03.103413
Total Time: 4:50:36.997838
```

Unit tests
```
pytest . -W ignore
============================ test session starts ============================
collected 303 items                                                         

test_alignments.py .                                                  [  0%]
test_arc.py .                                                         [  0%]
test_archive.py ..                                                    [  1%]
test_arcimage.py ..                                                   [  1%]
test_biasframe.py ..                                                  [  2%]
test_bitmask.py ....                                                  [  3%]
test_bpmimage.py ..                                                   [  4%]
test_bspline.py ...............                                       [  9%]
test_buildimage.py ..                                                 [ 10%]
test_calibrations.py .......                                          [ 12%]
test_chk_calibs.py ..                                                 [ 13%]
test_coadd.py .                                                       [ 13%]
test_collate_1d.py ............                                       [ 17%]
test_cooked.py .                                                      [ 17%]
test_coremeta.py .                                                    [ 18%]
test_datacontainer.py ........                                        [ 20%]
test_detector.py ...                                                  [ 21%]
test_display.py .                                                     [ 22%]
test_edgetrace.py .                                                   [ 22%]
test_fitting.py ....                                                  [ 23%]
test_flatfield.py .                                                   [ 24%]
test_flexure.py ..                                                    [ 24%]
test_flux.py ....                                                     [ 26%]
test_fluxspec.py .....                                                [ 27%]
test_frametype.py ..                                                  [ 28%]
test_history.py ....                                                  [ 29%]
test_load.py .                                                        [ 30%]
test_load_images.py ................                                  [ 35%]
test_maskdesign_matching.py ...                                       [ 36%]
test_masterframe.py .                                                 [ 36%]
test_match.py ....                                                    [ 37%]
test_metadata.py ......                                               [ 39%]
test_moment.py ......                                                 [ 41%]
test_msgs.py ..                                                       [ 42%]
test_opticalmodel.py ....                                             [ 43%]
test_parse.py ...                                                     [ 44%]
test_pca.py ........                                                  [ 47%]
test_processimage.py ..                                               [ 48%]
test_procimg.py .....                                                 [ 49%]
test_pydl.py .                                                        [ 50%]
test_pypeit.py ..                                                     [ 50%]
test_pypeitimage.py ..                                                [ 51%]
test_pypeitpar.py ................................                    [ 62%]
test_pypeitsetup.py ........                                          [ 64%]
test_qa.py .                                                          [ 65%]
test_runpypeit.py ..                                                  [ 65%]
test_scienceimage.py ...                                              [ 66%]
test_scripts.py .................                                     [ 72%]
test_setups.py .....................                                  [ 79%]
test_skysub.py ..                                                     [ 79%]
test_slitmask.py ...                                                  [ 80%]
test_slittrace.py ....                                                [ 82%]
test_spec2dobj.py ......                                              [ 84%]
test_specobj.py .......                                               [ 86%]
test_specobjs.py .....                                                [ 88%]
test_spectrographs.py ....................                            [ 94%]
test_syncspec.py .                                                    [ 95%]
test_telluric.py ..                                                   [ 95%]
test_traceimage.py .                                                  [ 96%]
test_utils.py ...                                                     [ 97%]
test_vcorr.py ..                                                      [ 97%]
test_wavemodel.py .                                                   [ 98%]
test_wavetilts.py ....                                                [ 99%]
test_wvcalib.py ..                                                    [100%]

===================== 303 passed in 2181.93s (0:36:21) ======================
```

debora-pe:
merging

Issue #1283, opened 2021-10-30
Correct "Flexure Correction" documentation


tbowers7:
The flexure correction documentation (https://pypeit.readthedocs.io/en/latest/flexure.html#flexure) states that spectral flexure correction is enabled by default.

However, the default FlexurePar key `spec_method` actually defaults to `skip`.  This parameter is set to either `boxcar` or `slitcen` for the following spectrographs only:
* Gemini/GMOS (N&S)
* Keck DEIMOS
* Keck KCWI
* Keck LRIS (all)
* LBT MODS (all)
* MMT Binospec
* NTT EFOSC2
* Shane KAST (all)
* VLT FORS2

debora-pe:
Addressed in PR #1285 
Closing this.

tbowers7:
Thanks for adding this!

PR #1285, opened 2021-11-03
Fix flatfield bitmask bug and address issues #1281 and #1283 

debora-pe:
In `flatfield.py`, even if the flat field was not successful, the bitmask was not flagged as `BADFLATCALIB`. This is now fixed.

Also, addresses issues #1281 and #1283

### DIFF
--- a/doc/flexure.rst
+++ b/doc/flexure.rst
@@ -76,20 +76,33 @@ This will:
 Spectral
 ========
 
-By default, the code will calculate a flexure shift based on the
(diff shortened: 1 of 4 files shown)


debora-pe:
For issue #1281 

Screen Shot 2021-11-02 at 3 04 24 PM

profxj (approved):
Thx for this!

kbwestfall (approved):
Excellent!

PR #1303, opened 2021-12-11
Disallow fz files to be used for the reduction

rcooke-ast:
fz files change the extension where the data and header information are loaded, so I've fixed the soar goodman spectrograph file, and added an error for those that include fz files. There is a related PR in the dev suite

### DIFF
--- a/pypeit/pypeitsetup.py
+++ b/pypeit/pypeitsetup.py
@@ -25,6 +25,9 @@ from pypeit.par.util import parse_pypeit_file, make_pypeit_file
 from pypeit.spectrographs.util import load_spectrograph
 
 
+bad_ext = [".fz"]
+
+
 # TODO: Instantiation should just automatically trigger the run
 # method...
 class PypeItSetup:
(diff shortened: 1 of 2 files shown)


pypeit/pypeitsetup.py
```diff
@@ -25,6 +25,9 @@
 from pypeit.spectrographs.util import load_spectrograph
 
 
+bad_ext = [".fz"]
```
kbwestfall, line 28: Please put this inside the `check_file_list` function.

kbwestfall (approved):
My only request was to move where you define `bad_ext` to inside the main function.  Other than that, I had the same question as @profxj about the raw format provided by the telescope.  Given that it's just a change in the extension number, it would be nice if this could be something the code handles automatically.  But a quick look at the control flow suggests this is easier said than done...  Along the same lines, I'm a little hesitant to place a blanket ban of the `fz` format for all instruments, but making it specific to `soar_goodman` also seems to be non-trivial.

rcooke-ast:
OK - I'll revert these changes. It seems that they recommend downloading and using `.fz` files. I'll add a note in the docs that only `.fits.fz` files can be used to reduce soar+goodman data.

rcooke-ast:
You're quite right, @kbwestfall - it's not trivial to handle both `.fits.fz` and `.fits`, as I tried doing that first  :-)

PR #1305, opened 2021-12-16
Fixing an overflow error when polynomial fits get too large.

jhennawi:
This one line PR fixes a failure mode that impacts spectral scaling and telluric fitting codes where we exp(polynomial) to enforce positivity. In cases where the polynomial initialises to a bad location this can lead to overflow. The clipping introduced in this one line PR fixes this failure mode.

### DIFF
--- a/pypeit/core/coadd.py
+++ b/pypeit/core/coadd.py
@@ -7,6 +7,7 @@ Coadding module.
 """
 
 import os
+import sys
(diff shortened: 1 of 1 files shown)


profxj (approved):
looks fine to me.

feigewang (approved):
Looks good!

PR #1312, opened 2022-01-11
Make bottleneck an optional dependency

finagle29:
Installing Bottleneck via pip is occasionally troublesome, so this PR makes Bottleneck an optional dependency. If it's installed, then we'll use it for a fast moving median, otherwise reverting to previous python code. The approach I took is similar to the bspline module.

The bug in the pure python code I identified in #1164, where an even window size results in an incorrect median calculation, is resolved by splitting the logic into two functions, one for even window size and one for odd window size.


pypeit/move_median/mmpy.py
```diff
@@ -0,0 +1,59 @@
+from collections import deque
+from bisect import insort, bisect_left
+import itertools
+
+def move_median(seq_pad, window_size):
+    if window_size % 2 == 0:
+        return move_median_even_window(seq_pad, window_size)
+    else:
+        return move_median_odd_window(seq_pad, window_size)
```
kbwestfall: This is nit-picky ...  Is there any loss in efficiency if you make these two functions the same and put the check for even vs. odd behavior where it's needed?  I.e., there are only two lines that are different between the two functions, and you would change them to:
```
result.append(s[len(d)//2] if odd else (s[len(d)//2] + s[len(d)//2 - 1])/2)
result.append(s[m] if odd else (s[m] + s[m-1])/2)
```
where you would set `odd = window_size % 2 == 0` at the beginning of the function.

kbwestfall (changes requested):
Thanks for doing this!  I have one nit-picky comment on the odd vs. even functions.

Other than that, my one request is that you add some text to the installation doc (`doc/installing.rst`) in the "Install via pip`` section that briefly describes the optional dependency and how to install it.  I.e., people should just be able to do `pip install "pypeit[bottleneck]"`, right?  Also, should we add `bottleneck` to the `dev` dependencies?

pypeit/move_median/mmpy.py
jhennawi: I agreed with @kbwestfall suggestion here.

pypeit/move_median/move_median.py
```diff
@@ -0,0 +1,11 @@
+import warnings
+
+try:
+    from bottleneck import move_median as move_median_
+    def move_median(seq, window):
+        return move_median_(seq, window=window)
+    warnings.warn("using bottleneck")
+except:
+    warnings.warn('Unable to load bottleneck moving median.  Try reinstalling bottleneck.  In the '
+                  'meantime, falling back to pure python code.')
```
jhennawi: falling back on the slower pure python code

setup.cfg
```diff
@@ -57,6 +56,8 @@ pyside2 =
     pyside2>=5.0
 pyqt5 =
     pyqt5
+bottleneck =
```
jhennawi, line 59: Can we update the docs on installation to reflect these changes. In other words, we need to mention bottleneck there, we should mention how to install (both pip and developer etc.) with bottleneck, and also without bottleneck.

jhennawi (changes requested):
Thanks for this @finagle29. Minor code changes requested and a doc update.

pypeit/move_median/mmpy.py
finagle29: Good call! I thought that my original way would be better-performing, but I ran some quick performance tests and the difference is at the 0.1% level, so I've consolidated the even/odd behavior into one function.

kbwestfall:
Tests pass after removing the `shapely` dependency:
```
======================= 318 passed in 1285.51s (0:21:25) =======================
```

kbwestfall (approved):
Approving now.  But we should wait and try to work out other installation kinks here before merging.

jhennawi (approved):
Thanks!

PR #1319, opened 2022-01-25
Skymask noise bug

debora-pe:
This deals with a bug during the sky subtraction. Here are the changes:

- sky model is included in the 2nd pass of global sky subtraction (not for IR redux)
- skymask is now computed also for the maskdef_extract objects
- Improvements in the maskdef extraction:
              -- dedicated fwhm and boxcar_radius for maskdef_extract objects. They can be input as parsets. Also fwhm (if not provided by the user) will be computed using the average fwhm of the detected objects
              -- added a new parset `skip_serendip`. If a detected object is considered a serendipitous detection the user can decide not to extract it.
- `find_fwhm` is back to a smaller value for DEIMOS (find_fwhm = 7pixels ~ 0.8'')
- (unrelated) added pypeit_version to the pypeit file header (it's been in the todo list for awhile)

Associated Dev Suite PR is here: https://github.com/pypeit/PypeIt-development-suite/pull/186

PR #1321, opened 2022-01-27
Coadd2d maskdef

debora-pe:
This PR makes maskdef matching available for `pypeit_coadd_2dspec`, like it already is for the main reduction (only for the supported instruments).  Many more improvements are also added (see below). 
_Note_: **(1)** I had to pull in some of the changes I made in PR #1319 (mostly in `slittrace.py`), so please refer to that PR for those. **(2)** I made a change to the `Spec2DObj` data model, so `pypeit_coadd_2dspec` will not work with a older version `Spec2DObj`.

**_Changes:_**
- `maskdef_id` assigned to each slit
- assigning object's name, ra and dec to detected objects is now available
- force extract undetected objects is now available
- `maskdef_offset` can be use as offsets in the coadd
- coadding only a specific sets of slits is now possible with the parset `--only_slits`
- if the user inputs a list of offsets, the weights can still be computed if a bright object is found, otherwise uniform weigths will be used
- fixed manual extraction bug
- various improvements in the flow of the code
- spec1d*.txt is now produced also for coadd2d

Associated Dev Suite PR is here: https://github.com/pypeit/PypeIt-development-suite/pull/185


debora-pe:
Dev Suite passes
```
Testing Started at 2022-01-28T17:36:03.634963
Testing Completed at 2022-01-29T09:01:39.652428
Total Time: 15:25:36.017465
```

pypeit/core/coadd.py
```diff
@@ -2855,7 +2859,7 @@ def compute_coadd2d(ref_trace_stack, sciimg_stack, sciivar_stack, skymodel_stack
 
     # TODO -- If weights is a numpy.ndarray, how can this not crash?
     #   Maybe the doc string above is inaccurate?
-    if 'uniform' in weights:
+    if  weights == 'uniform':
```
kbwestfall: Can we change the default to `weights=None` instead of `weights='uniform'`, and make the relevant change here?  I think it would make more sense that way.

pypeit/core/coadd.py
```diff
@@ -2920,11 +2924,21 @@ def compute_coadd2d(ref_trace_stack, sciimg_stack, sciivar_stack, skymodel_stack
         dspat_img_fake = spat_img_coadd + dspat_mid[0]
         dspat[np.invert(outmask)] = dspat_img_fake[np.invert(outmask)]
 
+    # update maskdef_objpos and maskdef_slitcen with the new value in the new slit
+    new_maskdef_objpos = None
+    new_maskdef_slitcen = None
+    if (maskdef_dict['maskdef_objpos'] is not None) and (maskdef_dict['maskdef_slitcen']):
```
kbwestfall: Nit picking, but you don't need the parentheses here.

pypeit/scripts/coadd_2dspec.py
```diff
@@ -87,9 +85,11 @@ def main(args):
             # the PypeIt run that extracted the objects?  Why are we not
             # just passing the pypeit file?
             # JFH: The reason is that the coadd2dfile may want different reduction parameters
-            spectrograph_def_par = spectrograph.default_pypeit_par()
-            parset = par.PypeItPar.from_cfg_lines(cfg_lines=spectrograph_def_par.to_config(),
-                                                    merge_with=config_lines)
+            # DP: I think config_specific_par() is more appropriate here. default_pypeit_par()
+            # is included in config_specific_par()
+            spectrograph_cfg_lines = spectrograph.config_specific_par(spec2d_files[0]).to_config()
```
kbwestfall, line 90: If it works, great!  But I thought `config_specific_par` typically expected a raw file.  We might want to add a comment here that this works with the spec2d files specifically because of how we construct the header of those files.  I.e., if we suddenly lose the required header parameters in the spec2d files, this will fault.  Regardless, I agree that we want the result of `config_specific_par`, not just the `default_pypeit_par`.

pypeit/scripts/coadd_2dspec.py
```diff
@@ -99,20 +99,17 @@ def main(args):
             head0 = fits.getheader(spec2d_files[0])
             spectrograph_name = head0['PYP_SPEC']
             spectrograph = load_spectrograph(spectrograph_name)
-            parset = spectrograph.default_pypeit_par()
+            spectrograph_cfg_lines = spectrograph.config_specific_par(spec2d_files[0]).to_config()
```
kbwestfall, line 104: Same thought about adding a relevant comment here.

pypeit/scripts/coadd_2dspec.py
```diff
@@ -99,20 +99,17 @@ def main(args):
...
-                                              merge_with=parset.to_config())
+            return msgs.error('You must provide either a coadd2d file (--file) or an object name (--obj)')
 
         # If detector was passed as an argument override whatever was in the coadd2d_file
         if args.det is not None:
             msgs.info("Restricting reductions to detector={}".format(args.det))
-            parset['rdx']['detnum'] = par.util.eval_tuple(args.det.split(','))
+            # TODO this needs to be adjusted if we want to pass (as inline command) more than one detector
+            #  and also if we are combining mosaic detectors
+            parset['rdx']['detnum'] = int(args.det)
```
kbwestfall, line 115: My guess is that I put in that line.  I checked and there's no `coadd_2dspec` test for `gemini_gmos`.  But, as annoying as it will be, I think we should add such a test to make sure the 2D coadds work for mosaics.  We should discuss whether we do that now or in a different PR.

pypeit/scripts/coadd_2dspec.py
```diff
@@ -140,7 +137,7 @@ def main(args):
         # Write the par to disk
         par_outfile = basename+'_coadd2d.par'
         print("Writing the parameters to {}".format(par_outfile))
-        parset.to_config(par_outfile)
+        parset.to_config(par_outfile, exclude_defaults=True, include_descr=False)
```
kbwestfall, line 143: Nice :)

pypeit/slittrace.py
```diff
@@ -957,12 +958,13 @@ def assign_maskinfo(self, sobjs, plate_scale, slits_left, TOLER=1.):
             cut_sobjs = sobjs[on_det]
             if cut_sobjs.nobj == 0:
                 msgs.warn('NO detected objects.')
-                return
+                return sobjs
         else:
             msgs.warn('NO detected objects.')
-            return
+            return sobjs
```
kbwestfall: Edit the docstring to include a description of the returned object.

pypeit/slittrace.py
```diff
@@ -1035,17 +1040,21 @@ def assign_maskinfo(self, sobjs, plate_scale, slits_left, TOLER=1.):
...
-                # Assign
-                sobj.RA = new_obj_coord.ra.value
-                sobj.DEC = new_obj_coord.dec.value
-                sobj.MASKDEF_OBJNAME = 'SERENDIP'
-                sobj.MASKDEF_EXTRACT = False
-                sobj.hand_extract_flag = False
+                if skip_serendip:
+                    remove_idx.append(ss)
+                    msgs.info('Skipping SERENDIP object at pixel {}'.format(cut_sobjs[ss].SPAT_PIXPOS))
+                else:
```
kbwestfall: I would do:
```
if skip_serendip:
    remove_idx.append(ss)
    msgs.info('Skipping SERENDIP object at pixel {}'.format(cut_sobjs[ss].SPAT_PIXPOS))
    continue

sobj = cut_sobjs[ss]
...
```
I.e., I would use `continue` to force the loop to go to the next index without executing the rest of the code, meaning you don't need the `else` indentation.

pypeit/slittrace.py
```diff
@@ -1294,6 +1305,49 @@ def get_maskdef_offset(self, sobjs, slits_left, platescale, slitmask_off, bright
                       f'Slitmask offset cannot be estimated in {self.detname}.')
             self.maskdef_offset = 0.0
 
+        return
```
kbwestfall: If a function doesn't return anything, you don't need the `return` statement at the end.

pypeit/coadd2d.py
```diff
@@ -169,39 +173,117 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
+            msgs.error('Not all of your exposures have the same spectral dimension. Check your inputs')
         self.nspec = self.stack_dict['slits_list'][0].nspec
+
+        # Check that binning is the same for all the exposures
+        binspec_list = [slits.binspec for slits in self.stack_dict['slits_list']]
+        binspat_list = [slits.binspat for slits in self.stack_dict['slits_list']]
+        if not len(set(binspec_list)) == 1:
+            msgs.error('Not all of your exposures have the same spectral binning. Check your inputs')
+        if not len(set(binspat_list)) == 1:
+            msgs.error('Not all of your exposures have the same spatial binning. Check your inputs')
```
kbwestfall, line 217: Nice input checks!

pypeit/coadd2d.py
```diff
@@ -169,39 +173,117 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
+        Returns:
+            :obj:`float`: Number of slits of the coadded frame
+
+        """
+        # this is the number of slits of the coadded frame
+        if self.good_slits is None:
+            msgs.error('Cannot determine nslits_coadded because `self.good_slits` is `None`. '
+                       'Run `self.good_slits = self.good_slitindx(only_slits=only_slits)` first')
+        else:
+            return self.good_slits.size
```
kbwestfall: You don't need the `else`; the `msgs.error` call halts the control flow.

pypeit/coadd2d.py
```diff
@@ -169,39 +173,117 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
+                List of slits to combine. It must be `slitord_id`
+
+        Returns:
+            `numpy.ndarray`_: array of index of good slits in the un-coadded frames
+        """
+        # TODO We should be checking the bitmask for the reductions or something here??
+        #  Yes, definitely - DONE?
+
+        only_slits = [only_slits] if (only_slits is not None and
+                                      isinstance(only_slits, (int, np.int, np.int64, np.int32))) else only_slits
```
kbwestfall: You can use `isinstance(only_slits, (int, np.integer))` to catch all the possible numpy integer types as well as the built-in `int`.

pypeit/coadd2d.py
```diff
@@ -169,39 +173,117 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
+
+        # This creates a unified bpm common to all frames
+        slits0 = self.stack_dict['slits_list'][0]
+        # bpm for the first frame
+        reduce_bpm = (slits0.mask > 0) & (np.invert(slits0.bitmask.flagged(slits0.mask,
+                                                                           flag=slits0.bitmask.exclude_for_reducing)))
```
kbwestfall, line 247: I think these are boolean arrays are degenerate.  By selecting mask values that are `>0`, you're automatically selecting on all bits.  Then you select only the ones that are flagged with the bits specified by `slits0.bitmask.exclude_for_reducing`, which should just be a subset of the ones selected by the first condition, meaning the bits that meet both conditions is identical to the set the only meets the second condition.  All of this is to say that I don't think you need the first `(slits0.mask > 0)` condition...

pypeit/coadd2d.py
```diff
@@ -169,39 +173,117 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
+        for i in range(1, self.nexp):
+            # update bpm with the info from the other frames
+            slits = self.stack_dict['slits_list'][i]
+            reduce_bpm & (slits.mask > 0) & (np.invert(slits.bitmask.flagged(slits.mask,
```
kbwestfall: Does this line do anything?  There's no assignment.  Should it be `reduce_bpm &= ...` ?

Also, for these bpms, should you be using `&` or `|`?

pypeit/coadd2d.py
```diff
@@ -169,39 +173,117 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
+                                                                             flag=slits.bitmask.exclude_for_reducing)))
+        # this are the good slit index according to the bpm mask
+        good_slitindx = np.where(np.invert(reduce_bpm))[0]
```
kbwestfall: Nit-picking ...  I've been trying to use `np.logical_not` instead of `np.invert` because of the more stable behavior of the former.

pypeit/coadd2d.py
```diff
@@ -169,39 +173,117 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
+
+        # If we want to coadd all the good slits
+        if only_slits is None:
+            return good_slitindx
+
+        # If instead we want to coadd only a selected (by the user) number of slits
+        else:
```
kbwestfall: Don't need the else block because the `return` statements interrupts the control flow.

pypeit/coadd2d.py
```diff
@@ -278,49 +359,38 @@ def coadd(self, only_slits=None, interp_dspat=True):
...
-            # (nslit, nexp) array?
-            # This one line deals with the different weighting strategies between MultiSlit echelle. Otherwise, we
-            # would need to copy this method twice in the subclasses
-            if 'auto_echelle' in self.use_weights:
-                rms_sn, weights = self.optimal_weights(slitord_id, self.objid_bri)
+            # maskdef info
+            maskdef_dict = self.get_maskdef_dict(slit_idx, ref_trace_stack)
+
+            # weights
+            if (not isinstance(self.use_weights, str)) and (self.use_weights.ndim > 2):
```
kbwestfall: Don't need the parentheses.

pypeit/coadd2d.py
```diff
@@ -516,6 +615,22 @@ def reduce(self, pseudo_dict, show=None, show_peaks=None, basename=None):
             sciImage.image, show_peaks=show_peaks, save_objfindQA=True,
             manual_extract_dict=manual_dict)
 
+        # maskdef stuff
+        if parcopy['reduce']['slitmask']['assign_obj'] is True and slits.maskdef_designtab is not None:
```
kbwestfall: Don't need the `is True` part.

pypeit/coadd2d.py
```diff
@@ -581,21 +696,44 @@ def snr_report(self, snr_bar, slitid=None):
...
+    #         only_slits:
+    #
+    #     Returns:
+    #
+    #     """
+    #
+    #     only_slits = [only_slits] if (only_slits is not None and
+    #                                     isinstance(only_slits, (int, np.int, np.int64, np.int32))) else only_slits
+    #     good_slits = np.arange(self.nslits) if only_slits is None else only_slits
+    #     return good_slits
```
kbwestfall: Remove the commented code?

pypeit/coadd2d.py
```diff
@@ -581,21 +696,44 @@ def snr_report(self, snr_bar, slitid=None):
...
 
         Returns:
 
         """
 
-        only_slits = [only_slits] if (only_slits is not None and
-                                        isinstance(only_slits, (int, np.int, np.int64, np.int32))) else only_slits
-        good_slits = np.arange(self.nslits) if only_slits is None else only_slits
-        return good_slits
+        if (offsets_method is not None) and (offsets is not None):
```
kbwestfall: Don't need the parentheses.

pypeit/coadd2d.py
```diff
@@ -753,32 +895,39 @@ def load_coadd2d_stacks(self, spec2d, chk_version=False):
...
+        Returns:
+            :obj:`list` or `numpy.ndarray`_: User input values
         """
-        pass
+        if isinstance(input, (list, np.ndarray)):
+            if len(input) != self.nexp:
+                msgs.error(f'If {type} are input it must be a list/array with same number of elements as exposures')
+            return np.atleast_1d(input)
+        else:
+            msgs.error(f'Unrecognized format for {type}')
```
kbwestfall: Don't need the `else`.

pypeit/coadd2d.py
```diff
@@ -839,132 +1002,148 @@ def __init__(self, spec2d_files, spectrograph, par, det=1, offsets=None, weights
...
+        # or Yes bright object but the user wants still to use uniform weights
+        if ((self.objid_bri is None) and (weights in ['auto', 'uniform'])) or \
+                ((self.objid_bri is not None) and (weights == 'uniform')):
+            # use uniform weights
+            self.use_weights = 'uniform'
+            if weights == 'auto':
+                # warn if the user had put `auto` in the parset
+                msgs.warn('Weights cannot be computed because no unique reference object '
+                          'with the highest S/N was found. Using uniform weights instead.')
+            if weights == 'uniform':
```
kbwestfall: Use `elif`?

pypeit/coadd2d.py
```diff
@@ -839,132 +1002,148 @@ def __init__(self, spec2d_files, spectrograph, par, det=1, offsets=None, weights
...
+                traces_rect[:, iexp] = sobjs_exp.TRACE_SPAT
+            # Now deterimine the offsets. Arbitrarily set the zeroth trace to the reference
+            med_traces_rect = np.median(traces_rect,axis=0)
+            offsets = med_traces_rect[0] - med_traces_rect
+            if self.debug_offsets:
+                for iexp in range(self.nexp):
+                    plt.plot(traces_rect[:, iexp], linestyle='--', label='original trace')
+                    plt.plot(traces_rect[:, iexp] + offsets[iexp], label='shifted traces')
+                    plt.legend()
+                plt.show()
```
kbwestfall, line 1116: Is this worth formalizing into a QA plot?

pypeit/coadd2d.py
```diff
@@ -1021,11 +1201,11 @@ def get_brightest_obj(self, specobjs_list, spat_ids):
...
         objid = objid_max[slitid, :]
         if (snr_bar_mean == -np.inf):
-            msgs.error('You do not appear to have a unique reference object that was traced as the highest S/N '
+            msgs.warn('You do not appear to have a unique reference object that was traced as the highest S/N '
                        'ratio on the same slit of every exposure')
-
-        self.snr_report(snr_bar, slitid=spat_ids[slitid])
-        return objid, slitid, spat_ids[slitid], snr_bar
+            return None, None, None, None
+        else:
```
kbwestfall: Don't need the `else`.

pypeit/coadd2d.py
```diff
@@ -1044,6 +1224,65 @@ def reference_trace_stack(self, slitid, offsets=None, objid=None):
...
+            shifted according the frames offsets. Shape is (nspec, nimgs).
+
+        Returns:
+            :obj:`dict`: Dictionary containing all the maskdef info. The quantities saved
+            are: maskdef_id, maskdef_objpos, maskdef_slitcen, maskdef_designtab. To learn what
+            they are see :class:`~pypeit.slittrace.SlitTraceSet` datamodel.
+
+        """
+        # maskdef info
+        if (self.par['calibrations']['slitedges']['use_maskdesign'] is True) and \
```
kbwestfall: Don't need the parentheses or the `is True`.

pypeit/coadd2d.py
```diff
@@ -1077,29 +1316,82 @@ def __init__(self, spec2d_files, spectrograph, par, det=1, offsets=None, weights
...
-        if offsets is None:
-            self.objid_bri, self.slitid_bri, self.snr_bar_bri = self.get_brightest_obj(self.stack_dict['specobjs_list'], self.nslits)
-        else:
-            # TODO -- Check the input offsets list matches the science images
-            pass
+        # get offsets and weights
+        self.compute_offsets_and_weights(weights, offsets)
 
-        self.use_weights = self.parse_weights(weights)
+    def compute_offsets_and_weights(self, weights, offsets):
```
kbwestfall: There are parts of this function that are common to both the MultiSlit and Echelle versions?  Can you strip out the common parts into the base class?

pypeit/coadd2d.py
```diff
@@ -1130,21 +1422,25 @@ def get_brightest_obj(self, specobjs_list, nslits):
...
             snr_bar[iexp] = snr_bar_vec[snr_bar_vec.argmax()]
-
-        self.snr_report(snr_bar)
-
-        return objid, None, snr_bar
+        if 0 in snr_bar:
+            msgs.warn('You do not appear to have a unique reference object that was traced as the highest S/N '
+                      'ratio for every exposure')
+            return None, None, None
+        else:
```
kbwestfall: Don't need the `else`.

kbwestfall (changes requested):
Lot's of nit-picky stuff from me.  The only comment that I'd like a specific response to is the issue of the `detnum` parsing.  We should test how this works for mosaic'd data, but I'm fine with pushing that off to a later PR.

pypeit/coadd2d.py
debora-pe: Hi @kbwestfall ! Thanks for this comments. I remember spending a lot of time in the past trying to figure out how bitmask works and I don't think I still fully understand it. But what you says does not work here because of the following example:
```
In [17]: slits.mask
Out[17]: array([ 0, 82,  0], dtype=int16)

In [18]: (slits.mask > 0) & (np.invert(slits.bitmask.flagged(slits.mask, flag=slits.bitmask.exclude_for_reducing)))
Out[18]: array([False,  True, False])

In [19]: slits.mask > 0
Out[19]: array([False,  True, False])

In [20]: np.invert(slits.bitmask.flagged(slits.mask, flag=slits.bitmask.exclude_for_reducing))
Out[20]: array([ True,  True,  True])
```
In this case I definitely want to mask the middle slit and the combination of the conditions above gives me the right mask, as it does only `slits.mask > 0`, but the second condition with `slits.bitmask.flagged` alone does not give me the right masking. Maybe it is the second one that is redundant, but there may be slits with one of the flags specified in `slits.bitmask.exclude_for_reducing` that we don't want to mask. I don't know but it's very confusing.

pypeit/coadd2d.py
debora-pe: Thanks!

pypeit/coadd2d.py

pypeit/coadd2d.py
debora-pe: Thanks! Changed.

pypeit/coadd2d.py
debora-pe: Thanks! I forgot the `=`. now added.

pypeit/coadd2d.py
debora-pe: Changed

pypeit/coadd2d.py

pypeit/coadd2d.py

pypeit/coadd2d.py

pypeit/coadd2d.py
debora-pe: removed

pypeit/coadd2d.py
debora-pe: removed

pypeit/coadd2d.py
debora-pe: removed

pypeit/coadd2d.py

pypeit/coadd2d.py
debora-pe: removed

pypeit/coadd2d.py

pypeit/coadd2d.py
debora-pe: It's good point. If you don't mind I'll do that in another PR.

pypeit/coadd2d.py
debora-pe: removed

pypeit/core/coadd.py
debora-pe: That is a good point. Do you mind if I do that in another PR?

pypeit/core/coadd.py
debora-pe: Removed

pypeit/scripts/coadd_2dspec.py
debora-pe: Thanks for pointing it out. I added a comment

pypeit/scripts/coadd_2dspec.py
debora-pe: Added a comment

pypeit/scripts/coadd_2dspec.py
debora-pe: Much needed :)

pypeit/slittrace.py
debora-pe: Added

pypeit/slittrace.py
debora-pe: Done

pypeit/slittrace.py
debora-pe: removed

pypeit/scripts/coadd_2dspec.py
debora-pe: The logic I follow here is the one currently used for `run_pypeit`. If we provide the detector in the inline call (`run_pypeit --det X`) currently we cannot pass a mosaic, see https://github.com/pypeit/PypeIt/blob/develop/pypeit/scripts/run_pypeit.py#L135
This is why I added the TODO. For now coadd of mosaic should work if we specify it in the `.pypeit` file (in this case `.coadd2d` file), but I think we should make it possible also for inline commands. But we can do it in another PR.

kbwestfall: I'm fine with punting this to another PR.  But, instead of deleting the previous code, would you mind commenting it out for now?

pypeit/coadd2d.py
kbwestfall: Thanks for the detailed response!  And sorry for not realizing you were inverting the second selection.  In that case, the second statement is finding all bits not masked with the `exclude_for_reducing` flags, which includes bits that are not masked at all.

If I understand correctly, you're trying to find all the slits flagged by anything *except* the flags in `exclude_for_reducing`.  Of course, the way you've done it works, but you could also create the relevant list of flags, which is admittedly requires some digging to figure out how to do.  I'm fine with you leaving this as is, but just for reference, here's how to do it with a single call to `flagged`:

```
>>> from pypeit.slittrace import SlitTraceBitMask
>>> bm = SlitTraceBitMask()
>>> import numpy as np
>>> mask = np.array([0, 82, 0], dtype=np.int16)
>>> np.setdiff1d(bm.keys(), bm.exclude_for_reducing)
array(['BADFLATCALIB', 'BADREDUCE', 'BADTILTCALIB', 'BADWVCALIB',
       'BOXSLIT', 'SHORTSLIT', 'USERIGNORE'], dtype='<U13')
>>> bm.flagged(mask, flag=np.setdiff1d(bm.keys(), bm.exclude_for_reducing))
array([False,  True, False])
```

kbwestfall (approved):
Thanks for all the hard work!

pypeit/core/coadd.py
```diff
@@ -2747,7 +2747,8 @@ def get_spat_bins(thismask_stack, trace_stack, spat_samp_fact=1.0):
 
 def compute_coadd2d(ref_trace_stack, sciimg_stack, sciivar_stack, skymodel_stack,
                     inmask_stack, tilts_stack,
-                    thismask_stack, waveimg_stack, wave_grid, spat_samp_fact=1.0,
+                    thismask_stack, waveimg_stack, maskdef_dict,
```
jhennawi: Please make maskdef_dict an optional and not a required argument.

pypeit/core/coadd.py
```diff
@@ -2920,11 +2924,21 @@ def compute_coadd2d(ref_trace_stack, sciimg_stack, sciivar_stack, skymodel_stack
         dspat_img_fake = spat_img_coadd + dspat_mid[0]
         dspat[np.invert(outmask)] = dspat_img_fake[np.invert(outmask)]
 
+    # update maskdef_objpos and maskdef_slitcen with the new value in the new slit
```
jhennawi: This sequence should only be executed if maskef_dict is passed in. You can leave the hook for masked_designtab in the return dictionary, and set it to None if no maskdef_dict was passed in.

pypeit/par/pypeitpar.py
```diff
@@ -1698,9 +1698,9 @@ class SlitMaskPar(ParSet):
 
 
     """
-    def __init__(self, obj_toler=None, assign_obj=None, nsig_thrshd=None,
+    def __init__(self, obj_toler=None, assign_obj=None, skip_serendip=None, nsig_thrshd=None,
```
jhennawi: skip_serendip needs to be removed. It makes no sense.

pypeit/par/pypeitpar.py
```diff
@@ -1698,9 +1698,9 @@ class SlitMaskPar(ParSet):
...
                  slitmask_offset=None, use_dither_offset=None, bright_maskdef_id=None, extract_missing_objs=None,
-                 use_alignbox=None):
+                 missing_objs_fwhm=None, missing_objs_boxcar_rad=None, use_alignbox=None):
```
jhennawi, line 1703: I don't see why these parameters missing_objs_fwhm and missinb_objs_boxcar_rad are specific to coadd2d. They are not. If you want to be able to specify these things that is a general question for normal single frame reductions. These parameters should thus be added to the parsers for objfind or extract and not for coadd2d.

pypeit/par/pypeitpar.py
```diff
@@ -1726,6 +1726,10 @@ def __init__(self, obj_toler=None, assign_obj=None, nsig_thrshd=None,
         dtypes['assign_obj'] = bool
         descr['assign_obj'] = 'If SlitMask object was generated, assign RA,DEC,name to detected objects'
 
+        defaults['skip_serendip'] = False
```
jhennawi: Let's discuss if this is needed. I feel strongly it makes no sense given how local is performed.

pypeit/par/pypeitpar.py
```diff
@@ -1763,9 +1767,22 @@ def __init__(self, obj_toler=None, assign_obj=None, nsig_thrshd=None,
         defaults['extract_missing_objs'] = False
         dtypes['extract_missing_objs'] = bool
         descr['extract_missing_objs'] = 'Force extraction of undetected objects at the location expected ' \
-                                        'from the slitmask design. PypeIt will try to determine the FWHM from ' \
-                                        'the flux profile (by using ``find_fwhm`` in `FindObjPar` as initial guess). ' \
-                                        'If the FWHM cannot be determined, ``find_fwhm`` will be assumed.'
+                                        'from the slitmask design.'
```
jhennawi: These params should be moved to find_objects or extraction parsers, but not here.

pypeit/pypeit.py
```diff
@@ -902,6 +901,11 @@ def extract_one(self, frames, det, sciImg, global_sky, sobjs_obj, skymask):
         spec_flex_table['spat_id'] = self.caliBrate.slits.spat_id
         spec_flex_table['sci_spec_flexure'] = self.redux.slitshift
 
+        # pull out maskdef_designtab from caliBrate.slits
```
jhennawi, line 913: These lines seem peculiar. Why is this being done? Your comment does not explain why.

pypeit/scripts/coadd_2dspec.py
jhennawi: Good catch!

pypeit/scripts/coadd_2dspec.py
```diff
@@ -189,6 +190,20 @@ def main(args):
 #            set(np.arange(spectrograph.ndet) + 1) - set(detectors)])))
         msgs.info(f'Detectors to work on: {detectors}')
 
+        # Only_slits?
```
jhennawi, line 192: only_slits = [int(item) for item in args.only_slits.split(',')] if args.only_slits is not None else None

pypeit/scripts/coadd_2dspec.py
```diff
@@ -189,6 +190,20 @@ def main(args):
...
+            only_slits = [int(item) for item in args.only_slits.split(',')]
+        else:
+            only_slits = None
+
+        # container for specobjs
+        all_specobjs = specobjs.SpecObjs()
+        # container for spec2dobj
+        all_spec2d = spec2dobj.AllSpec2DObj()
+        # set some meta
+        all_spec2d['meta']['ir_redux'] = ir_redux
```
jhennawi: This will conflict with the next PR, i.e. the ir_redux is replaced by bg_redux.

pypeit/scripts/coadd_2dspec.py
```diff
@@ -227,11 +242,42 @@ def main(args):
                 sci_dict[coadd.detname]['waveimg'] \
                     = coadd.reduce(pseudo_dict, show=args.show, show_peaks=args.peaks, basename=basename)
 
+            # Tack on detector (similarly to pypeit.extract_one)
+            for sobj in sci_dict[coadd.detname]['specobjs']:
+                sobj.DETECTOR = sci_dict[coadd.detname]['detector']
+                # iwv = np.where(self.caliBrate.wv_calib.spat_ids == sobj.SLITID)[0][0]
```
jhennawi: Can you remove the commented out lines please.

pypeit/scripts/coadd_2dspec.py
```diff
@@ -227,11 +242,42 @@ def main(args):
...
+                # sobj.WAVE_RMS =self.caliBrate.wv_calib.wv_fits[iwv].rms
+
+            # fill the specobjs container
+            all_specobjs.add_sobj(sci_dict[coadd.detname]['specobjs'])
+
+            # fill the spec2dobj container but first ...
```
jhennawi, line 248: Again these lines are confusing. Same comment as in PypeIt. At the very least please put in a comment to indicate what you are doing here and the logic behind it. It is add that you are assigning this thing and then zeroing it out in slits.

pypeit/scripts/ql_keck_mosfire.py
```diff
@@ -148,7 +148,8 @@ def run_pair(A_files, B_files, caliBrate, spectrograph, det, parset, show=False,
                                       vel_corr=None,
                                       vel_type=parset['calibrations']['wavelengths']['refframe'],
                                       tilts=tilts,
-                                      slits=copy.deepcopy(caliBrate.slits))
+                                      slits=copy.deepcopy(caliBrate.slits),
+                                      maskdef_designtab=None)
```
jhennawi, line 156: Can you just make the default to these instantiations maskdef_designtab=None and hence just omit this optional variable in both cases here?

pypeit/slittrace.py
```diff
@@ -912,19 +912,20 @@ def mask_add_missing_obj(self, sobjs, fwhm, slits_left, slits_right):
...
+            # Vette
+            for sobj in sobjs:
+                if not sobj.ready_for_extraction():
+                    msgs.error("Bad SpecObj.  Can't proceed")
 
         # Return
         return sobjs
 
-    def assign_maskinfo(self, sobjs, plate_scale, slits_left, TOLER=1.):
+    def assign_maskinfo(self, sobjs, plate_scale, slits_left, TOLER=1., skip_serendip=False):
```
jhennawi: skip_serendip needs to be removed. It does not make sense.

pypeit/coadd2d.py
```diff
@@ -140,13 +139,19 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
         """
 
-        ## Use Cases:
-        #  1) offsets is None -- auto compute offsets from brightest object, so then default to auto_weights=True
-        #  2) offsets not None, weights = None (uniform weighting) or weights is not None (input weights)
-        #  3) offsets not None, auto_weights=True (Do not support)
-        if offsets is not None and 'auto' in weights:
-            msgs.error("Automatic weights cannot be computed for input offsets. "
-                       "Set weights='uniform' or input an array of weights with shape (nexp,)")
+        # Use Cases:
```
jhennawi, line 145: Very nice!

pypeit/coadd2d.py
jhennawi: Agreed!

pypeit/coadd2d.py
```diff
@@ -169,39 +173,114 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
             msgs.error('Not all of your exposures have the same number of slits. Check your inputs')
-        # TODO: Do the same check above but for the shape and binning
-        # of the input images?
-        self.nslits = nslits_list[0]
-        self.nexp = len(self.stack_dict['specobjs_list'])
+        # This is the number of slits of the single (un-coadded) frames
+        self.nslits_single = nslits_list[0]
+
+        # Check that nspec is the same for all the exposures
+        nspec_list = [slits.nspec for slits in self.stack_dict['slits_list']]
```
jhennawi, line 203: This variable is exactly the same as nslits_list define above. You have replicated that line and defined a new variable.

pypeit/coadd2d.py
```diff
@@ -169,39 +173,114 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
-        self.maskdef_objpos = np.array([slits.maskdef_objpos for slits in self.stack_dict['slits_list']])
-        self.maskdef_slitcen = np.array([slits.maskdef_slitcen for slits in self.stack_dict['slits_list']])
+    @property
+    def nslits_coadded(self):
+        """
+        This gives the number of slits of the coadded frame.
+        `self.good_slits` must be defined first.
+
+        Returns:
+            :obj:`float`: Number of slits of the coadded frame
```
jhennawi: I guess this should be an int right, not a float?

pypeit/coadd2d.py
```diff
@@ -169,39 +173,114 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
 
         # If smoothing is not input, smooth by 10% of the spectral dimension
         self.sn_smooth_npix = sn_smooth_npix if sn_smooth_npix is not None else 0.1*self.nspec
 
-        # maskdef info
-        self.maskdef_id = np.array([slits.maskdef_id for slits in self.stack_dict['slits_list']])
-        self.maskdef_offset = np.array([slits.maskdef_offset for slits in self.stack_dict['slits_list']])
-        self.maskdef_objpos = np.array([slits.maskdef_objpos for slits in self.stack_dict['slits_list']])
-        self.maskdef_slitcen = np.array([slits.maskdef_slitcen for slits in self.stack_dict['slits_list']])
+    @property
```
jhennawi: This seems pretty redundant. You are defining a property that just equal to an attribute of the class. Avoid state dependent properties. They just create a confusing control flow, but add no real value.

pypeit/coadd2d.py
```diff
@@ -169,39 +173,114 @@ def __init__(self, spec2d, spectrograph, par, det=1, offsets=None, weights='auto
...
+
+        """
+        # this is the number of slits of the coadded frame
+        if self.good_slits is None:
+            msgs.error('Cannot determine nslits_coadded because `self.good_slits` is `None`. '
+                       'Run `self.good_slits = self.good_slitindx(only_slits=only_slits)` first')
+        return self.good_slits.size
+
+    def good_slitindx(self, only_slits):
+        """
```
jhennawi, line 227: Please add more documentation indicating what this routine does.

pypeit/coadd2d.py
```diff
@@ -278,49 +356,38 @@ def coadd(self, only_slits=None, interp_dspat=True):
...
-        # TODO We should be checking the bitmask for the reductions or something here??
-        #  Yes, definitely
-        if only_slits is None:
-            slits = self.stack_dict['slits_list'][0]
-            reduce_bpm = (slits.mask > 0) & (np.invert(slits.bitmask.flagged(
-                slits.mask, flag=slits.bitmask.exclude_for_reducing)))
-            good_slits = np.where(np.invert(reduce_bpm))[0]
-        else:
-            raise NotImplementedError('Slit selection (only_slits) not implemented yet!')
+        self.good_slits = self.good_slitindx(only_slits=only_slits)
```
jhennawi, line 348: Please add a comment explaining what is being done here.

pypeit/coadd2d.py
```diff
@@ -278,49 +356,38 @@ def coadd(self, only_slits=None, interp_dspat=True):
...
             ref_trace_stack = self.reference_trace_stack(slit_idx, offsets=self.offsets,
                                                          objid=self.objid_bri)
             thismask_stack = np.abs(self.stack_dict['slitmask_stack'] - self.stack_dict['slits_list'][0].spat_id[slit_idx]) <= self.par['coadd2d']['spat_toler']
 
-            # TODO Can we get rid of this one line simply making the weights returned by parse_weights an
-            # (nslit, nexp) array?
-            # This one line deals with the different weighting strategies between MultiSlit echelle. Otherwise, we
-            # would need to copy this method twice in the subclasses
-            if 'auto_echelle' in self.use_weights:
-                rms_sn, weights = self.optimal_weights(slitord_id, self.objid_bri)
```
jhennawi, line 309: I'm a but confused bout what happened to this "auto_echelle" functionality. It appears that you have removed it.

pypeit/coadd2d.py
```diff
@@ -278,49 +356,38 @@ def coadd(self, only_slits=None, interp_dspat=True):
...
+                                               self.stack_dict['sciivar_stack'],
+                                               self.stack_dict['skymodel_stack'],
+                                               self.stack_dict['mask_stack'] == 0,
+                                               self.stack_dict['tilts_stack'],
                                                thismask_stack,
-                                           self.stack_dict['waveimg_stack'],
-                                           self.wave_grid, self.spat_samp_fact,
-                                           weights=weights, interp_dspat=interp_dspat)
+                                               self.stack_dict['waveimg_stack'],
+                                               maskdef_dict,
```
jhennawi: Can you make maskdef_dict optional please.  compute_coadd2d is a core method and you should not be adding required arguments that it may never need. Make them optional.

pypeit/coadd2d.py
```diff
@@ -400,20 +475,38 @@ def create_pseudo_image(self, coadd_list):
             spec_max1[islit] = nspec_vec[islit]-1
             spat_left = spat_righ + nspat_pad
 
+            # maskdef info
```
jhennawi, line 468: Nice!

pypeit/coadd2d.py

pypeit/coadd2d.py
debora-pe: Right!

pypeit/coadd2d.py
jhennawi: Actually, this code was added here and is completely redundant. I also think it does not make any sense for echelon since it was not a method in echelle before. I think it would be more efficient to chat.

jhennawi (changes requested):
Thanks for this Herculean effort. Everything looks good, but there is a snag in pypeit/coadd2d.py that we need to discuss. The inheritance logic has changed and lots of code is being replicated. I'm concerned about both of those, so let's schedule a time to discuss the path forward.

pypeit/coadd2d.py
jhennawi: Let's add a TODO.

pypeit/coadd2d.py
jhennawi: Just add a todo that we will break this up into two methods compute_offsets and compute_weights, and we will avoid replicated code as much as possible.

pypeit/pypeit.py
jhennawi: @kbwestfall I think you understand how these data models work. Can you please provide some help here. The problem is the maskdef_designtab is somehow not be written in the slits object, even thought it is an attribute there. For that reason @deborape is making this an explicit argument to the data model and then adding it explicitly. But the slits are already part of the data model so why don't nested data models work.

jhennawi (approved):
Please run all tests before you merge.

pypeit/coadd2d.py
debora-pe: Added

pypeit/coadd2d.py
debora-pe: Nice! Thanks for the explanation. I will keep it in mind for the future for sure!

pypeit/coadd2d.py
debora-pe: Done

pypeit/coadd2d.py

pypeit/coadd2d.py
debora-pe: Added

pypeit/core/coadd.py
debora-pe: Done

pypeit/core/coadd.py
debora-pe: Done

pypeit/par/pypeitpar.py
debora-pe: Removed

pypeit/scripts/coadd_2dspec.py
debora-pe: Done

pypeit/scripts/coadd_2dspec.py
debora-pe: Done

pypeit/scripts/coadd_2dspec.py
debora-pe: Done

pypeit/coadd2d.py
debora-pe: Broke up the method into `compute_offsets` and `compute_weights`. Added a TODO to move part of the code that is common to multislit and echelle in the parent class

pypeit/pypeit.py
kbwestfall: Thanks for pointing this out, @jhennawi .  Nested `DataContainer`s can work, just not always out of the box.  They (often) require some hand-holding, particularly if the datamodels are non-trivial.  If these lines exist basically to get the I/O in `Spec2DObj` to work, then I'd like to help figure this out.  @deborape maybe we can chat about this today?

debora-pe:
Tests pass.

Dev Suite
```
Testing Started at 2022-02-10T11:01:08.991904
Testing Completed at 2022-02-10T15:57:59.836444
Total Time: 4:56:50.844540
```

And unit test
```
============ 318 passed in 1218.03s (0:20:18) =========
```

debora-pe:
Merging

PR #1326, opened 2022-01-31
Remove CentOS 8 from test matrix since it's now EOL

tepickering:
All of the package mirrors for CentOS 8 have been moved now that it has been EOL for a month. This PR removes testing under CentOS 8, but keeps CentOS 7 which is still supported for a while longer and is still being used within the PypeIt community.


kbwestfall (approved):

profxj (approved):
thx!

PR #1333, opened 2022-02-08
Pattern modelling

rcooke-ast:
It turns out that the pattern noise in KCWI is not constant for a given exposure, as previously assumed. It gradually varies, but can be well fit by a linear function. The image shows the worst affected regions of the detector with the old algorithm (left panels) and new algorithm (right panels).

**NOTE:** This is branched off `skymask_noise_bug`.

image


kbwestfall (approved):
Looks good to me!

jhennawi (approved):

debora-pe:
merging

PR #1342, opened 2022-02-15
Update conda configuration and Apple Silicon related tweaks to installation docs

tepickering:
This PR updates `environment.yml` so that `conda` can do a full installation of all required and optional dependencies along with `pypeit` itself.  I specifically added `pyqt` and `scikit-image` , pinned `python` to be `< 3.10`, and added `pypeit` as a package to install via `pip`. This allows users to set up a working `pypeit` installation with a single command. This method was tested successfully on an Apple Silicon MacBook Pro using the Apple Silicon-native `miniconda` installer. 

The installation docs were updated to reflect the changes and hopefully made clearer. Wording was added to note that `conda` is the only way known to work on Apple Silicon-based machines. Links were added to where the correct `miniconda` installers can be found.

I also added `scikit-image` to `setup.cfg` as an optional dependency. It was mentioned as such in the docs, but `setup.cfg` hadn't been updated.


kbwestfall (approved):
Looks great! Thanks @tepickering !

jhennawi (approved):
Thanks Timothy, this is great work!

tepickering:
thanks for the comments. clicked 'update branch' to make sure it's truly mergeable. should be gtg once tests pass.

finagle29:
I was able to follow the instructions and get PypeIt working in a fresh `miniconda` installation, but not in my `miniforge` installation. I think this is because the M1 pyqt package is hosted by Anaconda in their `repo.anaconda.com` repository, which is part of the default channels in `miniconda`, but not `miniforge`.

Adding `pkgs/main` under `conda-forge` in `environment.yml` let me get PypeIt working in a `miniforge` install.

environment.yml
```diff
@@ -2,7 +2,7 @@ name: pypeit
 channels:
   - conda-forge
```
finagle29, line 3: Please add the `pkgs/main` channel under `conda-forge` to fix `miniforge` issues.

finagle29 (changes requested):

environment.yml
tepickering: thanks for testing! i will do so now...

tepickering: should i specify `pkgs/main` or just use `defaults`?

finagle29: Just tested it and `defaults` also works for me, so I think that `defaults` is good.

tepickering: adding both channels to the list should be good enough for now. if we run into problems, we can specify which channel to use on a per-package basis.

finagle29 (approved):

Issue #1251, opened 2021-08-18
Some feedback from a first time PypeIt user...


rcooke-ast:
I received the following feedback from someone new to PypeIt, and some of the concerns they had during the reduction.
* Suppress warning messages for frame types that were not found (e.g. there are no "align" frames for longslit)
* Suppress verify warning from astropy package
* define object and background regions (maybe interactively)?
* Allow users to run the reduction step by step and perform checks at each step

The first two are somewhat related to #1167

EDIT: The second point above is for warnings such as this:
`WARNING: VerifyWarning: Keyword name 'NOISE_FLOOR' is greater than 8 characters or contains characters not allowed by the FITS standard; a HIERARCH card will be created. [astropy.io.fits.card]`

rcooke-ast:
These appear to be all resolved now.

PR #1365, opened 2022-03-01
Silence VerifyWarning for "Keyword name ... is greater than 8..."

tbowers7:
This particular Warning (raised by `astropy.io.fits.card`) does not seem to be helpful for either the user or developer because the PypeIt team made the conscious decision to use long keyword names for reduction metadata added to the FITS headers of data products.

The user can't do anything about the long keywords, and the code base would require extensive overhaul to remove them.  Furthermore, there is universal support among modern software for handling the HIERARCH keyword as it has been around for more than 20 years (cf. https://www.eso.org/sci/software/esomidas/doc/user/18NOV/vola/node116.html and https://heasarc.gsfc.nasa.gov/fitsio/c/f_user/node28.html).

These warnings also make it challenging to follow the flow of PypeIt messages during reduction for debugging and troubleshooting.

Partially addresses #1251 .

### DIFF
--- a/pypeit/io.py
+++ b/pypeit/io.py
@@ -22,6 +22,7 @@ from configobj import ConfigObj
 
 from astropy.io import fits
 from astropy.table import Table
+from astropy.io.fits.verify import VerifyWarning
 
 from pypeit import msgs
 from pypeit import par
@@ -657,6 +658,10 @@ def write_to_hdu(d, name=None, hdr=None, force_to_bintbl=False):
             types.
 
     """
+    # Silence the "Keyword name ... is greater than 8 characters ... a HIERARCH
+    #   card will be created" Warnings from [astropy.io.fits.card]
+    warnings.simplefilter('ignore', VerifyWarning)
+
     if isinstance(d, dict):
         return dict_to_hdu(d, name=name, hdr=hdr, force_to_bintbl=force_to_bintbl)
     if isinstance(d, Table):


profxj (approved):
Thanks for this!

If you can also mention this to the `ginga` team, that would be wonderful.
Actually, I can raise an Issue..

pypeit/io.py
```diff
@@ -657,6 +658,10 @@ def write_to_hdu(d, name=None, hdr=None, force_to_bintbl=False):
             types.
 
     """
+    # Silence the "Keyword name ... is greater than 8 characters ... a HIERARCH
+    #   card will be created" Warnings from [astropy.io.fits.card]
+    warnings.simplefilter('ignore', VerifyWarning)
```
kbwestfall, line 663: Do we need the `from astropy.io.fits.verify import VerifyWarning` import?  Can we instead set this to `warnings.simplefilter('ignore', fits.verify.VerifyWarning)` since we already have `fits` imported?

kbwestfall (approved):
Thanks!  @profxj , I think this is an `astropy.io.fits` issue, not a `ginga` issue, right?

profxj:
yes, but they need to make the same fix to reduce the behaviour I have.

pypeit/io.py
tbowers7: @kbwestfall I'll fix this as part of #1360.

PR #1368, opened 2022-03-18
Add rms threshold ot collate_1d, issue 1317

badpandabear:
This pull request is to resolve issue [1317](https://github.com/pypeit/PypeIt/issues/1317)

Unit tests:
============================ 324 passed, 325 warnings in 991.46s (0:16:31) =======================================

I didn't run the dev suite as pypeit_collate_1d is tested entirely in unit tests but I can if requested.


pypeit/scripts/collate_1d.py
```diff
@@ -505,6 +514,11 @@ def get_parser(cls, width=None):
                                  'F|                        "ra/dec"\n'
                                  'F|  dry_run               If set the matches are displayed\n'
                                  'F|                        without any processing\n'
+                                 'F|  flux                  Flux calibrate using archived sensfuncs.\n'
+                                 'F|  ignore_flux           Ignore any flux calibration information in\n'
+                                 'F|                        spec1d files.\n'
+                                 'F|  rms_thresh            If set any objects with rms > than this are\n'
+                                 'F|                        skipped.\n'
```
bpholden: I would suggest to rephrase as "If set, any objects with a rms > than the input value are skipped, else all are accepted."

bpholden (approved):

pypeit/scripts/collate_1d.py
badpandabear: I went with " If set, any objects with a rms > than the input value are skipped, else all rms values are accepted." in order to clarify that objects might still be rejected for other reasons.

debora-pe (approved):
Hey Dusty, this looks good. Thank you fo doing it!
Just a really nit-picking thing: do you mind making more clear in the name of the parameter and/or in the description that we are talking about wavelength RMS?

PR #1369, opened 2022-04-03
GTC/OSIRIS

rcooke-ast:
A few updates to:
* properly process arc frames from multiple lamps
* enhance slit edges when the slit edge is "blurred" or vignetted (i.e. not sharp)
* includes some updated wavelength solutions for GTC/OSIRIS.


pypeit/core/trace.py
```diff
@@ -97,6 +100,10 @@ def detect_slit_edges(flux, bpm=None, median_iterations=0, min_sqm=30., sobel_mo
 
     # Filter with a Sobel
     filt = ndimage.sobel(sqmstrace, axis=1, mode=sobel_mode)
+    # Enhance blurred edges
+    if sobel_enhance:
+        nsum = 3
```
kbwestfall: Does this need to be parameter?  I.e., instead of having the parameter be a boolean, would it make sense for the `sobel_enhance` parameter to be this integer?

pypeit/par/pypeitpar.py
```diff
@@ -281,6 +281,11 @@ def __init__(self, trim=None, apply_gain=None, orient=None,
                                'sinusoidal along one direction, with a frequency that is ' \
                                'constant across the detector.'
 
+        defaults['use_continuum'] = False
+        dtypes['use_continuum'] = bool
+        descr['use_continuum'] = 'Subtract off the continuum level fromt he image. This correction is helpful ' \
+                                 'if you are combining arcs with multiple different lamps and exposure times. '
+
```
kbwestfall, line 289: I have a philosophical objection to this being in the basic image processing.  Is it possible to handle the combination of the multiple arc frames in the wavelength calibration step instead?

Expecting the answer is no, I think we should at least make this description of the parameter more ominous, to ward people off from thinking they can use it more generally.  Something like: "... For arc images from GTC/OSIRIS only."

kbwestfall (changes requested):
Thanks, @rcooke-ast !  I'm unsure about the inclusion of the continuum subtraction in the basic image processing module.  If we keep it there, I'd like to change the description of the `use_continuum` parameter.  And we should also update the basic image processing doc.  Everything else looks great.

pypeit/core/trace.py
rcooke-ast: Yes - that's a good point. I was going to do that, but I worried that this was out of style with the other parameters used here!

pypeit/par/pypeitpar.py
rcooke-ast: This is a step that has to be done before the individual frames are combined, so it can't be done on the combined image. The only PypeItPar that gets passed into CombineImage is a ProcessImagesPar. I'll make it more ominous, but not explicitly mention that this is for GTC/OSIRIS, as I can see other instruments/setups may utilise this.

doc/image_proc.rst
```diff
@@ -291,6 +291,29 @@ variations (``use_pixelflat``), the slit illumination profile
...
+with different lamps.
+
+.. warning::
+
+    If you set this parameter to True, you should also set the following parameters:
+
+    calibrations arcframe process clip = False
+    calibrations arcframe process combine = 'mean'
+    calibrations tiltframe process clip = False
+    calibrations tiltframe process combine = 'mean'
```
kbwestfall: Thanks for these new docs!  Can we include a formatting statement and put these in the configobj format?  I.e., 
```
.. code-block:: ini

    [calibrations]
        [[arcframe]]
            [[[process]]]
                clip = False
                combine = mean
        [[tiltframe]]
            [[[process]]]
                clip = False
                combine = mean
```

kbwestfall (approved):
Thanks, @rcooke-ast !  I've got one lingering request on the docs, but I'm approving now.

doc/image_proc.rst
rcooke-ast: Thanks @kbwestfall for the code - I didn't know this was possible until now! I've added this to the docs

pypeit/spectrographs/gtc_osiris.py
```diff
@@ -116,6 +118,10 @@ def default_pypeit_par(cls):
         par['calibrations']['arcframe']['exprng'] = [None, None]  # Long arc exposures
         par['calibrations']['arcframe']['process']['clip']=False
         par['calibrations']['arcframe']['process']['combine']='mean' #Multiple arcs with different lamps, so can't median combine
+        par['calibrations']['arcframe']['process']['use_continuum']=True #Multiple arcs with different lamps, so can't median combine
+        par['calibrations']['tiltframe']['process']['clip']=False
+        par['calibrations']['tiltframe']['process']['combine']='mean' #Multiple arcs with different lamps, so can't median combine
```
profxj: Please update the comments here and below

profxj (approved):
Looks good -- are there now corresponding tests
of these two new processing steps (sobel and 
continuum) in the DevSuite?

If not please add those.  Thanks!

pypeit/spectrographs/gtc_osiris.py
rcooke-ast: Done!

rcooke-ast:
2 tests added, and all tests pass
image

PR #1371, opened 2022-04-06
Changed format of coadd1d file names

badpandabear:
This PR has some improvements to 1d coadd file naming. 

First it implements a standard naming convention for coadd outputs for pypeit_coadd_1dspec:
`coadd1d_<target>_<instrument name>_<YYYYMMDD>-<YYYYMMDD>.fits`
The new name is only used when coaddfile is not given in the input file and has the format. The `-<YYYYMMDD>` is optional and is only included if the coadded spec1ds span multiple days.
The file is placed along side the first spec1d used in coadding.

Second it changes pypeit_collate_1d so the names of the coadded files it creates can specify a date range similar to above. 

Unit Tests:
============ 324 passed, 3956 warnings in 947.84s (0:15:47) ========

Dev Suite:
--- PYPEIT DEVELOPMENT SUITE PASSED 138/138 TESTS (Masters ignored) ---
Testing Started at 2022-04-04T19:48:17.800889
Testing Completed at 2022-04-05T01:55:26.865851
Total Time: 6:07:09.064962


bpholden (approved):
When you say 
` If the coaddfile is not given the output file will be
                        placed along side the first spec1d file.`
I am guessing you mean that it will be placed in the same directory with the automatically generated filename?

badpandabear:
> When you say ` If the coaddfile is not given the output file will be placed along side the first spec1d file.` I am guessing you mean that it will be placed in the same directory with the automatically generated filename?

Yes. I've updated the documentation and command line help to be more clear.

bpholden:
> > When you say ` If the coaddfile is not given the output file will be placed along side the first spec1d file.` I am guessing you mean that it will be placed in the same directory with the automatically generated filename?
> 
> Yes. I've updated the documentation and command line help to be more clear.

Great! Thanks

kbwestfall (approved):
Looks great!  My only question is whether it's worth the effort making the name different between whether or not the coadd crosses multiple dates.  In terms of parsing the filename, it's always easier when they have a consistent format.  So I'm wondering if the format should always be `coadd1d_<target>_<instrument name>_<YYYYMMDD>-<YYYYMMDD>.fits`, even when the first and last date are the same.

pypeit/scripts/coadd_1dspec.py
```diff
@@ -83,6 +84,34 @@ def read_coaddfile(ifile):
...
+    The filename convention is coadd1d_<target>_<instrument name>_<YYYYMMDD>.fits or
+    coadd1d_<target>_<instrument name>_<YYYYMMDD>-<YYYYMMDD>.fits if the coadd included more than
+    one day's worth of data. The default location of the file will be along side the first spec1d file.
+
+    Currently instrument_name is taken from spectrograph.camera
+
+    Returns: 
+        str:  The name of the coadd output file.
+    """
+    mjd_list = [float(fits.getheader(f)['MJD']) for f in spec1dfiles]
```
debora-pe: I'm concerned about how this is handled if for some reason the `MJD` keyword is missing. I don't remember if `fits.getheader(f)['MJD']` will give an error if `MJD` is not present or it will give you a `None`. I guess in both cases it would be good to have a way to handle it.

debora-pe (approved):
@badpandabear I have only a small comment, so I'm approving. Thanks for doing this.
I saw Kyle's comment on having a consistent format for the name. I think it's valid point, and we could discuss it at the next meeting.

pypeit/scripts/coadd_1dspec.py
bpholden: This is an excellent point, and I am glad you caught this @deborape!

If the MJD does not exist, astropy.io.fits will throw an exception. 

`KeyError:` "Keyword 'MJD' not `found."`

So, instead of a list comprehension, that needs a for loop with a try/except. If the except is caught, I think there are a couple of choices. 

If there is some sort of date in the files, you can use 
`astropy.time.Time`
to convert the ISOT date string in the header to a MJD (this is two lines of code that I seem to use a lot.) 

The other option would be to punt on a date at all and use the KOA filenames (which are date based.)

PR #1372, opened 2022-04-06
Spec2d hotfix

rcooke-ast:
The headers of spec2d files are written out with:

```
BIASMKEY= 'A_1_DE  '
ARCMKEY = 'A_1_DE  '
TRACMKEY= 'A_1_DE  '
FLATMKEY= 'A_1_DE  '
```

This PR has a simple fix to remove the `DE`. The master key can be reconstructed with `BIASMKEY + spec.get_det_name(det)`

### DIFF
--- a/pypeit/core/datacube.py
+++ b/pypeit/core/datacube.py
@@ -17,7 +17,7 @@ from scipy.interpolate import interp1d
 import numpy as np
 
 from pypeit import msgs
-from pypeit import spec2dobj, alignframe
+from pypeit import spec2dobj, alignframe, masterframe
 from pypeit.core.flux_calib import load_extinction_data, extinction_correction
 from pypeit.core.flexure import calculate_image_offset
 from pypeit.core import parse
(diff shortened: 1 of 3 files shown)


kbwestfall (changes requested):
Good catch!  Although I think there are a couple deeper issues (with the significant caveat that the code that creates and parses the master keys is in desperate need of a clean-up):

- The master_key should be the whole string.  Is there as specific use case for removing the detector?
- Of all of these keys, I could only find one instance of the keys being used in the code (although there were more uses in some deprecated code).  That usage is of `TRACMKEY` in `pypeit/core/datacube.py` when used to construct `alignfile`.  I haven't dug into the code to understand why this is being done, but I expect the name is actually wrong.  This occurrence of the master key was something I missed in the introduction of the mosaicing.  The new master keys include 'DET' or 'MSC' in the naming convention.  I.e., a name like `A_1_01` is now invalid and it now should be `A_1_DET01`. 

So instead of removing parts of the key when assigning it to the header, I would assign the full string, and then I would adjust the code in `pypeit/core/datacube.py` accordingly.

rcooke-ast:
OK - I think I agree with that. My only worry was why the `01` suffix was being removed in the first place. Since a spec2d file can only contain the output from one DET (I think!), then keeping the whole MKEY makes sense, so I've just made that change. Indeed, I only realised this when updating the `datacube.py` module... PR forthcoming about that, in the future...

kbwestfall:
> OK - I think I agree with that. My only worry was why the `01` suffix was being removed in the first place. Since a spec2d file can only contain the output from one DET (I think!), then keeping the whole MKEY makes sense, so I've just made that change. Indeed, I only realised this when updating the `datacube.py` module... PR forthcoming about that, in the future...

Yeah, I don't know why the detector was being removed in the first place.  Is it worth making the minor adjustment to the assignment of `alignfile` in `datacube.py` based on this change in this PR?

rcooke-ast:
Sure thing - I've just pushed some new changes. `datacube.py` has also had an overhaul, so this will change again, but at least it should be good for now.  Thanks for the good suggestions!

kbwestfall (approved):
Thanks, @rcooke-ast !  I don't think we need to run the dev-suite on this, but can you confirm the unit-tests pass?

rcooke-ast:
Sounds good @kbwestfall - I've just run the unit tests, and needed to make a change to `test_setups.py` in order to get all tests to pass. Here's the screenshot:
image

profxj (approved):
approved

PR #1374, opened 2022-04-08
Allow curved traces for user-added slit edges

kbwestfall:
When users add slits, the code currently forces the pseudo-slit edge traces to be linear; i.e., the spatial pixel of the edge trace is the same at every spectral pixel.  This PR allows the user to define how the spectral form of the trace is predicted.  They can either specify "straight" (mimics the current behavior), "nearest", or "pca".  Selection of "nearest" and "pca" is effectively identical to how the automatic tracing code inserts missing left or right traces when syncing the traces into slit pairs.  The code works well in a single test case, shown below.

Poor result when using "add_predict = straight":

straight_test

Better result when using "add_predict = nearest"

nearest_test

or "add_predict = pca"

pca_test

Tests pass.  Dev suite pending.

### DIFF
--- a/pypeit/edgetrace.py
+++ b/pypeit/edgetrace.py
@@ -3946,7 +3946,7 @@ class EdgeTraceSet(DataContainer):
         elif method == 'pca':
             if self.is_empty:
                 msgs.error('No edge traces currently exist.  Cannot insert user slits with a '
-                           'shape based on the PCA decomposition of the existing slit edges!
+                           'shape based on the PCA decomposition of the existing slit edges!  '
                            'Set add_predict = straight.')
             if self.pcatype is None:
                 if not self.can_pca():
--- a/pypeit/edgetrace.py
+++ b/pypeit/edgetrace.py
@@ -3933,7 +3933,8 @@ class EdgeTraceSet(DataContainer):
         elif method == 'nearest':
             if self.is_empty:
                 msgs.error('No edge traces currently exist.  Cannot insert user slits with a '
-                           'shape based on the nearest existing slit edges!')
+                           'shape based on the nearest existing slit edges!  '
+                           'Set add_predict = straight.')
             # Use the measured edges if the functional forms don't exist (yet)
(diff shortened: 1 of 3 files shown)


kbwestfall:
I'll investigate why CI failed, but at least on my machine:
```
======================= 324 passed in 1295.25s (0:21:35) =======================
```

jhennawi (approved):
Hi @kbwestfall, 

Thanks for this. Can you fix the CI issue and make sure this passes dev suite before merging.  Also, you did not mention what the default is. I think it definitely should not be straight, but will leave whether it is nearest or pca up to you. I would try to do the thing that will be the most stable.

kbwestfall:
> Hi @kbwestfall,
> 
> Thanks for this. Can you fix the CI issue and make sure this passes dev suite before merging. Also, you did not mention what the default is. I think it definitely should not be straight, but will leave whether it is nearest or pca up to you. I would try to do the thing that will be the most stable.

Thanks @jhennawi ! The CI issue is fixed (I just needed to re-run the job).  The default is "add_predict = nearest".  The only time this will fail is when you're adding a slit that does not include any automatically identified edges.  In that case, the code will raise an exception.  I've now added to the error message to say that the user should change to "add_predict = straight".

I'm still working on the dev-suite run.

debora-pe (approved):
@kbwestfall thank you for doing this. It looks good. I'm approving.

Just a note. If you have a failure in the Dev Suite with  `keck_lris_red_mark4`, I already got that failure and fixed in the Dev Suite branch `mosfire_wavecalib`, which I'm about to open a PR on.

kbwestfall:
Dev-suite passed.
```
--- PYPEIT DEVELOPMENT SUITE PASSED 138/138 TESTS (Masters ignored) ---
Testing Started at 2022-04-11T19:47:22.266301
Testing Completed at 2022-04-12T01:08:58.519976
Total Time: 5:21:36.253675
```
Merging.

PR #1375, opened 2022-04-11
Mosfire wavecalib

debora-pe:
Created wavelength templates for MOSFIRE. There are 4 templates (one per each filter) using OH lines and 4 templates using arc frames.
The OH line templates are the default.
The templates with arc lines (Ar, Ne) are used by default only when the `long2pos_specphot` mask is used (this is the only case in which I saw PypeIt completely failing at performing the wavelength calibration using OH line). However, the user can modify the pypeit file to use the arc frames anytime (instruction added in the docs).

Companion Dev-Suite PR: https://github.com/pypeit/PypeIt-development-suite/pull/193

Running the tests now.


debora-pe:
Unit tests pass:

```
pytest . -W ignore
============================ test session starts =============================
platform darwin -- Python 3.9.7, pytest-7.1.1, pluggy-1.0.0
collected 324 items                                                          

test_alignments.py .                                                   [  0%]
test_arc.py .                                                          [  0%]
test_archive.py ..                                                     [  1%]
test_arcimage.py ..                                                    [  1%]
test_biasframe.py ....                                                 [  3%]
test_bitmask.py ....                                                   [  4%]
test_bpmimage.py ..                                                    [  4%]
test_bspline.py ...............                                        [  9%]
test_buildimage.py ..                                                  [ 10%]
test_calibrations.py .......                                           [ 12%]
test_chk_calibs.py ..                                                  [ 12%]
test_coadd.py .                                                        [ 13%]
test_collate_1d.py ..........                                          [ 16%]
test_cooked.py .                                                       [ 16%]
test_coremeta.py .                                                     [ 16%]
test_datacontainer.py .......                                          [ 19%]
test_detector.py ....                                                  [ 20%]
test_display.py .                                                      [ 20%]
test_edgetrace.py .                                                    [ 20%]
test_fitting.py ....                                                   [ 22%]
test_flatfield.py .                                                    [ 22%]
test_flexure.py ..                                                     [ 23%]
test_flux.py ....                                                      [ 24%]
test_fluxspec.py .....                                                 [ 25%]
test_frametype.py ..                                                   [ 26%]
test_history.py ....                                                   [ 27%]
test_load.py .                                                         [ 28%]
test_load_images.py ................                                   [ 33%]
test_manual.py ..                                                      [ 33%]
test_maskdesign_matching.py ...                                        [ 34%]
test_masterframe.py .                                                  [ 34%]
test_match.py ....                                                     [ 36%]
test_metadata.py ......                                                [ 37%]
test_moment.py ......                                                  [ 39%]
test_mosaic.py ...                                                     [ 40%]
test_msgs.py ..                                                        [ 41%]
test_opticalmodel.py ....                                              [ 42%]
test_parse.py ...                                                      [ 43%]
test_pca.py ........                                                   [ 45%]
test_processimage.py ..                                                [ 46%]
test_procimg.py ........                                               [ 49%]
test_pydl.py .                                                         [ 49%]
test_pypeit.py .                                                       [ 49%]
test_pypeitimage.py ..                                                 [ 50%]
test_pypeitpar.py ................................                     [ 60%]
test_pypeitsetup.py ........                                           [ 62%]
test_qa.py .                                                           [ 62%]
test_runpypeit.py ..                                                   [ 63%]
test_scienceimage.py ...                                               [ 64%]
test_scripts.py ....................                                   [ 70%]
test_sensfilearchive.py ...                                            [ 71%]
test_setups.py ......................                                  [ 78%]
test_skysub.py ..                                                      [ 79%]
test_slitmask.py ...                                                   [ 79%]
test_slittrace.py ....                                                 [ 81%]
test_spec2dobj.py ......                                               [ 83%]
test_specobj.py .......                                                [ 85%]
test_specobjs.py .....                                                 [ 86%]
test_spectrographs.py .......................                          [ 93%]
test_syncspec.py .                                                     [ 94%]
test_telluric.py ..                                                    [ 94%]
test_traceimage.py .                                                   [ 95%]
test_transform.py ...                                                  [ 95%]
test_utils.py ....                                                     [ 97%]
test_vcorr.py ..                                                       [ 97%]
test_wavemodel.py .                                                    [ 98%]
test_wavetilts.py ....                                                 [ 99%]
test_wvcalib.py ..                                                     [100%]

====================== 324 passed in 2016.79s (0:33:36) ======================
```

jhennawi (changes requested):
Hi @deborape, 

Thanks for this PR, which is badly needed. I have two questions: 

1) I cannot imagine a situation where arc lamps work better than OH sky lines for wavelength solution. Can you elaborate what you mean by the wavelength calibration failed? Do you mean the holy-grail algorithm failed or do you mean that you could not calibrate long2pos data with your existing templates. The latter makes no sense to me. 

2) Your IR wavelength file says it comes from the Keck webpage. Can you confirm that these wavelengths are in vacuum? Also, the arcsines prototocol is to get the wavelengths from NIST, so you probably want to run this by @profxj and @rcooke-ast. However, again I don't really know why one would ever use arcs when the OH sky lines work so much better. However, it is still useful to have the arc line templates since some observers may choose to use them.

debora-pe:
Hi @jhennawi. Thanks for your comments.
1. The slitmask I am referring to is `long2pos_specphot`, not `long2pos`.  `long2pos` works fine with the OH lines. `long2pos_specphot`, instead, is a strange `long2pos` slitmask with a wider aperture in the center (see image below, the red lines show the edges of the slit). Because the slits in this slitmask will have OH lines with changing FWHM along the spatial direction, the wavelength calibration fails (both with the holy-grail and the template). This is why I think the arc frames are needed in this case. Also, as you said, some user may want to use the arc frames anyway. I found a lot of arc frames in the archive.
2. The wavelengths of the arc lines are in vacuum. I'll add a comment in the files. To create these lists I followed what was already done for GNIRS (https://github.com/pypeit/PypeIt/blob/develop/pypeit/data/arc_lines/lists/Ar_IR_GNIRS_lines.dat), where the source of the line list was "gemini_webpage". Also, I have compared a few of those lines with the ones in NIST and they are consistent.
Screenshot 2022-04-11 at 13 35 02

debora-pe:
Dev Suite passes:

```
--- PYPEIT DEVELOPMENT SUITE PASSED 137/137 TESTS (Masters ignored) ---
Testing Started at 2022-04-11T11:00:43.678635
Testing Completed at 2022-04-11T15:58:30.247812
Total Time: 4:57:46.569177
```

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -179,6 +179,41 @@ def config_specific_par(self, scifile, inp_par=None):
                 # assume that the main target is always detected, i.e., skipping force extraction
                 par['reduce']['slitmask']['extract_missing_objs'] = False
 
+        # wavelength calibration using OH lines
+        if 'long2pos_specphot' not in self.get_meta_value(headarr, 'decker') and \
+                self.get_meta_value(headarr, 'filter1') in ['Y', 'J', 'H', 'K']:
```
kbwestfall: Because you grab them more than once, I would save the filter and decker to variables and then use those throughout the function.  I.e.:
```
decker = self.get_meta_value(headarr, 'decker')
band = self.get_meta_value(headarr, 'filter1')
if 'long2pos_specphot' not in decker and band in ['Y', 'J', 'H', 'K']:
    ...
```

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -179,6 +179,41 @@ def config_specific_par(self, scifile, inp_par=None):
...
+            par['calibrations']['wavelengths']['sigdetect'] = 10.
+
+            # templates
+            if self.get_meta_value(headarr, 'filter1') == 'Y':
+                par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_mosfire_OH_Y.fits'
+            if self.get_meta_value(headarr, 'filter1') == 'J':
+                par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_mosfire_OH_J.fits'
+            if self.get_meta_value(headarr, 'filter1') == 'H':
+                par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_mosfire_OH_H.fits'
+            if self.get_meta_value(headarr, 'filter1') == 'K':
```
kbwestfall: This block is mostly why I'm suggesting saving the filter name instead of grabbing it from the header multiple times.  Also, I would use `elif` and then add an `else` that catches when the filter isn't known.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -179,6 +179,41 @@ def config_specific_par(self, scifile, inp_par=None):
...
+
+            # templates
+            if self.get_meta_value(headarr, 'filter1') == 'Y':
+                par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_mosfire_arcs_Y.fits'
+            if self.get_meta_value(headarr, 'filter1') == 'J':
+                par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_mosfire_arcs_J.fits'
+            if self.get_meta_value(headarr, 'filter1') == 'H':
+                par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_mosfire_arcs_H.fits'
+            if self.get_meta_value(headarr, 'filter1') == 'K':
+                par['calibrations']['wavelengths']['reid_arxiv'] = 'keck_mosfire_arcs_K.fits'
```
kbwestfall, line 222: Same comment here as above.

kbwestfall (approved):
Looks great, @deborape !  Some minor comments from me, so I'm approving now.

jhennawi:
Hi @deborape this sounds great. I think having those archives in there is good. However, note that we have code that synthesizes an OH line list for every resolution using the full resolution line strengths and line locations and a model of the thermal K-band sky. This is in Pypeit somewhere, but I'm suddenly not able to find it. This would be the most robust way to create linelists for MOSFIRE. That would allow you to use the correct resolution for both MOSFIRE and this wider slit. I think it is probably worth making this line list anyway and archiving it since you are using OH_NIRES whereas MOSFIRE has higher resolution.

debora-pe:
Examples of wavelength calibration:

`long2pos_specphot` mask:
Arc_1dfit_D_1_DET01_S0946
Arc_1dfit_D_1_DET01_S1122

Y filter:
Arc_1dfit_A_1_DET01_S1476

K filter:
Arc_1dfit_A_1_DET01_S0656

H filter:
Arc_1dfit_B_1_DET01_S1034

J filter:
Arc_1dfit_A_7_DET01_S0705

debora-pe:
I have re-run the Dev Suite for MOSFIRE, after a few minor changes and passed:

```
--- PYPEIT DEVELOPMENT SUITE PASSED 10/10 TESTS (Masters ignored) ---
Testing Started at 2022-04-13T18:31:34.399192
Testing Completed at 2022-04-13T19:07:17.158787
Total Time: 0:35:42.759595
```

pypeit/wavemodel.py
```diff
@@ -717,7 +717,7 @@ def create_linelist(wavelength, spec, fwhm, sigdetec=2.,
 
 def create_OHlinelist(resolution, waveminmax=(0.8,2.6), dlam=40.0, flgd=True, nirsky_outfile=None,
                       fwhm=None, sigdetec=3., line_name='OH', file_root_name=None, iraf_frmt=False, 
-                      debug=False):
+                      vacuum=False, debug=False):
     """Create a synthetic sky spectrum at a given resolution, extract significant lines, and
     store them in a PypeIt compatibile file. The skymodel is built from nearIR_modelsky and
     includes black body at 250K, OH lines, and H2O lines (but only at lambda>2.3microns).
```
jhennawi: I'm lost as to why vacuum is False here. All pypeit wavelengths and linelists are always in vacuum.  Can you please explain?

pypeit/data/arc_lines/lists/OH_MOSFIRE_Y_lines.dat
```diff
@@ -0,0 +1,182 @@
+# Created by DP on 2022-Apr-13. R=3388 (from MOSFIRE website) and dv=40km/s
```
jhennawi, line 1: To all of these files can you explicitly state that the wavelengths are vacuum in the file header.

jhennawi (changes requested):
One very important comment about the whether things are in air or vacuum. It should always be vacuum.

pypeit/wavemodel.py
debora-pe: @jhennawi this is related to what I talked to you about a week ago when I was making the OH linelists. When I first run the `create_OHlinelist` script, the wavelengths of the OH lines were consistently different (~2Ang) than the ones that we have in other linelists (e.g, NIRES). Then, I figured out that the sky lines from Rousselot2000 should be already in vacuum, and because the script `create_OHlinelist` was converting those into vacuum by default, the wavelengths of the OH lines were double corrected. This is why I introduced the `vacuum`  parameter and set it to `False`.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -171,14 +171,56 @@ def config_specific_par(self, scifile, inp_par=None):
...
-            if 'long2pos' in self.get_meta_value(headarr, 'decker'):
+            if 'long2pos' in decker:
                 # exclude the random slits outside the long2pos from slit tracing
                 pix_start, pix_end = self._long2pos_pos()
                 par['calibrations']['slitedges']['exclude_regions'] = ['1:0:{}'.format(pix_start),
                                                                        '1:{}:2040'.format(pix_end)]
                 # assume that the main target is always detected, i.e., skipping force extraction
                 par['reduce']['slitmask']['extract_missing_objs'] = False
 
+        # wavelength calibration
```
jhennawi, line 182: Is there a reason why you didn't just create one OH line list for all of the MOSFIRE setups? I think there is no difference except less code, and possibly less chance for confusion from lines not within the filter, which for full_template should basically never happen. The main upside is that if there were another IR instrument with comparable resolution to MOSFIRE, we could use that list without re-running the wave model code again.

pypeit/wavemodel.py
```diff
@@ -680,7 +680,7 @@ def create_linelist(wavelength, spec, fwhm, sigdetec=2.,
         if True, the file is written in the IRAF format (i.e. wavelength,
         ion name, amplitude).
     vacuum (bool):
-        If True, write the wavelengths in vacuum
+        If True, convert the wavelengths of the created linelist from air to vacuum
```
jhennawi, line 683: Can we please rename this to something like airtovac. Vacuum is a bad name for this option.

pypeit/wavemodel.py
```diff
@@ -716,8 +716,8 @@ def create_linelist(wavelength, spec, fwhm, sigdetec=2.,
 
 
 def create_OHlinelist(resolution, waveminmax=(0.8,2.6), dlam=40.0, flgd=True, nirsky_outfile=None,
-                      fwhm=None, sigdetec=3., line_name='OH', file_root_name=None, iraf_frmt=False, 
-                      debug=False):
+                      fwhm=None, sigdetec=3., line_name='OH', file_root_name=None, iraf_frmt=False,
```
jhennawi, line 719: convert_air_to_vac should never be an option. I think you misunderstood what I wanted here. Happy to chat more. The vacuum option in create_linelist should be renamed to convert_air_to_fac. In the code to create linelists, there should be no convert_air_to_vac option anymore since all the input linelists (Rousselot, HITRAN) are in vacuum.

jhennawi (approved):

jhennawi (changes requested):
Still some minor comments.

jhennawi (approved):
Great thanks!

debora-pe:
Merging

PR #1376, opened 2022-04-12
Fix to GNIRS XD 32 wavelengths for bluest order

jhennawi:
Fixed an issue with continuum subtraction and threshold determination for arc line spectra that have significant zero-padding. The arc line peak threshold was being set to be too high.

We probably need to run the full test suites and I could use some help with that since I'm not yet running in the cloud.


feigewang (approved):
I think running test is necessary for such change!

pypeit/core/arc.py
```diff
@@ -879,7 +879,7 @@ def detect_lines(censpec, sigdetect=5.0, fwhm=4.0, fit_frac_fwhm=1.25, input_thr
 
     bpm: `numpy.ndarray`_, optional
         Bad-pixel mask for input spectrum. If None, all pixels
-        considered good.
+        considered good. If passed in shape must match that of censpec
```
kbwestfall, line 882: Thanks for the doc improvements!

pypeit/core/arc.py
```diff
@@ -916,25 +916,27 @@ def detect_lines(censpec, sigdetect=5.0, fwhm=4.0, fit_frac_fwhm=1.25, input_thr
...
-    xrng = np.arange(detns.size, dtype=np.float)
+    # OLD CODE
+    #if len(censpec.shape) == 3:
+    #    detns = censpec[:, 0].flatten()
+    #else:
+    #    detns = censpec.copy()
+    #detns = detns.astype(np.float)
+
+    bpm_out = censpec == 0.0 if bpm is None else bpm
+    xrng = np.arange(censpec.size, dtype=np.float)
```
kbwestfall, line 929: Is there a case where the calling function passes a 3D array?  If this works, I'd suggest removing the commented code and the TODO comment above.

kbwestfall (commented):
Looks good to me.  I'm only leaving a "comment" for now until we've checked that tests and the dev-suite pass.

kbwestfall:
```======================= 324 passed in 1421.55s (0:23:41) =======================```

kbwestfall (approved):

PR #1379, opened 2022-04-16
HOTFIX: QA path

rcooke-ast:
As titled

### DIFF
--- a/pypeit/core/qa.py
+++ b/pypeit/core/qa.py
@@ -339,6 +339,18 @@ def html_exp_pngs(exp_name, det):
     # Return
     return links, body
 
+
+def gen_qa_dir(qa_path):
+    """ Make the QA directory if it doesn't already exist
+
+    Args:
+        qa_path (str):
+            Path to the QA folder
+    """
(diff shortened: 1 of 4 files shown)


profxj (approved):
Looks fine to me, but am not sure what problem it is fixing.

Can you add a unit test that would fail with the previous version
of the code?  Not necessary, but would be helpful.

pypeit/core/qa.py
```diff
@@ -339,6 +339,18 @@ def html_exp_pngs(exp_name, det):
...
+    Args:
+        qa_path (str):
+            Path to the QA folder
+    """
+    if not os.path.exists(qa_path):
+        os.makedirs(qa_path)
+
+
 def gen_mf_html(pypeit_file, qa_path):
     """ Generate the HTML for a MasterFrame set
```
rcooke-ast, line 356: The crash happens on line 384 (`with open(MF_filename,'w') as f:`) if the directory `qa_path` doesn't already exist.

kbwestfall (approved):
Fine with me!  Thanks for putting this in!

PR #1380, opened 2022-04-18
Remote-Hosting of Certain Package Data

tbowers7:
The main thrust of this PR is to reduce the size of the PyPI-hosted release packages by omitting various package-included data such as `reid_arxiv` and `sensfunc` files.  

Major Changes:
* Add download/caching for various package-included data using AstroPy's [`download_file()`](https://docs.astropy.org/en/stable/api/astropy.utils.data.download_file.html#astropy.utils.data.download_file) functionality.  Remove data files from package export in the `MANIFEST.in` file.
* Use this framework to directly pull Telluric Atmospheric Grid files from the cloud into the user's cache without separate download from Google Drive.
* Also use this framework to save the output of `pypeit_identify` (`wvarxiv.fits`) into the user's cache as a way to directly and quickly use manual wavelength identifications in reductions.
* Add unit tests to ensure the downloading introduced in this PR continues to work.

Minor Changes:
* Tweak the `pypeit_identify` GUI to display the spectrum on log scale by default (linear scale selectable by command line argument).
* Address #1345, since I was tweaking wavelength stuff anyways.

Further Discussion:
Especially since most users will not need most of the files in the `pypeit.data` directories (and they are somewhat large), this PR adds functionality to download them from the GitHub repo at runtime when needed.

The downloading and caching process takes advantage of AstroPy's [`download_file()`](https://docs.astropy.org/en/stable/api/astropy.utils.data.download_file.html#astropy.utils.data.download_file) functionality, along with user-space caching.  Downloaded files are nominally kept in `~/.pypeit/cache`, but this can be changed using the [Astropy configuration system](https://docs.astropy.org/en/stable/config/index.html).  For files being kept out of the release distribution, PypeIt now checks first the package directory, then the cache, then downloads the file if not already found.

Taking advantage of this new framework, this PR also revamps the `pypeit_install_telluric` script to download a specified Telluric Atmospheric Grid file from the cloud into the cache for use.  This alleviates the need for separately downloading said files from the Google Drive and installing them into the package directory.

A feature requested by `LDT/DeVeny` users is the ability to quickly and easily use manual wavelength calibrations for their reductions.  This PR uses the new caching functionality to save a copy of the output `wvarxiv.fits` from `pypeit_identify` (with date/time in the filename) for direct reuse with the `full_template` method.

Unit Tests Pass:
```
platform linux -- Python 3.9.12, pytest-7.1.1, pluggy-1.0.0
3.7.0, doctestplus-0.12.0, hypothesis-6.41.0
collected 329 items                                                            
======================= 329 passed in 1868.46s (0:31:08) =======================
```

Dev Suite (branch: `develop`) is getting caught on the newer `keck_lris_red_mark4` files -- also a `gemini_gnirs` error "No matching objects for OBJ0659-DET01", but otherwise passes.  May need to run the `dev-suite` in the cloud.


tbowers7:
The CI tests are revealing some issues with pathnames and filenames in this new scheme.  I will need to address them before this PR can be reviewed.

pypeit/core/gui/identify.py
```diff
@@ -269,7 +276,14 @@ def initialise(cls, arccen, lamps, slits, slit=0, par=None, wv_calib_all=None,
         fig, ax = plt.subplots(figsize=(16, 9), facecolor="white")
         plt.subplots_adjust(bottom=0.05, top=0.85, left=0.05, right=0.65)
         ax.add_line(spec)
-        ax.set_ylim((0.0, 1.1 * spec.get_ydata().max()))
+        if y_log:
```
profxj, line 283: Nice added feature

pypeit/par/pypeitpar.py
```diff
@@ -2476,18 +2476,6 @@ def __init__(self, reference=None, method=None, echelle=None, ech_fix_format=Non
         descr['sigrej_final'] = 'Number of sigma for rejection for the final guess to the ' \
                                 'wavelength solution.'
 
-        # TODO: Not used
-        # Backwards compatibility with basic and semi_brute algorithms
-        defaults['wv_cen'] = 0.0
-        dtypes['wv_cen'] = float
-        descr['wv_cen'] = 'Central wavelength. Backwards compatibility with basic and ' \
```
profxj, line 2490: well pruned

pypeit/spectrographs/gtc_osiris.py
```diff
@@ -276,58 +276,58 @@ def config_specific_par(self, scifile, inp_par=None):
 
         # Wavelength calibrations
         if self.get_meta_value(scifile, 'dispname') == 'R300B':
-            par['calibrations']['wavelengths']['wv_cen'] = 4405.
-            par['calibrations']['wavelengths']['disp'] = 4.96
+            # par['calibrations']['wavelengths']['wv_cen'] = 4405.
```
profxj, line 279: @rcooke-ast -- you might confirm this is ok

pypeit/spectrographs/wht_isis.py
```diff
@@ -171,8 +171,8 @@ def default_pypeit_par(cls):
         par['calibrations']['wavelengths']['n_first'] = 3
         par['calibrations']['wavelengths']['n_final'] = 5
         par['calibrations']['wavelengths']['sigdetect'] = 10.0
-        par['calibrations']['wavelengths']['wv_cen'] = 4859.0
-        par['calibrations']['wavelengths']['disp'] = 0.2
+        # par['calibrations']['wavelengths']['wv_cen'] = 4859.0
```
profxj: also a quick check for @rcooke-ast

pypeit/tests/test_pkgdata.py
```diff
@@ -0,0 +1,55 @@
+"""
```
profxj, line 1: Thanks for the tests!!

profxj (approved):
Great stuff.

Should we consider a test in the DevSuite too?
Or do you feel what you've generated is enough?

Am approving either way.  But maybe wait for @rcooke-ast to comment on the elmination of semi-brute and basic.  Thanks!

pypeit/core/gui/identify.py
```diff
@@ -269,7 +276,14 @@ def initialise(cls, arccen, lamps, slits, slit=0, par=None, wv_calib_all=None,
...
+            ax.set_yscale('log')
+            ax.set_ylim( (max(1., spec.get_ydata().min()),
+                        4.0 * spec.get_ydata().max()))
+        else:
+            ax.set_ylim((0.0, 1.1 * spec.get_ydata().max()))
```
rcooke-ast, line 289: Perhaps be explicit, and add:
`ax.set_yscale('linear')`

I can see reasons why we might want to toggle between log and linear during the fitting (e.g. the user... likely me... forgets to put the command line flag)

pypeit/core/gui/identify.py
```diff
@@ -701,17 +715,27 @@ def store_solution(self, final_fit, binspec, rmstol=0.15,
...
+                cachename = f"manual_{self.specname}_{date_str}.fits"
+                data.write_file_to_cache("wvarxiv.fits",
+                                         cachename,
+                                         "arc_lines/reid_arxiv")
+
+                msgs.info("Your arxiv solution has been written to ./wvarxiv.fits\n")
+                msgs.info(f"Your arxiv solution has been cached.{msgs.newline()}"
+                          f"Use 'reid_arxiv = {cachename}' and{msgs.newline()}"
+                          f"'method = full_template' in your PypeIt{msgs.newline()}"
+                          "Reduction File to utilize this wavelength solution.")
```
rcooke-ast: Should we add the explicit lines of code the user needs to copy paste into their reduction file?

```
[calibrations]
  [[wavelengths]]
    reid_arxiv = {cachename}
    method = full_template
```

pypeit/par/pypeitpar.py
```diff
@@ -2554,7 +2542,7 @@ def valid_methods():
         Return the valid wavelength solution methods.
         """
         # TODO: Remove from this list anything that is not valid!
-        return ['simple', 'semi-brute', 'basic', 'holy-grail', 'identify', 'reidentify',
+        return ['simple', 'holy-grail', 'identify', 'reidentify',
```
rcooke-ast, line 2553: Yes! Be gone `basic` and `semi-brute`!  I don't think they are even used...

pypeit/spectrographs/wht_isis.py
rcooke-ast: Yes, this can be deleted

pypeit/spectrographs/gtc_osiris.py
rcooke-ast: Yes, fine to comment it out, but maybe don't delete just yet - thanks!

tbowers7:
> Should we consider a test in the DevSuite too?
> Or do you feel what you've generated is enough?

@profxj : I think an independent `dev-suite` run in the cloud will be very useful given the new feature of needing to download certain data from a remote server.  I found several issues just with the CI tests, and I would appreciate having a separate `dev-suite` run not on my own machine.

pypeit/par/pypeitpar.py
tbowers7: They are not used... I did a thorough check of all non-deprecated code and found they are commented out.  I chose not to delete the commented code sections entirely, but maybe we can consider moving them to `deprecated/` somehow and remove the commented code from `pypeit/core/arc.py`.

tbowers7:
@tepickering : I'm interested in your view of how I implemented AstroPy's [`download_file()`](https://docs.astropy.org/en/stable/api/astropy.utils.data.download_file.html#astropy.utils.data.download_file) functionality, and if it conforms to standard usage.

pypeit/core/gui/identify.py
tbowers7: Added.

pypeit/core/gui/identify.py
tbowers7: Added.

pypeit/data/utils.py
```diff
@@ -121,6 +132,413 @@ class Paths(metaclass=Paths_meta):
...
+
+        # Raise the appropriate error message
+        msgs.error(err_msg)
+
+
+def write_file_to_cache(filename, cachename, filetype, remote_host="github"):
+    """write_file_to_cache Use `astropy.utils.data` to save local file to cache
+
+    This function writes a local file to the PypeIt cache as if it came from a
+    remote server.  This is useful for being able to use lcoally created files
```
tepickering: small typo

tepickering (approved):
really great stuff! i heartily approve!

tepickering:
> @tepickering : I'm interested in your view of how I implemented AstroPy's [`download_file()`](https://docs.astropy.org/en/stable/api/astropy.utils.data.download_file.html#astropy.utils.data.download_file) functionality, and if it conforms to standard usage.

looks good to me. like i said in the slack convo, ours is the exact use case that it was built for and you're using it exactly as intended.

tbowers7:
Cloud run of `dev-suite` passes with only 1 (well-documented) error:
```
And it is our (not so favorite):
End of Log:
  File "/PypeIt/pypeit/pypmsgs.py", line 185, in error
    raise PypeItError(msg)
pypeit.pypmsgs.PypeItError: No matching objects for OBJ0211-DET01.  Odds are you input the wrong OBJID
```

I will add a little more documentation about how to run PypeIt without internet access (_i.e._, how to pre-download necessary files) before merging.

tbowers7:
@kbwestfall : I would appreciate your eye on these last few changes, in addition to any comments you have on the PR overall.

Some more doc changes forthcoming.

doc/installing.rst
```diff
@@ -229,43 +240,36 @@ Atmospheric Model Grids
...
+The downloaded file will exist in the ``PypeIt`` cache, and will persist through
+upgrades of your installation via `pip`_ or `conda`_.  To force the update of a
+telluric model grid file to the latest version, simply run ``pypeit_install_telluric``
+with the ``--force_update`` option.
+
+If you require a telluric grid that is not presently hosted in the cloud, the code will
+instruct you to download the file separately from the `PypeIt dev-suite Google Drive`_.
+Users may select any of the files in the Google Drive for their telluric correction,
+download them sepatately, then install them using the ``--local`` option to
+``pypeit_install_telluric``.
```
kbwestfall, line 279: Awesome!

doc/wave_calib.rst
```diff
@@ -181,20 +181,26 @@ quit the GUI to see if you want to save the solution. Note,
...
+
+   .. code-block:: console
+
+      [INFO]    :: Your arxiv solution has been written to ./wvarxiv.fits
+      [INFO]    :: Your arxiv solution has also been cached.
+                  To utilize this wavelength solution, insert the
+                  following block in your PypeIt Reduction File:
+                  [calibrations]
+                     [[wavelengths]]
+                        reid_arxiv = manual_ldt_deveny_20220422T1006.fits
```
kbwestfall: Should this be `wvarxiv.fits` to match the example in the text?

pypeit/data/utils.py
```diff
@@ -121,6 +133,430 @@ class Paths(metaclass=Paths_meta):
     """
 
 
+# Remote-fetch functions for package data not distributed via PyPI ===========#
+def get_reid_arxiv_filepath(arxiv_file, copy_to_pkgdir=False):
+    """get_reid_arxiv_filepath Return the full path to the `reid_arxiv` file
```
kbwestfall: You don't need the function name in the docstring.

pypeit/data/utils.py
```diff
@@ -121,6 +133,430 @@ class Paths(metaclass=Paths_meta):
...
+    As most users will need only a small number of `reid_arxiv` files for thier
+    particular reductions, the remote fetch will only occur once per file (per
+    version of PypeIt).
+
+    Args:
+        arxiv_file: str
+          The base filename of the `reid_arxiv` file to be located
+        copy_to_pkgdir: bool, optional
+          Force copy any cached downloaded file into the package directory
+          [Default: False]
```
kbwestfall: I don't think the above will render correctly.  Check the syntax and distinction between Google and Numpy styles: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#google-vs-numpy .  You can pick either, but the syntax has to consistently conform to one of the two within any given docstring.

pypeit/sensfilearchive.py
```diff
@@ -67,11 +67,12 @@ class DEIMOSSensFileArchive(SensFileArchive):
     """SensFileArchive subclass specifically for keck_deimos SensFuncs."""
     spec_name = "keck_deimos"
 
-    def get_archived_sensfile(self, fitsfile):
+    def get_archived_sensfile(self, fitsfile, unit_test=False):
```
kbwestfall: Can we change this to be explicitly `copy_to_pkgdir`?  I definitely appreciate the intention to test this with a unit test.  But, as much as possible, functions should not "know" whether or not they're being executed as part of a unit test.

kbwestfall (changes requested):
Amazing work, @tbowers7 !!  Thanks for all your effort on this.  I have a couple minor requests, though:

 - Please check the docstring syntax.  I don't think Sphinx will correctly render what you have in many of the `data.utils` functions.
 - I realize I'm in the minority, but the use of `as` is a pet peeve of mine.  Unless it's an "industry standard" (like `import numpy as np`), I think we should almost always avoid using this.  Specifically, instead of `from astropy.utils import data as astropy_data`, I'd prefer `import astropy.utils.data`; and instead of `from pypeit import __version__ as pypeit_version`, I'd prefer `from pypeit import __version__`.  And then propagate the relevant changes.

Three other thoughts/questions not necessarily for this PR:

- Similar to your excellent changes to the telluric installation script, is it useful to update the QL masters installation script?
- Can you remind me of the use case for the `copy_to_pkgdir` option?  Is it possible/useful to instead have a symlink installed in the package directory?  I'm wary of having multiple files on disk, particularly for the large telluric files.
- For the telluric files, can we eventually get all the telluric files in the cloud?  I think I remember that being the goal, but I don't remember the hang-up.

profxj:
Regarding @kbwestfall 's comments on the QL Masters' files, 
  I am going to generate Cloud scripts for `release` and `develop`
 that will run `build_ql_masters` and push the outputs to `s3`.  
 And expose them.

I'll then ask @tbowers7 for help to integrate them into the 
  framework he build for this PR.  Bottom line, defer the masters files
  until a (not so) future PR.  Thx!

doc/wave_calib.rst
tbowers7: I've addressed this in 2 ways.
  1. No longer writing to generic `wvarxiv.fits`, but instead writing to `wvarxiv_{spec}_{datetime}.fits` in the local directory.
  2. I've corrected the text above this code block to indicate what is saved in the cache is labeled with the spectrograph and timestamp of creation for uniqueness.  This is in case a user is doing manual IDs for multiple setups and wants to be able to use all of them.

tbowers7:
@kbwestfall :
> * Can you remind me of the use case for the `copy_to_pkgdir` option?  Is it possible/useful to instead have a symlink installed in the package directory?  I'm wary of having multiple files on disk, particularly for the large telluric files.

This option is only necessary for a single unit test case to appease an `assert` statement regarding the name of a `sensfunc` file.  I've removed the option from the `reid_arxiv` and `skisim` functions, and have converted the remaining instance to a symlink situation.

> * For the telluric files, can we eventually get all the telluric files in the cloud?  I think I remember that being the goal, but I don't remember the hang-up.

@profxj has uploaded to the cloud the telluric files specifically called out in spectrograph parameter lists.  I'll defer to him the question of putting the remaining files there.

profxj:
There is no problem putting all the `telluric` files in the cloud.
  I simply prioritized the ones required for redux

kbwestfall:
Sorry to harp on this, @tbowers7 , but the docstring syntax still won't render correctly.  Here's an example for the `get_reid_arxiv_filepath` function:

Current:
```
    Args:
        arxiv_file: str
          The base filename of the ``reid_arxiv`` file to be located

    Returns:
        calibfile: str
          The full path to the ``reid_arxiv`` file
```
Renders as:
incorrect_rendering

Corrected Google style:
```
    Args:
        arxiv_file (str):
            The base filename of the ``reid_arxiv`` file to be located

    Returns:
        str: The full path to the ``reid_arxiv`` file
```
Renders as:
correct_google_rendering

Corrected Numpy Style:
```
    Parameters
    ----------
    arxiv_file: str
        The base filename of the ``reid_arxiv`` file to be located

    Returns
    -------
    calibfile: str
        The full path to the ``reid_arxiv`` file
```
correct_numpy_rendering

Everything else looks great!

tbowers7:
@kbwestfall : Thanks for your patience with the docstring syntax.  Sphinx is still on my "to learn" list.

kbwestfall:
Thanks for being willing to bear with me...  There's one final issue, which is a subtle difference between the return syntax for Google and numpy styles.  Google style doesn't allow you define the name of the returned object, whereas you can with numpy format; i.e., the fact that I did not include the name of the returned object in the corrected syntax above is important for the former.  You're only allowed to define the type and provide a description in Google style.  This means that when you're returning multiple objects, the return type is always `tuple`, and you provide the components of the tuple in the description.  If you think it's important to provide the object names in the return statement, you have to use Numpy style.

tbowers7:
If it's still not right, maybe you can expose this branch on readthedocs so I can continue tweaking until it's right.

I usually use Numpy style (but not with Sphinx) for my docstrings, but was trying to match existing PypeIt style.

kbwestfall:
The quick way to check how things render is to make the html locally:
```
cd doc
make clean ; make html
open _build/html/index.html
```
and then navigate to the relevant page.  All the function docs are accessed through the `Python API` link in the navigation bar.  The `make html` command also runs a bunch of scripts to create some of the auto-documented files (like the datamodel and parameters).  If you don't need to remake these (i.e., if you're just editing function doc strings), you can speed up the process by just calling `make htmlonly`, instead of `make clean ; make html`.  This still takes a while, but it should be faster than waiting for an exposed branch on readthedocs to rebuild.

The "PypeIt style" is a bit all over the place.  I prefer Google style, Joe prefers Numpy style, and there are still many functions with syntax errors.  If you're more comfortable with Numpy style, I would stick with it and just make sure that you use syntax that Sphinx can correctly render in html.

I'll go ahead and approve now, but it looks like there are some CI failures that should be fixed before merging.

Thanks again for all your effort on this!

kbwestfall (approved):

PR #1381, opened 2022-04-19
KCWI flux calibration

rcooke-ast:
This PR brings the Keck/KCWI data reduction to completion, and fixes a few minor bugs (e.g. mask_abs_line was not being passed).

Tests pass:
image

I've run this on the `Cooked` dev suite, and generated a new Cooked file...


pypeit/core/coadd.py
```diff
@@ -762,11 +762,10 @@ def sn_weights(waves, fluxes, ivars, masks, sn_smooth_npix, const_weights=False,
...
         refscale = (sn_val[:, ref_spec] > 0) / (sn_val[:, ref_spec] + (sn_val[:, ref_spec] == 0))
         for iexp in range(nstack):
-            if iexp != ref_spec:
-                # Compute the relative (S/N)^2 and update the mask
-                sn2[iexp] /= sn2[ref_spec]
-                mask_stack[:, iexp] *= (mask_stack[:, ref_spec]) | (sn_val[:, ref_spec] != 0)
-                sn_val[:, iexp] *= refscale
+            # Compute the relative (S/N)^2 and update the mask
+            sn2[iexp] /= sn2[ref_spec]
+            mask_stack[:, iexp] *= (mask_stack[:, ref_spec]) | (sn_val[:, ref_spec] != 0)
```
jhennawi: Formally should this not be a boolean operation rather than *=? I'm sure this does the same thing but is less clear since mask_stack is a boolean array.

pypeit/core/datacube.py
```diff
@@ -58,10 +58,13 @@ class DataCube(datamodel.DataContainer):
             Build from PYP_SPEC
 
     """
-    version = '1.0.1'
+    version = '1.0.2'
 
-    datamodel = {'flux': dict(otype=np.ndarray, atype=np.floating, descr='Flux array in units of counts/s/Ang or 10^-17 erg/s/cm^2/Ang'),
+    datamodel = {'flux': dict(otype=np.ndarray, atype=np.floating, descr='Flux array in units of counts/s/Ang/arcsec^2'
```
jhennawi, line 63: I'm not following why this is either or?

pypeit/core/datacube.py
```diff
@@ -267,6 +270,228 @@ def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, re
     return ra_diff, dec_diff
 
 
+def gaussian2D_cube(tup, amplitude, xo, yo, dxdz, dydz, sigma_x, sigma_y, theta, offset):
```
jhennawi: Please add documentation.

pypeit/core/datacube.py
```diff
@@ -267,6 +270,228 @@ def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, re
...
+    a = (np.cos(theta)**2)/(2*sigma_x**2) + (np.sin(theta)**2)/(2*sigma_y**2)
+    b = -(np.sin(2*theta))/(4*sigma_x**2) + (np.sin(2*theta))/(4*sigma_y**2)
+    c = (np.sin(theta)**2)/(2*sigma_x**2) + (np.cos(theta)**2)/(2*sigma_y**2)
+    norm = 1/(2*np.pi*np.sqrt(a*c-b*b))
+    g = offset + norm*amplitude*np.exp( - (a*((x-xo)**2) + 2*b*(x-xo)*(y-yo)
+                            + c*((y-yo)**2)))
+    return g.ravel()
+
+
+def gaussian2D(tup, amplitude, xo, yo, sigma_x, sigma_y, theta, offset):
```
jhennawi: Documentation please. Inline comments are also strongly encouraged. Right now it does not look anyone else can maintain or modify this code.

pypeit/core/datacube.py
```diff
@@ -267,6 +270,228 @@ def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, re
...
+    b = -(np.sin(2*theta))/(4*sigma_x**2) + (np.sin(2*theta))/(4*sigma_y**2)
+    c = (np.sin(theta)**2)/(2*sigma_x**2) + (np.cos(theta)**2)/(2*sigma_y**2)
+    norm = 1/(2*np.pi*np.sqrt(a*c-b*b))
+    g = offset + norm*amplitude*np.exp( - (a*((x-xo)**2) + 2*b*(x-xo)*(y-yo)
+                            + c*((y-yo)**2)))
+    return g.ravel()
+
+
+def rebinND(a, shape):
+    sh = shape[0],a.shape[0]//shape[0],shape[1],a.shape[1]//shape[1]
```
jhennawi: Docs, comments.

pypeit/core/datacube.py
```diff
@@ -267,6 +270,228 @@ def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, re
...
+        # oprof = model.repeat(numwave, axis=2).reshape((numim, numwave)).T
+        # sobj = specobj.SpecObj('IFU', 'DET01', SLITID=0)
+        # extract.extract_optimal(sciimg, ivar, optmask, waveimg, skyimg, thismask, oprof, sobj)
+        # opt_flux, opt_var, opt_gpm = sobj.OPT_COUNTS, sobj.OPT_COUNTS_SIG**2, sobj.OPT_MASK
+        # # Setup the return values
+        # ret_flux, ret_var, ret_gpm = opt_flux, opt_var, opt_gpm
+    else:
+        msgs.error("Unknown extraction method: ", method)
+
+    # Convert to counts/s/A
```
jhennawi: I'm not following why this is counts per second per angstrom and not just counts per second. Where have you divided by the wavelength grid?

pypeit/core/datacube.py
```diff
@@ -575,69 +889,155 @@ def coadd_cube(files, spectrograph=None, parset=None, overwrite=False):
...
+        relscl = spec2DObj.scaleimg/relScaleImg
+        sciimg = (spec2DObj.sciimg-spec2DObj.skymodel)*relscl  # Subtract sky
+        ivar = spec2DObj.ivarraw / relscl**2
         waveimg = spec2DObj.waveimg
         bpmmask = spec2DObj.bpmmask
 
         # Grab the slit edges
         slits = spec2DObj.slits
 
+        # embed()
```
jhennawi: please clean up commented regions if you no longer need them.

pypeit/core/flux_calib.py
```diff
@@ -66,17 +66,13 @@ def blackbody_func(a, teff):
         a (float):
```
jhennawi, line 66: Can you add the units of a?

pypeit/core/flux_calib.py
```diff
@@ -66,17 +66,13 @@ def blackbody_func(a, teff):
...
             flux normalisation factor
         teff (float):
-            Effective temperature of the blackbody
+            Effective temperature of the blackbody (in units of K)
```
jhennawi, line 69: If I understand what you've done in this function teff is not a float, but rather an astropy Quantity. Please avoid making satrapy quantities inputs to functions like this. I would make the inputs floats and then clearly specify what their units are in the docs. We have a convention in PypeIt which is that we don't use astropy Quantities except internally in functions and then we also pass back floats (i.e. .value) and not quantities. This is done for very good reasons, namely computations with astropy Quantities can slow down floating point operations by a factor of 10 or more. My suggestion is thus that you pass in floats and pass out floats, and only use the quantities and units internally. I love this functionality in astropy, but I think this is the best approach given how slow it is.

pypeit/core/flux_calib.py
```diff
@@ -210,7 +209,7 @@ def find_standard_file(ra, dec, toler=20.*units.arcmin, check=False):
                 std_dict['flux'] = std_dict['flux'][np.logical_not(mask)]
             elif sset == 'blackbody':
                 # TODO let's add the star_mag here and get a uniform set of tags in the std_dict
-                waves, flam = blackbody_func(star_tbl[_idx]['a_x10m23'], star_tbl[_idx]['T_K'])
+                waves, flam = blackbody_func(star_tbl[_idx]['a_x10m23']*1.0E-23, star_tbl[_idx]['T_K']*units.K)
```
jhennawi: maybe define 1.0e-23 with a global constant similar to PYPEIT_FLUX_SCALE?

pypeit/core/flux_calib.py
```diff
@@ -724,29 +735,9 @@ def get_sensfunc_factor(wave, wave_zp, zeropoint, exptime, tellmodel=None, extin
     return senstot/exptime/delta_wave
 
 
-# JFH TODO This code needs to be cleaned up. The telluric option should probably be removed. Logic is not easy to follow.
-def fit_zeropoint(wave, counts, counts_ivar, counts_mask, exptime, airmass, std_dict, longitude, latitude,
-                  mask_abs_lines=True, polyorder=4, balm_mask_wid=10.0, nresln=20., resolution=3000.,
-                  trans_thresh=0.9, polycorrect=True, polyfunc=False, debug=False):
-
+def prep_zeropoint(wave, counts, counts_ivar, counts_mask, exptime, airmass, longitude, latitude):
```
jhennawi: How about we rename this counts2Nlam or something like that.

pypeit/core/flux_calib.py
```diff
@@ -761,12 +752,49 @@ def fit_zeropoint(wave, counts, counts_ivar, counts_mask, exptime, airmass, std_
...
+                  mask_abs_lines=True, polyorder=4, balm_mask_wid=10.0, nresln=20., resolution=3000.,
+                  trans_thresh=0.9, polycorrect=True, polyfunc=False, debug=False):
+
+    """
+    Function to generate the sensitivity function. This function fits
+    a bspline to the 2.5*log10(flux_std/flux_counts). The break
+    points spacing, which determines the scale of variation of the
+    sensitivity function is determined by the nresln parameter.
+
+    Args:
```
jhennawi, line 794: Please document the arrays Nlam_star, Nlam_ivar_star, and gpm_star. And also specify their units.

pypeit/core/flux_calib.py
```diff
@@ -786,24 +814,12 @@ def fit_zeropoint(wave, counts, counts_ivar, counts_mask, exptime, airmass, std_
             If you have significant telluric absorption you should be using telluric.sensnfunc_telluric. default = 0.9
 
     Returns:
-            zeropoint (`numpy.ndarray`_): Sensitivity function with same shape as wave (nspec,)
-            mask_sens (`numpy.ndarray`_): Good pixel mask for sensitivity function with same shape as wave (nspec,)
+        zeropoint_data (`numpy.ndarray`_): Sensitivity function with same shape as wave (nspec,)
+        zeropoint_data_gpm (`numpy.ndarray`_): Good pixel mask for sensitivity function with same shape as wave (nspec,)
```
jhennawi, line 818: good pixel mask for zerpoint

pypeit/par/pypeitpar.py
```diff
@@ -645,7 +645,7 @@ def __init__(self, method=None, pixelflat_file=None, spec_samp_fine=None,
 
         defaults['slit_illum_relative'] = False
         dtypes['slit_illum_relative'] = bool
-        descr['slit_illum_relative'] = 'Generate an image of the relative spectral illumination' \
+        descr['slit_illum_relative'] = 'Generate an image of the relative spectral illumination ' \
```
jhennawi, line 648: Can you add to these docs that this functionality is currently only used for IFUs?

jhennawi (changes requested):
Looks good Ryan. Great work here. My comments are relatively minor and mainly relate to documentation. Dev-suite and unit tests definitely need to be run here.

pypeit/core/coadd.py
rcooke-ast: OK - done

pypeit/core/datacube.py
rcooke-ast: It depends whether the cube is flux calibrated or not. The combined standard star datacubes are given in counts.

pypeit/core/datacube.py
rcooke-ast: Done!

pypeit/core/datacube.py
rcooke-ast: Too right - both implemented!

pypeit/core/datacube.py
rcooke-ast: Done!

pypeit/core/datacube.py
rcooke-ast: I've expanded the comment. We are integrating over the spatial coordinates, and each pixel is a surface brightness. The input is already per Angstroms.

pypeit/core/datacube.py
rcooke-ast: Thanks for pointing this out - I've fixed this, and a couple more throughout...

pypeit/core/flux_calib.py
rcooke-ast: Yep - it's dimensionless  :-)

pypeit/core/flux_calib.py
rcooke-ast: That's a really good point - simple to update, and done!

pypeit/core/flux_calib.py
rcooke-ast: Sure - I'm not sure how useful this will be for other routines, as it's just how the table data are stored.

pypeit/core/flux_calib.py
rcooke-ast: Sounds good to me - done!

pypeit/core/flux_calib.py
rcooke-ast: Done

pypeit/par/pypeitpar.py
rcooke-ast: I have added this, but I think @profxj and @marlageha may be using this to account for vignetting of DEIMOS (but I could be mistaken).

rcooke-ast (commented):
Thanks @jhennawi for the quick and solid review!  Comments addressed in all cases. Let me know if you have anything else to add.

pypeit/core/datacube.py
jhennawi: Okay, that is fine but then counts are not counts/s/ang unless you have divided by the exposure time (easy) and the delta_lambda (more complicated since in general it varies with pixel). So has that been done here (perhaps I missed it) or is the label of the units incorrect?

jhennawi (approved):
Hi @rcooke-ast, 

Thanks for addressing everything. Given that you modified the core fluxing routines the full dev suite needs to be run here. If you are not able to do that yourself, another developer can help you run in the cloud.  I'm approving, but please don't merge until the full dev suite passes.

pypeit/core/datacube.py
rcooke-ast: Thanks for checking - the units label is correct here. The exposure time and delta wave is done around line 1200. The delta wave is done on a per pixel basis (as you mention, it's not constant for every pixel, and this is accounted for).

pypeit/core/datacube.py
```diff
@@ -267,6 +270,306 @@ def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, re
...
+    c = (np.sin(theta)**2)/(2*sigma_x**2) + (np.cos(theta)**2)/(2*sigma_y**2)
+    # Normalise so that the integrated flux is a parameter, instead of the amplitude
+    norm = 1/(2*np.pi*np.sqrt(a*c-b*b))
+    gtwod = offset + norm*intflux*np.exp(-(a*((x-xo)**2) + 2*b*(x-xo)*(y-yo) + c*((y-yo)**2)))
+    return gtwod.ravel()
+
+
+def rebinND(img, shape):
+    """
+    Rebin a 2D image to a smaller shape. For example, if img.shape=(100,100),
```
profxj, line 397: Maybe note that the rebinning adopts the `mean`

pypeit/core/datacube.py
```diff
@@ -267,6 +270,306 @@ def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, re
...
+    flxcube = stdcube['FLUX'].data.T.copy()
+    varcube = stdcube['VARIANCE'].data.T.copy()
+    numwave = flxcube.shape[2]
+
+    # Setup the WCS
+    stdwcs = wcs.WCS(stdcube['FLUX'].header)
+    wcs_wav = stdwcs.wcs_pix2world(np.vstack((np.zeros(numwave), np.zeros(numwave), np.arange(numwave))).T, 0)
+    wave = wcs_wav[:, 2] * 1.0E10 * units.AA
+
+    # Generate a whitelight image
```
profxj: while this is a pretty simple 4 lines of code, I wonder whether it should be a method somewhere.  Just consider it

pypeit/core/datacube.py
```diff
@@ -267,6 +270,306 @@ def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, re
...
+    nrmval = np.sum(flxcube != 0.0, axis=2)
+    nrmval[nrmval == 0.0] = 1.0
+    wl_img = np.sum(flxcube, axis=2) / nrmval
+    wlscl = np.max(wl_img)  # Need to make sure the value is of order 1, so it's the same order of magnitude as the other parameters
+
+    # Estimate the centroid and width of the standard star
```
profxj: Similar to the above.  If this is a generally useful method, consider having it stand-alone for wider usage.

pypeit/core/datacube.py
```diff
@@ -267,6 +270,306 @@ def dar_correction(wave_arr, coord, obstime, location, pressure, temperature, re
...
+        global_sky = np.zeros_like(box_sciimg)
+        # Find objects and then extract
+        sobj = findobj_skymask.objs_in_slit(box_sciimg, thismask, slit_left, slit_right)
+        skysub.local_skysub_extract(box_sciimg, box_sciivar, tilts, waveimg, global_sky, thismask, slit_left,
+                             slit_right, sobj, model_noise=False)
+        opt_flux, opt_var, opt_gpm = sobj.OPT_COUNTS[0,:], sobj.OPT_COUNTS_SIG[0,:]**2, sobj.OPT_MASK[0,:]
+        # Setup the return values
+        ret_flux, ret_var, ret_gpm = opt_flux, opt_var, opt_gpm
+
+        ###  OLD VERSION OF OPTIMAL USING 2D PROFILE
```
profxj: Remove?

pypeit/flatfield.py
```diff
@@ -1277,7 +1277,7 @@ def illum_profile_spectral(rawimg, waveimg, slits, slit_illum_ref_idx=0, smooth_
     slit_illum_ref_idx : int
         Index of slit that is used as the reference.
     smooth_npix : int
-        smoothing used for determining smoothly varying S/N ratio weights by sn_weights
+        smoothing used for determining smoothly varying relative weights by sn_weights
```
profxj, line 1287: is this optional?  just checking.

profxj (approved):
Whoa quite the PR!

Is there going to be a test of the fluxing in the DevSuite?
Am hoping so.   But am approving now anyhow.

pypeit/core/datacube.py
rcooke-ast: Done!

pypeit/core/datacube.py
rcooke-ast: Done!

pypeit/flatfield.py
rcooke-ast: Yep - added a note

pypeit/core/datacube.py
rcooke-ast: OK - I've made this its own routine, and beefed it up a little to make it more general and useful...

pypeit/core/datacube.py
rcooke-ast: Done!

rcooke-ast (commented):
Thanks @profxj - comments addressed. Just running some final checks before we run the dev suite

rcooke-ast:
> Whoa quite the PR!
> 
> Is there going to be a test of the fluxing in the DevSuite? Am hoping so. But am approving now anyhow.

Thanks @profxj - I've added a test of the flux calibration in the unit tests. Hope that's OK...

rcooke-ast:
Tests pass...
image

rcooke-ast:
Tests and devsuite pass... merging!

image

image
(the three dev suite failures are expected failures that are fixed with an update to the dev suite)

PR #1386, opened 2022-04-24
Remote package data yscale

rcooke-ast:
Added log/linear toggling for identify y-axis scale


tbowers7 (approved):
Looks great!  I like the ability to toggle between the linear and log scales during the fit.

PR #1401, opened 2022-05-11
turn off astrometric+grating corrections

rcooke-ast:
A fix to `test_coadd`... the other option is to include the master alignments and master flats in Cooked, or refactor the coadd3d file... Probably the latter needs to be done eventually, but this fix is enough for now.


jhennawi (approved):
Thanks Ryan!

PR #1403, opened 2022-05-12
GTC/OSIRIS -- added BPM, datasec, oscansec

rcooke-ast:
As titled... A few updates to GTC/OSIRIS.

Also, there appears to be a bug in specobjs that I encountered. Fixed here. The bug is that OPT_COUNTS is a vector of vectors, and not a matrix as the code currently assumes. Not sure why this hasn't been encountered before, so maybe I'm missing something!


rcooke-ast:
Yep... same problems with develop... and the added problem of "cooked"
image

pypeit/edgetrace.py
```diff
@@ -3672,9 +3672,11 @@ def nudge_traces(self, trace_cen):
             `numpy.ndarray`_: The nudged traces.
         """
         # Check input
-        if self.par['max_nudge'] is not None and self.par['max_nudge'] <= 0:
-            # Nothing to do
-            return trace_cen
+        if self.par['max_nudge'] is None:
```
jhennawi: I'm not following the logic here. At the very least the documentation needs to make clear what the behavior is if max_nudge is None. Further downstream it looks as if the routine still does stuff if max_nudge is None, so I'm not following. I think @kbwestfall should take a look.

pypeit/specobjs.py
```diff
@@ -262,10 +262,14 @@ def get_std(self, multi_spec_det=None):
         pypeline = (self.PYPELINE)[0]
         if 'MultiSlit' in pypeline or 'IFU' in pypeline:
             # Have to do a loop to extract the counts for all objects
-            if self.OPT_COUNTS[0] is not None:
-                SNR = np.median(self.OPT_COUNTS*np.sqrt(self.OPT_COUNTS_IVAR), axis=1)
-            elif self.BOX_COUNTS[0] is not None:
-                SNR = np.median(self.BOX_COUNTS*np.sqrt(self.BOX_COUNTS_IVAR), axis=1)
+            if len(self.OPT_COUNTS) != 0 and self.OPT_COUNTS[0] is not None:
```
jhennawi: I am not following why these changes were necessary. Can you please elaborate. It is odd that you are encountering a bug here that was not encountered before.

jhennawi (changes requested):
Thanks for this PR. I have some questions.

pypeit/edgetrace.py
rcooke-ast: The default value of `max_nudge` is "None". Previously, a "None" value wasn't caught by this if statement. I think it should be `if max_nudge is None or max_nudge <=0` then do nothing and return. My update is first checking if max_nudge is None, and if not, check if max_nudge is positive.

pypeit/specobjs.py
rcooke-ast: The bug is that OPT_COUNTS is a vector of vectors, and not a matrix, so you can't use the "axis" keyword. I agree though, I'm surprised this was never encountered before... I'm not sure if this is something due to a recent PR, and not captured by the dev suite?

Perhaps in all previous reductions only a single standard star was identified? In the case of the data that I'm reducing, there are multiple stars detected in the standard star exposure. Otherwise, I'm stumped why this is something previously not encountered???

profxj: I think your guess is a good one, i.e. multiple standards

pypeit/spectrographs/gtc_osiris.py
```diff
@@ -61,7 +62,8 @@ def get_detector_par(self, det, hdu=None):
             numamplifiers   = 1,
             gain            = np.atleast_1d([0.95]),
             ronoise         = np.atleast_1d([4.5]),
-            datasec         = np.atleast_1d('[1:4102,52:1880]')
+            datasec         = np.atleast_1d('[1:4102,280:2048]'),
```
profxj, line 65: Did the detector change?  Or just our understanding of it.
Just checking.

profxj (approved):
Looks good.  Thx for this.
Please confirm it runs smoothly on the DevSuite for this spectrograph.

pypeit/spectrographs/gtc_osiris.py
rcooke-ast: The left edge is vignetted, so data outside this vignetted region is useless and can affect the edge tracing, so much cleaner to just chop it off.

rcooke-ast:
> Looks good. Thx for this. Please confirm it runs smoothly on the DevSuite for this spectrograph.

Will do!  I'll run the dev-suite shortly... (unless it is preferred to run this in tandem with other PRs (as discussed in the Slack).

pypeit/edgetrace.py
kbwestfall: This changes the behavior of the code.  In the original logic, a value of `max_nudge = None` means "nudge the trace as far as you want", whereas now it means "don't nudge the traces."  The latter was achieved in the previous logic by setting `max_nudge = 0` (or a negative number.  The description in `EdgeTracePar` attempts to explain this, but it could be more clear.  Unless there's a failure, I suggest we revert the logic.

rcooke-ast: OK - this makes sense, in which case we need to update the info msg on line 3687 - I'll do this  :-)  Thanks for the clarification @kbwestfall!

kbwestfall: Good point!  Thanks for correcting the info message.

rcooke-ast:
Tests pass:
image

pypeit/edgetrace.py
jhennawi: Thanks @kbwestfall. That was my concern that the behavior was being changed.

pypeit/specobjs.py
jhennawi: Hi @rcooke-ast, sorry to be picky here. Can you just use np.atleast_2d or np.reshape to cast the opt_counts to a 2d array and then use the axis keyword and forego the loop. That is less code and easier to read, but then I'm not sure I follow the shape of opt_counts as a vector of vectors so this may not work.

rcooke-ast: Strange... I'm no longer getting multiple standards... so I just reverted the code and had no issues... Nothing obvious has changed in recent PRs, so it's a mystery to me...

jhennawi (approved):
Thanks @rcooke-ast

rcooke-ast:
Tests passing
image

PR #1410, opened 2022-05-21
MOSFIRE flatlampoff

debora-pe:
Introduction fo the `thermalflat` frame type (or lamp off flat).  The `thermalflat` frame is subtracted from the `pixelflat` and `illumflat` frames. The automatic identification of `thermalflat` frames is currently only available for MOSFIRE.

Documentation updated.
MOSFIRE K band pypeit files updated in the Dev Suite, see (and review) PR:  https://github.com/pypeit/PypeIt-development-suite/pull/206


doc/mosfire.rst
```diff
@@ -116,9 +116,11 @@ flag **bound_detector** is set to **True**.
 
 Flat Fielding
 -------------
-Although ``PypeIt`` is able to recognize flat frames taken with the lamps off (and those frame will be
-assigned the frame type ``pixelflat``, ``trace``, but not ``illumflat``), currently the lamps off flats
-are not used to correct the dome flats for the thermal emission generated by the dome during K-band observations.
+``Pypeit`` is able to recognize flat frames taken with the lamps off, and it will assign them the
+``thermalflat`` frame type. Whenever ``thermalflat`` frames are identified in the PypeIt file, ``Pypeit``
+will subtract them from the Master Pixel Flat and Master Illumination Flat. It is responsibility of the
```
jhennawi: These docs are incorrect. We are not differencing Master frames which would sort of defeat the entire purpose of using lampooff flats. We are subtracting the lampoff flat from the flat images before constructing the masters. Please update.

pypeit/calibrations.py
```diff
@@ -562,10 +576,22 @@ def get_flats(self):
...
         # Only build illum_flat if the input files are different from the pixel flat
         if not pix_is_illum and len(illum_image_files) > 0:
+            msgs.info('Creating Master Illumination Flat using files: ')
+            for f in illum_image_files:
+                msgs.info(f'{os.path.basename(f)}')
             illum_flat = buildimage.buildimage_fromlist(self.spectrograph, self.det,
                                                         self.par['illumflatframe'],
                                                         illum_image_files, dark=self.msdark,
                                                         bias=self.msbias, bpm=self.msbpm)
+            if len(flatLoff_image_files) > 0:
```
jhennawi, line 586: I think requires a bit more thought. Suppose the user uses sky spectra as illumination flat,  which is sometimes done. Another possibility is two different lamps used to construct pixel flats and illumination flats. Then I don't want to subtract the lampoff flat from the sky spectrum. Again, another example where we need a unique association algorithm. Anyway, I think what you are doing only makes sense for the case where the dome flat and illumflat come from the same set of flats, i.e. if pix_is_illum is True. My suggestion is to just state that as a limitation of the current approach. I don't see a path forward unless you 1) create lampoffillumflats, or 2) solve the problem of calibration frame association that we have discussed.

pypeit/core/framematch.py
```diff
@@ -38,6 +38,7 @@ def __init__(self):
                      ('pinhole', 'Pinhole observation used for tracing slit centers'),
                    ('pixelflat', 'Flat-field exposure used for pixel-to-pixel response'),
                    ('illumflat', 'Flat-field exposure used for illumination flat'),
+                   ('thermalflat', 'Flat-field exposure with lamps off used to remove thermal emission from dome'),
```
jhennawi: Let's discuss the name. As I've said, thermal emission from the dome is only done in the k-band. In the Y through H one uses these flats to remove persistence. It would be more clear if your naming and if your descriptions were more general than the specific K-band usage case. 

I suggest calling them lampoffflats, and here write "to remove persistence from lamp on flat exposures and/or  thermal emission from the telescope and dome."

jhennawi (changes requested):
Minor changes requested.

pypeit/core/framematch.py
debora-pe: changed to `lampoffflats`

pypeit/calibrations.py
debora-pe: Added a note in the documentation.

doc/mosfire.rst
debora-pe: Fixed

jhennawi (approved):
Thanks @deborape. I touched up your docs a bit. I suspect we want a full dev suite run here since calibrations is pretty core functionality and has been modified significantly.

debora-pe:
Dev Suite and unit tests pass.

```
Testing Started at 2022-06-01T15:24:11.304207
Testing Completed at 2022-06-01T23:40:33.113813
Total Time: 8:16:21.809606
```


`=========== 328 passed in 1284.77s (0:21:24) =========`

doc/mosfire.rst
```diff
@@ -116,9 +116,22 @@ flag **bound_detector** is set to **True**.
...
+GUI only for the K-band, but we recommend taking these ``lampoffflats`` for all MOSFIRE spectroscopic
+observations. The purpose of the ``lampoffflats`` is to remove the increase and/or variation in zero level caused
+by persistence from the high counts in the flats and/or thermal emission from the telescope/dome (in the K-band). 
+``Pypeit`` is able to recognize flat frames taken with the lamps off, and it will assign them the
+``lampoffflats`` frame type. Whenever ``lampoffflats`` frames are identified in the PypeIt file, ``Pypeit``
+will subtract them from the frames taken with the lamps on before creating the Master Edges, Master Pixel Flat and
+Master Illumination Flat. It is responsibility of the user to ensure that the ``lampoffflats`` frames
+in the PypeIt file have the same exposure time of the ``trace``, ``pixelflat`` and ``illumflat`` frames.
+
+.. note::
```
jhennawi, line 129: Thanks for updating this!

jhennawi (approved):
Great work @deborape!

profxj (approved):
Looks great to me!

debora-pe:
Re-ru the Dev Suite and unit tests after making some changes and adding the J2 changes:

```
Testing Started at 2022-06-03T16:48:22.855769
Testing Completed at 2022-06-03T23:10:38.387584
Total Time: 6:22:15.531815
```

`============ 328 passed in 1279.89s (0:21:19) ===========`

debora-pe:
merging

PR #1414, opened 2022-05-24
KCWI - detector structure correction

rcooke-ast:
Updates to the `flatfield` module to include structured pixel-to-pixel sensitivity variations in the pixel flat.


rcooke-ast:
Tests pass, fwiw...
image

pypeit/calibrations.py
```diff
@@ -559,14 +559,16 @@ def get_flats(self):
...
         # Only build illum_flat if the input files are different from the pixel flat
         if not pix_is_illum and len(illum_image_files) > 0:
             illum_flat = buildimage.buildimage_fromlist(self.spectrograph, self.det,
-                                                        self.par['illumflatframe'],
-                                                        illum_image_files, dark=self.msdark,
-                                                        bias=self.msbias, bpm=self.msbpm)
-            # Initialise the pixel flat
+                                                        self.par['illumflatframe'], illum_image_files,
+                                                        dark=self.msdark, bias=self.msbias,
+                                                        flatimages=self.flatimages, bpm=self.msbpm)
```
jhennawi, line 587: So you are flattening the illumflat image now? We were not doing this before.

pypeit/core/datacube.py
```diff
@@ -1247,8 +1247,8 @@ def coadd_cube(files, spectrograph=None, parset=None, overwrite=False):
         if cubepar['grating_corr'] and flatfile not in flat_splines.keys():
             msgs.info("Calculating relative sensitivity for grating correction")
             flatimages = flatfield.FlatImages.from_file(flatfile)
-            flatframe = flatimages.pixelflat_model
-            flatframe /= flatimages.fit2illumflat(slits, frametype='pixel', initial=True, flexure_shift=flexure)
+            flatframe = flatimages.illumflat_raw
```
jhennawi: flatframe = flatimages.illumflat_raw/flatimages.fit2illumflat(slits, frametype='illum', initial=True, flexure_shift=flexure)

Would be cleaner since then a new image is created, whereas right now you are modifying flatimages.illumflat_raw in place with a pointer. It is fine I guess if you don't use flatimages.illumflat_raw again, but this pointer stuff often introduces bugs in python.

pypeit/flatfield.py
```diff
@@ -448,7 +447,30 @@ def run(self, debug=False, show=False):
         """
         # Fit it
         # NOTE: Tilts do not change and self.slits is updated internally.
-        self.fit(spat_illum_only=self.spat_illum_only, debug=debug)
+        if not self.flatpar['flatfield_structure']:
```
jhennawi: Please provide some inline comments about what flatfield_structure is doing to document this functionality. Maybe also point out that it is IFU specific.

pypeit/flatfield.py
```diff
@@ -487,6 +509,17 @@ def build_mask(self):
                 bpmflats[bpm] = self.slits.bitmask.turn_on(bpmflats[bpm], flag)
         return bpmflats
 
+    def build_waveimg(self):
```
jhennawi, line 516: We execute these steps in several places and are replicating that code evyrehere, i.e. extract, build_wavecalib etc. Should we have a core utility routine somewhere that executes them?

pypeit/par/pypeitpar.py
```diff
@@ -703,6 +703,17 @@ def __init__(self, method=None, pixelflat_file=None, spec_samp_fine=None,
...
                                           'where nspec is the number of spectral pixels.'
 
+        defaults['flatfield_structure'] = False
+        dtypes['flatfield_structure'] = bool
+        descr['flatfield_structure'] = 'Set this variable to True if you want to compute and ' \
+                                       'account for the detector structure in the flatfield image. ' \
+                                       'Note that ``detector structure`` refers to pixel sensitivity ' \
+                                       'variations that depend on both the spatial and spectral coordinates. ' \
+                                       'In most cases, the default 2D bspline is sufficient to account ' \
+                                       'for detector structure (i.e. set this parameter to False). Note ' \
```
jhennawi: Can you please add a note to the docs indicating that currently this functionality is only used for KCWI.

pypeit/par/pypeitpar.py
```diff
@@ -703,6 +703,17 @@ def __init__(self, method=None, pixelflat_file=None, spec_samp_fine=None,
                                           'relative weights is given by ``nspec/slit_illum_smooth_npix``, ' \
                                           'where nspec is the number of spectral pixels.'
 
+        defaults['flatfield_structure'] = False
```
jhennawi: Maybe a better name for this parameter that conveys that you are doing a 2D fit or something. The 2d bispline is not a real 2d fit. Just flatfield_structure does not seem to get across what is done differently here.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -921,4 +929,44 @@ def get_datacube_bins(self, slitlength, minmax, num_wave):
...
         return xbins, ybins, spec_bins
 
+    def flatfield_structure(self, ff_struct, gpmask):
+        r"""
+        Perform a model fit to the flatfield structure of KCWI data.
+        A few different setups were inspected (BH2 & BM with different
+        grating angles), and a very similar structure was found for all
+        setups, indicating that this structure is something to do with
+        the detector. The starting parameters and functional form are
+        assumed to be sufficient for all setups.
```
jhennawi, line 939: According to Keck folks, this is some kind of transient feature. You may want to chat with them about that (Max) and also add that to the docs here.

pypeit/spectrographs/spectrograph.py
```diff
@@ -1245,14 +1245,30 @@ def get_datacube_bins(self, slitlength, minmax, num_wave):
...
+        Args:
+            ff_struct (`numpy.ndarray`_):
+                An image of the flatfield structure.
+            gpmask (`numpy.ndarray`_):
+                Good pixel mask (True=good), the same shape as ff_struct.
+
+        Returns:
+            `numpy.ndarray`_: A model fit to the flatfield structure.
+        """
+        msgs.warn("Flatfield structure is not implemented for spectrograph: {0:s}".format(self.name))
```
jhennawi: Not sure we want a warning that will show up every time. Maybe better to print out a message if the non-default behavior is executed for KCWI.

jhennawi (changes requested):
Thanks @rcooke-ast. Very minor changes requested.

pypeit/calibrations.py
rcooke-ast: That's correct - the option is now available to correct pixel-to-pixel sensitivity variations in the illumflat.

pypeit/core/datacube.py
rcooke-ast: Good point - done

pypeit/flatfield.py
rcooke-ast: Done. Technically, this correction is not IFU specific, it's just instrument specific, and we've only implemented this for KCWI. But, I'll update the notes about this in the parset.

pypeit/par/pypeitpar.py
rcooke-ast: Done!

pypeit/spectrographs/spectrograph.py
rcooke-ast: This message is only printed if the user requests a structure correction, but it is not implemented for the spectrograph they are reducing data for, so I think this is OK here.

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: I discussed this over email with a few folks at Keck, and they mentioned that this is worse after changing the binning, and that it depends on time. I've run this code on a few different grating setups (i.e. covering different wavelength ranges), and taken years apart, and the structure appears to be more-or-less identical. So, I haven't noticed it as a transient feature. So, at least based on the tests I have done, I believe it is a multiplicative correction, and this appears to work extremely well for the 4 setups I have tested so far.

pypeit/flatfield.py
rcooke-ast: That's right - I have consolidated all of the cases that are in this `flatfield.py` but I suspect you are suggesting something more general that can be used by all pypeit code?

pypeit/par/pypeitpar.py
rcooke-ast: OK - I've updated this based on our Slack discussion.

rcooke-ast (commented):
Addresses JFH comments

rcooke-ast:
Tests still passing (I've only run KCWI on the dev suite, at the moment):
image
image

profxj (approved):
Approving

jhennawi (approved):
Thanks @rcooke-ast. Please make sure you've run all tests given the changes to the core flat fielding methods.

rcooke-ast:
Tests pass 
image

rcooke-ast:
dev suite tests pass:

image

PR #1415, opened 2022-05-26
Update lris.rst due to changes in TILSOTUA

jchowk:
Updated the text describing TILSOTUA to describe how to use the autoslit ".file3" files in place of the UCO FITS files.

Note: I removed this text (slightly edited), as I wasn't sure it applied to the steps needed to run TILSOTUA. It may be needed later in the description?

One feeds it
 #. The mask design file (either the FITS or ".file3" files).
 #. The object table that the user fed to AUTOSLIT to generate the mask.
 #. The table of targeted objects generated by AUTOSLIT (has the extension ".file1" by default).


jhennawi (changes requested):
Hi @jchowk should this PR be to develop instead of release?

jhennawi (approved):
After discussions, I'm fine with his going through.

profxj (approved):

PR #1421, opened 2022-06-02
Adding Mosfire J2

dmyang42:
small modification for reducing Mosfire/J2 data


pypeit/spectrographs/keck_mosfire.py
```diff
@@ -532,6 +532,23 @@ def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table,
             #    plt.legend()
             #    plt.show()
             return wave, counts, counts_ivar, gpm
+        elif 'J2-spectroscopy' in meta_table['DISPNAME']:
+            # TODO: how to set wave_blue and wave_red here
```
jhennawi: You can remove this TODO.

jhennawi (approved):
Minor comments. @deborape can you please run the Dev-Suite test for just J2 as @topologyYDM does not have the full dev suite running yet.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -532,6 +532,22 @@ def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table,
             #    plt.legend()
             #    plt.show()
             return wave, counts, counts_ivar, gpm
+        elif 'J2-spectroscopy' in meta_table['DISPNAME']:
```
debora-pe, line 514: This part is repetitive since the only thing that is changing here with respect to the first `if` is `wave_blue` and `wave_red`. It's fine to keep it like this for now, but maybe you want to add a TODO to simplify this later

debora-pe (approved):
Looks good. I'll run the Dev Suite for this.

pypeit/spectrographs/keck_mosfire.py
dmyang42: Hi Debora, thanks for the review! I made some simplifications

debora-pe: Great. Thanks for the quick fix.

PR #1427, opened 2022-06-10
Miscellaneous fixes

debora-pe:
- Found a bug in coadd2d. The coadd bpm wasn't fully taking into account the bpm of the single frames. Fixed that.

- The maximum shift in the spectral flexure correction was not fully constrained (even if a parameter to constrain it exist) and the value of the shift could get reducibly large (~5000 pixels). I added a fix to ensure that this will not happen.

### DIFF
--- a/pypeit/coadd2d.py
+++ b/pypeit/coadd2d.py
@@ -249,8 +249,8 @@ class CoAdd2D:
         for i in range(1, self.nexp):
             # update bpm with the info from the other frames
             slits = self.stack_dict['slits_list'][i]
-            reduce_bpm &= (slits.mask > 0) & (np.invert(slits.bitmask.flagged(slits.mask,
-                                                                             flag=slits.bitmask.exclude_for_reducing)))
+            reduce_bpm |= (slits.mask > 0) & (np.invert(slits.bitmask.flagged(slits.mask,
+                                                                              flag=slits.bitmask.exclude_for_reducing)))
         # this are the good slit index according to the bpm mask
         good_slitindx = np.where(np.logical_not(reduce_bpm))[0]
 
--- a/pypeit/core/flexure.py
+++ b/pypeit/core/flexure.py
@@ -272,15 +272,17 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
         fit = fitting.PypeItFit(xval=subpix_grid, yval=corr[subpix_grid.astype(np.int)],
                                 func='polynomial', order=np.atleast_1d(2))
         fit.fit()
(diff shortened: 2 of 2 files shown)


pypeit/coadd2d.py
```diff
@@ -249,8 +249,8 @@ def good_slitindx(self, only_slits=None):
         for i in range(1, self.nexp):
             # update bpm with the info from the other frames
             slits = self.stack_dict['slits_list'][i]
-            reduce_bpm &= (slits.mask > 0) & (np.invert(slits.bitmask.flagged(slits.mask,
-                                                                             flag=slits.bitmask.exclude_for_reducing)))
+            reduce_bpm |= (slits.mask > 0) & (np.invert(slits.bitmask.flagged(slits.mask,
```
kbwestfall, line 252: Good catch!

pypeit/core/flexure.py
```diff
@@ -289,7 +291,7 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_lines, mxshft=20):
 
     return dict(polyfit=fit, shift=shift, subpix=subpix_grid,
                 corr=corr[subpix_grid.astype(np.int)], sky_spec=obj_skyspec, arx_spec=arx_skyspec,
-                corr_cen=corr.size/2, smooth=smooth_sig_pix, success=success)
+                corr_cen=corr.size/2, smooth=smooth_sig_pix)
```
kbwestfall, line 294: How were the calling functions handling the case when `success = False`?  Do they now know how to handle if the returned value is `None`?

kbwestfall (approved):
Approving, but I had a comment on the change in flexure.py.

pypeit/core/flexure.py
debora-pe: From looking at this code, I found that the calling function `spec_flexure_slit` doesn't actually use the `success` info at all. This is why I removed `success` from the dict. It deals instead with `None`. So if the returned value is `None` the flexure correction is not performed, which (in my opinion) is much better than getting a 5000 pixels shift in the wavelength direction.

kbwestfall: Agreed!!

profxj (approved):
Approved!

debora-pe:
Just to make absolutely sure, I ran Dev Suite and unit tests. All good.


```
Testing Started at 2022-06-13T17:50:39.820993
Testing Completed at 2022-06-14T02:08:55.125660
Total Time: 8:18:15.304667
```

`=================== 328 passed in 1295.27s (0:21:35) ================`

debora-pe:
merging

PR #1431, opened 2022-06-16
Hotfix for TNG dolores - use decimal degrees

rcooke-ast:
This is untested, but it shouldn't have gone wrong  :-)


profxj (approved):
Nice catch

kbwestfall (approved):
Nice!

rcooke-ast:
Looking good to me...
image

PR #1441, opened 2022-06-30
Coadd3d inputfiles

rcooke-ast:
Tested and it works... there's a dev suite PR as well, which I hope fixes the issue there (untested).

I'll update the docs...


profxj (approved):
thx for all of this!

PR #1444, opened 2022-07-13
Keck/NIRSPEC fix

rcooke-ast:
I was reducing some low resolution Keck NIRSPEC data, and the frame typing was not working. It seems this was due to a type issue, but also, the one line statements in `spectrograph.lamps()` were throwing errors about unknown variables.. Something that must be new in python 3 🤷


kbwestfall (approved):
Nice!  And thanks for adding the unit test!

profxj (approved):
Thx for this.

Should our tests have caught this?

rcooke-ast:
Nope, the tests wouldn't have caught this previously... The dev suite doesn't test frame typing (at least for nirspec low), and the pypeit tests don't test this either. The test I've implemented should test this going forward, but there may be other instruments that use the same frame-typing code, so that still needs to be checked.

PR #1445, opened 2022-07-16
MOSFIRE calibration and combination association

debora-pe:
This addresses the last 2 major Keck requirements for MOSFIRE:

- correctly associate calibrations with science data. Often (for longslit and long2pos) calibrations are taken with masks that have names that differs from the mask names of the science frames. This PR deals with that.
- automatically turn the dithering pattern into a reduction sequence. So, now PypeIt attempts to automatically assign `comb_id` and `bkg_id`.

Documentation is included.

Most changes are related only to MOSFIRE, but one more general change is that now more than one setup can be assigned to the same calibration frame, allowing to associate the same calibration frames to different science/standard
frames, if desired.

Companion Dev-Suite PR: https://github.com/pypeit/PypeIt-development-suite/pull/214


debora-pe:
Tests pass, except a strange failure for GMOS because of a missing fits file.
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  197 passed, 1277 warnings in 154.94s (0:02:34) ---
--- PYTEST UNIT TESTS PASSED  117 passed, 1861 warnings in 835.70s (0:13:55) ---
--- PYTEST VET TESTS PASSED  24 passed, 1570 warnings in 829.57s (0:13:49) ---
--- PYPEIT DEVELOPMENT SUITE FAILED 1/149 TESTS  ---
Failed tests:
    gemini_gmos/GS_HAM_R400_700 pypeit_sensfunc
Skipped tests:
    gemini_gmos/GS_HAM_R400_700 pypeit_flux
    gemini_gmos/GS_HAM_R400_700 pypeit_coadd_1dspec
    gemini_gmos/GS_HAM_R400_700 pypeit_tellfit
Testing Started at 2022-07-16T12:33:52.301016
Testing Completed at 2022-07-16T19:41:38.257748
Total Time: 7:07:45.956732
```

kbwestfall (approved):
Looks great to me!  Thanks for all the doc updates and additions.  My only request is that you merge in the `release` and add/update the `CHANGES.rst` file.  I'm approving now though.

debora-pe:
Re-ru the tests. All good.

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  198 passed, 1295 warnings in 160.02s (0:02:40) ---
--- PYTEST UNIT TESTS PASSED  117 passed, 1906 warnings in 843.19s (0:14:03) ---
--- PYTEST VET TESTS PASSED  24 passed, 1570 warnings in 831.12s (0:13:51) ---
Testing Started at 2022-07-18T15:57:51.320787
Testing Completed at 2022-07-18T22:58:44.035477
Total Time: 7:00:52.714690
```

pypeit/core/meta.py
```diff
@@ -132,6 +132,7 @@ def define_additional_meta(nlamps=20):
                        'calpos': dict(dtype=str, comment='Position of calibration system (KCWI)'),
                        'datasec': dict(dtype=str, comment='Data section (windowing)'),
                        'dateobs': dict(dtype=str, comment='Observation date'),
+                       'decker_basename': dict(dtype=str, comment='Slit/mask/decker basename'),
```
jhennawi: Can you explain why we need both decker and decker_basename?

pypeit/core/meta.py
```diff
@@ -153,6 +154,7 @@ def define_additional_meta(nlamps=20):
                        'pressure': dict(dtype=float, comment='Pressure (units.bar) at observation time'),
                        'seq_expno': dict(dtype=int, comment='Number of exposure in observing sequence'),
                        'slitwid': dict(dtype=float, comment='Slit width, sometimes distinct from decker'),
+                       'slitlength': dict(dtype=float, comment='Slit length, used only for long slits'),
```
jhennawi, line 163: Why do we need slitlength, decker, and decker_basename? Can you explain

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -269,6 +270,10 @@ def init_meta(self):
         self.meta['idname'] = dict(card=None, compound=True)
         self.meta['frameno'] = dict(ext=0, card='FRAMENUM')
         self.meta['object'] = dict(ext=0, card='OBJECT')
+        self.meta['decker_basename'] = dict(card=None, compound=True)
```
jhennawi: It seems to me that you could be using decker instead of decker_basename here and then we don't need this additional meta variable.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -347,7 +372,214 @@ def configuration_keys(self):
             and used to constuct the :class:`~pypeit.metadata.PypeItMetaData`
             object.
         """
-        return ['decker', 'dispname', 'filter1']
+        return ['decker_basename', 'slitlength', 'slitwid', 'dispname', 'filter1']
+
+    def modify_config(self, fitstbl, cfg):
```
jhennawi, line 389: Provide some context for why this method is necessary in the docs. I presume this is something to do with reducing long2pos data with calibrations taken with a different decker. If that is the case explicitly state please.

pypeit/spectrographs/keck_mosfire.py
```diff
@@ -347,7 +372,214 @@ def configuration_keys(self):
...
+                and 'standard' not in fitstbl['frametype']:
+            cfg2 = copy.deepcopy(cfg)
+            cfg2['decker_basename'] = 'long2pos'
+            return cfg2
+        return cfg
+
+    def get_comb_group(self, fitstbl):
+        """
+
+        This method modify comb_id and bkg_id metas for a specific instrument.
```
jhennawi: Docs are too terse. Explain that this routine parses the dither pattern etc. to set the ids. It is doing more than just modifying ids.

jhennawi: Also, please look at the parse dither pattern method that is also defined for mosfire, and figure out if you can combine that codes functionality with what you are doing here, i.e. it appears you are also parsing dither patterns from header cards to set comb and bg_ids. We do not want multiple pieces of code that do similar things. Perhaps we can get rid of parse_dither_pattern.

pypeit/spectrographs/spectrograph.py
```diff
@@ -661,6 +681,25 @@ def config_independent_frames(self):
         """
         return {'bias': None, 'dark': None}
 
+    def get_comb_group(self, fitstbl):
+        """
+
```
jhennawi, line 686: See my point above about docs.

jhennawi (changes requested):
Hi @deborape thanks for this. Minor changes requested.

pypeit/core/meta.py
debora-pe: I added some comments in the code, but basically `decker_basename` is necessary for the reduction of `LONGSLIT` masks that use calibrations taken with a different  `decker`. I could define `decker` in the same way I defined `decker_basename` and get rid of `decker_basename`, but I believe it's important to propagate the information on the full name of the slitmask (`decker`), which is not preserved in `decker_basename` for `LONGSLIT` masks.

pypeit/core/meta.py
debora-pe: They are necessary for the reduction of `LONGSLIT` masks. I added some comments in the code.

pypeit/spectrographs/keck_mosfire.py
debora-pe: See my response to your previous comment about `decker` and `decker_basename`.

pypeit/spectrographs/keck_mosfire.py
debora-pe: Added.

pypeit/spectrographs/keck_mosfire.py
debora-pe: Added more docs.

debora-pe: I don't think it's possible at these stage to combine this method with `parse_dither_pattern`, without additional substantial changes. 
`parse_dither_pattern` is a little more generic in the sense that takes as input a list of files, while `get_comb_group` uses the fitstbl for the selected calibration group. 
As it is now, `parse_dither_pattern` is useful to be used also outside of the main reduction. Moreover, `parse_dither_pattern` is also defined for `vlt_fors`.

jhennawi: Hi @deborape I'm not satisfied with your response here. The docs still don't state that this routine sets the comb and bg_ids by parsing the dither pattern.  That needs to be added to the docs. Also, I'm not convinced by your argument that this routine and parse_dither_pattern cannot be merged. It is a flaw that parse_dither_pattern is written to take a list of files which is admittedly my fault.  You could easily compute the fitstable where the files are being used. I don't like that we have two routines that are parsing dither patterns in different ways. Furthermore we need the information that you figured out here for places where parse_dither_pattern is used, i.e. for quicklook and I presume also for 2d coadding where we need the offsets. It would take someone a lot of time to figure that out in the future, whereas now you have the mosfire header cards and dither patterns at your fingertips.  

This does not look like a very difficult change since I think all the information is in the fitstable. I realize a few other instruments are involved, but that is part of how we code pypeit, i.e. to not just solve one instruments but all instruments.

pypeit/core/meta.py
jhennawi: Okay fair enough. I think people are going to get confused by the naming though which is so similar to decker.  Should we think of something better than decker basename that more clearly indicates when decker_basename is going to be used? Maybe decker_calibration or decker_secondary or decker_calib_secondary and then in the comment you need to clearly spell out how this differs from decker.

jhennawi (changes requested):
Thanks @deborape. I would still like you to address the parse_dither_pattern issue. See my comments there. Happy to also discuss over Slack or at the telecom.

pypeit/spectrographs/keck_mosfire.py
debora-pe: @jhennawi you are not looking at the latest changes. I do clearly state that this method, specifically for MOSFIRE, parse the dither pattern.

I think you got a different understanding of the philosophy of this method. When I wrote it, I did not mean it to be a dither pattern parsing, but it's meant to change the `comb_id` and `bkg_id` (how it's stated in the first sentence of the docs), since those values are assigned as default in `metadata.set_combination_groups`. In a generic usage, this method can assign  `comb_id` and/or `bkg_id` following the dither pattern or not and adapt to any other logic that the instrument requires. 
This is the reason why I don't want to replace `parse_dither_pattern` with this method. I could see using `parse_dither_pattern`  inside this method, but I will have to make substantial changes to  `parse_dither_pattern` and all the quicklook scripts that use it, and this is beyond the scope of this PR.  We can add a TODO for the future.

jhennawi: Hi @deborape perhaps this is an issue with GitHub, but I'm looking here: 

https://github.com/pypeit/PypeIt/blob/bfb3b1b6475d37ee744d3e1169f4b29bee57e21d/pypeit/spectrographs/keck_mosfire.py

And I see no mention of parsing dither patterns in the documentation for get_comb_group.  This needs to be clearly stated in the get_comb_group Docstring as well as in the dummy method in the spectrograph.py. 

As for the parse_dither_pattern issue, I'm fine with making that a to do for the future.

debora-pe: @jhennawi please look here: https://github.com/pypeit/PypeIt/blob/72b28ffb472649172f55877213b379d6a6f5a9d0/pypeit/spectrographs/keck_mosfire.py#L431

I did not add it in the dummy method in the spectrograph.py for the reason I explained above. This method does not need to use the dither pattern. It does for MOSFIRE, but it may not use it for other spectrographs.

I'll add a TODO for  `parse_dither_pattern`

jhennawi: Okay thanks, it is strange that GitHub was not updated so I could see your docs. Sorry about that!

jhennawi (approved):
Thanks!

PR #1452, opened 2022-08-01
slitmask and extraction hotfixes

debora-pe:
This fixes 2 bugs found while running the reduction for the ADAP:

- (probably rare) byte decoding issue when retrieving the object name from the slitmask design.
-  crash in optimal extraction when "eligible_pixels" for extraction is equal to zero.

I'll run the tests.


kbwestfall (approved):
Looks good to me!

bpholden (approved):
looks good

PR #1458, opened 2022-08-14
IFU flexure + fine spatial illumination correction

rcooke-ast:
Implemented:
* spectral flexure + heliocentric correction for IFU reductions
* fine spatial illumination correction (QA plots for Shane+Kast shown below)... I'll post a KCWI QA plot later...
* minor fixes, typos, cleanup

Tests pass:
image

image


rcooke-ast:
KCWI QA... Looks noisy because it's a sky flat...

image

pypeit/core/flexure.py
```diff
@@ -505,6 +507,46 @@ def spec_flexure_slit(slits, slitord, slit_bpm, sky_file, method="boxcar", speco
     return flex_list
 
 
+def spec_flexure_slit_global(sciImg, waveimg, global_sky, par, slits, slitmask, trace_spat, gd_slits, pypeline, det):
+    # TODO :: Need to think about spatial flexure - is the appropriate spatial flexure already included in trace_spat via left/right slits?
```
profxj, line 789: Need a `doc` string

pypeit/flatfield.py
```diff
@@ -262,16 +306,61 @@ def fit2illumflat(self, slits, frametype='illum', initial=False, flexure_shift=N
...
                                                       slitid_img=_slitid_img,
-                                                      flexure_shift=flexure_shift)
+                                                      flexure_shift=spat_flexure)
             illumflat[onslit] = spat_bsplines[slit_idx].value(spat_coo[onslit])[0]
         # TODO -- Update the internal one?  Or remove it altogether??
         return illumflat
 
+    def fit2illumflat_fine(self, slits, tilts=None, initial=False, spat_flexure=None, frametype='illum'):
+        """
+        Main algorithm to build the wavelength image
```
profxj: wavelength image?

profxj (approved):
Very good!

pypeit/core/flexure.py
rcooke-ast: Thanks - added!

pypeit/flatfield.py
rcooke-ast: Oops - thanks for spotting that - fixed!

rcooke-ast:
Tests still pass after merging in the conflicting PR #1465 (also, reduction of kast+kcwi both checked)

image

debora-pe:
@rcooke-ast looks good to me!

pypeit/core/datacube.py
```diff
@@ -795,6 +795,8 @@ def make_whitelight_frompixels(all_ra, all_dec, all_wave, all_sci, all_wghts, al
             Number of RA spaxels in the reference white light image
         numdec (int, optional):
             Number of DEC spaxels in the reference white light image
+        trim (int, optional):
```
jhennawi, line 798: Please try to indicate shapes of arrays in the docs for both inputs and outputs.

pypeit/core/flexure.py
```diff
@@ -749,6 +751,80 @@ def spec_flexure_slit(slits, slitord, slit_bpm, sky_file, method="boxcar", speco
     return flex_list
 
 
+def spec_flexure_slit_global(sciImg, waveimg, global_sky, par, slits, slitmask, trace_spat, gd_slits, wv_calib, pypeline, det):
+    """Calculate the spectral flexure for every slit
```
jhennawi, line 755: How does this modification interface with @debora-pe's PR?

pypeit/extraction.py
```diff
@@ -317,8 +313,12 @@ def extract(self, global_sky, model_noise=None, spat_pix=None):
         # Return
         return self.skymodel, self.objmodel, self.ivarmodel, self.outmask, self.sobjs
 
-    def prepare_extraction(self):
+    def prepare_extraction(self, global_sky):
```
jhennawi: This method is going to disappear in my JWST branch I'm a bit worried about conflicts there that I cannot easily resolve. Let's discuss how to proceed here. This PR precedes that one so should go forward, but maybe you can think about how you are going to restructure what you are doing here in that context.

pypeit/images/rawimage.py
```diff
@@ -806,7 +806,9 @@ def flatfield(self, flatimages, slits=None, force=False, debug=False):
         # Generate the illumination flat, as needed
         illum_flat = 1.0
         if self.par['use_illumflat']:
-            illum_flat = flatimages.fit2illumflat(slits, flexure_shift=self.spat_flexure_shift)
+            illum_flat = flatimages.fit2illumflat(slits, spat_flexure=self.spat_flexure_shift)
+            # TODO :: We don't have tilts here yet... might be ever so slightly better, especially on very tilted slits
```
jhennawi, line 809: Yes I agree with this, we could change the calibration order to have tilts by this point.

pypeit/pypeit.py
```diff
@@ -858,19 +858,26 @@ def extract_one(self, frames, det, sciImg, objFind, initial_sky, sobjs_obj):
             basename=self.basename)
 
         if not self.par['reduce']['extraction']['skip_extraction']:
-            skymodel, objmodel, ivarmodel, outmask, sobjs, waveImg, \
-                tilts = self.exTract.run(final_global_sky, ra=self.fitstbl["ra"][frames[0]],
-                                         dec=self.fitstbl["dec"][frames[0]], obstime=self.obstime)
+            # Perform the extraction
+            self.exTract.run(final_global_sky)
```
jhennawi: Please don't change this. I prefer that run return values rather than just assigning form the class.

jhennawi: Let's discuss @rcooke-ast. I think a control flow where there are return values is much more clear. My suggestion is that if you run without actually extracting that you returning None.

pypeit/pypeit.py
```diff
@@ -858,19 +858,26 @@ def extract_one(self, frames, det, sciImg, objFind, initial_sky, sobjs_obj):
...
+            sobjs = self.exTract.sobjs
+            # Apply a reference frame correction to each object and the waveimg
+            self.exTract.refframe_correct(self.fitstbl["ra"][frames[0]], self.fitstbl["dec"][frames[0]], self.obstime,
+                                          sobjs=self.exTract.sobjs)
         else:
-            # Although exrtaction is not performed, still need to prepare some masks and the tilts
-            self.exTract.prepare_extraction()
+            # Although extraction is not performed, still need to prepare some masks and the tilts, and apply heliocentric correction
+            self.exTract.prepare_extraction(final_global_sky)
+            self.exTract.refframe_correct(self.fitstbl["ra"][frames[0]], self.fitstbl["dec"][frames[0]], self.obstime)
```
jhennawi, line 869: Prepare extraction is going away. The init from extraction will do what that did. Let's think about how to merge there.

jhennawi, line 869: This reframe correct should not modify things in place. Let's get into the habit of returning things from functions so as to not obfuscate control flow and manage scope.

pypeit/flatfield.py
```diff
@@ -262,16 +306,61 @@ def fit2illumflat(self, slits, frametype='illum', initial=False, flexure_shift=N
...
+        flat_finecorr = np.ones(self.shape())
+        # Get the slit properties
+        _slitid_img = slits.slit_img(flexure=spat_flexure, initial=initial)
+        left, right, _ = slits.select_edges(initial=initial, flexure=spat_flexure)
+        # Loop through all slits
+        for slit_idx, slit_spat in enumerate(slits.spat_id):
+            onslit_tweak = _slitid_img == slit_spat
+            ww = np.where(onslit_tweak)
+            cut = (ww[0], ww[1])
+            xpos = (cut[1] - left[cut[0],slit_idx]) / (right[cut[0],slit_idx] - left[cut[0],slit_idx])
```
jhennawi: It seems some state variable should be set that indicates whether tilts were used in the fit or not. In other words, I can imagine a situation where tilts were not used in the fit, but then accidentally passed in and then the code is going to use the tilts even though the fit was not done with the fits.

pypeit/flatfield.py
```diff
@@ -996,12 +1098,18 @@ def fit(self, spat_illum_only=False, debug=False):
                 ax.set_title('Illumination Function Fit for slit={:d}'.format(slit_spat))
                 plt.show()
 
+            # Perform a fine correction to the spatial illumination profile
```
jhennawi, line 1101: Thanks for the great QA.

jhennawi (changes requested):
Thanks @rcooke-ast. Very minor comments and change requests. Perhaps the most significant question is how the pypeit_chk_flats script now works in the context of this new illumination flat fit. I presume that script calls the show method of the class and so it will just show the right flat?

jhennawi (changes requested):
Thanks @rcooke-ast. Very minor comments and change requests. Perhaps the most significant question is how the pypeit_chk_flats script now works in the context of this new illumination flat fit. I presume that script calls the show method of the class and so it will just show the right flat?

pypeit/core/datacube.py
rcooke-ast: I think the only update was to `all_ivar` here... everything else is covered already.

pypeit/core/flexure.py
rcooke-ast: It's the exact same code as in `extraction.py`. I've just moved this code into a new function in `flexure.py` so that: (1) The XSpectrum1D dependency is contained within `flexure.py`; and (2) so that I can call this same code from `find_objects.py`.

pypeit/pypeit.py
rcooke-ast: I tend to agree with you... However, I'm not changing anything about how `refframe_correct` operates, I'm merely putting it in a different place in the code. Previously it was "hidden" in `extraction.py`. I personally still think my changes are an improvement... previously we were passing in a bunch of variables into `self.extract.run()` and most of those variables are only used for the reframe correction. So, separating that makes more sense to me.

So, I've left everything as is.

pypeit/pypeit.py
rcooke-ast: This seems to be a bigger issue than just this PR... PypeIt is largely based on using classes as "containers", and that is no different here. I tend to disagree with returning a bunch of objects within the class. I think the whole point of having `self.exTract` is so that these variables are contained and not floating around (i.e. no idea of where they came from). In this way, I've made it clear in this code (`pypeit.py`) where all variables come from. The control flow is no more complicated than it was before, imho. But, perhaps this is an issue that should be discussed more generally.

Two things I think we agree with are:
(1) passing a class into a function, and you have no idea if the class variables have been updated or not.
(2) irreversibly updating a variable within a class with a call to one of its functions. I suspect this happens often in PypeIt, and we should avoid that, but I haven't made any changes in this PR that affects it.

pypeit/pypeit.py
rcooke-ast: OK - I'll leave it in for now, and we can discuss how to merge it later.

pypeit/extraction.py
rcooke-ast: I agree - we should move this to the init (or get_instance) method. It would avoid us passing around the global_sky, and incorrectly assigning it in the self.run method.

pypeit/flatfield.py
rcooke-ast: I'm going to remove the tilts dependency for now, and will do a separate PR and try to include this in the zeroth correction to the spatial illumination profile, too.

rcooke-ast:
> Thanks @rcooke-ast. Very minor comments and change requests. Perhaps the most significant question is how the pypeit_chk_flats script now works in the context of this new illumination flat fit. I presume that script calls the show method of the class and so it will just show the right flat?

The script still works the same as before. It displays the fine correction to the spatial illumination profile as a separate image to the original spatial illumination profile.

rcooke-ast (commented):
Thanks @jhennawi - comments addressed!

rcooke-ast:
I've removed the tilts dependency (to be reintroduced in a later PR). I've checked that Shane/Kast data reduce, and the output looks virtually unchanged from the previous version:
image

rcooke-ast:
I've also checked that the tests pass. I'll run the dev suite when all final comments are in.
image

pypeit/pypeit.py
jhennawi: @rcooke-ast I agree that we need to discuss this in general in pypeit but it needs to be clear what the run method returns. For that reason I'm asking that you not modify run to not return variables. After you've changed that back, I'll approve. If you are running self.run differently in the IFU child such that run does not return anything then: 1)  I'd suggest maybe splitting up what run does into multiple methods and only run the sub-method that you need for IFU, 2) returning None in places where you don't need what run returns. I don't have the top-level view of how IFU using this class, but I'm very much against the object oriented approach we've adopted everywhere in PypeIt and trying to retain some semblance of functional programming in places is very important to me, which is why I designed many of these methods to (somewhat redudnantly) return variables so it is clear what the control flow is.

rcooke-ast:
At long last... all tests pass

image

jhennawi (approved):

rcooke-ast:
Docs included - merging!

PR #1464, opened 2022-09-10
updated vcorr

rcooke-ast:
Addresses #1462 - should this hotfix be directed to release?

### DIFF
--- a/pypeit/telescopes.py
+++ b/pypeit/telescopes.py
@@ -1,27 +1,33 @@
 # encoding: utf-8
 """
 Define the telescopes parameters used by Pypit.
+
+NOTE: Longitudes are measured increasing to the east, so west longitudes are negative.
 """
 from pypeit.par.pypeitpar import TelescopePar
+from astropy.coordinates import EarthLocation
+from astropy import units
 
 #TODO: Remove 'Par' from class name?
 
 class GTCTelescopePar(TelescopePar):
     def __init__(self):
+        loc = EarthLocation.of_site('Roque de los Muchachos')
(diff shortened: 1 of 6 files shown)


profxj (approved):
Thanks, @rcooke-ast !

Looks like a unit test may still be failing...
But I am approving

pypeit/telescopes.py
```diff
@@ -1,171 +1,193 @@
...
+                                              longitude=loc.lon.to(units.deg).value,
+                                              latitude=loc.lat.to(units.deg).value,
+                                              elevation=loc.height.to(units.m).value,
                                               diameter=4.2)
                                 
 class APFTelescopePar(TelescopePar):
     def __init__(self):
         super(APFTelescopePar, self).__init__(name='APF',
-                                              longitude=121.642778,
+                                              longitude=-121.642778,
```
kbwestfall: Can we use `loc = EarthLocation.of_site('Lick Observatory')`?

kbwestfall (approved):
Nice!  Thanks for the fix.  I checked and there is a "Lick Observatory" site.  Did you try that for APF?

Also, I think it would be fine if this goes to release, but does it carry with it many other changes from previous merges to `develop`?  If so, we should probably keep it as going to develop.

pypeit/telescopes.py
rcooke-ast: Given that it's predominantly used by planet hunters (I presume), I didn't dare change it from the default value... I was being overly cautious, perhaps... Happy to update this...

rcooke-ast:
OK - I've checked that this at least reduces kast data  :-)

Merging...

PR #1466, opened 2022-09-13
Fix location of LDT

tbowers7:
AstroPy's listing for `EarthLocation.of_site('Lowell Observatory')` is our Anderson Mesa station, some 30 miles distant from the LDT.  A PR to AstroPy was submitted some time ago to change this, but...

The proper site in the AstroPy database is 'Discovery Channel Telescope'.  For those wondering, DCT --> Lowell Discovery Telescope name change happened in early 2020.

	modified:   pypeit/telescopes.py

### DIFF
--- a/pypeit/telescopes.py
+++ b/pypeit/telescopes.py
@@ -184,7 +184,7 @@ class BokTelescopePar(TelescopePar):
 
 class LDTTelescopePar(TelescopePar):
     def __init__(self):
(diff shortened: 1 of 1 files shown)


rcooke-ast (approved):
Good catch!!  Thanks for fixing this  :-)

kbwestfall (approved):
Nice catch!

PR #1480, opened 2022-10-08
JWST sky regions updates

rcooke-ast:
Requested updates to the skyregions and removal of `caliBrate` from `find_objects()`

I have checked that this still reduces IFU data. Some work will be needed to merge this PR with #1458


pypeit/find_objects.py
```diff
@@ -88,7 +87,7 @@ class FindObjects:
 
     # Superclass factory method generates the subclass instance
     @classmethod
```
jhennawi, line 89: I don't think wv_calib should be a required argument here since only the IFU child needs it. Instead I think it should be optional, and well documented that it is only needed for the IFU.  Furthermore, I think it is cleaner and better practice to require waveimg rather than wv_calib. All you are doing in the IFU child is generating waveimg from wv_calib. Better to do that outside of this class since that gives on the flexibility to use this class in situations where a PypeIt wv_calib object was not generated (i.e this is the case for JWST and for 2d coadds). So my request is: 
1) make it optional, 2) the variable should be waveimg, not wv_calib.  I would be fine with letting either one be passed in or something along those lines. 

The general theme is that it is important that these base classes do not presume that a full PypeIt calibration run has been performed, which is why I removed the dependencies on calibrate.

pypeit/extraction.py
```diff
@@ -66,7 +66,7 @@ class Extract:
 
     # Superclass factory method generates the subclass instance
     @classmethod
-    def get_instance(cls, sciImg, slits, sobjs_obj, spectrograph, par, objtype,
+    def get_instance(cls, sciImg, slits, sobjs_obj, spectrograph, par, objtype, global_sky=None,
```
jhennawi, line 69: Good call to make this an init parameter. See further discussion in the #extract channel in PypeIt slack.

jhennawi (changes requested):
This looks great. Minor change requested on the required objects for find_objects.

pypeit/find_objects.py
rcooke-ast: Fair enough - I've put wv_calib as an optional argument for now, and we can think how best to incorporate the joint skysub for IFU data later. One way is to add a `joint_skysub` function to all children of FindObjects, and pass in wv_calib directly to this function. For multislit and echelle, this would just return the input global_sky, but for the IFU child, it would perform a joint skysub.

jhennawi (approved):

Issue #1479, opened 2022-10-07
In matplotlib >3.3, set_window_title method is depreciated, and causes an error when using skysub gui


prappleizer:
With matplotlib version >3.3, the `set_window_title` method has been depreciated. The latest version of pypeit installed from git still uses this method with the manual sky definition gui, throwing an error on launch: 

```
'FigureCanvasQTAgg' object has no attribute 'set_window_title'
```

I was able to resolve the problem by downgrading to matplotlib 3.3, but in the future this should be fixed as we are past matplotlib 3.6 at this point. 

From the perspective of the codebase, I found one piece of advice: A quick fix is to replace `fig.canvas.set_window_title` with `fig.canvas.manager.set_window_title`, but there may be more elegant ways to get around this as well.

rcooke-ast:
Addressed by PR #1482 - Thanks for submitting the issue @prappleizer!

PR #1482, opened 2022-10-11
HOTFIX: skysub gui

rcooke-ast:
Addresses Issue #1479

### DIFF
--- a/pypeit/core/gui/skysub_regions.py
+++ b/pypeit/core/gui/skysub_regions.py
@@ -217,6 +217,10 @@ class SkySubGUI(object):
 
         return srgui
 
+    def finalize(self):
+        plt.rcdefaults()
+        plt.close()
+
     def region_help(self):
         print("You can enter the regions in the text box, as a comma separated")
         print("list of percentages. For example, typing  :10,35:65,80:  in the")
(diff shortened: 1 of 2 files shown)


profxj (approved):
Am assuming this was tested. :)

rcooke-ast:
Yes, tested... but only on matplotlib 3.5... (not 3.6) - the solution implemented is the common approach used as described online  🤞

kbwestfall (approved):
Thanks for addressing the issue!  I'm approving and happy for you to merge, but two thoughts:
- Should this go to `release` instead of `develop`?  Maybe it's not worth it because it likely means you'd have to redo the edits in a new branch that branches from `release` instead of `develop` (to avoid including all the other edits already in `develop`).
- I recently had to manipulate the `keymap` items in `rcParams`.  Is there any reason not to turn them all off?  I did:
```
for key in plt.rcParams.keys():
    if 'keymap' in key:
        plt.rcParams[key] = []
```
Also, I'm not sure it matters, but you might want to also add `plt.rcdefaults()` at the end to make sure the defaults are restored.

rcooke-ast:
Good points @kbwestfall - I have added a call to `finalize` that resets to the matplotlib defaults. I've also used your code to unset all keymaps, and then just initialise the ones that matter to me

I don't think this script is used all that often, so pushing to release is probably not critical, but I'm happy to cherry pick this commit there - shouldn't be any other conflicts. There's also a workaround that users can specify `user_regions`, which is less "clicky" but still available...

PR #1483, opened 2022-10-11
Datacube skyframe

rcooke-ast:
This PR allows the user to specify sky frames (i.e. off target) to be used for sky subtraction and the relative spectral illumination correction.


jhennawi (approved):
Approved pending tests.

rcooke-ast:
Thanks @jhennawi!  Some final doc updates and tests pass:

image

pypeit/core/datacube.py
```diff
@@ -1239,6 +1239,26 @@ def coadd_cube(files, opts, spectrograph=None, parset=None, overwrite=False):
...
+    if cubepar['skysub_frame'] in [None, 'none', '', 'None']:
+        skysub_default = "none"
+        skysubImgDef = np.array([0.0])  # Do not perform sky subtraction
+    elif cubepar['skysub_frame'].lower() == "image":
+        skysub_default = "image"
+    else:
+        msgs.info("Loading default image for sky subtraction:" +
+                  msgs.newline() + cubepar['skysub_frame'])
+        try:
+            spec2DObj = spec2dobj.Spec2DObj.from_file(cubepar['skysub_frame'], detname)
```
kbwestfall, line 1255: Can you have the try except block only around this line?  I.e., errors in the other lines (insofar as that's possible) will cause the exception block to execute, which could be unrelated to whether or not the file was successfully opened.  Or do you need the last line to also be part of the try/except?

pypeit/core/datacube.py
```diff
@@ -1270,11 +1290,47 @@ def coadd_cube(files, opts, spectrograph=None, parset=None, overwrite=False):
...
+                skysubImg = spec2DObj.skymodel
+                this_skysub = "image"  # Use the current spec2d for sky subtraction
+            elif opts['skysub_frame'][ff].lower() == 'none':
+                skysubImg = np.array([0.0])
+                this_skysub = "none"  # Don't do sky subtraction
+            else:
+                try:
+                    # Load a user specified frame for sky subtraction
+                    msgs.info("Loading skysub frame:"+msgs.newline()+opts['skysub_frame'][ff])
+                    spec2DObj_sky = spec2dobj.Spec2DObj.from_file(opts['skysub_frame'][ff], detname)
```
kbwestfall, line 1318: Same comment here about limiting how many lines are in the try block.

pypeit/core/datacube.py
```diff
@@ -1330,17 +1386,18 @@ def coadd_cube(files, opts, spectrograph=None, parset=None, overwrite=False):
         if dspat is None:
             dspat = max(pxscl, slscl)
         if pxscl > dspat:
-            msgs.warn("Spatial scale requested ({0:f}'') is less than the pixel scale ({1:f}'')".format(dspat,pxscl))
-        if pxscl > dspat:
-            msgs.warn("Spatial scale requested ({0:f}'') is less than the slicer scale ({1:f}'')".format(dspat, slscl))
+            msgs.warn("Spatial scale requested ({0:f}'') is less than the pixel scale ({1:f}'')".format(3600.0*dspat, 3600.0*pxscl))
+        if slscl > dspat:
+            msgs.warn("Spatial scale requested ({0:f}'') is less than the slicer scale ({1:f}'')".format(3600.0*dspat, 3600.0*slscl))
```
kbwestfall: Add units to these warnings?

kbwestfall (approved):
Looks good.  I merged in `develop` and `release`, and I made a couple of direct doc changes.  My only substantive comment was to check if you can limit the number of lines in a couple of try blocks.  The more lines in a given block can make it harder to debug and lead to error messages that aren't related to what's actually causing the problem.  Regardless, I'm approving now.

pypeit/core/datacube.py
rcooke-ast: Good point - fixed!

PR #1487, opened 2022-10-13
JWST docs update + sky regions

rcooke-ast:
Updates to the load_skyregions including doc updates


jhennawi (approved):
Thanks @rcooke-ast looks great!

Issue #1463, opened 2022-09-07
Coadd2D HOWTO example uses the old / deprecated file format


tomgreene:
The Coadd2D HOWTO at https://pypeit.readthedocs.io/en/latest/coadd2d_howto.html gives an example of making a Coadd2D parameter file in the old / deprecated format. It does not contain the 'filename' or 'path' keywords.
It also sets the snr threshold with the work keyword (should be snr_thresh in current pypeit v 1.10).

The main docs on Coadd2D at https://pypeit.readthedocs.io/en/latest/coadd2d.html#coadd2d-file do give the new / correct format.

debora-pe:
Addressed in #1490
Closing this.

PR #1490, opened 2022-10-19
HOWTO DOCS

debora-pe:
1) Update the Coadd2D HOWTO docs (address issue #1463 )
2) Added MOSFIRE HOWTO docs (for the workshop and addressing [PM-27](https://keckobservatory.atlassian.net/browse/PM-27) in JIRA)

Also changed a MOSFIRE pypeit file in the Dev Suite to be consistent with the example in  MOSFIRE HOWTO. Please review this too: https://github.com/pypeit/PypeIt-development-suite/pull/224


doc/mosfire_howto.rst
```diff
@@ -0,0 +1,548 @@
+.. include:: include/links.rst
+
+.. _mosfire_howto:
+
+==================
+Keck-MOSIFRE HOWTO
```
kbwestfall: MOSFIRE mispelled

doc/mosfire_howto.rst
```diff
@@ -0,0 +1,548 @@
...
+    m121128_0107.fits |              lampoffflats |         7.8 |        45.0 |       unknown | K-spectroscopy | ic348_TK_M03A |     1,1 | 56259.14262084 | 1.41291034 |  14.5479 |       K |        off |    none |    none |     0.0 |     107 |     0 |      -1 |     -1
+    m121128_0112.fits | pixelflat,illumflat,trace |         7.8 |        45.0 |       unknown | K-spectroscopy | ic348_TK_M03A |     1,1 | 56259.14425684 | 1.41291034 |  14.5479 |       K |         on |    none |    none |     0.0 |     112 |     0 |      -1 |     -1
+    m121128_0113.fits | pixelflat,illumflat,trace |         7.8 |        45.0 |       unknown | K-spectroscopy | ic348_TK_M03A |     1,1 | 56259.14450569 | 1.41291034 |  14.5479 |       K |         on |    none |    none |     0.0 |     113 |     0 |      -1 |     -1
+    m121128_0114.fits | pixelflat,illumflat,trace |         7.8 |        45.0 |       unknown | K-spectroscopy | ic348_TK_M03A |     1,1 | 56259.14479678 | 1.41291034 |  14.5479 |       K |         on |    none |    none |     0.0 |     114 |     0 |      -1 |     -1
+    data end
```
kbwestfall, line 122: Is this exactly what `pypeit_setup` produces without any hand editing?  If so, it might be useful for the docs to generate this automatically so that any future code changes are reflected.  See the `shane_kast` example in `pypeit/doc/scripts/make_example_files.py`.

... Although it may not be worth it given that you show the hand-edited version below.

doc/mosfire_howto.rst
```diff
@@ -0,0 +1,548 @@
...
+
+Other possible edits to the :ref:`pypeit_file:Data Block` are related to the ``comb_id`` and ``bkg_id``
+columns, which instruct PypeIt on the desired frame combination and background subtraction.
+For Keck/MOSFIRE data, PypeIt tries to automatically set the ``comb_id`` and ``bkg_id`` using the dither
+information (reported here in the ``dithpat``, ``dithpos``, and ``dithoff`` columns) recorded in the header
+of the science frames (see :ref:`mosfire_combid_bkgid`); however,
+the user can edit these columns according to the preferred reduction (see :ref:`a-b_differencing` and
+:ref:`2d_combine` for more info).
+
+Finally, in this example, we also edit the :ref:`pypeit_file:Parameter Block` adding the following lines::
```
kbwestfall: I recommend using
```
... the following lines:

.. code-block:: ini

...
```
to get the syntax highlighting.

kbwestfall (changes requested):
Nice!  Just a few comments from me.  I suggest we merge this *before* #1488 , then I can deal with any conflicts after in that PR afterwards.

I pulled down the branch and compiled the docs and note that there are a few warnings to be fixed.  I can help with that, if needed.  For debugging the docs, note the `make htmlonly` option, which builds the docs without running all the scripts that generate the automated tables that get included.  It still takes a while, but this helps cut down on the compile time when you're just making simple edits to the main `rst` files.

profxj (approved):
Nothing for me to add here.

I'd only consider our asking a WMKO staff member to
give the HOWTO a try and see if they can make it work.
e.g. Josh W.

doc/mosfire_howto.rst
debora-pe: THANK YOU!

doc/mosfire_howto.rst
debora-pe: Thank you for the suggestion. Yeah, because of the hand-edited version below I think (at least for now) I prefer to leave it like this, so that the two would be consistent.

doc/mosfire_howto.rst
debora-pe: Done. Thank you! 
I have a question though. I got the colors a bit off when I compile it on my machine (one line is green and one red, see below). Do you get the same thing? Am I doing something wrong?

Screen Shot 2022-10-19 at 2 46 28 PM

debora-pe (commented):
> Nice! Just a few comments from me. I suggest we merge this _before_ #1488 , then I can deal with any conflicts after in that PR afterwards.
> 
> I pulled down the branch and compiled the docs and note that there are a few warnings to be fixed. I can help with that, if needed. For debugging the docs, note the `make htmlonly` option, which builds the docs without running all the scripts that generate the automated tables that get included. It still takes a while, but this helps cut down on the compile time when you're just making simple edits to the main `rst` files.

Thanks for the review @kbwestfall . Yes, I'm OK with merging it before #1488.
I fixed a warning that I haven't noticed before, but the two other warning I get are:
```
/Users/dpelliccia/PypeIt/PypeIt/doc/fluxing.rst:366: WARNING: Unknown target name: "sensitivity function".
/Users/dpelliccia/PypeIt/PypeIt/doc/mosfire_howto.rst:535: WARNING: Unknown target name: "linetools".
```

Do you get the same ones? 
I decided not to resolve these warnings because they are actually fixed in your PR on documentation, so I thought it would have been easier to leave them. But let me know if you prefer me to deal with them.

doc/mosfire_howto.rst
kbwestfall: Sounds good.

doc/mosfire_howto.rst
kbwestfall: Interesting.  Maybe try upgrading sphinx?  It might also be the indentation?  Sphinx can sometimes be super picky that you're, e.g., using 4 spaces for each tab.  If you display `_build/html/pypeit_par.html#instrument-specific-default-configuration`, does that have all the variable names as green and the values as red?

kbwestfall:
> I fixed a warning that I haven't noticed before, but the two other warning I get are:
> 
> ```
> /Users/dpelliccia/PypeIt/PypeIt/doc/fluxing.rst:366: WARNING: Unknown target name: "sensitivity function".
> /Users/dpelliccia/PypeIt/PypeIt/doc/mosfire_howto.rst:535: WARNING: Unknown target name: "linetools".
> ```
> 
> Do you get the same ones? I decided not to resolve these warnings because they are actually fixed in your PR on documentation, so I thought it would have been easier to leave them. But let me know if you prefer me to deal with them.

I think the first one is dealt with in many PRs :)

For the second one, I think this should be fixed if you add:

```
.. include:: include/links.rst
```

at the top of the file.

kbwestfall (approved):
Approving!

kbwestfall:
> For the second one, I think this should be fixed if you add:
> 
> ```
> .. include:: include/links.rst
> ```
> 
> at the top of the file.

Oops, sorry!  It's already there.  Yes, this should be fixed by my PR.  Thanks!

doc/mosfire_howto.rst
debora-pe: I think you are right. It's a rendering problem in my machine. I get the same problem with colors in other docs that I did not edit and that look OK in the PypeIt webpage. I'll try re-installing Sphinx. Thanks!

debora-pe:
Merging!

PR #1505, opened 2022-11-10
GTC/OSIRIS quicklook

rcooke-ast:
The `pypeit_ql_mos` script doesn't work for spectrographs that require multiple different arc lamp frames to build the master arc for wavelength calibration. A simple additional child class allows for an easy change to the reduction steps. I also window down the detector here, since most observations with GTC/OSIRIS focus on this very small window range for the targets.

On my laptop, once the calibrations are pre-prepared, the reduction of a single science frame is ~6s... pretty amazing...

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,7 @@
 - Allow separate sky frame to be used for sky subtraction with IFU data
 - Limit the images written to the MasterEdges file to only the trace
   image, mask, and detector.
+- Added quicklook reductions for single objects observed with GTC/OSIRIS
 
 1.11.0 (21 Oct 2022)
 --------------------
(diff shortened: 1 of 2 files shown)


jhennawi (approved):
Looks good.

kbwestfall (commented):
I think this breaks our concept for the Spectrograph classes.  I'm not familiar enough with the QL code to have a better suggestion, but could we instead have functions within the main `GTCOSIRISSpectrograph` class that are specific to providing the reduction and detector parameters specific to quicklook?

rcooke-ast:
I agree @kbwestfall - this is more of a hack than a solution... I know the QL code is going to be updated soon. The key point here is to window down the detector.

Viewed in a different way, I wonder if another solution might be to see this as a windowed detector... I suppose we should wait for the QL updates to occur, and then see how best to implement it..

PR #1507, opened 2022-11-11
Update installing.rst

stinyanont:
The telluric file name that works should be TelFit_MaunaKea_3100_26100_R20000.fits and not TelFit_MaunaKea_3100_26100_R200000.fits


jhennawi (approved):
Looks good. Thanks for fixing this!

kbwestfall (approved):
Thanks for the fix!

PR #1515, opened 2022-12-03
Optimize bspline C code

tepickering:
as i was reducing a bunch of binospec data on my snazzy new mac studio, i became irked that it was taking so long and only using one of the machine's many cores for the vast majority of the time. out of curiosity, i ran `run_pypeit` with `cProfile` loaded to see where it was spending the time. the answer was that 60% was being spent in the `solution_arrays()` function in `pypeit/bspline/src/bspline.c`. all that's really happening in there is populating some large arrays via nested `for` loops. 

this PR cleans up that C code and optimizes it in the following ways:

- refactor the array population loops and do away with the convenience arrays. this improves memory utilization noticeably when dealing with large images like binospec's. the refactor also removes the need to do `sqrt()` calls. 
- move the transposing and fortran-formatting of the `action` array up into `pypeit/bspline/utilc.py` and use `np.transpose()` and `np.asfortranarray()` to do it. unsure if this is actually faster, but it is cleaner and more explicit about what's happening. also use `np.zeros` to initialize `alpha` and `beta` rather than doing so in a C loop.
- add `#pragma` lines to parallelize the loop in `solution_arrays()` if the compiler supports OpenMP. this is a huge benefit for machines with many cores/threads. 

the catch is that Apple's `clang` compiler doesn't support OpenMP. GCC, on the other hand, has supported it fully since version 9 and at least partially since 4.2. i added a check to `pypeit/bspline/setup_package.py` to see if the compiler being used supports it and add the appropriate compiler flag if it does. if it doesn't, the code will still compile and work, albeit in single-threaded mode (i.e., the `#pragma` statements just get ignored). i added a check to `pypeit_c_enabled` to report if OpenMP is available or not and added a blurb in the docs about OpenMP and these caveats. 

without these changes, it took `run_pypeit` 2h15m to run through the binospec data i've been testing on. with these changes it now completes in 1h26m with an average CPU utilization of 4.27 cores. the `solution_arrays()` function is now using 40% of the total runtime. this is significant progress, but i do think there is more to be made. for example, the `np.asfortranarray()` step is using about 5% if the total runtime.


doc/installing.rst
```diff
@@ -393,6 +393,22 @@ Some notes if you have problems installing the C code:
...
+or ``macports`` and specify its use when installing ``pypeit``. For example, if you installed
+GCC 12.x via ``homebrew``, you would get ``pypeit`` to use it by doing, for example:
+
+.. code-block:: console
+
+    CC=gcc-12 pip install "pypeit[pyqt5]"
+
+Basically, ``pypeit`` checks the ``CC`` environment variable for what compiler to use so configure
+that as needed to use your desired compiler. The ``pypeit_c_enabled`` script can be used to check if
+your compiler has OpenMP support.
```
kbwestfall, line 410: Nice.  Thanks for including this explanation, and I'm happy with this solution.  But, out of curiosity, do you know if there's a way to do this within the `setup.cfg` options?  I.e., could we enable something like `pip install "pypeit[pyqt5,openmp]"` and have that be the equivalent of the above?

pypeit/bspline/setup_package.py
```diff
@@ -1,16 +1,63 @@
...
+
+    os.chdir(curdir)
+    # clean up test code
+    shutil.rmtree(tmpdir)
+
+    # return code from compiler process is 0 if it completed successfully
+    if result == 0:
+        return True
+    else:
+        return False
```
kbwestfall, line 47: `return result == 0`

pypeit/bspline/utilc.py
```diff
@@ -164,17 +164,20 @@ def solution_arrays(nn, npoly, nord, ydata, action, ivar, upper, lower):
...
-    alpha = np.empty((bw, nfull+bw), dtype=float)
-    beta = np.empty((nfull+bw,), dtype=float)
+
+    alpha = np.zeros((bw, nfull+bw), dtype=float)
+    beta = np.zeros((nfull+bw,), dtype=float)
     upper = np.array(upper, dtype=np.int64)
     lower = np.array(lower, dtype=np.int64)
+
+    # need to convert action to fortran-style order and contiguous layout.
+    action_t = np.asfortranarray(np.transpose(action))
```
kbwestfall, line 174: Is this necessary?  It's difficult to dig through the code, but I think the `action` matrix is allocated in column-major order in the top-level calling functions, like `core.fitting.bspline_profile`.

kbwestfall (commented):
Awesome!  Thanks for doing this, Tim.  I have one question on the memory order for the `action` matrix, but am ready to approve.

pypeit/bspline/setup_package.py
tepickering: oh yeah, duh. that's a lot cleaner...

doc/installing.rst
tepickering: maybe? openmp isn't a python thing, though. it's compiler level. testing for it in a cross-platform way requires running a compiler so i don't see how to easily fold that into `setup.cfg`. it could go in `setup.py`, though. that's basically how the now-deprecated `astropy-helpers` package did it.

pypeit/bspline/utilc.py
tepickering: well, the original C code definitely did a transpose of the action matrix and `asfortranarray` is definitely required for tests to pass. in `bspline_profile`, order is set to `'F'` which sets up the memory format to be fortran-contiguous, but doesn't change the numpy indexing. `transpose` also creates a new array and doesn't provide the `order` option. so that's why `asfortranarray` is required. i'll have a deeper look/think and see if there's some way to do this more cleanly.

kbwestfall (approved):

profxj (approved):
Thanks for this!

Please merge after DevSuite tests pass

doc/installing.rst
tepickering: it looks like `conda-forge` uses `clang` as the compiler for the `osx` platform. don't see an obvious way to specify `gcc`, but will look a little deeper...

kbwestfall:
Tests pass:
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  195 passed, 10 skipped, 709 warnings in 203.54s (0:03:23) ---
--- PYTEST UNIT TESTS PASSED  118 passed, 1239 warnings in 833.97s (0:13:53) ---
--- PYTEST VET TESTS PASSED  28 passed, 979 warnings in 801.84s (0:13:21) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 155/155 TESTS  ---
Testing Started at 2022-12-05T23:38:36.102716
Testing Completed at 2022-12-06T09:39:14.754327
Total Time: 10:00:38.651611
```

PR #1524, opened 2023-01-02
Moments hotfix

rcooke-ast:
A pretty minor hotfix (I think)...

### DIFF
--- a/pypeit/core/moment.py
+++ b/pypeit/core/moment.py
@@ -374,7 +374,7 @@ def moment1d(flux, col, width, ivar=None, bpm=None, fwgt=None, row=None, weighti
             raise ValueError('First axis of col must match length of row.')
         if _row.ndim == 2 and _col.ndim == 2 and _col.shape != _row.shape:
             raise ValueError('col and row must have the same shape.')
-        if len(_row) == 1 and _col.ndim == 2 or _row.ndim == 2 and _col.ndim == 1:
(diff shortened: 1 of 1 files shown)


profxj (approved):
Approving but let's get @kbwestfall 's eyes on this before merging.

kbwestfall (changes requested):
Admittedly, figuring out all the possible input combinations is non-trivial, but this change breaks catching one of the failure modes.  If `col` is 2D, the code interprets that as the user defining specific columns for the calculation for each row.  If `row` is a single integer (i.e., it's length after converting to an array is 1), it makes no sense for `col` to be 2D because there's only one row.  This changes the test to check for the dimensionality of `row`, not it's length.  A change to `if _row.size == 1` might make sense, given:
```
>>> x = numpy.arange(4).reshape(2,2)
>>> x
array([[0, 1],
       [2, 3]])
>>> len(x)
2
```
but earlier checks should catch if `_row` is 2d.

rcooke-ast:
OK - sounds good... Admittedly, I'm using PypeIt's core routines to reduce data from an unsupported instrument, when I encountered this issue (and my proposed change fixed it... I presumed the definitions here should have been symmetric), so let me dive into see what the issue might be and if there's anything I can do to fix it - thanks @kbwestfall!

kbwestfall:
Sounds good, @rcooke-ast .  Let me know if you run into any problems.  It could be there's a use case that's not being caught or passed correctly.

rcooke-ast:
It was an issue on my end, and it's now resolved... closing.

PR #1526, opened 2023-01-07
DOC hotfix

rcooke-ast:
The link to spectrographs was broken... fixed here.


profxj (dismissed):
might be better to use `release` than `latest` but the latter
is pointed at the former.  :)

kbwestfall (dismissed):
Good catch!  Thanks!

profxj (approved):
Now on `develop`

kbwestfall (approved):

PR #1530, opened 2023-01-30
Updating numpy.int

SilviaOnorato:
Into pypeit/core/arc.py there were two np.int(...) left that I've changed to int(...).


profxj (approved):
Thx for fixing this!

kbwestfall:
I found a bunch more.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -681,6 +681,7 @@ def calc_pattern_freq(self, frame, rawdatasec_img, oscansec_img, hdu):
             # Deal with the different locations of the overscan regions in 2- and 4- amp mode
             if num_amps == 2:
                 cmin = 1+np.max(pixs[0])
+                # TODO: Can we just use float here?
```
rcooke-ast: Yes - I think so - I think numpy uses float64 by default (?)

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -691,6 +692,7 @@ def calc_pattern_freq(self, frame, rawdatasec_img, oscansec_img, hdu):
                     pixalt = np.where((rawdatasec_img == amp-2) | (oscansec_img == amp-2))
                     cmax = 1+np.min(pixs[0])
                     cmin = (np.max(pixalt[0]) + cmax)//2
+                # TODO: Can we just use float here?
```
rcooke-ast: same here - should be fine to just use `float`

pypeit/utils.py
```diff
@@ -898,7 +898,7 @@ def polyval2d(x, y, m):
         z += a * x**i * y**j
     return z
 
-
+# TODO: Any reason to keep this?
```
rcooke-ast: I'd say this is safe to delete... the more modern version is stored in: `core.fitting.robust_fit`

pypeit/core/skysub.py
```diff
@@ -1471,6 +1471,8 @@ def generate_mask(pypeline, skyreg, slits, slits_left, slits_right, spat_flexure
 
     # Now that we have sky region traces, utilise the SlitTraceSet to define the regions.
     # We will then use the slit_img task to create a mask of the sky regions.
+    # TODO: Is the datatype correct here?  It should be set by the SlitTraceSet
```
rcooke-ast: You're right... I think this should just be `np.integer` - I see no harm in changing it  :-)

pypeit/tests/test_history.py
```diff
@@ -95,6 +95,7 @@ def test_add_reduce():
     spectrograph_def_par = spectrograph.default_pypeit_par()
     ftype_bitmask = FrameTypeBitMask()
 
+    # TODO: Does this need to use np.int_?  Can we just use 'int'?
```
badpandabear, line 98: The bitmask.py code calls "astype" on the argument, which doesn't work on int but does on np.int_.

kbwestfall: Hmmm.... Okay.  Leaving it for now.  I don't remember why these type casting calls are there.

kbwestfall (approved):

PR #1537, opened 2023-02-09
HOTFIX: Add correct exception to ``pypeit.data._get_s3_hostname()``

tbowers7:
Following a question in the User's Slack, it was realized that the try/except block that was supposed to catch the situation of a machine being offline (e.g., a cluster node) was missing the proper exception from the list.  This fix adds ``requests.exceptions.ConnectionError`` to that list.

	modified:   pypeit/data/utils.py


kbwestfall (approved):
Nice!  Thanks for the quick fix!  Can you note this in the CHANGES doc (you might need to add a new 1.12.1dev section at the top)?

profxj (approved):
Thanks!

PR #1539, opened 2023-02-10
Added support for SOAR/Goodman blue (M1 only)

rcooke-ast:
As titled - this adds support for SOAR/Goodman blue M1, and pushed some code into the parent class.

There is a related PR in the dev suite. For the record, here's the output from the wavelength calibration:
image


profxj (changes requested):
Thanks for this!

A few minor requests:

1.  Can you update the spectrograph docs to indicate these updates?
2.  Post tests when complete

rcooke-ast:
Tests pass
image

profxj (approved):
All good

kbwestfall (approved):
Looks good!

PR #1552, opened 2023-03-02
Add wavelength calibration templates to MMTO's Blue Channel and Binospec

tepickering:
This PR started out as adding `reid_arxiv` templates to Binospec for the G600 and G1000 gratings and expanded into doing the same for all of the commonly used gratings for Blue Channel. In the process of testing on various datasets, several bugs were found in the Blue Channel support (e.g. not grouping by grating angle) and some more enhancements were made (e.g. parse arc lamps from the FITS headers). A companion PR adds coverage of these changes to the Dev Suite: https://github.com/pypeit/PypeIt-development-suite/pull/242. 

Running `update_docs` pulled in a lot of unrelated changes. A bug is fixed in `gemini_gmos.rst` that was generating a bunch of warnings. There are four remaining warnings, but I can't figure out why or where they're coming from:
```
include/links.rst:3: WARNING: Bullet list ends without a blank line; unexpected unindent.
include/links.rst:3: WARNING: Bullet list ends without a blank line; unexpected unindent.
include/links.rst:7: WARNING: Bullet list ends without a blank line; unexpected unindent.
include/links.rst:1: WARNING: Bullet list ends without a blank line; unexpected unindent.
```
Happy to fix them here, but also pretty certain they're unrelated to the substantive changes I've made.


kbwestfall:
@tepickering You can ignore those doc warnings.  I should have fixed them in the NIRES Tutorial PR.  (They were particularly annoying because they had nothing to do with the `links.rst` file!)

doc/help/pypeit_cache_github_data.rst
```diff
@@ -25,7 +25,7 @@
                       vlt_xshooter_uvb, vlt_xshooter_vis, wht_isis_blue,
                       wht_isis_red
     
-    optional arguments:
+    options:
```
kbwestfall, line 28: Is this a python version issue?  I'm running python 3.9.

pypeit/spectrographs/mmt_bluechannel.py
```diff
@@ -301,17 +401,43 @@ def check_frame_type(self, ftype, fitstbl, exprng=None):
         good_exp = framematch.check_frame_exptime(fitstbl['exptime'], exprng)
         if ftype == 'bias':
             return fitstbl['idname'] == 'zero'
-        if ftype in ['science', 'standard']:
-            return good_exp & (fitstbl['lampstat01'] == 'off') & (fitstbl['idname'] == 'object')
-        if ftype in ['arc', 'tilt']:
+        if ftype == 'dark':
+            return fitstbl['idname'] == 'dark'
+        if ftype in ['science']:
```
kbwestfall: I would change these `in` conditionals to `==` when there is only one option.  I.e., change to `if ftype == 'science':`.

kbwestfall (approved):
Looks good to me.  Only a couple of minor things.

doc/help/pypeit_cache_github_data.rst
tepickering: oh! good point. i was in my 3.11 env when i ran `update_docs`...

pypeit/spectrographs/mmt_bluechannel.py
tepickering: yeah, i was moving things in and out of lists as i was finding new edge cases in the archived data. i'll clean that up.

tepickering: that said, the optimal solution is to use `match`-`case` and require python >=3.10...

tepickering:
> @tepickering You can ignore those doc warnings. I should have fixed them in the NIRES Tutorial PR. (They were particularly annoying because they had nothing to do with the `links.rst` file!)

omg, i did some digging to try to figure it out, but soon came to realize that would lead to madness...

pypeit/spectrographs/mmt_bluechannel.py
kbwestfall: Ha!  I had exactly the same thought, but decided against the suggestion :)

tepickering: it's worth discussing making the move to require 3.10+. 3.11 is consistently at least 5% faster than 3.9 for every reduction i've compared. maybe target the `1.13.0` release for that.

pypeit/spectrographs/mmt_bluechannel.py
```diff
@@ -178,18 +193,20 @@ def default_pypeit_par(cls):
         # 1D wavelength solution
         par['calibrations']['wavelengths']['rms_threshold'] = 0.5
         par['calibrations']['wavelengths']['sigdetect'] = 5.
-        par['calibrations']['wavelengths']['fwhm']= 5.0
-        # HeNeAr is by far most commonly used, though ThAr is used for some situations.
-        par['calibrations']['wavelengths']['lamps'] = ['ArI', 'ArII', 'HeI', 'NeI']
+        par['calibrations']['wavelengths']['fwhm_fromlines'] = True
+
+        # Parse the lamps used from the image header.
+        par['calibrations']['wavelengths']['lamps'] = ['use_header']
```
jhennawi, line 199: So currently this says below that the wavelength calibration is done with holy-grail. You overwrite this below in config_specific_par, but I'm wondering what the most transparent approach is here. In other words, do you want to default to holy-grail if config_specific_par does not find a match? I think this does that, and one can debate about how to make that most explicit. But perhaps it would be more transparent to the user if an exception was raised in config_specific_par if the grating setup is not calibrated. Currently it will just run with holy-grail in that case which may or may not work, and if it fails it will throw a cryptic exception.

jhennawi: The desired behavior here sort of depends on how likely such a corner case is to occur.

jhennawi (approved):
Thanks @tepickering. I have a relatively minor comment on the "holy-grail" method default. You are ready to merge after tests pass. I am not sure whether we need a full dev-suite run here, but the team can help you run that. It is never a bad idea to test the code.

pypeit/spectrographs/mmt_bluechannel.py
tepickering: yes, the intention here is to default to `holy-grail` and override in `config_specific_par` if there is an available template. this is how `mmt_binospec` works currently and until this PR gets merged. in reality, the blue channel templates here cover all cases i've found in the archived data going back to 2014. the only missing dispersers are an old 600 l/mm grating and the echellette mode. the former has been replaced by the 832 l/mm grating which is vastly superior in every way. the latter hasn't been used in over 20 years and i don't think was ever re-commissioned after changing the detector format. if for some weird reason someone did choose to use the 600 l/mm grating (and assuming we can find where it's stored), there is reasonable expectation that `holy-grail` would work. it works pretty well out of the box with the 800 and 500 gratings at similar central wavelength settings.

PR #1555, opened 2023-03-06
Datacube resample

rcooke-ast:
This (long overdue) PR includes support for new algorithms for the coadd3d module. I've included support for resampling the 3D voxels (very slow), and a faster subsampling approach, which is now the default option. This will need to survive the dev suite, and I'll run that once folks are happy with these changes.

I've also removed the skimage dependency from one part of the code.


pypeit/alignframe.py
```diff
@@ -323,3 +324,96 @@ def show_alignment(alignframe, align_traces=None, slits=None, clear=False):
...
+
+        Parameters
+        ----------
+        traces : `numpy.ndarray`
+            The alignments (traces) of the slits. This allows different slices
+            to be aligned correctly. Ideally, this variable will be assigned the
+            value of alignments.traces. However, this can also be assigned the
+            left and right slit edges:
+            traces = np.append(left.reshape((left.shape[0],1,left.shape[1])),
+                   right.reshape((left.shape[0],1,left.shape[1])), axis=1)
```
kbwestfall: Put these lines in a code block?  And give the shape of the traces array.

pypeit/alignframe.py
```diff
@@ -323,3 +324,96 @@ def show_alignment(alignframe, align_traces=None, slits=None, clear=False):
...
+        slitnum : `int`
+            Slit number
+        spatpix : `numpy.ndarray`
+            Detector pixel coordinate (spatial direction)
+        specpix : `numpy.ndarray`
+            Detector pixel coordinate (spectral direction)
+
+        Returns
+        -------
+        tuple : There are
```
kbwestfall: Doc incomplete.

pypeit/core/extract.py
```diff
@@ -693,6 +693,9 @@ def fit_profile(image, ivar, waveimg, thismask, spat_img, trace_in, wave,
     waveimg: `numpy.ndarray`_
         Floating-point wavelength image. Must have the same shape as ``image``,
         :math:`(N_{\rm spec}, N_{\rm spat})`.
+    thismask : `numpy.ndarray`_
+        Boolean image indicating which pixels are on the slit/order in question.
+        Must have the same shape as ``sciimg``, :math:`(N_{\rm spec}, N_{\rm spat})`.
```
kbwestfall, line 691: Good catch!

pypeit/par/pypeitpar.py
```diff
@@ -1388,6 +1388,30 @@ def __init__(self, slit_spec=None, relative_weights=None, combine=None, output_f
                                    'ranges. If combine=False, the individual spec3d files will have a suffix ' \
                                    '"_whitelight".'
 
+        defaults['method'] = "subsample"
+        dtypes['method'] = str
+        descr['method'] = 'What method should be used to generate the datacube. There are currently three options:' \
+                          '(1) "subsample" (default) - this algorithm subsamples each pixel in the spec2d frames ' \
+                          'and assigns each subpixel into a voxel of the datacube. Note that the resample algorithm ' \
+                          'exactly maps pixels to voxels, the subsample algorithm is much faster and performs just ' \
```
kbwestfall: "... voxels, the ... " -> "... voxels, but the ..." ?

pypeit/slittrace.py
```diff
@@ -498,46 +462,14 @@ def get_radec_image(self, wcs, alignments, tilts, locations,
...
-                evalpos = (pixdst[:, 0] - 0.5) * slitlength
-            else:
-                evalpos = onslit_init[1] - trace_cen[onslit_init[0], slit_idx]
-            minmax[:, 0] = np.min(evalpos)
-            minmax[:, 1] = np.max(evalpos)
+                msgs.error(f"Slit {spatid} ({slit_idx+1}/{self.spat_id.size}) is masked. Cannot generate RA/DEC image.")
+            # Retrieve the pixel offset from the central trace
+            evalpos = alignSplines.transform(slit_idx, onslit_init[1], onslit_init[0])
+            minmax[slit_idx, 0] = np.min(evalpos)
+            minmax[slit_idx, 1] = np.max(evalpos)
```
kbwestfall, line 469: Nice!  Much cleaner.

pypeit/core/datacube.py
```diff
@@ -164,6 +170,40 @@ def to_file(self, ofile, primary_hdr=None, hdr=None, **kwargs):
...
+
+    Parameters
+    ----------
+    traces : `numpy.ndarray`
+        The alignments (traces) of the slits. This allows different slices
+        to be aligned correctly. Ideally, this variable will be assigned the
+        value of alignments.traces. However, this can also be assigned the
+        left and right slit edges:
+        traces = np.append(left.reshape((left.shape[0],1,left.shape[1])),
+               right.reshape((left.shape[0],1,left.shape[1])), axis=1)
```
kbwestfall: Put this in a code block.

pypeit/core/datacube.py
```diff
@@ -164,6 +170,40 @@ def to_file(self, ofile, primary_hdr=None, hdr=None, **kwargs):
...
+        In this case, locations=np.array([0,1])
+    tilts : `numpy.ndarray`
+        Spectral tilts.
+    locations : `numpy.ndarray`_, list
+        locations along the slit of the alignment traces. Must
+        be a 1D array of the same length as alignments.traces.shape[1]
+
+    Returns
+    -------
+    `pypeit.alignframe.AlignmentSplines` : An instance of the AlignmentSplines class that allows
```
kbwestfall: Wrong syntax for a Numpy style docstring.  Return entries should be identical to the parameter entries:
```
object_instance_name : type, etc
    description
```

pypeit/core/datacube.py
```diff
@@ -164,6 +170,40 @@ def to_file(self, ofile, primary_hdr=None, hdr=None, **kwargs):
         super(DataCube, self).to_file(ofile, primary_hdr=primary_hdr, hdr=hdr, **kwargs)
 
 
+def get_alignments(traces, locations, tilts):
```
kbwestfall: Is this function necessary?  All the input checks should also be checked when instantiating the `AlignmentSplines` object, and I think you only call this function once.  Can you instead instantiate the `AlignmentSplines` object directly at the appropriate spot?

pypeit/core/datacube.py
```diff
@@ -989,6 +1029,337 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
     return all_wghts
 
 
+def generate_cube_resample(outfile, frame_wcs, slits, fluximg, ivarimg, raimg, decimg, waveimg, slitimg, gpm,
+                           grid_nspat=5, grid_specsep=20,
+                           overwrite=False, output_wcs=None, blaze_wave=None, blaze_spec=None, fluxcal=False,
+                           sensfunc=None, specname="PYP_SPEC", debug=False):
```
kbwestfall: Why is the default name of the spectrograph `PYP_SPEC`?  Does the code fault if you set it to None?

pypeit/core/datacube.py
```diff
@@ -989,6 +1029,337 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
...
+        ivarimg (`numpy.ndarray`_):
+            Inverse variance of each pixel in the frame
+        raimg (`numpy.ndarray`_):
+            Right ascension of each pixel in the frame (units = decimal degrees)
+        decimg (`numpy.ndarray`_):
+            Declination of each pixel in the frame (units = decimal degrees)
+        waveimg (`numpy.ndarray`_):
+            Wavelength of each pixel in the frame (units = Angstroms)
+        slitimg (`numpy.ndarray`_):
+            Slit image. -1 is not on a sli, and all other
```
kbwestfall: "sli" -> "slit"

pypeit/core/datacube.py
```diff
@@ -989,6 +1029,337 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
...
+            pixels are labelled with their spatial IDs.
+        gpm (`numpy.ndarray`_):
+            Good pixel mask (bool). True = good pixel
+        grid_nspat (int):
+            Number of grid points in the spatial direction when evaluating the
+            voxel geometry in detector coordinates. This should be an odd number
+        grid_specsep (int):
```
kbwestfall: This one and the previous argument is optional.

pypeit/core/datacube.py
```diff
@@ -989,6 +1029,337 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
...
+        overwrite (bool, optional):
+            If the output file exists, it will be overwritten if this parameter is True.
+        output_wcs (`astropy.wcs.wcs.WCS`_, optional):
+            World coordinate system for the output datacube. If None, frame_wcs will be used.
+        blaze_wave (`numpy.ndarray`_, optional):
+            Wavelength array of the spectral blaze function
+        blaze_spec (`numpy.ndarray`_, optional):
+            Spectral blaze function
+        fluxcal (bool, optional):
+            Are the data flux calibrated?
```
kbwestfall: What are the units of `fluximg` if `fluxcal` is false?

pypeit/core/datacube.py
```diff
@@ -989,6 +1029,337 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
...
+        this_ra, this_dec, this_wave = raimg[wsl], decimg[wsl], waveimg[wsl]
+        _, spat_posn, _ = frame_wcs.wcs_world2pix(this_ra, this_dec, this_wave*1.0E-10, 0)
+        asrt = np.argsort(spat_posn)
+        ra0[sl] = np.interp(0.0, spat_posn[asrt], this_ra[asrt])
+        dec0[sl] = np.interp(0.0, spat_posn[asrt], this_dec[asrt])
+        # Generate the offsets
+        cosdec = np.cos(dec0[sl] * np.pi / 180.0)
+        diff_ra, diff_dec = (this_ra - ra0[sl]) * cosdec, this_dec - dec0[sl]
+        msgs.bug("There is sometimes a sign error that needs to be resolved here...")
+        msgs.error("Use another algorithm for the time being...")
```
kbwestfall: Doesn't this mean the code always faults when using this function?  If so, I think we shouldn't allow users to select this method.

pypeit/core/datacube.py
```diff
@@ -989,6 +1029,337 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
...
+    # Calculate an "offsets" image, which indicates the offset in arcsec from (RA_0, DEC_0)
+    # Create two splines of the offsets image: (1) offset predicts RA; (2) offset predicts Dec.
+    # Use these splines to calculate the RA and DEC of the voxels, combine this with the output wavelength grid.
+    # Generate all RA, DEC, WAVELENGTH triples (i.e. find the RA,DEC pairs along constant wavelength, for all wavelengths)
+    # Use the WCS (which contains the astrometric transform) to go from world to pix
+    #    i.e. need to invert this:
+    #    world_ra, world_dec, _ = wcs.wcs_pix2world(slitID, evalpos, tilts[onslit_init]*(nspec-1), 0)
+    # This gives us the x,y detector positions of the voxel geometry
+        from shapely.geometry import Polygon, box as shapelyBox
+        from shapely.strtree import STRtree
```
kbwestfall: These need to be imported at the top of the file, and we need a check early in any script workflow that checks that `shapely` is installed if a user selects the resample method.

pypeit/core/datacube.py
```diff
@@ -1128,13 +1501,30 @@ def coadd_cube(files, opts, spectrograph=None, parset=None, overwrite=False):
...
     # prep
     numfiles = len(files)
     combine = cubepar['combine']
+    method = cubepar['method'].lower()
+
+    # Determine what method is requested
+    if method == "resample":
+        if shapely is None:
+            msgs.error("To use the 'resample' algorithm to make your datacube, you need to install shapely.")
+        else:
```
kbwestfall: Don't need the else (`msgs.error` breaks the code flow).

pypeit/core/datacube.py
```diff
@@ -989,6 +1029,337 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
...
+
+    # Save a residuals cube, if requested
+    if debug:
+        outfile_resid = outfile.replace(".fits", "_resid.fits")
+        msgs.info("Saving residuals datacube as: {0:s}".format(outfile_resid))
+        hdu = fits.PrimaryHDU(residcube.T, header=hdr)
+        hdu.writeto(outfile_resid, overwrite=overwrite)
+
+
 def generate_cube_ngp(outfile, hdr, all_sci, all_ivar, all_wghts, pix_coord, bins,
```
kbwestfall: Given the overlap with the new subsample function, do we still need this ngp function?

kbwestfall (changes requested):
Thanks @rcooke-ast .  Mostly minor stuff from me and some clean-up requests.  Two more significant requests are:

1. Tests: We need some unit tests and/or vet tests for the resampling functionality.  Is there an associated dev-suite PR incoming?
2. Docs: It would be really helpful to have a complete explanation of how the datacube reconstruction is performed.  I'm not sure how much of this already exists, but it would be good to have an explanation in the main doc pages describing the different algorithms available and their pros/cons.

pypeit/par/pypeitpar.py
MNBrod: Is there really a reason to leave resampling in as an option if the result of subsampling is essentially identical and faster? If so, documenting that reason (e.g. "this weird edge case" or "if you really care about ___") would be welcome

pypeit/alignframe.py
rcooke-ast: Great - done!  But I may not have done this right - please let me know if anything is out of place  :-)

pypeit/alignframe.py
rcooke-ast: Whoops - thanks for spotting that!

pypeit/par/pypeitpar.py
rcooke-ast: done

pypeit/core/datacube.py
rcooke-ast: Nope... it's pointless... I'll remove it - thanks!!

pypeit/core/datacube.py
rcooke-ast: Fixed - Changed to None. The code works just fine with `PYP_SPEC=None`

pypeit/core/datacube.py
rcooke-ast: Added some docs for this

pypeit/core/datacube.py
rcooke-ast: It does... I remember now that this is a work in progress, that took a lot of effort to get working and there's a bug I can't resolve. For now, I'll comment out the `resample` option, but I'd like to keep the code here for the time-being.

pypeit/core/datacube.py
rcooke-ast: That's right... I was still debugging this code, but for now, I've commented it all out.

pypeit/core/datacube.py
rcooke-ast: Resample is now gone as an option!

pypeit/par/pypeitpar.py
rcooke-ast: Good point, thanks!  I've also realised that the resampling code has a bug that I haven't managed to track down, so I'm removing this support for now, anyway. I don't really think there is an edge case, tbh...

rcooke-ast:
Thanks @kbwestfall and @MNBrod for the quick reviews!

> Tests: We need some unit tests and/or vet tests for the resampling functionality. Is there an associated dev-suite PR incoming?

There is already a test in the dev-suite for the subsample algorithm, and I've now removed the resample algorithm.

> Docs: It would be really helpful to have a complete explanation of how the datacube reconstruction is performed. I'm not sure how much of this already exists, but it would be good to have an explanation in the main doc pages describing the different algorithms available and their pros/cons.

Agreed... I'll write that now and push the changes.

pypeit/core/datacube.py
rcooke-ast: Yes... for the moment... This function is used when combining multiple cubes together. The subsample function is only used for generating a NGP cube when converting single spec2d files into a single datacube. I can't find a resource friendly way of doing this to multiple frames, yet.

rcooke-ast (commented):
Thanks for the comments!

pypeit/alignframe.py
kbwestfall: The syntax looks right, but you might need some empty lines above and below the code in the block.  I.e., something like:
```
    arg (type):
        Here's the description, and some code:
        
        .. code-block:: python

            this = 'the code'

    next_arg (type):
        ...
```
I would compile the docs, and check what the result looks like, assuming you don't get any warnings.

kbwestfall (approved):
Thanks for addressing all my comments!  Have you run any of the tests?

pypeit/alignframe.py
rcooke-ast: OK - it compiled the docs just fine (but I decided to add the space anyway, since it looks a lot clearer like this)

rcooke-ast:
> Thanks for addressing all my comments! Have you run any of the tests?

Not yet - I figured I would wait for a review from @jhennawi, implement all of the suggested changes, and then fix any problems.

pypeit/core/datacube.py
```diff
@@ -24,6 +24,12 @@
 from pypeit.core.procimg import grow_mask
 from pypeit.spectrographs.util import load_spectrograph
 
+# Shapely is needed if using the resample algorithm
```
jhennawi: Remove?

pypeit/core/datacube.py
```diff
@@ -989,6 +995,339 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
     return all_wghts
 
 
+# def generate_cube_resample(outfile, frame_wcs, slits, fluximg, ivarimg, raimg, decimg, waveimg, slitimg, gpm,
```
jhennawi: Move to deprecated?

pypeit/core/datacube.py
```diff
@@ -989,6 +995,339 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
...
+    # Prepare the header, and add the unit of flux to the header
+    hdr = output_wcs.to_header()
+    if fluxcal:
+        hdr['FLUXUNIT'] = (PYPEIT_FLUX_SCALE, "Flux units -- erg/s/cm^2/Angstrom/arcsec^2")
+    else:
+        hdr['FLUXUNIT'] = (1, "Flux units -- counts/s/Angstrom/arcsec^2")
+
+    # Write out the datacube
+    msgs.info("Saving datacube as: {0:s}".format(outfile))
+    final_cube = DataCube(datacube.T, varcube.T, specname, blaze_wave, blaze_spec, sensfunc=sensfunc, fluxed=fluxcal)
```
jhennawi: We use ivar instead of var everywhere in PypeIt for the reason that if ivar = 0 that is better than var = 0 to describe bad pixels. Of course we also have masks that deal with that. But my point is that you are switching convention here by outputting a variance instead of an inverse variance.

pypeit/core/datacube.py
```diff
@@ -989,6 +995,339 @@ def compute_weights(all_ra, all_dec, all_wave, all_sci, all_ivar, all_idx, white
...
+#     msgs.info("Saving datacube as: {0:s}".format(outfile))
+#     final_cube = DataCube(datcube.T, varcube.T, specname, blaze_wave, blaze_spec, sensfunc=sensfunc, fluxed=fluxcal)
+#     final_cube.to_file(outfile, hdr=hdr, overwrite=overwrite)
+
+
+def generate_cube_subsample(outfile, output_wcs, all_sci, all_ivar, all_wghts, all_wave, tilts, slits, slitid_img_gpm,
+                            astrom_trans, bins, subsample=10, overwrite=False, blaze_wave=None, blaze_spec=None,
+                            fluxcal=False, sensfunc=None, specname="PYP_SPEC", debug=False):
+    """
+    Save a datacube using the subsample algorithm. This algorithm splits
```
jhennawi: Maybe add to the docs that the if subsample > 10 the errors will be correlated and the covariance is not being tracked, so the errors won't be right. That is fine, tracking the covariance would be massively painful and this is the tradeoff one has to make between sampling and better looking cubes, and poor sampling and better behaved errors. I would just add a note. Finally, in coadd2d, we called subsample spec_fact and spat_fact. And I think that definition is the inverse of what you have here. It might be best to try to adopt consistent definitions. In other words, spat_fact = 2.0 performs a coarse rebidding, whereas spat_fact = 0.5 would be your subsample = 2. Finally, shouldn't there be an option here to have a different sampling in the spectral direction (is that your z) as opposed to the spatial directions (x,y)?

pypeit/core/datacube.py
```diff
@@ -1011,15 +1350,17 @@ def generate_cube_ngp(outfile, hdr, all_sci, all_ivar, all_wghts, pix_coord, bin
             pixel in the processed spec2d frames. After setting up an astropy WCS, pix_coord is
```
jhennawi: See my comment above. Would it not be best to distinguish between spectral directions and spatial directions in your documentation to make things more clear.

pypeit/par/pypeitpar.py
```diff
@@ -1388,6 +1388,29 @@ def __init__(self, slit_spec=None, relative_weights=None, combine=None, output_f
...
                                    '"_whitelight".'
 
+        defaults['method'] = "subsample"
+        dtypes['method'] = str
+        descr['method'] = 'What method should be used to generate the datacube. There are currently two options:' \
+                          '(1) "subsample" (default) - this algorithm subsamples each pixel in the spec2d frames ' \
+                          'and assigns each subpixel into a voxel of the datacube. Flux is conserved, but voxels ' \
+                          'are correlated, and the error spectrum does not account for covariance between adjacent ' \
+                          'voxels. ' \
+                          '(2) "NGP" (nearest grid point) - this algorithm is effectively a 3D histogram. Flux is ' \
```
jhennawi, line 1398: Nice docs. I suggest adding what you wrote here to the core routine to address my comment above.

jhennawi: Maybe clarify that when multiple pixels hit the same bin for NGP, you are taking the average flux. I think that SB and hence flux are still conserved, but maybe worth a note since your docs suggest it would just add the flux.

jhennawi (changes requested):
Looks great @rcooke-ast. I have some rather minor comments about definitions and docs. Otherwise you are good to go.

pypeit/par/pypeitpar.py
rcooke-ast: Yes, good catch... I'll add that in.

pypeit/core/datacube.py
rcooke-ast: Removed

pypeit/core/datacube.py
rcooke-ast: Yes, good idea - moved (and I've moved the shapely import, too).

pypeit/core/datacube.py
rcooke-ast: OK - I'm willing to concede that `varcube` is a poor choice. However, the default thing that we output in the spec1d files include things like `OPT_FLAM`, `OPT_FLAM_IVAR`, and `OPT_FLAM_SIG`. I'm more than happy to output the IVAR and the SIG cubes, but these are usually quite big files. Also, I suspect `SIG` might be more useful to users than `IVAR` with the caveat that you mention about 0 values in the `SIG` cubes. I'm also not outputting a mask for the datacubes... Maybe we can discuss this on Slack.

pypeit/core/datacube.py
rcooke-ast: Good point - I've added an important note to the top of the doc string.

Regarding the sampling factor, I can see the benefit of it being consistent, however, it really should be an integer in this case (i.e. you want to break up a pixel into an integer number of small pixels. Also, in datacube, subsample cannot be less than 1. So it needs to be an integer >= 1). We could consider refactoring coadd2d to account for this, but I feel this might be more effort than it's worth... Happy to hear your thoughts on this. One possibility is to adopt the same formalism as in `coadd2d`, but convert this to an (allowed) integer within datacube. Seems like a messy way of doing it, though...

Good idea to have separate x and y subsampling. Note that the subsampling done here is on each individual detector pixel, not quite the spatial and wavelength coordinates (due to the tilts). In my latest refactoring, I've adopted the names `spec_subsample` and `spat_subsample`.

pypeit/core/datacube.py
rcooke-ast: I'm not sure I understand this comment in the context of what is written on this line. I've tried to clarify this part of the docstring, but if I've missed the point, let me know!

rcooke-ast (commented):
Addressed comments from JFH

pypeit/core/datacube.py
jhennawi: Okay, I see your point about this being different since it has to be an integer. In that case, casting an integer to 1/integer is a bad idea, so let's leave your definitions and variables as they are. So shouldn't this be called supersample though? You are not subsampling if your final cube has higher resolution than your individual frames, you are super-sampling right?

jhennawi (approved):
@rcooke-ast I'm approving but please see my last comment. Also as per the Slack, I'm still for including the ivar as an output, but we can discuss there.

pypeit/core/datacube.py
rcooke-ast: So, the final cube doesn't have higher spatial resolution. I've just divided each detector pixel into N subpixels (say 25), and a fraction of those 25 subpixels are put into a single voxel. For example, of the 25 subpixels, 2 subpixels end up in voxelA, 16 subpixels end up in voxelB, and the remaining 7 subpixels end up in voxelC... with NGP, all 25 subpixels (i.e. the 1 pixel) end up in voxelB. So, this is a way to share the flux from 1 detector pixel into several voxels, depending on where the voxel boundaries are.

... and you're right - subsample is the incorrect word... I've now opted for subpixel.

rcooke-ast:
> @rcooke-ast I'm approving but please see my last comment. Also as per the Slack, I'm still for including the ivar as an output, but we can discuss there.

As an alternative, instead of including `ivar` as an output, can we instead include it as a @property. Here is an example:

image

So, the user won't know anything different (unless they go digging into the fits files), and it saves on output filesize.

rcooke-ast:
Merging - and will reconsider adding ivar in a future PR

PR #1563, opened 2023-03-17
Flipped slitpa bug

debora-pe:
Address #1551


rcooke-ast (approved):
Looks good to me!

profxj (changes requested):
Great stuff!

Is it possible to add a `vet` test
in the DevSuite?  I think we have tilted slits there.

profxj (approved):
all set now!

PR #1565, opened 2023-03-23
Skysub hotfix

rcooke-ast:
Due to a bug identified on the Users Slack, I've cleaned up the code (significantly reduced!) to make it much easier to maintain.


profxj (approved):
All good.

Should we run the DevSuite to make sure removing `scripts/utils.py` is ok?

rcooke-ast:
> Should we run the DevSuite to make sure removing `scripts/utils.py` is ok?

Yes - I'll do that, especially since this is going into release... but I'm pretty sure that I wrote this script a long time ago, and no one ever used it (and it was only used for the skysub regions GUI)!!

Thanks for the review.

profxj (approved):
thx for the redirect

merge once tests pass

rcooke-ast:
Tests passing...
image

debora-pe (approved):
Looks good to me!

PR #1570, opened 2023-03-27
Hotfix for calibration-only run

kbwestfall:
Fixes two bugs when running calibrations only:

 - If all calibration groups are set to `all`, the code wasn't identifying any calibration groups.  This wasn't tripped for a normal pypeit run because science frames can only belong to a single calibration group.

 - Calibration group numbers are (currently) sequential numbers starting with 0.  So if all calibration groups are set to 1, there are actually 2 calibration groups (0,1) and one of them is empty.  This causes a fault in the `calib_all` function because it hits a group that has no associated frames.


profxj (approved):
thx for the quick fix

jhennawi (approved):
Thanks!

PR #1572, opened 2023-03-28
Coadd different size inputs

rcooke-ast:
Several minor updates

* Includes a new wavelength solution for a Shane/Kast red grating
* Included functionality to coadd input arrays of different sizes (in this case, supporting the coaddition of Kast blue + Kast red).


pypeit/coadd1d.py
```diff
@@ -112,7 +112,20 @@ def load_arrays(self):
                 if 'RA' in sobjs[indx][0].keys() and 'DEC' in sobjs[indx][0].keys():
                     header_out['RA_OBJ']  = sobjs[indx][0]['RA']
                     header_out['DEC_OBJ'] = sobjs[indx][0]['DEC']
-
+            # Check if the arrays need to be padded
```
jhennawi, line 116: I'm fine with leaving this in, but it will eventually be solved with lists. Maybe please just add a todo here indicating that a better way is to use lists rather than arrays and add a TODO for the future so we remember to remove this.

jhennawi (approved):
Minor comment. Please address, but otherwise approved.

pypeit/coadd1d.py
rcooke-ast: Sounds good - Added TODOs and also added an issue as a reminder to address this in the future: #1575

debora-pe (approved):
Looks good to me.  I'm approving.
Maybe it would be good to run some tests.

rcooke-ast:
This PR also includes a fix to a bug in the VET tests due to changes in the datacube datamodel.

rcooke-ast:
Tests mostly pass... broken tests are unrelated to this PR, and will be fixed with a subsequent PR.

image

PR #1577, opened 2023-04-04
fwhm in pixels

rcooke-ast:
FWHM should be in pixels, not Angstroms


profxj (approved):
I assume this is right :)

Shall we put a test on it?

PR #1578, opened 2023-04-05
Uniformly set FWHM to be pixels

rcooke-ast:
Proposal to set FWHM in pixels for both archive and object spectra.


pypeit/core/flexure.py
```diff
@@ -668,23 +667,23 @@ def spec_flexure_slit(slits, slitord, slit_bpm, sky_file, method="boxcar", speco
             continue
 
         # get spectral FWHM (in Angstrom) if available
```
debora-pe: please change this comment: `(in Angstrom)`--> `(in pixels)`, just to avoid confusion

debora-pe (approved):
It looks OK to me. It does look more clean and uniform like this. 
The only thing is that `wv_calib.wv_fits[iwv].cen_disp` and `obj_disp` are computed in a slightly different way, so it would be good to run the tests to check that everything looks still good.
Thanks.

rcooke-ast:
Good idea @debora-pe -- nothing appears to be broken, and the results are consistent with what we were using before  :-)

**Old version:**
image

**New Version:**
image

pypeit/core/flexure.py
rcooke-ast: Good catch - updated!

profxj (approved):
thx for this!

rcooke-ast:
Thanks for the approvals - I will merge #1572 first, and then merge this PR...

PR #1580, opened 2023-04-06
Added chk_version option to override different version of PypeIt crash.

jhennawi:
This minor PR adds an option for the user to disable version checking for coadding 1d spectra. I would have liked to also add this as an option to the pypeit_coadd1d_spec script, but instead opted to just make it a parameter in the parset.


pypeit/par/pypeitpar.py
```diff
@@ -1164,6 +1164,13 @@ def __init__(self, ex_value=None, flux_value=None, nmaskedge=None,
         dtypes['coaddfile'] = str
         descr['coaddfile'] = 'Output filename'
 
+
+        # Version checking
+        defaults['chk_version'] = True
+        dtypes['chk_version'] = bool
+        descr['chk_version'] = 'If True enforce strict PypeIt version checking to ensure that spec1d*.fits files were created' \
+                               'with the current version of PypeIt'
+
```
kbwestfall, line 1173: It isn't strictly the pypeit version that matters.  It's the version of the data model for the SpecObj object.  So how about instead:
```
descr['chk_version'] = 'If True, enforce strict PypeIt version checking.  Beware that setting this to False may cause other obscure faults or silent bugs.'
```
And a general note, when you break strings like this, make sure you add relevant spacing.  The two strings are directly combined meaning that, as you've written it, the resulting spring will include the word `createdwith`.

kbwestfall (approved):
One minor comment from me, but looks good otherwise.

profxj (approved):
all good

PR #1581, opened 2023-04-10
Extra QA + code cleanup

rcooke-ast:
* Added some QA for the detector structure correction (of the flatfield) -- currently only used by KCWI. See below for an example of the QA
* improved the code flow for the structure correction
* Iterated over the fine correction for the spatial illumination correction (so that the detector structure is accounted for). See the before and after example below.
* made a convenience function for the flexure correction

**New Detector structure QA**
image

**Iterating over flatfield improves the fine spatial illumination profile correction**
image


profxj (approved):
Love the new QA!

debora-pe (approved):
Looks good to me!

PR #1582, opened 2023-04-14
Hotfix : Datacube alignment frames

rcooke-ast:
The tilts need some padding when using the alignment frames for the astrometric correction. Running dev suite tests now...


tbowers7:
This commit will let the PR pass CI tests.

rcooke-ast:
The key thing is that the VET tests pass. There are two dev suite failures:
* ldt_deveny -- this bug is fixed in PR #1567
* keck_hires -- this bug cannot be related to the current PR (the only change here is one line in the `keck_kcwi.py` spectrograph file)
image

kbwestfall (approved):

profxj (approved):
thx!

PR #1587, opened 2023-04-21
Hotfix: SOAR goodman red fine corr

rcooke-ast:
Implements a hot fix for the fine correction. There was an issue with the tweaked edges. This commit sherry picks from another branch I'm working on that resolves this issue.

FWIW - the fine correction should probably be turned off for this spectrograph, as the fine correction is fitting to some unaccounted for detector structure.

Eventually, we should use inverse variance weights, and possible a 2D bspline fit...


rcooke-ast:
Closing, as discussed on Slack.

PR #1593, opened 2023-05-10
JWST setup fix

kbwestfall:
Fix for running `pypeit_setup` on JWST files.  The root of the failure mode was that PypeIt couldn't set the frametype for the JWST file, and that meant it couldn't assign them to a calibration group.  There are two possible fixes, either of which will fix the failure mode but I've actually implemented both:

1. The safest approach may be to fix the frame typing for JWST.  In `jwst_nirspec.py`, I just set everything to a science frame.

2. The more general fix is in `metadata.py`.  The existing code will only set the calibration group of a frame if it has been successfully typed.  I simply removed that requirement and that fixed the failure.  I'm a little worried about this fix, though, because I'm not sure why we required the frame to be successfully typed to add it to a default calibration group.  It makes sense to do this, but I don't think it's necessarily a programmatic requirement.

I'm fine with either or both fixes, but the 2nd one will require more testing to make sure we're not introducing other bugs.

### DIFF
--- a/pypeit/metadata.py
+++ b/pypeit/metadata.py
@@ -1095,7 +1095,8 @@ class PypeItMetaData:
         # The configuration must be present to determine the calibration
         # group
         if 'setup' not in self.keys():
-            msgs.error('Must have defined \'setup\' column first; try running set_configurations.')
+            msgs.error('CODING ERROR: Must have defined \'setup\' column first; try running '
+                       'set_configurations.')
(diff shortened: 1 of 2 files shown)


jhennawi (approved):
Thanks, @kbwestfall this is much appreciated. I think we want to prevent crashes if typing fails and just print a warning so I would be in favor of both changes here. More appropriate is if poorly constructed pypeit files crash than if pypeit_setup or pypeit_obslog crashes, since this prevents users from figuring out the information about their frames.

PR #1594, opened 2023-05-11
Update parameters for LDT/DeVeny

tbowers7:
The primary focus of this PR is to update reduction parameters (and internal methods) for the LDT/DeVeny spectrograph, including adding new neon lines (from NIST) to the linelist.

Along the way, I've made several small changes, corrections, and documentation updates.  Indeed, all changes to the spectrograph modules _other than LDT/DeVeny_ are docstring-only, plus a conformal change to a parameter in GTC/Osiris.

**This PR is predicated on #1583 clearing (_i.e._, I've pulled in the `ginga_stuff` branch here), so proper review should occur after that PR merges.**

All changes included in this PR:
Major:
1. Updates to reduction parameters and internal methods for the LDT/DeVeny spectrograph and the LDT telescope definition.
2. Addition of 11 neon lines (from NIST) consistently recovered with our 600L grating:
ldt_deveny_600_Ne

Minor:
3. Allow the user to pass in a list of `-r` directories to `pypeit_setup` (_e.g._, by using wildcards).  This includes changes in `pypeit/scripts/setup.py` and `pypeit/io.py`.
4. Produce more helpful error messages to the user when a `.pypeit` file contains frames with `None` or other unrecognized frame types.  This includes changes to `pypeit/bitmask.py`, `pypeit/metadata.py`, and `pypeit/tests/test_bitmask.py`.
5. Clarify that the `FlexurePar` parameter `spec_maxshift` must be an integer because it is used directly in a slice in the flexure code, and returning the correct lag from the correlation.  This required changes to `pypeit/core/flexure.py`, as well as the parameter definitions and a conformal change in GTC/Osiris.
6. Allow for spaces in pathnames when parsing input files (_e.g._, a `.pypeit` file).  The parser in `pypeit/inputfiles.py` implicitly assumed the only space in the path was right after `path`.  This change preserves the entire string following `path `.
7. Docstring updates to make sphinx happy -- this includes most of the spectrograph modules.


pypeit/bitmask.py
```diff
@@ -19,6 +19,8 @@
 import os
 import textwrap
 
+from pypeit import msgs
+
```
kbwestfall, line 23: I'd prefer that you revert these changes.  `bitmask.py` is something I have in many repositories that I work on.  And, although I constantly fail, I try to keep them up-to-date with each other.  All this means that I'd like to keep `bitmask.py` independent of the larger pypeit code.

pypeit/core/flexure.py
```diff
@@ -313,7 +313,7 @@ def spec_flex_shift(obj_skyspec, arx_skyspec, arx_fwhm_pix, spec_fwhm_pix=None,
 
     return dict(polyfit=fit, shift=shift, subpix=subpix_grid,
                 corr=corr[subpix_grid.astype(int)], sky_spec=obj_skyspec, arx_spec=arx_skyspec,
-                corr_cen=corr.size/2, smooth=smooth_fwhm_pix, method=method)
+                corr_cen=lag0, smooth=smooth_fwhm_pix, method=method)
```
kbwestfall, line 316: It would be good to have someone else more familiar with this part of the code to look at this.  @profxj maybe?  Note that `lag0 = corr.size//2` is slightly different from `corr.size/2`, but I don't know if this makes a difference.

pypeit/inputfiles.py
```diff
@@ -267,7 +267,9 @@ def _read_data_file_table(lines):
         paths = []
         for l in lines:
             # Strip allows for preceding spaces before path
-            prs = l.strip().split(" ")
+            l = l.strip()
+            # Split at only the first space to allow paths that contain spaces
+            prs = [l[:l.find(' ')], l[l.find(' ')+1:]]
```
kbwestfall, line 272: Paths with spaces... ick :)

pypeit/io.py
```diff
@@ -865,8 +865,19 @@ def files_from_extension(raw_path:str,
...
+        # Loop over the list
+        retlist = []
+        for rpath in raw_path:
+            # Grab the list of files
+            dfname = os.path.join(rpath, f'*{extension}*') \
+                        if os.path.isdir(rpath) else f'{rpath}*{extension}*'
+            retlist += sorted(glob.glob(dfname))
+        return retlist
+    else:
+        msgs.error(f"Incorrect type {type(raw_path)} for raw_path (must be str or list)")
```
kbwestfall, line 883: Remove the type hint for raw_path in the function call, and update the doc string.  You could also do this using recursion instead:
```
if isinstance(raw_path, list):
    return np.concatenate([files_from_extension(p, extension=extension) for p in raw_path]).to_list()
```
And, unless you find it easier to read, the control flow means you don't need the `elif` or `else`:
```
if isinstance(raw_path, str):
   ...
   return ...

if isinstance(raw_path, list):
    ....
    return ...

msgs.error('...')
```

pypeit/metadata.py
```diff
@@ -16,7 +16,7 @@
 import numpy as np
 import yaml
 
-from astropy import table, coordinates, time, units
+from astropy import table, time
```
kbwestfall, line 19: Nice.  Thanks for cleaning up!

pypeit/metadata.py
```diff
@@ -1323,13 +1324,16 @@ def get_frame_types(self, flag_unknown=False, user=None, merge=True):
...
                 else:
-                    raise ValueError('The user-provided dictionary does not match table length.')
+                    msgs.error('The user-provided dictionary does not match table length.')
             msgs.info('Using user-provided frame types.')
             for ifile,ftypes in user.items():
                 indx = self['filename'] == ifile
-                type_bits[indx] = self.type_bitmask.turn_on(type_bits[indx], flag=ftypes.split(','))
+                try:
+                    type_bits[indx] = self.type_bitmask.turn_on(type_bits[indx], flag=ftypes.split(','))
+                except PypeItError as err:
```
kbwestfall, line 1335: Now I see why you made the change to `bitmask.py`.  I would instead make this `except ValueError as err`.

pypeit/par/pypeitpar.py
```diff
@@ -836,7 +836,7 @@ def __init__(self, spec_method=None, spec_maxshift=None, spectrum=None,
                           'Options are: None, {0}'.format(', '.join(options['spec_method']))
 
         defaults['spec_maxshift'] = 20
-        dtypes['spec_maxshift'] = [int, float]
+        dtypes['spec_maxshift'] = int
```
kbwestfall, line 839: Similar comment as above.  Need input from someone who knows the flexure code better.

pypeit/par/pypeitpar.py
```diff
@@ -3899,7 +3899,8 @@ def __init__(self, boxcar_radius=None, std_prof_nsigma=None, sn_gauss=None,
 
         defaults['std_prof_nsigma'] = 30.
         dtypes['std_prof_nsigma'] = float
-        descr['std_prof_nsigma'] = 'prof_nsigma parameter for Standard star extraction.  Prevents undesired rejection.'
+        descr['std_prof_nsigma'] = 'prof_nsigma parameter for Standard star extraction.  Prevents undesired rejection. ' \
+                                   'NOTE: Not consumed by the code at present.'
```
kbwestfall, line 3903: Nice catch.  Can we remove or comment out this parameter if it's not being used?

pypeit/scripts/identify.py
```diff
@@ -31,7 +32,7 @@ def get_parser(cls, width=None):
         parser.add_argument("--pixtol", type=float, default=0.1,
                             help="Pixel tolerance for Auto IDs")
         parser.add_argument('--test', default=False, action='store_true',
-                            help="Testing functionality, do not show plots")
+                            help=argparse.SUPPRESS)
```
kbwestfall, line 35: Nice!

pypeit/scripts/setup.py
```diff
@@ -23,7 +23,7 @@ def get_parser(cls, width=None):
         parser.add_argument('-s', '--spectrograph', default=None, type=str,
                             help='A valid spectrograph identifier: {0}'.format(
                                     ', '.join(available_spectrographs)))
-        parser.add_argument('-r', '--root', default='current working directory', type=str,
+        parser.add_argument('-r', '--root', default='current working directory', type=str, nargs='*',
```
kbwestfall, line 26: I would change this to `nargs='+'`, meaning that at least one argument must be provided if the `-r` tag is used.

A more general comment is that we should discuss as a team whether it makes sense to enable `pypeit_setup` to run simultaneously on multiple directories.  Do you have a specific use-case for this?

pypeit/spectrographs/bok_bc.py
```diff
@@ -115,12 +115,12 @@ def raw_header_cards(self):
         downstream output files for configuration identification.
 
         The list of raw data FITS keywords should be those used to populate
-        the :meth:`~pypeit.spectrograph.Spectrograph.configuration_keys`
-        or are used in :meth:`~pypeit.spectrograph.Spectrograph.config_specific_par`
+        the :meth:`~pypeit.spectrographs.spectrograph.Spectrograph.configuration_keys`
+        or are used in :meth:`~pypeit.spectrographs.spectrograph.Spectrograph.config_specific_par`
```
kbwestfall, line 119: Nice catch.  I'm not sure what `meth` does, but I would use `func` in this case.  I'm not sure they're any different...

pypeit/spectrographs/bok_bc.py
```diff
@@ -115,12 +115,12 @@ def raw_header_cards(self):
...
         for a particular spectrograph, if different from the name of the
-        PypeIt metadata keyword.
+        ``PypeIt`` metadata keyword.
```
kbwestfall, line 121: I was doing this before to make sure that the I in PypeIt had crossbars to differentiate it from a lower-case L.  But I've since changed the global font for the docs such that that is true for general text, meaning we don't need this terminal-type typesetting anymore.

tbowers7:
Replaced by #1595 .

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -176,12 +176,12 @@ def raw_header_cards(self):
...
-        the :meth:`~pypeit.spectrograph.Spectrograph.configuration_keys`
-        or are used in :meth:`~pypeit.spectrograph.Spectrograph.config_specific_par`
+        the :meth:`~pypeit.spectrographs.spectrograph.Spectrograph.configuration_keys`
+        or are used in :meth:`~pypeit.spectrographs.spectrograph.Spectrograph.config_specific_par`
         for a particular spectrograph, if different from the name of the
-        PypeIt metadata keyword.
+        ``PypeIt`` metadata keyword.
 
-        This list is used by :meth:`~pypeit.spectrograph.Spectrograph.subheader_for_spec`
+        This list is used by :meth:`~pypeit.spectrographs.spectrograph.Spectrograph.subheader_for_spec`
```
kbwestfall, line 184: Same comments for these and similar changes to the other spectrograph files.

pypeit/spectrographs/ldt_deveny.py
```diff
@@ -1,10 +1,23 @@
...
+and known as the White Spectrograph. It had a long career at the #1 36-inch
+and 84-inch telescopes there before being retired; Lowell Observatory acquired
+the spectrograph from KPNO on indefinite loan in 1998. A new CCD camera was
+built for it, and the instrument was further modified for installation on the
+72-inch Perkins telescope in 2005. Following 8 years of service there, it was
+removed in 2013 for upgrades for installation on the Lowell Discovery Telescope
+(LDT) instrument cube. It has been in service since February 2015. The
+spectrograph was designed for f/7.5 telescope optics, and new re-imaging
+optics were designed and fabricated to match the spectrograph with LDT's
+f/6.1 beam.
```
kbwestfall, line 14: Nice!

pypeit/tests/test_bitmask.py
```diff
@@ -99,13 +97,13 @@ def test_wrong_bits():
...
-#    out = image_bm.flagged(mask, 'JUNK')
     
-    with pytest.raises(ValueError):
+    # Fails with all bad flags
+    with pytest.raises(PypeItError):
         out = image_bm.flagged(mask, flag='JUNK')
 
-    with pytest.raises(ValueError):
+    # Fails with mix of good and bad flags
+    with pytest.raises(PypeItError):
```
kbwestfall, line 106: Same comment about the changes to `bitmask.py`.

kbwestfall (changes requested):
Looks good @tbowers7 !  Thanks for all the clean-up.  It would be good to have someone more familiar with the flexure correction code to weigh in on the changes there.  Other than that, mostly minor requests from me.

pypeit/inputfiles.py
tbowers7: I agree, but I came across this while working with someone else's data.

pypeit/bitmask.py
tbowers7: Reverted.

tbowers7: Reverted.

pypeit/metadata.py
tbowers7: Corrected.

pypeit/par/pypeitpar.py
tbowers7: The extraction code allows for input of this parameter, but it is not being done.  Perhaps I should also insert a TODO where this might be used?

pypeit/scripts/setup.py
tbowers7: Yes, I have a specific case of data for a single large project over multiple nights, where some calibrations are shared between nights.

I tried `nargs='+'` -- for exactly the reason you cite -- and was having weird issues with the lists being passed.  I can try to play with this more if you think the `+` is required.

pypeit/spectrographs/bok_bc.py
tbowers7: They aren't different in how it shows up in HTML.  Someone else fussed at me about using `meth` over `func` in the fluxing documentation.  Since these are class methods instead of stand-alone functions, I chose `meth`.

See, for example, https://stackoverflow.com/questions/34848253/whats-the-difference-in-behaviour-between-func-and-meth-roles-in-python-sph

pypeit/scripts/setup.py
kbwestfall: Weird.  In that case, you could instead add a check in the code that `args.root` is not an empty list.

pypeit/tests/test_bitmask.py
tbowers7: Reverted.

pypeit/io.py
tbowers7: I think I've fixed this.

pypeit/par/pypeitpar.py
tbowers7: Added a TODO in the appropriate location.

pypeit/spectrographs/bok_bc.py
tbowers7: Changed.

pypeit/io.py
tbowers7: It seems the `concatenate` trick solved whatever problem I had had with `nargs='+'` previously.

pypeit/scripts/setup.py
tbowers7: It seems the `concatenate` trick solved whatever problem I had had with `nargs='+'` previously.

PR #1596, opened 2023-05-13
spat flexure

rcooke-ast:
A TODO item in `pypeit.py` correctly points out a bug in the code... I've attached a fix here before 1.13

### DIFF
--- a/pypeit/images/rawimage.py
+++ b/pypeit/images/rawimage.py
@@ -303,8 +303,7 @@ class RawImage:
         return utils.inverse(var)
 
     def estimate_readnoise(self):
-        """
-        Estimate the readnoise (in electrons) based on the overscan regions of
+        """ Estimate the readnoise (in electrons) based on the overscan regions of
         the image.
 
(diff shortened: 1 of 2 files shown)


profxj (approved):
Nice catch and fix!

debora-pe (approved):
Nice catch @rcooke-ast !
All looks good.
I have 2 requests, if possible.

1. I found an ineffectual bug [here](https://github.com/pypeit/PypeIt/blob/f987706093abbc647cd2e77fcbad36057978a8ee/pypeit/images/rawimage.py#L751). There are 2 `return`. I don't think it was messing up anything, but it's good to remove one.
2.  Could you run some tests? It's not necessary all of them, but at least a few to see if we need to update any test.

I'm approving anyway.
Thanks!

rcooke-ast:
Thanks! I've removed the duplicate return statement, and run the unit tests (all passing, see below). I have also tested that the code reduces SOAR+Kast+KCWI data that either do or do not use this functionality, and all seems OK.

image

PR #1599, opened 2023-05-23
Update pyqt and pyside dependencies

tepickering:
This PR makes `pyqt6` and `pyside6` the defaults since `pyqt5` support is becoming problematic and `pyside2` even more so. `conda` still installs `pyqt5` when you do `conda install pyqt`, but there are not binary `pip` wheels for all platforms for it anymore. The sections in `setup.cfg` have been renamed to remove version information. The installation docs and `tox.ini` have been updated to reflect the changes to `setup.cfg`. Building fresh environments works with these changes on both Ubuntu and Apple Silicon. By "works" I mean the install works properly and `pytest` runs the local unit tests successfully. Manual spot checks of gui functionality work as far as I've tried. Mostly `pypeit_identify` and the various `ginga`-based tools. People more well-versed in possible `Qt6` compatibility issues should vet this more deeply.


profxj (approved):
thx for this!

doc/installing.rst
```diff
@@ -106,16 +106,16 @@ PypeIt has a few optional dependencies that improve and/or expand functionality.
 
       .. code-block:: console
 
-        pip install "pypeit[pyside2,scikit-image]"
+        pip install "pypeit[pyside,scikit-image]"
```
kbwestfall: Can we change this to `pyqt` instead?  It emphasizes that `pyqt` should be the default.

kbwestfall (approved):
Thanks!  Only one minor comment from me.

tepickering:
ok, tests happy now that centos has been given the heave-ho. have a peek at the docs to make sure i didn't miss anything.

PR #1602, opened 2023-06-02
Tiny typo in tellfit script

freddavies:
One letter off which causes the script to crash when run with ```objmodel = star```.


profxj (approved):
all good

rcooke-ast (approved):

PR #1604, opened 2023-06-04
Improvements to HIRES, JWST, fluxing, and coadding

jhennawi:
This PR implements the following changes. It is not finalized with regard to testing, documentation, and other loose ends, but I'm posting now so that developers can started on finalizing it. 

- Improvements to wavelength grids and masking in coadd routines. 
- Fixed a bug in echelle coadding where the wrong coadded spectra were being
used in final stacks. 
- Sensitivity function models can now be computed relative to the blaze spectrum. 
- Refactored coadding routines to work with lists to support coadding
data from different setups.
- Changes to how masking is dealt with in extraction to fix a bug in how masks
were being treated for echelle data
- Various fixes and changes required to add more support for Keck/HIRES and JWST

PR #1608, opened 2023-06-15
Datacube, further cleaning up...

rcooke-ast:
This is actually half a PR (with several TODO items that will be included in a forthcoming PR once #1604 clears). Therefore, this PR shouldn't significantly impact any of the other currently open PRs. 

The main improvements of this PR include:
* a general tidy up and improvement of the `datacube.py` routine.
* removing the reference frame correction from the `extraction.py` class (As part of this PR, we should discuss where this should best be placed).

Currently running the dev-suite...


rcooke-ast:
Dev suite results. I've since fixed the KCWI vet test failure, and I'm investigating the MMT reduction error in the meantime.
image

pypeit/extraction.py
```diff
@@ -447,7 +445,7 @@ def run(self, model_noise=None, spat_pix=None):
                 self.slits.mask[reduce_masked], 'BADEXTRACT')
 
         # Return
-        return self.skymodel, self.objmodel, self.ivarmodel, self.outmask, self.sobjs, self.waveimg, self.tilts
```
jhennawi, line 450: Is the point of returning self.slits here that they have been modified in the routine? My point is that the slits are an input argument, but seems the bitmask is being modified based on whether the extraction is successful or not, which means the input argument can change.  I think it is indeed better practice to return a new set of slits from this method so that input arguments are immutable, but I just wanted to understand the motivation. Also, can you please update the docs on the return values of the run method to reflect this change.

pypeit/pypeit.py
```diff
@@ -961,34 +960,42 @@ def extract_one(self, frames, det, sciImg, objFind, initial_sky, sobjs_obj):
             slits.mask[flagged_slits] = \
                 slits.bitmask.turn_on(slits.mask[flagged_slits], 'BADSKYSUB')
 
-        msgs.info("Extraction begins for {} on det={}".format(self.basename, det))
```
jhennawi, line 964: I like the way you restructured this. Indeed if extraction is skipped, there is no need to instantiate the Extract object just to correct the reframe.

jhennawi (approved):
Looks good @rcooke-ast. A minor question about the return values of Extract. Other than that you are ready to go after all tests pass.

pypeit/extraction.py
rcooke-ast: That's right... the bitmask is getting updated, so it really should be returned. Technically, `slits` is the input argument, and `self.slits` is modified during extraction. So, I think `self.slits` needs to be returned.

Edit: Docs updated, and TODO removed

jhennawi: Yep I agree. It is better coding practice for arguments that are modified to be immutable, and the modified object to be returned.

rcooke-ast:
All tests pass, except for the gemini_gnirs spectrum that is fixed in PR #1609 
image

profxj (approved):
all good!

PR #1609, opened 2023-06-19
Hotfix for rebinning

rcooke-ast:
This PR offers a hotfix for rebinning. It cleans up the ugly evlist version of rebinning (with a speedup - see timings below) and removes the old "rebin" function which is really just slicing, and not rebinning (i.e. the old version doesn't conserve flux).

image

### DIFF
--- a/pypeit/utils.py
+++ b/pypeit/utils.py
@@ -477,7 +477,8 @@ def rebinND(img, shape):
     Rebin a 2D image to a smaller shape. For example, if img.shape=(100,100),
     then shape=(10,10) would take the mean of the first 10x10 pixels into a
     single output pixel, then the mean of the next 10x10 pixels will be output
-    into the next pixel
+    into the next pixel. Note that img.shape must be an integer multiple of the
+    elements in the new shape.
 
     Args:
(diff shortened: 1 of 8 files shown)


jhennawi (approved):
Looks good @rcooke-ast. Thanks. I think we should probably still run tests.

pypeit/utils.py
```diff
@@ -501,20 +502,36 @@ def rebin(a, newshape):
...
 
+    Args:
+        img (`numpy.ndarray`_):
+            A 2D input image
+        shape (:obj:`tuple`):
+            The desired shape to be returned. The elements of img.shape
+            should be an integer multiple of the elements of shape.
+
+    Returns:
+        img_out (`numpy.ndarray`_): The input image rebinned to shape
```
profxj: I think the syntax is without naming the variable, at least left of the `:`

profxj (approved):
All good;  one minor doc edit

pypeit/utils.py
rcooke-ast: Quite right... I'll update this!

rcooke-ast:
All tests are passing...
image

Issue #842, opened 2020-03-06
Single element lists in pypeitpar.py


rcooke-ast:
There appears to be a bug in `pypeitpar.py`, when the value provided is a single element list. For example, in `BarPar()` the following works absolutely fine:

`defaults['locations'] = [0.0, 0.5, 1.0]`

but this doesn't work:
`defaults['locations'] = [0.5]`

Somewhere in the processing, the code is converting single element lists to floats. Not sure if this is intended, but I haven't been able to find where this bug is creeping in...

Strangely, while trying to diagnose this with `test_metadata.test_read_combid()`, it seems that BarPar gets called twice. When the arguments are all `None` the default value is read correctly as a single element array. On the second pass, the array gets converted into a float... not sure if that helps...

jhennawi:
@rcooke-ast can you touch base with @kbwestfall about this. We don't quite follow.

rcooke-ast:
OK... I think I've finally found a simple way to reproduce this error... It seems to be a problem with single element lists when calling ConfigObj. Here are some simple scripts to reproduce the bug:

**WRONG:**
```
tmp = ['[calibrations]',
       '    [[alignment]]',
       '        locations = 0.0']
cfg = ConfigObj(tmp)
print(cfg)

{'calibrations': {'alignment': {'locations': '0.0'}}}
```

**CORRECT:**
```
tmp = ['[calibrations]',
       '    [[alignment]]',
       '        locations = 0.0,1.0']
cfg = ConfigObj(tmp)
print(cfg)

{'calibrations': {'alignment': {'locations': ['0.0', '1.0']}}}
```

**CORRECT:**
```
tmp = ['[calibrations]',
       '    [[alignment]]',
       '        locations = 0.0,']
cfg = ConfigObj(tmp)
print(cfg)

{'calibrations': {'alignment': {'locations': ['0.0']}}}
```

So, I think we need to add a trailing comma when it's a list. I'll submit a PR...

rcooke-ast:
Resolved by PR #1610

PR #1610, opened 2023-06-19
Hotfix for parset

rcooke-ast:
This hotfix fixes the single element list error in the parset. For further details, see Issue #842...


profxj (changes requested):
Looks good, but can we add a new unit test to check it?

jhennawi (approved):

rcooke-ast:
> Looks good, but can we add a new unit test to check it?

Good idea... several checks added!

profxj (approved):
thx!

PR #1615, opened 2023-06-22
GNIRS IFU implementation

rcooke-ast:
This is mostly a refactor of the `gemini_gnirs.py` file to accommodate the IFU modes in the future. No on-sky IFU (LR or HR) data currently exist, so a future PR will ingest data into the dev-suite and update this file as appropriate. Thanks go to Ray Sharples for contributing to this PR as well!!

I've also included a script that can be used to convert a wavelength solution file to something that can be used in the reid archive.

Running the dev-suite to make sure nothing broke in the process...


pypeit/core/wavecal/autoid.py
```diff
@@ -2763,14 +2763,14 @@ def report_prelim(self, slit, best_patt_dict, best_final_fit):
         # Report on the best preliminary result
         if best_final_fit is None:
             msgs.warn('---------------------------------------------------' + msgs.newline() +
-                      'Preliminary report for slit {0:d}/{1:d}:'.format(slit, self._nslit-1) + msgs.newline() +
+                      'Preliminary report for slit {0:d}/{1:d}:'.format(slit+1, self._nslit) + msgs.newline() +
```
profxj, line 2766: thx for this;  always annoyed me :)

pypeit/scripts/arxiv_solution.py
```diff
@@ -0,0 +1,58 @@
+"""
```
profxj, line 1: Nice!

profxj (approved):
Great stuff!

Post tests clearing before merging.
Thx!

pypeit/spectrographs/gemini_gnirs.py
```diff
@@ -435,44 +567,250 @@ def spec_min_max(self):
...
+        Define the list of keys to be output into a standard ``PypeIt`` file.
+
+        Returns:
+            :obj:`list`: The list of keywords in the relevant
+            :class:`~pypeit.metadata.PypeItMetaData` instance to print to the
+            :ref:`pypeit_file`.
+        """
+        pypeit_keys = super().pypeit_file_keys()
+        # TODO :: Think about this... do we even need these extra file keys if skysub is done during the reduction
+        #pypeit_keys += ['calib', 'comb_id', 'bkg_id']
```
debora-pe: The keys `comb_id` and `bkg_id` are added when we use the `-b` flag with `pypeit_setup`, we don't add them by default. I believe  `calib` is now always added.
Could you add here the key `filter`? It seems it's a useful information to have in the pypeit file

debora-pe (approved):
Thanks for this work @rcooke-ast . All looks good. I just wanted to let you know a few things.
On-sky low resolution IFU data does exist. It's the first light data taken almost a year ago. You can find all of them [here](https://archive.gemini.edu/searchform/GN-2022A-ENG-132/includeengineering/not_site_monitoring/NotFail/cols=CTOWBEQ), but I also selected a subset of those and put them in the Dev Suite Google Drive (in case you want to add a test in the dev suite - I also added a pypeit file for those). 
I did try to run the reduction and it run through fine except that I had to commented out [these lines](https://github.com/pypeit/PypeIt/blob/49e1897015794d9d741df435c88cd6069a11bd75/pypeit/find_objects.py#L1230-L1232) 
```
        if np.any(global_sky_sep[skymask] == 0):
            # Cannot continue without a sky model for all slits
            msgs.error("Global sky subtraction has failed for at least one slit.")
```

because for GNIRS there are some pixels outside of the slits that have global_sky == 0, so it seems that this condition is too restrictive.
I also tried to run `pypeit_coadd_datacube` and it run through fine, but I am not sure how to check the quality of the reduction.

That's everything. Thanks!

pypeit/spectrographs/gemini_gnirs.py
rcooke-ast: Good idea - updated.

rcooke-ast:
> I just wanted to let you know a few things.

Thanks @debora-pe for the review and topping up the dev-suite!  I'll include this in the dev-suite PR...

profxj:
Feel free to merge this @rcooke-ast if you are happy

rcooke-ast:
Yep - I'll merge when dev-suite checks out!

rcooke-ast:
> I did try to run the reduction and it run through fine except that I had to commented out [these lines](https://github.com/pypeit/PypeIt/blob/49e1897015794d9d741df435c88cd6069a11bd75/pypeit/find_objects.py#L1230-L1232)

Hi @debora-pe - Thanks again for providing the test data in the dev-suite. I've fixed this error, but upon checking all of the QA, I noticed a slight issue in the illumflat determination. I think this is fine for now, but I'll look into this.

I also need to check the datacubes looks good  :-)

debora-pe:
> > I did try to run the reduction and it run through fine except that I had to commented out [these lines](https://github.com/pypeit/PypeIt/blob/49e1897015794d9d741df435c88cd6069a11bd75/pypeit/find_objects.py#L1230-L1232)
> 
> Hi @debora-pe - Thanks again for providing the test data in the dev-suite. I've fixed this error, but upon checking all of the QA, I noticed a slight issue in the illumflat determination. I think this is fine for now, but I'll look into this.
> 
> I also need to check the datacubes looks good :-)

Sounds good! Thanks @rcooke-ast !

rcooke-ast:
Tests passing...
image

PR #1621, opened 2023-07-04
Allow same bias for multiple configs

rcooke-ast:
GTC/OSIRIS+ hotfix to allow bias frames to be used independent of the grating and slit width

### DIFF
--- a/pypeit/metadata.py
+++ b/pypeit/metadata.py
@@ -871,7 +871,7 @@ class PypeItMetaData:
                     for c in range(len(self.table['setup'][indx])):
                         if cfg_key in self.table['setup'][indx][c]:
                             new_cfg_key[c] = self.table['setup'][indx][c]
-                        if self.table['setup'][indx][c] == 'None':
+                        elif self.table['setup'][indx][c] == 'None':
                             new_cfg_key[c] = cfg_key
                         else:
                             new_cfg_key[c] = self.table['setup'][indx][c] + ',{}'.format(cfg_key)
@@ -894,7 +894,7 @@ class PypeItMetaData:
                 for c in range(len(self.table['setup'][indx])):
                     if cfg_key in self.table['setup'][indx][c]:
                         new_cfg_key[c] = self.table['setup'][indx][c]
-                    if self.table['setup'][indx][c] == 'None':
+                    elif self.table['setup'][indx][c] == 'None':
                         new_cfg_key[c] = cfg_key
                     else:
                         new_cfg_key[c] = self.table['setup'][indx][c] + ',{}'.format(cfg_key)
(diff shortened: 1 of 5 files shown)


kbwestfall (changes requested):
A couple of requests from me:
  - There's a separate functionality that should ignore the configuration assignments for certain frames.  This is the `config_independent_frames` function in the spectrograph class, actually just below the function you added.  I'd like to better understand the use case and why this new `modify_config` function is needed.
  - Assuming we keep the function: 
      - Please update the docstring of the function.
      - Given there's only one if statement, you actually don't need the return statement inside it.

rcooke-ast:
@kbwestfall is correct - the original solution I proposed doesn't actually fix the underlying problem. There appears to be two bugs that I've now fixed:

(1) When configuring the setups of all frames, ignore the frames of a certain type (as specified by `configuration_independent_frames`).
(2) During the next step (assigning setups to ignored frames), we were previously just writing the last setup if the configuration independent frames had a "None" value.

pypeit/metadata.py
```diff
@@ -870,9 +866,16 @@ def set_configurations(self, configs=None, force=False, fill=None):
                 if not np.any(indx):
                     continue
                 if metakey is None:
-                    # No matching meta data defined, so just set all
-                    # the frames to this (first) configuration
-                    self.table['setup'][indx] = cfg_key
+                    # No matching meta data defined, so just set all the frames to all of the configurations
+                    new_cfg_key = np.full(len(self.table['setup'][indx]), 'None', dtype=object)
+                    for c in range(len(self.table['setup'][indx])):
+                        if cfg_key in self.table['setup'][indx][c]:
```
badpandabear, line 872: I'm not sure that this if statement is necessary. `indx` is built using `not_setup` above, so any values in `self.table['setup'][indx][c]` would have been set by  a previous iteration of the loop on 852. So it couldn't contain `cfg_key` as `_configs.keys()` should not have duplicates. But if I'm wrong and it is needed, then it should be followed by an `elif` instead of an `if` on line 874, to prevent `cfg_key` from being added twice by the `else` on lines 876-877.

badpandabear (changes requested):
Looks good except for one bit of code I have concerns about.

pypeit/metadata.py
rcooke-ast: Excellent catch @badpandabear! I had copied this from a line below. This should be an `elif` (and I fixed it in the line below). The first `if` statement catches frames that could be two types (e.g. arc and tilt), and avoids setting it twice. It's pretty rare that this `if` would be used... Thanks!

badpandabear (approved):

debora-pe (approved):
Nice catch @rcooke-ast . I checked that these changes do not interfere with the changes in the `lris_wave` branch, and all looks good!
Yes, please, add a vet test for this. Thanks!

kbwestfall (approved):

PR #1626, opened 2023-07-09
Change default behaviour of pypeit_setup

rcooke-ast:
There is currently a "conflict" between `config_independent_frames` and `configuration_keys`. This is particularly relevant for instruments with multiple arms (e.g. `X-SHOOTER`), and is due to the recent fix to allow `config_independent_frames` to be applied to multiple different setups.

The proposed solution here is that by default there are no `config_independent_frames`, but this means that every spectrograph would need to specify `bias` and `dark` to None in the `config_independent_frames`.


debora-pe:
@rcooke-ast I'm OK with this change, but I would like to know what @kbwestfall thinks about it. 
Also, could you give a specific example of the conflict, so just I can understand it better. Thanks!

rcooke-ast:
Sure, @debora-pe! It's a little subtle, though...

The conflict is between two different ways of using PypeIt. In both cases, the user dumps all data into a single Raw directory:
(1) multiple setups that have some setup-independent frames (e.g. bias, dark).
(2) multiple arms of one spectrograph that have setup-dependent frames (i.e. bias from each individual detector).

Previously, if users had multiple setups (or different spectrograph arms) and dumped all of the data in one Raw directory, running `pypeit_setup` would do the following (respectively to the numbered list above):

(1) only assign bias frames to the last setup considered
(2) appropriately assign bias frames to each individual detector/arm

A fix was put in place (PR #1621) to fix (1), but this broke (2), and bias frames were cross-assigned to different spectrograph arms. This is due to a "conflict" between `config_independent_frames` and `configuration_keys`.

This PR addresses both issues, however, it means that in some spectrographs, we will need to manually add the following code to the associated spectrograph file so that setup independent frames can be correctly assigned. In the example below, `bias` and `dark` frames will be assigned to all setups identified by `pypeit_setup`:

```
    def config_independent_frames(self):
        """
        Define frame types that are independent of the fully defined
        instrument configuration.

        This method returns a dictionary where the keys of the dictionary are
        the list of configuration-independent frame types. The value of each
        dictionary element can be set to one or more metadata keys that can
        be used to assign each frame type to a given configuration group. See
        :func:`~pypeit.metadata.PypeItMetaData.set_configurations` and how it
        interprets the dictionary values, which can be None.

        Returns:
            :obj:`dict`: Dictionary where the keys are the frame types that
            are configuration-independent and the values are the metadata
            keywords that can be used to assign the frames to a configuration
            group.
        """
        return {'bias':None, 'dark':None}
```

This is the code that I have added to `shane_kast.py` as an example in this PR (as this is required by one of the unit tests I added in PR#1621).

debora-pe:
@rcooke-ast Thank you for explaining this. It seems to me that instead of changing `config_independent_frames()` in `pypeit/spectrographs/spectrograph.py` we could just change it (or add it) in the specific spectrographs. So, in `shane_kast.py` we don't need to have `config_independent_frames()` because the default works. For other instruments that have setup-dependent frames (for examples different arms), we add `config_independent_frames()` returning `{'bias':'arm', 'dark':'arm'}` or returning `{}` if the setup defined by the `configuration_keys` works for `bias` and `dark` as well. A similar example is already in `keck_deimos.py` where we return `{'bias': 'dateobs', 'dark': 'dateobs'}`

Could that work?

rcooke-ast:
> It seems to me that instead of changing config_independent_frames() in pypeit/spectrographs/spectrograph.py we could just change it (or add it) in the specific spectrographs.

You are quite right @debora-pe - this is a much better solution. I've checked that this works for X-SHOOTER and Kast. I'll see if there's any other instruments that this might apply to.

pypeit/spectrographs/keck_hires.py
```diff
@@ -208,6 +208,26 @@ def compound_meta(self, headarr, meta_key):
...
+        :func:`~pypeit.metadata.PypeItMetaData.set_configurations` and how it
+        interprets the dictionary values, which can be None.
+
+        Returns:
+            :obj:`dict`: Dictionary where the keys are the frame types that
+            are configuration-independent and the values are the metadata
+            keywords that can be used to assign the frames to a configuration
+            group.
+        """
+        return {'bias': 'binning', 'dark': 'binning'}
```
kbwestfall: `binning` isn't one of the configuration keys for HIRES.  It doesn't need to be for its use here to associate biases and darks with specific configurations, but should we add it to `configuration_keys`?

pypeit/spectrographs/p200_dbsp.py
```diff
@@ -88,6 +88,26 @@ def compound_meta(self, headarr: List[fits.Header], meta_key: str):
...
+        :func:`~pypeit.metadata.PypeItMetaData.set_configurations` and how it
+        interprets the dictionary values, which can be None.
+
+        Returns:
+            :obj:`dict`: Dictionary where the keys are the frame types that
+            are configuration-independent and the values are the metadata
+            keywords that can be used to assign the frames to a configuration
+            group.
+        """
+        return {'bias': ['binning', 'dichroic'], 'dark': ['binning', 'dichroic']}
```
kbwestfall: Why does the dichroic matter, or this just a useful proxy for something?

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -83,6 +83,26 @@ def compound_meta(self, headarr, meta_key):
...
+        interprets the dictionary values, which can be None.
+
+        Returns:
+            :obj:`dict`: Dictionary where the keys are the frame types that
+            are configuration-independent and the values are the metadata
+            keywords that can be used to assign the frames to a configuration
+            group.
+        """
+        return {'bias': 'arm', 'dark': 'arm'}
+
```
kbwestfall, line 105: I think it makes more sense for this to return None.  The only configuration key is arm meaning that the biases and dark shouldn't be treated differently from any other frame type.  You still need this method to overwrite the default in Spectrograph, but here you can just `return None` (I think).

kbwestfall (changes requested):
Looking through all the spectrographs that you changed, it does feel like we could add binning as a default configuration key, and set the default config_independent_frames to use that for the biases and darks.  But I don't actually know if that means you'll have to edit fewer files (particularly given the edits you've already made)!  I caught a couple of other minor issues along the way.

pypeit/spectrographs/keck_hires.py
rcooke-ast: Yes, I believe so... I have added it.

pypeit/spectrographs/p200_dbsp.py
rcooke-ast: I just noticed that this was in the `configuration keys`, and given that this is a double arm spectrograph, I wonder what is determining the red vs blue camera, in this case 🤷‍♂️ I've removed the dichroic, in any case...

pypeit/spectrographs/vlt_xshooter.py
rcooke-ast: Based on the X-SHOOTER data that I have, this can't be `None` (incidentally, this is how I first noticed the issue!), and returning `arm` solves the issue in this case.

rcooke-ast (commented):
Thanks @kbwestfall!  All comments addressed, I think.

pypeit/spectrographs/keck_hires.py
debora-pe: I guess you can remove `config_independent_frames` for hires since it's equal to the default.

pypeit/spectrographs/vlt_xshooter.py
debora-pe: I agree with @kbwestfall that, since `arm` is the only key in `configuration_keys`, putting it here is kind of redundant and also confusing, since the frames that go through this method are the one that are treated differently from the rest. I think `return {}` should work here, because if the dictionary is empty it means that there are no frames that are treated differently.

debora-pe (approved):
Nice work on this @rcooke-ast . I have 2 minor comments, but I am approving anyway.

pypeit/spectrographs/vlt_xshooter.py
rcooke-ast: Yes of course...  🤦‍♂️  I have updated this - thanks for the suggestion!

pypeit/spectrographs/keck_hires.py
rcooke-ast: Yes - that's a good point - done!

kbwestfall (approved):
Thanks!

Issue #1622, opened 2023-07-04
pypeit_chk_tilts fails for X-shooter arcs because of resizing issue


jhennawi:
The following failure is occurring for binned X-shooter arcs. For x-shooter we analyze arcs of different shapes because of the way ESO takes calibrations. This is dealt with by methods in core.arc.resize_slits2arc and core.arc.aresize_mask2arc in e.g. wavecalib.py and wavetilts.py. I think the problem is the show_tilts method is not executing these steps. 

(pypeit) dhcp181:VIS_2x2 joe$ pypeit_chk_tilts Calibrations/Tilts_A_0_DET01.fits 
[INFO]    :: datamodel.py 1533 from_file() - Loading WaveTilts from /Users/joe/python/PypeIt-development-suite/REDUX_OUT/vlt_xshooter/VIS_2x2/Calibrations/Tilts_A_0_DET01.fits
[INFO]    :: datamodel.py 1533 from_file() - Loading SlitTraceSet from /Users/joe/python/PypeIt-development-suite/REDUX_OUT/vlt_xshooter/VIS_2x2/Calibrations/Slits_A_0_DET01.fits.gz
[INFO]    :: datamodel.py 1533 from_file() - Loading TiltImage from /Users/joe/python/PypeIt-development-suite/REDUX_OUT/vlt_xshooter/VIS_2x2/Calibrations/Tiltimg_A_0_DET01.fits
Traceback (most recent call last):
  File "/Users/joe/miniconda3/envs/pypeit/bin/pypeit_chk_tilts", line 33, in <module>
    sys.exit(load_entry_point('pypeit', 'console_scripts', 'pypeit_chk_tilts')())
  File "/Users/joe/python/PypeIt/pypeit/scripts/scriptbase.py", line 112, in entry_point
    cls.main(cls.parse_args())
  File "/Users/joe/python/PypeIt/pypeit/scripts/chk_tilts.py", line 39, in main
    tilts.show(in_ginga=np.logical_not(pargs.mpl), show_traces=pargs.show_traces)
  File "/Users/joe/python/PypeIt/pypeit/wavetilts.py", line 204, in show
    tilt_img = tilt_img_dict.image * (slitmask > -1) if slits is not None else tilt_img_dict.image
ValueError: operands could not be broadcast together with shapes (4000,2048) (2000,1024)

Issue #1617, opened 2023-06-29
Object finding QA plots are not being written when no spec1d file is generated


jhennawi:
Going through the NIRES tutorial that is here: 

https://pypeit.readthedocs.io/en/release/tutorials/nires_howto.html

I found a bug in PypeIt on both release and develop. Namely, the object finding QA for 

pos_s190519_0067

i.e. the QA associated with the file s190519_0067.fits is never written to disk. Note that the spec1d is also not written since no objects were found. This is problematic since one wants to look at the QA to see how the object finding threshold should be adjusted to identify the object. Object finding QA should always be written to disk even if no objects are found.

PR #1630, opened 2023-07-13
Issues #1622 & #1617

debora-pe:
Addressed issues:
- https://github.com/pypeit/PypeIt/issues/1622
- https://github.com/pypeit/PypeIt/issues/1617


profxj (approved):
nice nice!

pypeit/wavetilts.py
```diff
@@ -188,34 +188,42 @@ def show(self, waveimg=None, wcs_match=True, in_ginga=True, show_traces=False):
...
-                tilts = self.fit2tiltimg(slitmask, flexure=self.spat_flexure)
+                tilts = self.fit2tiltimg(_slitmask, flexure=self.spat_flexure)
                 waveimg = wv_calib.build_waveimg(tilts, slits, spat_flexure=self.spat_flexure)
             else:
                 msgs.warn('Could not load Wave image to show with tilts image.')
 
         # Show
+        # tilt image
+        # resize slitmask to match tilt_img_dict.image.shape
+        slitmask = arc.resize_mask2arc(tilt_img_dict.image.shape, _slitmask)
```
kbwestfall: I would move this line to after the current line 207 (`gpm = mask == 0`) and adjust line 217 if necessary (i.e., should it use `_slitmask` or `slitmask`).  Otherwise, I think you run the risk of the slits file not existing and `_slitmask` being undefined when you hit this point in the code.

Or you could put this line and the next one in to an if/else block:
```
if slits is None:
    tilt_img = tilt_img_dict.image
else:
    slitmask = arc.resize_mask2arc(tilt_img_dict.image.shape, _slitmask)
    tilt_img = tilt_img_dict.image * (slitmask > -1)
```

kbwestfall (changes requested):
Looks good!  I have one request to look at a possible bug.

pypeit/wavetilts.py
debora-pe: Nice catch @kbwestfall! Thank you! I fixed it.

pypeit/wavetilts.py
```diff
@@ -188,34 +188,45 @@ def show(self, waveimg=None, wcs_match=True, in_ginga=True, show_traces=False):
...
         # get slits
+        slitmask = None
         if (Path(self.calib_dir).resolve() / self.slits_filename).exists():
             slits = slittrace.SlitTraceSet.from_file(Path(self.calib_dir).resolve() / self.slits_filename)
-            slitmask = slits.slit_img(initial=True, flexure=self.spat_flexure)
-            left, right, mask = slits.select_edges(flexure=self.spat_flexure)
-            gpm = mask == 0
+            _slitmask = slits.slit_img(initial=True, flexure=self.spat_flexure)
+            if tilt_img_dict.image.shape != _slitmask.shape:
+                msgs.warn('Tilt image and other calibration frames have different shapes. '
```
jhennawi: I don't think this is correct. Why are you throwing a warning here? The tilt_img shape can differ from the _slitmask shape for instruments where that is often the case (i.e. X-shooter where they don't take binned arcs), and it should not trigger a warning because nothing is going wrong. Furthermore, the resizing routine will just return the _slitmask if the shapes are the same so the if else is not needed.

jhennawi (changes requested):
Minor change requested based on functionality of arc_resizing routines.

pypeit/wavetilts.py
debora-pe: @jhennawi the warning here is about the slit traces and the wavelength image. arc resize routine allow me to resize  `_slitmask` but not `slits` (and therefore `waveimg`). This is why there is an if/else. If the sizes are different I don't want to use `slits` (because I cannot resize it). And if I don't use `slits` I cannot create the `waveimg`. Hence the warning, I cannot load those 2 things. I can remove the part where it mention the difference in shapes, but I prefer to leave the part where I warn that slits and waveimg cannot be loaded.

kbwestfall (approved):
Well done!

pypeit/wavetilts.py
```diff
@@ -188,34 +188,45 @@ def show(self, waveimg=None, wcs_match=True, in_ginga=True, show_traces=False):
...
+                          'Cannot load slits nor Wave image.')
+                # resize slitmask to match tilt_img_dict.image.shape
+                slitmask = arc.resize_mask2arc(tilt_img_dict.image.shape, _slitmask)
+                # but, still, we cannot really use slits, so we set it to None
+                slits = None
+            else:
```
jhennawi, line 209: There is a resizing routine for the slits as well. Again, you should not be throwing any warnings here because the slits being a different shape than the arc/tilt image is a common occurrence for X-shooter.  See the __init__.py method in wavetilts.py. There is a method called arc.resize_slits2arc.  

What you should be doing here is getting rid of the if/else because the routines are designed to work no matter what the shapes are as long as they are integer multiples of each other. 

_slitmask = slits.slit_img(initial=True, flexure=self.spat_flexure)
slitmask = arc.resize_mask2arc(tilt_img_dict.image.shape, _slitmask)
_left, _right, _mask = slits.select_edges(flexure=self.spat_flexure)
left = arc.resize_slits2arc(tilt_img_dict.image.shape, _slitmask.shape, _left)
right = left = arc.resize_slits2arc(tilt_img_dict.image.shape, _slitmask.shape, _right)

Then I'm a bit unclear on what you are doing with the mask, but you can I think use the arc.resize_mask2arc to reshape that as well. No if/then is needed and no warnings should be thrown. The resizing routines return the input image if the shapes are the same.

pypeit/wavetilts.py
jhennawi: There is a resizing routine for the slits as well. Again, you should not be throwing any warnings here because the slits being a different shape than the arc/tilt image is a common occurrence for X-shooter.  See the __init__.py method in wavetilts.py. There is a method called arc.resize_slits2arc.  

What you should be doing here is getting rid of the if/else because the routines are designed to work no matter what the shapes are as long as they are integer multiples of each other. 

_slitmask = slits.slit_img(initial=True, flexure=self.spat_flexure)
slitmask = arc.resize_mask2arc(tilt_img_dict.image.shape, _slitmask)
_left, _right, _mask = slits.select_edges(flexure=self.spat_flexure)
left = arc.resize_slits2arc(tilt_img_dict.image.shape, _slitmask.shape, _left)
right = left = arc.resize_slits2arc(tilt_img_dict.image.shape, _slitmask.shape, _right)

Then I'm a bit unclear on what you are doing with the mask, but you can I think use the arc.resize_mask2arc to reshape that as well. No if/then is needed and no warnings should be thrown. The resizing routines return the input image if the shapes are the same.

debora-pe: Done.

pypeit/wavetilts.py
```diff
@@ -188,34 +188,41 @@ def show(self, waveimg=None, wcs_match=True, in_ginga=True, show_traces=False):
...
+
         # get slits
+        slitmask = None
         if (Path(self.calib_dir).resolve() / self.slits_filename).exists():
             slits = slittrace.SlitTraceSet.from_file(Path(self.calib_dir).resolve() / self.slits_filename)
-            slitmask = slits.slit_img(initial=True, flexure=self.spat_flexure)
-            left, right, mask = slits.select_edges(flexure=self.spat_flexure)
-            gpm = mask == 0
+            _slitmask = slits.slit_img(initial=True, flexure=self.spat_flexure)
+            _left, _right, _mask = slits.select_edges(flexure=self.spat_flexure)
```
jhennawi, line 202: I think you may need to resize _mask as well. This should probably crash on x-shooter. Did you check. You can resize mask with the arc.resize_mask2arc

debora-pe: @jhennawi yes I checked with x-shooter and some other instrument and it works fine. `_mask` does not need to be resized because it's a 1D array with a size equal to the number of slits, i.e., it has only one value per slit.

jhennawi: Got it. I was confused.

jhennawi (approved):
Thanks @debora-pe. Approving!

PR #1633, opened 2023-07-16
Air2vac correction for XShooter standards

freddavies:
Finally addressing (at least one part of) https://github.com/pypeit/PypeIt/issues/941, as the issue has become even more obvious for the UVB arm of XShooter. 

Might be worth checking whether any of the other standard files also need this fix...

### DIFF
--- a/pypeit/core/flux_calib.py
+++ b/pypeit/core/flux_calib.py
@@ -27,6 +27,7 @@ from pypeit import io
 from pypeit.wavemodel import conv2res
 from pypeit.core.wavecal import wvutils
 from pypeit.core import fitting
+from pypeit.core import wave
 from pypeit import data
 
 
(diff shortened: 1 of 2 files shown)


jhennawi (changes requested):
Hi @freddavies can you please modify the README file at: 

~/PypeIt/pypeit/data/standards/xshooter

To explicitly note that the data stored in the files is in air. Also, does the reference say anything about it?

profxj (approved):
nice catch

freddavies:
README updated.

> Also, does the reference say anything about it?

Unfortunately it does not say anything explicitly, but one can guess it from the fact that they are using the ESO pipeline to reduce the data, which spits out air wavelengths by default. Otherwise the only other evidence is the much better agreement between the model and the (PypeIt-reduced) data in the absorption line cores...

jhennawi (approved):

PR #1635, opened 2023-07-19
Fix coadd2d in sensfunc_blaze_jwst_lists branch

debora-pe:
As titled

Vet test is here:https://github.com/pypeit/PypeIt-development-suite/pull/273


jhennawi (approved):
Thanks @debora-pe!

PR #1644, opened 2023-08-03
Hotfix for view_fits

jhennawi:
Documentation fixes for pypeit_view_fits and a minor doc change elsewhere in find objects parset. 

It appears that view_fits is still buggy, i.e. 

pypeit_view_fits gemini_gmos_north_ham N20190205S0035.fits --det mosaic

Only displays one detector. I'm not sure I follow what is going on here, but it seems that an unprocessed mosaic should be display.


kbwestfall (approved):
Looks good!

profxj (approved):
all good

PR #1647, opened 2023-08-08
PCA method for telluric corrections

freddavies:
Introduces a new method for computing telluric corrections that uses a PCA decomposition of a large suite of telluric absorption models. The specifics will (probably, someday) be written up in a paper, but long story short, I ran a latin hypercube of 2000 telluric models with varying atmospheric parameters (pressure, temperature, humidity, airmass, O2, O3, CO2) for the `MaunaKea`, `Paranal`, `MountGraham`, `LasCampanas`, `Palomar`, and `Lick` observatories, which span a wide range of latitudes and elevations. I then computed a PCA decomposition of the arcsinh of the atmospheric optical depth¹ from all 12000 models. The resulting "grid" of UV-to-NIR telluric spectra then consists of just 10 PCA components, with a filesize up to a thousand times smaller than the old telluric grids and with broad applicability to all observatories (as far as I can tell).

The corresponding "TellPCA" files have been uploaded to the telluric folder on the development suite drive, at varying spectral samplings corresponding to R=10000, 15000, and 25000; the re-sampling was done in a flux-conserving way prior to the PCA decomposition in arcsinh(tau) space. For testing on Keck/HIRES I have also included a beta version of an R=60000 model in the optical which is only derived from `MaunaKea` and `Paranal` models. 

My testing thus far has shown that this method provides comparably good telluric corrections to the old approach with a default of 4 PCA components used in the fit, with a similar runtime. The user can increase this number, given by a new parameter `ntell`, up to 10 if they desire a more accurate fit, though it usually makes little difference in practice (note that I have not tested this carefully at arbitrarily high S/N).

Due to its relative ease of use, this new PCA method has been set as the default telluric method. The original grid-based method can still be used by setting `teltype = grid` in the telluric parameter file, both to maintain reproducibility and to allow users to still use "actual" telluric absorption spectra if they are not comfortable with the pseudo-telluric spectra generated by the PCA method.

While I have done some testing on data from various instruments, before merging I think we need to do a more comprehensive comparison between the two methods to make sure that we are not compromising the quality of the telluric corrections. We could also have a meta-discussion as to whether there are unforeseen risks to using telluric corrections that are not strictly physical models with a particular set of parameters...

\
\
¹This may seem needlessly convoluted, but I swear there is a good reason! A decomposition in transmission space inevitably involves trimming at 0 and 1, where the former gives rise to nasty infinities in the correction. Optical depth space would otherwise work, except that the PCA decomposition puts too much weight in the highest optical depth pixels which have transmissions of ~0. Taking the log of the optical depth, conversely, puts way too much emphasis on regions with extremely *low* opacity where the transmission is ~1. It turns out that the arcsinh of the optical depth is a good middle ground, reducing the dynamic range at high values while maintaining linearity at small values.


CHANGES.rst
```diff
@@ -40,7 +40,9 @@
 - HIRES wavelength solution improvements galor
 - Added `redo_slits` option
 - Refactored ``load_line_lists()`` yet again!
+- Introduced PCA method for telluric corrections
 
+>>>>>>> develop
```
jhennawi: Looks like you generated conflicts when you merged. Please remove the >>>>>> develop.

pypeit/core/telluric.py
```diff
@@ -135,11 +134,69 @@ def pca_eval(theta,pca_dict):
...
+        wave_max (:obj:`float`):
+            Maximum wavelength at which the grid is desired.
+        pad_frac (:obj:`float`):
+            Percentage padding to be added to the grid boundaries if
+            ``wave_min`` or ``wave_max`` are input; ignored otherwise. The
+            resulting grid will extend from ``(1.0 - pad_frac)*wave_min`` to
+            ``(1.0 + pad_frac)*wave_max``.
+
+    Returns:
+        :obj:`dict`: Dictionary containing the telluric PCA components.
```
jhennawi, line 163: Please document the elements of the dict.

pypeit/core/telluric.py
```diff
@@ -135,11 +134,69 @@ def pca_eval(theta,pca_dict):
...
+    wave_grid = wave_grid_full[ind_lower:ind_upper]
+    pca_comp_grid = pca_comp_full[:,ind_lower:ind_upper]
+
+    dwave, dloglam, resln_guess, pix_per_sigma = wvutils.get_sampling(wave_grid)
+    tell_pad_pix = int(np.ceil(10.0 * pix_per_sigma))
+
+    return dict(wave_grid=wave_grid, dloglam=dloglam, resln_guess=resln_guess,
+            pix_per_sigma=pix_per_sigma, tell_pad_pix=tell_pad_pix, ncomp_tell_pca=ncomp,
+            tell_pca=pca_comp_grid, bounds_tell_pca=bounds, coefs_tell_pca=model_coefs)
+
```
jhennawi, line 196: Can you make it clear that the read_telluric grid is no longer the preferred approach as compared to the pca approach in the docs to read_telluric_grid

pypeit/core/telluric.py
```diff
@@ -321,7 +379,87 @@ def shift_telluric(tell_model, loglam, dloglam, shift, stretch):
...
+       
+       2. transformation of telluric PCA model from arsinh(tau) to transmission
+
+       3. convolution of the atmosphere model to the resolution set by
+          the spectral resolution.
+
+       4. (Optional) shift and stretch the telluric model.
+
+    Args:
+        theta_tell (`numpy.ndarray`_):
```
jhennawi, line 387: Please specify the shape of the arrays that are input and returned in the docs.

pypeit/core/telluric.py
```diff
@@ -321,7 +379,87 @@ def shift_telluric(tell_model, loglam, dloglam, shift, stretch):
...
+            Vector with the telluric PCA coefficients
+        tell_dict (:obj:`dict`):
```
jhennawi, line 392: Can you specify which parameters fill which indices in the theta_tell in the docs please.

pypeit/core/telluric.py
```diff
@@ -454,12 +598,13 @@ def tellfit(flux, thismask, arg_dict, init_from_last=None):
...
 
-                theta_obj = theta[:-7]
-                theta_tell = theta[-7:]
-
-            The telluric model theta_tell is currently hard wired to be six dimensional::
+                theta_obj = theta[:-(ntell+3)]
+                theta_tell = theta[-(ntell+3):]
 
-                pressure, temperature, humidity, airmass, resln, shift = theta_tell
+            The telluric model theta_tell includes a either user-specified
```
jhennawi: Can you be more specific in the docs and indicate what parameters live at which indices in the theta vector.

pypeit/core/telluric.py
```diff
@@ -1408,11 +1563,11 @@ def create_bal_mask(wave, bal_wv_min_max):
 
 
 
-def qso_telluric(spec1dfile, telgridfile, pca_file, z_qso, telloutfile, outfile, npca=8,
-                 pca_lower=1220.0, pca_upper=3100.0, bal_wv_min_max=None, delta_zqso=0.1,
+def qso_telluric(spec1dfile, telgridfile, teltype, pca_file, z_qso, telloutfile, outfile, npca=8,
```
jhennawi: I think teltype should be optional and we should simply default to the 'pca'

pypeit/core/telluric.py
```diff
@@ -1554,30 +1714,30 @@ def qso_telluric(spec1dfile, telgridfile, pca_file, z_qso, telloutfile, outfile,
...
+                        spectrograph=header['PYP_SPEC'], telluric=telluric, obj_model=qso_pca_model,
                         header=header, ex_value='OPT', overwrite=True)
 
     return TelObj
 
-def star_telluric(spec1dfile, telgridfile, telloutfile, outfile, star_type=None, star_mag=None,
-                  star_ra=None, star_dec=None, func='legendre', model='exp', polyorder=5,
-                  mask_hydrogen_lines=True, mask_helium_lines=False, hydrogen_mask_wid=10.,
-                  delta_coeff_bounds=(-20.0, 20.0),
+def star_telluric(spec1dfile, telgridfile, teltype, telloutfile, outfile, star_type=None,
```
jhennawi: Same comment here, make teltype optional.

pypeit/core/telluric.py
```diff
@@ -1678,12 +1839,12 @@ def star_telluric(spec1dfile, telgridfile, telloutfile, outfile, star_type=None,
 
     return TelObj
 
-def poly_telluric(spec1dfile, telgridfile, telloutfile, outfile, z_obj=0.0, func='legendre',
-                  model='exp', polyorder=3, fit_wv_min_max=None, mask_lyman_a=True,
+def poly_telluric(spec1dfile, telgridfile, teltype, telloutfile, outfile, z_obj=0.0, func='legendre',
```
jhennawi: Likewise here, teltype optional.

pypeit/core/telluric.py
```diff
@@ -1248,10 +1398,14 @@ def sensfunc_telluric(wave, counts, counts_ivar, counts_mask, exptime, airmass,
...
         File containing grid of HITRAN atmosphere models; see
         :class:`~pypeit.par.pypeitpar.TelluricPar`.
+    teltype : :obj:`str`
+        Method for evaluating telluric models, either `PCA` or `grid`.
     ech_orders : `numpy.ndarray`_, shape is (norders,), optional
         If passed, provides the true order numbers for the spectra provided.
     polyorder : :obj:`int`, optional, default = 8
         Polynomial order for the sensitivity function fit.
+    ntell : :obj:`int`, optional, default = 4
+        Number of telluric model parameters (must be 4 for teltype=`grid`, or <=10 for teltype=`PCA`)
```
jhennawi: I don't like this number being a free parameter that governs both. I would say just call this parameter telluric_npca or adapt it as such. Then teltype determines whether we use the fixed grid with X parameters, versus the PCA which uses telluric_npca telluric parameters.

pypeit/core/telluric.py
```diff
@@ -2203,9 +2385,9 @@ def empty_model_table(norders, nspec, n_obj_par=0):
             table.Column(name='WAVE_MAX', dtype=float, length=norders,
                          description='Maximum wavelength included in the fit')])
 
-    def __init__(self, wave, flux, ivar, gpm, telgridfile, obj_params, init_obj_model,
-                 eval_obj_model, log10_blaze_function=None, ech_orders=None, sn_clip=30.0, airmass_guess=1.5,
-                 resln_guess=None, resln_frac_bounds=(0.5, 1.5), pix_shift_bounds=(-5.0, 5.0),
+    def __init__(self, wave, flux, ivar, gpm, telgridfile, teltype, obj_params, init_obj_model,
```
jhennawi: please make teltype optional as per above.

pypeit/core/telluric.py
```diff
@@ -2263,8 +2447,23 @@ def __init__(self, wave, flux, ivar, gpm, telgridfile, obj_params, init_obj_mode
                 wave, flux, ivar, gpm)
         # 3) Read the telluric grid and initalize associated parameters
         wv_gpm = self.wave_in_arr > 1.0
-        self.tell_dict = read_telluric_grid(self.telgrid, wave_min=self.wave_in_arr[wv_gpm].min(),
-                                            wave_max=self.wave_in_arr[wv_gpm].max())
+        if teltype == 'pca' or teltype == 'PCA':
```
jhennawi: Can you instead wrap the reading functions into general readers and evaluators that take teletype as an argument. Then we could get rid of this relatively ugly if then statement.

pypeit/core/telluric.py
```diff
@@ -2263,8 +2447,23 @@ def __init__(self, wave, flux, ivar, gpm, telgridfile, obj_params, init_obj_mode
                 wave, flux, ivar, gpm)
         # 3) Read the telluric grid and initalize associated parameters
         wv_gpm = self.wave_in_arr > 1.0
```
jhennawi, line 2402: Use upper to convert the teltype string to upper, i.e. if upper(teltype) == 'PCA'

pypeit/par/pypeitpar.py
```diff
@@ -2102,6 +2102,19 @@ def __init__(self, telgridfile=None, sn_clip=None, resln_guess=None, resln_frac_
                                'must be downloaded from the GoogleDrive and installed in your PypeIt installation via ' \
                                'the pypeit_install_telluric script. NOTE: This parameter no longer includes the full ' \
                                'pathname to the Telluric Grid file, but is just the filename of the grid itself.'
+        
+        defaults['ntell'] = 4
```
jhennawi: See my point about renaming and redefining this parameter.

jhennawi (changes requested):
Minor changes requested. @freddavies I have a more technical question that I'll ask in the user slack.

CHANGES.rst
freddavies: Done. For some reason merging in CHANGES always gives me strange conflicts...

pypeit/core/telluric.py
freddavies: Done, and added similar documentation to `read_telluric_grid`

pypeit/core/telluric.py
freddavies: Done, I hope that was comprehensive enough...

pypeit/core/telluric.py
freddavies: I added a bit here but was somewhat more explicit in the docs to `tellfit`.

pypeit/core/telluric.py
freddavies: Made that change in all the `*_telluric` functions

pypeit/core/telluric.py
freddavies: Done, although the returned shape is complicated.

pypeit/core/telluric.py
freddavies: I wrapped the evaluators into one function since that makes sense, but left the readers separate, as they output dramatically different dicts

pypeit/core/telluric.py
freddavies: This took a bit more wrangling than you might have otherwise expected; there are many routines that need to know the shape of the parameter vector, and so then all those places also have to know the teltype. But it's done.

CHANGES.rst
```diff
@@ -52,7 +52,7 @@
     - Added a function ``check_spectrograph()`` (currently only defined for LRIS),
       that checks (during ``pypeit_setup``) if the selected spectrograph is the
       corrected one for the data used. 
-
+- Introduced PCA method for telluric corrections
```
rcooke-ast: Subsequent to this PR, the use of CHANGES has been deprecated, so I'll make a quick note of that here...

pypeit/par/pypeitpar.py
```diff
@@ -2107,6 +2107,17 @@ def __init__(self, telgridfile=None, sn_clip=None, resln_guess=None, resln_frac_
                                'must be downloaded from the GoogleDrive and installed in your PypeIt installation via ' \
                                'the pypeit_install_telluric script. NOTE: This parameter no longer includes the full ' \
                                'pathname to the Telluric Grid file, but is just the filename of the grid itself.'
+        
+        defaults['tell_npca'] = 4
+        dtypes['tell_npca'] = int
+        descr['tell_npca'] = 'Number of telluric PCA components used. Can be set to any number from 1 to 10.'
```
rcooke-ast, line 2241: This parameter should also be added to `validate()` below (on line ~2372).

pypeit/par/pypeitpar.py
```diff
@@ -2107,6 +2107,17 @@ def __init__(self, telgridfile=None, sn_clip=None, resln_guess=None, resln_frac_
...
+        
+        defaults['teltype'] = 'PCA'
```
rcooke-ast: Likewise, it might be good to add this to the `validate()` method

pypeit/core/telluric.py
```diff
@@ -39,7 +39,7 @@
 # TODO These codes should probably be in a separate qso_pca module. Also pickle functionality needs to be removed.
 # The prior is not used (that was the reason for pickling), so the components could be stored in fits format.
 # npca is not actually required here.
-def init_pca(filename,wave_grid,redshift, npca):
+def init_qso_pca(filename,wave_grid,redshift, npca):
```
rcooke-ast: For consistency with the other methods, should the prefix start with `qso_`, i.e. `qso_init_pca()`

pypeit/core/telluric.py
```diff
@@ -135,22 +134,73 @@ def pca_eval(theta,pca_dict):
...
+    """
+    Reads in the telluric PCA components from a file.
+
+    Optionally, this method also trims wavelength to be in within ``wave_min``
+    and ``wave_max`` and pads the data (see ``pad_frac``).
+
+    Args:
+        filename (:obj:`str`):
+            Telluric PCA filename
+        wave_min (:obj:`float`):
```
rcooke-ast: Add optional: `wave_min (:obj:'float', optional)`, and likewise for the other optional arguments

pypeit/core/telluric.py
```diff
@@ -135,22 +134,73 @@ def pca_eval(theta,pca_dict):
...
+    # load_telluric_grid() takes care of path and existance check
+    hdul = data.load_telluric_grid(filename)
+    wave_grid_full = hdul[1].data
+    pca_comp_full = hdul[0].data
+    ncomp = hdul[0].header['NCOMP']
+    bounds = hdul[2].data
+    model_coefs = hdul[3].data
+
+    ind_lower = np.argmin(np.abs(wave_grid_full - (1.0 - pad_frac)*wave_min)) \
+                    if wave_min is not None else 0
```
rcooke-ast, line 183: Should 0 instead be the minimum valid wavelength of the PCA routine?

pypeit/core/telluric.py
```diff
@@ -135,22 +134,73 @@ def pca_eval(theta,pca_dict):
...
+    ind_upper = np.argmin(np.abs(wave_grid_full - (1.0 + pad_frac)*wave_max)) \
+                    if wave_max is not None else nspec_full
```
rcooke-ast, line 185: Is `nspec_full` defined? Should this instead be the maximum valid wavelength of the models?

pypeit/core/telluric.py
```diff
@@ -135,22 +134,73 @@ def pca_eval(theta,pca_dict):
...
     Optionally, this method also trims the grid to be in within ``wave_min``
     and ``wave_max`` and pads the data (see ``pad_frac``).
 
-    .. todo::
-        List and describe the contents of the dictionary in the return
-        description.
-
     Args:
         filename (:obj:`str`):
            Telluric grid filename
```
rcooke-ast, line 207: Although this is not due to your PR, could you please add the missing `optional` in the docstring for optional variables?

pypeit/core/telluric.py
```diff
@@ -166,16 +216,15 @@ def read_telluric_grid(filename, wave_min=None, wave_max=None, pad_frac=0.10):
...
-            - tell_pad_pix=
-            - pressure_grid=
-            - temp_grid=
-            - h2o_grid=
-            - airmass_grid=
-            - tell_grid=
+            - wave_grid: Telluric model wavelength grid
+            - dloglam: Wavelength sampling of telluric model
+            - tell_pad_pix: Number of pixels to pad at edges for convolution
+            - pressure_grid: Atmospheric pressure values in telluric grid [mbar]
```
rcooke-ast, line 223: I flag this here about the default units used are different from the other units used throughout the code. I don't think this is an issue for this PR, since it's clearly documented.

EDIT : I've made changes in PR #1669 to keep the same units as you are using here.

pypeit/core/telluric.py
```diff
@@ -320,58 +362,66 @@ def shift_telluric(tell_model, loglam, dloglam, shift, stretch):
...
-
+    teltype = tell_dict['teltype']
+    # FD: Currently assumes that shift and stretch are on.
+    # TODO: Make this work without shift and stretch.
+    if teltype == 'PCA':
+        ntell = ntheta-3
+    elif teltype == 'grid':
+        ntell = 4
+    else:
+        msgs.error("Unsupported teltype, must be 'PCA' or 'grid'")
```
rcooke-ast: I think it's fine to catch this here, but I wonder if a better approach is to use the `validate()` method in `pypeitpar`, and then change the error message to read:
`msgs.error(f"Unsupported teltype: {teltype}")`
So that the options don't need to be updated here if there is ever a change to the allowed methods in `pypeitpar` (i.e. everything is then self-contained). In reality, this line should never be reached if it's in the `pypeitpar`, and the only reason it would be hit if a developer was adding a new option in the future.

pypeit/core/telluric.py
```diff
@@ -382,16 +432,34 @@ def eval_telluric(theta_tell, tell_dict, ind_lower=None, ind_upper=None):
...
+        tellmodel_hires = np.sinh(tellmodel_hires)
+        # It should generally be very rare, but trim negative optical depths here just in case.
+        clip = tellmodel_hires < 0
+        tellmodel_hires[clip] = 0
+        tellmodel_hires = np.exp(-tellmodel_hires)
+    elif teltype == 'grid':
+        # Interpolate within the telluric grid
+        tellmodel_hires = interp_telluric_grid(theta_tell[:ntell], tell_dict)
+    else:
+        msgs.error("Unsupported teltype, must be 'PCA' or 'grid'")
```
rcooke-ast: Same comment as above.

pypeit/core/telluric.py
```diff
@@ -419,14 +487,26 @@ def tellfit_chi2(theta, flux, thismask, arg_dict):
...
+    teltype = arg_dict['tell_dict']['teltype']
+
+    # TODO: make this work without shift and stretch?
+    # Number of telluric model parameters, plus shift, stretch, and resolution
+    if teltype == 'PCA':
+        nfit = arg_dict['tell_npca']+3
+    elif teltype == 'grid':
+        nfit = 4+3
+    else:
+        msgs.error("Unsupported teltype, must be 'PCA' or 'grid'")
```
rcooke-ast: Same comment as above

pypeit/core/telluric.py
```diff
@@ -452,14 +532,37 @@ def tellfit(flux, thismask, arg_dict, init_from_last=None):
             Parameter vector for the object + telluric model.
```
rcooke-ast: There doesn't appear to be a `theta` arg anymore?

pypeit/core/telluric.py
```diff
@@ -477,8 +580,8 @@ def tellfit(flux, thismask, arg_dict, init_from_last=None):
                 - ``arg_dict['flux_ivar']``:  Inverse variance for the
                   flux array
                 - ``arg_dict['tell_dict']``: Dictionary containing the
-                  telluric grid and its parameters read in by
-                  read_telluric_grid
+                  telluric model and its parameters read in by
+                  read_telluric_pca or read_telluric_grid
                 - ``arg_dict['ind_lower']``: Lower index into the
                   telluric model wave_grid to trim down the telluric
                   model.
```
rcooke-ast, line 578: Add `, optional` to `init_from_last` on line 597.

Also, do we still need the `**kwargs_opt`?

pypeit/core/telluric.py
```diff
@@ -525,6 +628,14 @@ def tellfit(flux, thismask, arg_dict, init_from_last=None):
...
     popsize = arg_dict['popsize'] # Note this does nothing if the init is done from a previous iteration or optimum
     nsamples = arg_dict['popsize']*nparams
+    teltype = arg_dict['tell_dict']['teltype']
+    # FD: Assumes shift and stretch are turned on.
+    if teltype == 'PCA':
+        ntheta_tell = arg_dict['tell_npca']+3 # Total number of telluric model parameters in PCA mode
+    elif teltype == 'grid':
+        ntheta_tell = 4+3 # Total number of telluric model parameters in grid mode
+    else:
+        msgs.error("Unsupported teltype, must be 'PCA' or 'grid'")
```
rcooke-ast: Same comment as previously

pypeit/core/telluric.py
```diff
@@ -905,17 +1017,17 @@ def init_qso_model(obj_params, iord, wave, flux, ivar, mask, tellmodel):
 
     """
 
-    pca_dict = init_pca(obj_params['pca_file'], wave, obj_params['z_qso'], obj_params['npca'])
-    pca_mean = np.exp(pca_dict['components'][0, :])
+    qso_pca_dict = init_qso_pca(obj_params['pca_file'], wave, obj_params['z_qso'], obj_params['npca'])
```
rcooke-ast: Change `init_qso_pca` -> `qso_init_pca`

pypeit/core/telluric.py
```diff
@@ -2263,8 +2413,19 @@ def __init__(self, wave, flux, ivar, gpm, telgridfile, obj_params, init_obj_mode
...
+                                               wave_max=self.wave_in_arr[wv_gpm].max())
+            if self.tell_npca > self.tell_dict['ncomp_tell_pca']:
+                msgs.error('Asked for more telluric PCA components ({}) ' \
+                           'than exist in the PCA file ({}).'.format(self.tell_npca,
+                                                                     self.tell_dict['ncomp_tell_pca']))
+        elif self.teltype == 'grid':
+            self.tell_dict = read_telluric_grid(self.telgrid, wave_min=self.wave_in_arr[wv_gpm].min(),
+                                                wave_max=self.wave_in_arr[wv_gpm].max())
+        else:
+            msgs.error('Invalid teltype -- must be `pca` or `grid`')
```
rcooke-ast: Same comment as before.

pypeit/core/telluric.py
rcooke-ast: Shouldn't this be handled in the `pypeitpar`?

pypeit/core/telluric.py
```diff
@@ -2380,8 +2542,12 @@ def run(self, only_orders=None):
...
                                               maxiter=self.maxiter, lower=self.lower,
                                               upper=self.upper, sticky=self.sticky)
-            self.theta_obj_list[iord] = self.result_list[iord].x[:-7]
-            self.theta_tell_list[iord] = self.result_list[iord].x[-7:]
+            if self.teltype == 'PCA':
+                self.theta_obj_list[iord] = self.result_list[iord].x[:-(self.tell_npca+3)]
+                self.theta_tell_list[iord] = self.result_list[iord].x[-(self.tell_npca+3):]
+            elif self.teltype == 'grid':
+                self.theta_obj_list[iord] = self.result_list[iord].x[:-(4+3)]
+                self.theta_tell_list[iord] = self.result_list[iord].x[-(4+3):]
```
rcooke-ast, line 2533: Perhaps it's not needed, but you might consider adding the `else` statement as well, including an error message similar to that described above.

pypeit/core/telluric.py
```diff
@@ -2460,6 +2626,11 @@ def assign_output(self, iord):
         gdwave = self.wave_in_arr[:,iord] > 1.0
         wave_in_gd = self.wave_in_arr[gdwave,iord]
         wave_grid_now = self.wave_grid[self.ind_lower[iord]:self.ind_upper[iord]+1]
+        
+        if self.teltype == 'PCA':
+            ntell = self.tell_npca
+        elif self.teltype == 'grid':
+            ntell = 4
```
rcooke-ast, line 2616: Likewise here, include an else statement and throw an error?

pypeit/core/telluric.py
```diff
@@ -2552,33 +2720,51 @@ def get_tell_guess(self):
...
+        if self.teltype == 'grid':
+            bounds.append((self.tell_dict['pressure_grid'].min(),
+                           self.tell_dict['pressure_grid'].max()))
+            bounds.append((self.tell_dict['temp_grid'].min(),
+                           self.tell_dict['temp_grid'].max()))
+            bounds.append((self.tell_dict['h2o_grid'].min(),
+                           self.tell_dict['h2o_grid'].max()))
+            bounds.append((self.tell_dict['airmass_grid'].min(),
+                           self.tell_dict['airmass_grid'].max()))
+        else:
```
rcooke-ast: Add a comment saying that this is the PCA approach within the else statement?

rcooke-ast (approved):
Thanks @freddavies for the comprehensive overhaul of the telluric correction! I have mostly minor comments, but something that is unclear to me is why you need several different resolution grids? Can't we just have a single very high resolution version and then convolve it to lower resolutions specific to the instrument (and/or the resolution of the data at a given wavelength)?

Since my comments are relatively minor, I'll approve for now. It would be nice to see some plots of the telluric correction applied to the data you've tested (for both the `grid` and `PCA` method for comparison, and to ensure that both options still work). Also, it would be a good idea to double check that the tests pass.

pypeit/par/pypeitpar.py
freddavies: Done -- I think I've done this right...

pypeit/par/pypeitpar.py
freddavies: Done.

pypeit/core/telluric.py
freddavies: That's probably true, changed.

pypeit/core/telluric.py
freddavies: Done.

pypeit/core/telluric.py
freddavies: This is an index not a wavelength, i.e. if wave_min is None then it uses the entire (valid) range of wavelengths in the PCA model.

pypeit/core/telluric.py
freddavies: Oops -- this was a carryover from `read_telluric_grid` but I forgot to copy over its definition. Same response as above applies.

pypeit/core/telluric.py
freddavies: Done.

pypeit/core/telluric.py
freddavies: Removed.

pypeit/core/telluric.py
freddavies: Removed.

pypeit/core/telluric.py
freddavies: Removed.

pypeit/core/telluric.py
freddavies: Indeed...not sure why it is here, actually! I have moved the description up to tellfit_chi2.

pypeit/core/telluric.py
freddavies: Done, and nope.

pypeit/core/telluric.py
freddavies: I've removed the else statements here (and in subsequent comments) in favor of the `validate()` routines in pypeitpar

freddavies:
Thanks for the comments, @rcooke-ast ! I'm finally getting around to pushing this through...

First, a couple of updates: I have added new PCA models to the google drive with R=120000 in the optical for HIRES, and one with R=60000 for 0.9-5.5 µm for high-resolution and long-wavelength modes of NIRSPEC.

> Thanks @freddavies for the comprehensive overhaul of the telluric correction! I have mostly minor comments, but something that is unclear to me is why you need several different resolution grids? Can't we just have a single very high resolution version and then convolve it to lower resolutions specific to the instrument (and/or the resolution of the data at a given wavelength)?

The question is whether you want to do that convolution before or after the transformation into flux space. The problem with the former is the non-linear nature of the PCA vectors -- while one could arbitrarily convolve down a high resolution version, that convolution would not conserve flux, so it becomes much less likely that you could still get a reasonable representation of a telluric spectrum.

One can always convolve down _after_ the transformation to flux, but that makes it run much more slowly, since you have to perform a bunch of operations on the high resolution grid. So I opted to make a handful of grids of different sizes to optimize the runtime (at least a little bit). It could probably be consolidated better, though, maybe into a single file which has all the different-sized grids.

> 
> Since my comments are relatively minor, I'll approve for now. It would be nice to see some plots of the telluric correction applied to the data you've tested (for both the `grid` and `PCA` method for comparison, and to ensure that both options still work). Also, it would be a good idea to double check that the tests pass.

Indeed, I'm working on this comparison, but I haven't had enough time to make a comprehensive and intelligible set of plots. I will post some figures in the developer Slack at a later date.

pypeit/par/pypeitpar.py
```diff
@@ -2234,6 +2234,17 @@ def __init__(self, telgridfile=None, sn_clip=None, resln_guess=None, resln_frac_
                                'must be downloaded from the GoogleDrive and installed in your PypeIt installation via ' \
                                'the pypeit_install_telluric script. NOTE: This parameter no longer includes the full ' \
                                'pathname to the Telluric Grid file, but is just the filename of the grid itself.'
+        
+        defaults['tell_npca'] = 5
+        dtypes['tell_npca'] = int
+        descr['tell_npca'] = 'Number of telluric PCA components used. Can be set to any number from 1 to 10.'
+        
+        defaults['teltype'] = 'pca'
+        dtypes['teltype'] = str
```
rcooke-ast, line 2245: I think you also need to add the following in:

`options['teltype'] = TelluricPar.valid_teltype()`

and then add the following function within this class

```
    @staticmethod
    def valid_teltype():
        """
        Return the valid telluric methods.
        """
        return ['pca', 'grid']
```

pypeit/par/pypeitpar.py
```diff
@@ -2484,6 +2495,15 @@ def validate(self):
         """
         Check the parameters are valid for the provided method.
         """
+        
+        if self.data['tell_npca'] < 1 or self.data['tell_npca'] > 10:
+            raise ValueError('Invalid value {:d} for tell_npca '.format(self.data['tell_npca'])+
+                             '(must be between 1 and 10).')
+                             
+        if self.data['teltype'].lower() != 'pca' and self.data['teltype'].lower() != 'grid':
```
rcooke-ast: Following on from the above comment, I think this might need to be replaced with:

`if self.data['teltype'] not in TelluricPar.valid_teltype():`

But, I think this approach is not uniformly used throughout the code, so no bother!

pypeit/par/pypeitpar.py
```diff
@@ -2234,6 +2234,17 @@ def __init__(self, telgridfile=None, sn_clip=None, resln_guess=None, resln_frac_
...
                                'the pypeit_install_telluric script. NOTE: This parameter no longer includes the full ' \
                                'pathname to the Telluric Grid file, but is just the filename of the grid itself.'
+        
+        defaults['tell_npca'] = 5
+        dtypes['tell_npca'] = int
+        descr['tell_npca'] = 'Number of telluric PCA components used. Can be set to any number from 1 to 10.'
+        
+        defaults['teltype'] = 'pca'
+        dtypes['teltype'] = str
+        descr['teltype'] = 'Method used to evaluate telluric models, either pca or grid. The grid option uses a ' \
```
rcooke-ast, line 2246: It might also be good to instead print the valid options from the `valid_teltype()` definition. Just in case someone later adds to this.

pypeit/par/pypeitpar.py
```diff
@@ -2484,6 +2495,15 @@ def validate(self):
...
+        
+        if self.data['tell_npca'] < 1 or self.data['tell_npca'] > 10:
+            raise ValueError('Invalid value {:d} for tell_npca '.format(self.data['tell_npca'])+
+                             '(must be between 1 and 10).')
+                             
+        if self.data['teltype'].lower() != 'pca' and self.data['teltype'].lower() != 'grid':
+            raise ValueError('Invalid teltype {} '.format(self.data['teltype'])+
+                             ', must be "pca" or "grid".')
+        
         pass
```
rcooke-ast: The `pass` can be removed here.

pypeit/par/pypeitpar.py
```diff
@@ -2484,6 +2495,15 @@ def validate(self):
         """
         Check the parameters are valid for the provided method.
         """
+
```
rcooke-ast: We should probably do this everywhere throughout the code, but if you're worried that users will sometimes set 'PCA' instead of 'pca', then you can add the following line to the beginning of validate:
`self.data['teltype'] = self.data['teltype'].lower()`

pypeit/par/pypeitpar.py
freddavies: It looks like a lot of the classes are missing `options`! I have added it, nevertheless, hopefully in the correct way.

pypeit/par/pypeitpar.py
freddavies: Done.

pypeit/par/pypeitpar.py
freddavies: Done.

pypeit/par/pypeitpar.py
freddavies: This was my worry, yes -- I originally had that transformation inside of the __init__ function for the Telluric class, but it probably makes more sense to put it here.

jhennawi (approved):
This is good to go. Thanks for the amazing work here @freddavies

profxj:
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  242 passed, 70 warnings in 358.78s (0:05:58) ---
--- PYTEST UNIT TESTS FAILED  2 failed, 131 passed, 171 warnings in 919.02s (0:15:19) ---
--- PYTEST VET TESTS PASSED  61 passed, 70 warnings in 1495.55s (0:24:55) ---
Testing Started at 2023-12-13T23:39:00.775595
Testing Completed at 2023-12-14T09:24:33.737452
Total Time: 9:45:32.961857

PR #1648, opened 2023-08-09
Doc fixes

kbwestfall:
Fixes warnings in the doc compilation that actually cause the ReadTheDocs build to fail.

Most of these changes are an artifact of me trying to find the failures.  The bottom line is that warnings were being caused by syntax like this:

```
:class:`~pypeit.alignframe.AlignmentSplines`_
```

This is a mix of an explicit hyperlink that requires the `_` at the end, and an internal class reference, which does not.  The solution is to remove the trailing `_`.

This also includes updates to the Sphinx version dependencies, and I now include `scikit-image` and `specutils` in the `dev` dependencies.


profxj (approved):
Thanks for this major clean up!

debora-pe (approved):
Thank you so much for this!!!

PR #1650, opened 2023-08-11
Keck KCRM

rcooke-ast:
This PR adds minimal support for Keck/KCRM data reduction, and some additional features to the sky subtraction to account for the non-uniform FWHM map of the KCWI+KCRM field of view.

A companion PR to the dev suite will follow, once I get approval to include some frames in the dev-suite. Here are the wavelength solutions currently implemented:

**RM1:**
KCRM_RM1_wavecal

**RM2:**
KCRM_RM2_wavecal


doc/coadd3d.rst
```diff
@@ -268,15 +268,51 @@ cubes covering different wavelength range, but it can coadd
...
+Combining multiple datacubes
+============================
 
 PypeIt is able to combine standard star frames for flux calibration, and
 should not have any difficulty with this. If your science observations are
 designed so that there is very little overlap between exposures, you should
-not expect the automatic combination algorithm to perform well. Instead, you
-should output individual data cubes and manually combine the cubes with some
-other purpose-built software.
+not assume that the automatic combination algorithm will perform well. Instead,
```
profxj, line 277: Thanks for these!

pypeit/data/arc_lines/lists/ArI_lines.dat
```diff
@@ -160,4 +160,9 @@
 | ArI |  9356.7870 |    1 |    89 |      3521 |  lrisr_400_8500_PYPIT.json |
 | ArI |  9660.4350 |    1 |    72 |     35000 |   deimos_830g_r_PYPIT.json |
 | ArI |  9787.1860 |    1 |    25 |      8197 |  lrisr_400_8500_PYPIT.json |
+| ArI | 10054.81   |    1 |     1 |       300 |       keck_mosfire_webpage |
```
profxj, line 163: are these in vacuum (should they be)?

profxj (approved):
this is great, thanks!

pypeit/data/arc_lines/lists/ArI_lines.dat
rcooke-ast: Yes, the new ones are in vacuum (as are the other lines in this file).

pypeit/core/wavecal/autoid.py
```diff
@@ -844,8 +842,12 @@ def map_fwhm(image, gpm, slits_left, slits_right, slitmask, npixel=None, nsample
             Arc image (nspec, nspat)
         gpm (`numpy.ndarray`_):
             Good pixel mask corresponding to the input arc image (nspec, nspat)
-        slits (:class:`~pypeit.slittrace.SlitTraceSet`):
-            Slit edges
+        slits_left (`numpy.ndarray`_):
```
jhennawi, line 846: Thanks for fixing this!

pypeit/find_objects.py
```diff
@@ -1120,6 +1119,68 @@ def illum_profile_spectral(self, global_sky, skymask=None):
...
 
+    def convolve_skymodel(self, input_img, fwhm_map, thismask, subpixel=5, nsample=10):
+        """Convolve an input image with a Gaussian function that ensures all pixels have
+        the same FWHM (i.e. the returned image has a spectral resolution corresponding to
+        the maximum FWHM specified in the input fwhm_map). To speed up the computation,
+        the input image is uniformly convolved by a grid of FWHM values, and the final
+        pixel-by-pixel convolved map is an interpolation of the grid point values.
+
+         Args:
+             input_img (`numpy.ndarray`_):
```
jhennawi: Please indicate the shapes of the arrays in the docs.

pypeit/find_objects.py
```diff
@@ -1143,56 +1204,41 @@ def joint_skysub(self, skymask=None, update_crmask=True, trim_edg=(0,0),
...
             if not self.par['reduce']['skysub']['global_sky_std']:
                 msgs.info('Skipping global sky-subtraction for standard star.')
-                return global_sky
-
+                return _global_sky
+
+        # Use the FWHM map determined from the arc lines to convert the science frame
+        # to have the same effective spectral resolution.
+        fwhm_map = self.wv_calib.build_fwhmimg(self.tilts, self.slits, initial=True, spat_flexure=self.spat_flexure_shift)
+        thismask = thismask & (fwhm_map != 0.0)
```
jhennawi, line 1149: Can you please add some documentation to this routine explaining how the joint sky subtraction works. I don't follow the philosophy here. Why you are convolving the image to a single fwhm and then fitting the global_sky. It looks like this is only used to compute the scaleimg. A detailed description of the control flow of the routine is needed in the docstring.

pypeit/find_objects.py
```diff
@@ -1120,6 +1119,68 @@ def illum_profile_spectral(self, global_sky, skymask=None):
         # Now apply the correction to the science frame
         self.apply_relative_scale(scaleImg)
 
+    def convolve_skymodel(self, input_img, fwhm_map, thismask, subpixel=5, nsample=10):
```
jhennawi: This routine should probably be moved to a core method wrapped by the method in the class. You are performing core functionality here as far as I can tell.

pypeit/flatfield.py
```diff
@@ -1946,6 +1947,76 @@ def illum_profile_spectral(rawimg, waveimg, slits, slit_illum_ref_idx=0, smooth_
...
+        Image data that will be used to estimate the spectral relative sensitivity
+    waveimg : `numpy.ndarray`_
+        Wavelength image
+    slits : :class:`~pypeit.slittrace.SlitTraceSet`
+        Information stored about the slits
+    model : `numpy.ndarray`_
+        A model of the rawimg data.
+    slit_illum_ref_idx : int
+        Index of slit that is used as the reference.
+    gpmask : `numpy.ndarray`_, None
```
jhennawi: specify that this is boolean and optional.

pypeit/flatfield.py
```diff
@@ -1946,6 +1947,76 @@ def illum_profile_spectral(rawimg, waveimg, slits, slit_illum_ref_idx=0, smooth_
...
+        Good pixel mask
+    thismask : `numpy.ndarray`_, None
```
jhennawi: same goes here specify data type.

pypeit/flatfield.py
```diff
@@ -1946,6 +1947,76 @@ def illum_profile_spectral(rawimg, waveimg, slits, slit_illum_ref_idx=0, smooth_
...
+    Use a polynomial fit to control points along the spectral direction to determine the relative
+    spectral illumination of all slits. Currently, this routine is only used for image slicer IFUs.
+
+    Parameters
+    ----------
+    rawimg : `numpy.ndarray`_
+        Image data that will be used to estimate the spectral relative sensitivity
+    waveimg : `numpy.ndarray`_
+        Wavelength image
+    slits : :class:`~pypeit.slittrace.SlitTraceSet`
```
jhennawi: Slits is an object. Instead pass in the boundaries, slit_left and slit_right which are arrays. That is our convention in core.

pypeit/utils.py
```diff
@@ -784,6 +783,81 @@ def boxcar_smooth_rows(img, nave, wgt=None, mode='nearest', replace='original'):
...
+    """
+    Convolve img with an input kernel using an FFT. Following the FFT,
+    a slower convolution is used to estimate the convolved image near
+    the masked pixels.
+
+    .. note::
+        For images following the PypeIt convention, this smooths the
+        data spectrally for each spatial position.
+
+    Args:
```
jhennawi, line 797: Specify the sizes of the images in the documentation please.

pypeit/utils.py
```diff
@@ -784,6 +783,81 @@ def boxcar_smooth_rows(img, nave, wgt=None, mode='nearest', replace='original'):
...
+    Returns:
+        `numpy.ndarray`_: The convolved image
+    """
+    nspec = img.shape[0]
+    kernsize = kernel.size
+    hwid = (kernsize - 1) // 2
+    harr = np.arange(-hwid, +hwid + 1)
+    fullsize = nspec + kernsize - 1
+    fsize = 2 ** int(np.ceil(np.log2(fullsize)))  # Use this size for a more efficient computation
+    conv = np.fft.fft(img, fsize, axis=0)
```
jhennawi: Should you be using this function: https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.fftconvolve.html
instead of doing it yourself. I suspect it is faster.

jhennawi (changes requested):
HI @rcooke-ast relatively minor comments and changes requested. Thanks for this PR.

pypeit/flatfield.py
rcooke-ast: Done!

pypeit/flatfield.py
rcooke-ast: Done!

pypeit/utils.py
rcooke-ast: Updated

pypeit/find_objects.py
rcooke-ast: Done!

pypeit/find_objects.py
rcooke-ast: Yes, you are correct - I have moved this routine to core.

pypeit/flatfield.py
rcooke-ast: Good point... Although this is not in core, it probably should be, and I've made the changes to pass in the arrays instead of the slits object.

pypeit/find_objects.py
rcooke-ast: Yes, you are exactly right. This is because once you get a model of the sky, all of the various deconvolution routines that I have attempted produced terrible results. The best was Weiner deconvolution. I think it works well for very small deviations from a uniform FWHM, but for the large deviations seen in KCWI/KCRM, this is virtually impossible to get a good result. I'm currently at a loss to get the convolution/deconvolution to work.

For now, I'll add a note to the documentation to describe how this is currently working. I think the only way forward is to have a single underlying "intrinsic" model that gets convolved with a grid of kernels, and evaluated. But, I fear this is far too computationally expensive. For KCWI/KCRM, I think the recommended approach is to have separate sky frames if you care about sky subtraction.

pypeit/utils.py
rcooke-ast: I benchmarked this, and you're right... so I've made this change.

```
from scipy import signal
import numpy as np

def fft_orig(img, kernel):
    nspec = img.shape[0]
    kernsize = kernel.size
    hwid = (kernsize - 1) // 2
    harr = np.arange(-hwid, +hwid + 1)
    fullsize = nspec + kernsize - 1
    fsize = 2 ** int(np.ceil(np.log2(fullsize)))  # Use this size for a more efficient computation
    conv = np.fft.fft(img, fsize, axis=0)
    conv *= np.fft.fft(kernel / np.sum(kernel), fsize)[:, None]
    img_conv = np.fft.ifft(conv, axis=0).real[hwid:hwid + nspec, :]
    return img_conv

def fft_scipy(img, kernel):
    return signal.fftconvolve(img, kernel, mode='same', axes=0)


img = np.random.normal(0.0,1.0,(511,511))
kernwid = 3
kernsize = 2 * int(5 * kernwid + 0.5) + 1
midp = (kernsize - 1) // 2
xkern = np.arange(kernsize+1, dtype=int) - midp
kern = np.exp(-0.5 * (xkern / kernwid) ** 2)
kern = kern / np.sum(kern)

tmp1 = fft_orig(img, kern)
kernrsh = kern.reshape(kern.size,1)
tmp2 = fft_scipy(img, kernrsh)

print(np.allclose(tmp1, tmp2))
```

```
# %timeit fft_orig(img, kern)
# %timeit fft_scipy(img, kernrsh)

511x511
11.4 ms ± 475 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
3.76 ms ± 212 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

512x512
11.5 ms ± 736 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
4.11 ms ± 225 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

513x513
8.79 ms ± 322 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
3.33 ms ± 48 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

rcooke-ast (commented):
Addresed PR comments

pypeit/flatfield.py
jhennawi: Okay my bad. Can you move it to core and wrap here?

pypeit/utils.py
jhennawi: Thanks for doing this!

jhennawi (approved):
Thanks!

pypeit/flatfield.py
rcooke-ast: Yep - I've already moved it to core  :-)
(note, this file is marked "outdated")

rcooke-ast:
Tests pass... including two new Keck KCRM tests

image

PR #1652, opened 2023-08-14
Dark Current updated for all spectrographs

rcooke-ast:
For many spectrographs, the dark current was defined as e-/s/pixel, but the DetectorContainer Data Model in PypeIt actually requires e-/hour/pixel. Therefore, there are multiple spectrographs that have had their base variance underestimated.


kbwestfall (changes requested):
Thanks for digging in on this!  My only request is that we perform the multiplications.  I.e., instead of setting the value to, e.g., `0.5*3600.`, I'd rather we put in `1800.`.

debora-pe (approved):
Looks good to me. Thank you for doing this.
Please, run the tests.

kbwestfall (approved):
Approving!

rcooke-ast:
Updated format, and made all units clear. Running dev suite now...

rcooke-ast:
Dev suite passes... merging...
image

PR #1654, opened 2023-08-18
wavecal for 600/7500

rcooke-ast:
This PR adds support for Shane/KASTr 600/7500 grating... Holy Grail worked so well for so long...

image


kbwestfall (approved):
Looks good to me.  Can you post the results from running the dev-suite on the relevant `shane_kast_red` setups?

rcooke-ast:
Sure thing - Here are the results:
image

profxj (approved):
Thanks thanks.

Can you confirm that the wavelength `vet` test 
is still passing before merging?

https://github.com/pypeit/PypeIt-development-suite/blob/develop/vet_tests/test_wavelengths.py#L18

rcooke-ast:
Here is the output of the command:
` python -m pytest test_wavelengths.py -k test_shane_kast_red`
The relevant kast test passes.

image

PR #1655, opened 2023-08-18
Slitmask matching error

kbwestfall:
Adds a display command used for debugging and throws an error if all of the slit edges are masked after the slit-design matching.  This catches an error from one of @SunilSimha's DEIMOS datasets.  The solution was to change the `edge_thresh`.  This improved error message hopefully points users in the right direction, compared to the old obscure one that was completely unrelated.

Changes to `slitmask.py` are due to an unrelated OCD moment.


debora-pe (approved):

debora-pe (approved):
Thanks @kbwestfall!

SunilSimha (approved):
Looks good. Thanks @kbwestfall, @debora-pe .

PR #1658, opened 2023-08-28
Hotfix for parameter name in slit tracing doc.

kbwestfall:
As titled.


profxj (approved):
approved

tbowers7 (approved):
Thanks for the fix!

PR #1661, opened 2023-09-03
KCWI scattered light

rcooke-ast:
This PR includes the functionality to generate a (spectrograph dependent) scattered light model, and subtract this light from the input image. The scattered light should be removed before correcting for spatial and spectral illumination, so it has to be included in the rawimage class. Here is an example for the KCWI correction. This is based on the left, right, and central parts of the detector (i.e. the not illuminated parts of the detector). This is a pixelflat.

image


pypeit/core/parse.py
```diff
@@ -124,6 +124,19 @@ def parse_binning(binning:str):
...
+
+            * string format
+                --  comma delimited string (e.g. '2,1')
+                --  x delimited string (e.g. '2x1')
+                --  space delimited string (e.g. '2 1')
+                --  'None' will always assume 1x1 binning
+            * tuple format
+                -- this must be of the form of tuple, for example: (2,1)
+            * numpy array
+                -- this must be of the form of tuple, for example: np.array([2,1])
```
kbwestfall: Please check this formats as you expect when the docs are built.

pypeit/core/skysub.py
```diff
@@ -1450,7 +1450,7 @@ def convolve_skymodel(input_img, fwhm_map, thismask, subpixel=5, nsample=10):
         xkern = np.arange(kernsize, dtype=int) - midp
         kern = np.exp(-0.5 * (xkern / kernwids[kk]) ** 2)
         kern = kern / np.sum(kern)
-        conv_allkern[:, :, kk] = utils.rebinND(utils.convolve_fft(_input_img, _input_msk, kern), input_img.shape)
+        conv_allkern[:, :, kk] = utils.rebinND(utils.convolve_fft(_input_img, kern, _input_msk), input_img.shape)
```
kbwestfall: Nice catch!

pypeit/find_objects.py
```diff
@@ -1244,6 +1244,10 @@ def global_skysub(self, skymask=None, update_crmask=True, trim_edg=(0,0),
         correction using the sky spectrum, if requested. See Reduce.global_skysub()
         for parameter definitions.
         """
+        if self.par['reduce']['findobj']['skip_skysub']:
+            msgs.info("Skipping global sky sub as per user request")
+            return np.zeros_like(self.sciImg.image)
+
```
kbwestfall, line 1250: Nice catch!

pypeit/images/rawimage.py
```diff
@@ -1106,6 +1111,27 @@ def subtract_continuum(self, force=False):
...
+        if self.steps[step]:
+            # Already pattern subtracted
+            msgs.warn("The scattered light has already been subtracted from the image!")
+            return
+
+        # Loop over the images
+        for ii in range(self.nimg):
+            binning = self.detector[0].binning
+            scatt_img = self.spectrograph.scattered_light(self.image[ii, :, :], binning)
+            self.image[ii, :, :] -= scatt_img
```
kbwestfall: For statements like this, I instead would do `self.image[ii] -= scatt_img`.  Is there a reason to include the colons (e.g., is there a subtle memory allocation issue)?

Another aside: If you don't know or don't want to specify the dimensionality of the remaining array dimensions, you can use and ellipsis: `self.image[ii,...] -= scat_img`.

pypeit/images/rawimage.py
```diff
@@ -1106,6 +1111,27 @@ def subtract_continuum(self, force=False):
         #cont = ndimage.median_filter(self.image, size=(1,101,3), mode='reflect')
         self.steps[step] = True
 
+    def subtract_scattlight(self):
+        """
+        Analyze and subtract the scattered light from the image.
+
+        This is primarily a wrapper for
+        :func:`~pypeit.spectrograph.procimg.subtract_pattern`.
```
kbwestfall: Fix this reference.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1017,6 +1023,39 @@ def get_rawimage(self, raw_file, det):
...
 
+    def scattered_light(self, frame, binning):
+        """
+        Calculate a model of the scattered light of the input frame.
+
+        Parameters
+        ----------
+        frame : `numpy.ndarray`_
+            Raw 2D data frame to be used to compute the scattered light.
+        binning : (str, `numpy.ndarray`_, tuple):
```
kbwestfall: Remove the parentheses.  I.e.,
```
binning : str, `numpy.ndarray`_, tuple
```

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1017,6 +1023,39 @@ def get_rawimage(self, raw_file, det):
...
+            Binning of the frame (e.g. '2x1' refers to a binning of 2 in the spectral
+            direction, and a binning of 1 in the spatial direction). For the supported
+            formats, refer to `~pypeit.core.parse.parse_binning`.
```
kbwestfall: Include the `:func:` syntax:
```
:func:`~pypeit.core.parse.parse_binning`
```

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1017,6 +1023,39 @@ def get_rawimage(self, raw_file, det):
...
+        -------
+        scatt_img : `numpy.ndarray`_
+            A 2D image of the scattered light determined from the input frame
+        """
+        spatbin = parse.parse_binning(binning)[1]
+        ym = 4096 // (2 * spatbin)
+        y0 = ym - 180 // spatbin
+        y1 = ym + 180 // spatbin
+        scattlightl = np.nanmedian(frame[:, :20//spatbin], axis=1)[:, None]
+        scattlightr = np.nanmedian(frame[:, -40//spatbin:], axis=1)[:, None]
```
kbwestfall, line 1049: Can we avoid hard-coding these numbers?  I.e., can we get them from the detector object?

pypeit/spectrographs/spectrograph.py
```diff
@@ -1879,6 +1879,28 @@ def calc_pattern_freq(self, frame, rawdatasec_img, oscansec_img, hdu):
...
+        Calculate a model of the scattered light of the input frame.
+
+        Parameters
+        ----------
+        frame : `numpy.ndarray`_
+            Raw 2D data frame to be used to compute the scattered light.
+        binning : (str, `numpy.ndarray`_, tuple):
+            Binning of the frame (e.g. '2x1' refers to a binning of 2 in the spectral
+            direction, and a binning of 1 in the spatial direction). For the supported
+            formats, refer to `~pypeit.core.parse.parse_binning`.
```
kbwestfall: Same fixes for the type and the `:func:` syntax.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1017,6 +1023,39 @@ def get_rawimage(self, raw_file, det):
...
+        scattlight = np.nanmedian(frame[:, y0:y1], axis=1)[:, None]
+        medl = np.median(scattlightl / scattlight)
+        medr = np.median(scattlightr / scattlight)
+        spatimg = np.meshgrid(np.arange(frame.shape[1]), np.arange(frame.shape[0]))[0]
+        scatt_img = np.outer(scattlight, np.ones(frame.shape[1]))
+        scatt_img[spatimg < ym] *= 1 + (medl - 1) * (spatimg[spatimg < ym] / ym - 1) ** 2
+        scatt_img[spatimg >= ym] *= 1 + (medr - 1) * (spatimg[spatimg >= ym] / (4095//spatbin - ym) - ym / (4095//spatbin - ym)) ** 2
+        return scatt_img
```
kbwestfall, line 1061: Please add some inline comments that explain the steps being taken.

pypeit/core/datacube.py
```diff
@@ -1953,7 +1953,8 @@ def coadd_cube(files, opts, spectrograph=None, parset=None, overwrite=False):
                     msgs.error("Could not load skysub image from spec2d file:" + msgs.newline() + opts['skysub_frame'][ff])
                 # TODO :: Consider allowing the actual frame (instead of the skymodel) to be used as the skysub image - make sure the BPM is carried over.
                 #      :: Allow sky data fitting (i.e. scale the flux of a skysub frame to the science frame data)
-                skyImg = spec2DObj_sky.skymodel * exptime / skysub_exptime  # Sky counts
+                skyImg = spec2DObj_sky.sciimg * exptime / skysub_exptime  # Sky counts
+                # skyImg = spec2DObj_sky.skymodel * exptime / skysub_exptime  # Sky counts
```
kbwestfall, line 1957: Can't really comment on the changes to this file, but are they worth commenting on in the CHANGES or keck_kcwi docs?

kbwestfall (changes requested):
A few minor comments from me, and two questions:

 - Can you add a description of this approach to the `keck_kcwi.rst` doc?
 - Can we save the scattered light image somewhere?  In either or both of the `Flat` calilbration frame and the `spec2d` files?

pypeit/find_objects.py
```diff
@@ -1244,6 +1244,10 @@ def global_skysub(self, skymask=None, update_crmask=True, trim_edg=(0,0),
         correction using the sky spectrum, if requested. See Reduce.global_skysub()
         for parameter definitions.
         """
+        if self.par['reduce']['findobj']['skip_skysub']:
```
jhennawi, line 1247: This should be upstream in the control flow where it is called. It is odd to call a skysub routine and then return zero behind the scenes. Better would be to just put this skip upstream so the user can see that sky subtraction is being skipped in the main control flow.

jhennawi (changes requested):
Very minor comments. Thanks!

pypeit/core/datacube.py
rcooke-ast: Good point - I have added a note in the release docs

pypeit/images/rawimage.py
rcooke-ast: Fixed (I think!)

pypeit/images/rawimage.py
rcooke-ast: I didn't know about ellipses - I've made your suggested change.

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Done - thanks!

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Good catch - thanks!

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: I'm afraid not, for now. I think the eventual goal is to base this on the slit edges, but this is a little tricky to include in `rawimage.py`, so I'm leaving this as a TODO for now... I'll add a GH issue to remind ourselves that this is what we need to eventually be doing.

I'm also working on an alternative model for the KCWI+KCRM scattered light.

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Comments added

pypeit/spectrographs/spectrograph.py
rcooke-ast: Fixed

pypeit/core/parse.py
rcooke-ast: OK - I had it totally wrong, but I've now fixed this and it looks grand... It seems I'm learning new things with docstrings all the time!

rcooke-ast (commented):
Thanks @jhennawi and @kbwestfall - I've addressed your comments.

In the process, I have also merged in in PR #1678 to take advantage of the new way to document changes.

rcooke-ast:
Sorry @kbwestfall, I forgot to address your other questions...

> * Can you add a description of this approach to the `keck_kcwi.rst` doc?

Yes, I will add this to Keck/KCWI docs.

> * Can we save the scattered light image somewhere?  In either or both of the `Flat` calilbration frame and the `spec2d` files?

I think this is not a good idea for the moment. I am working on a parameterised model that would need to store the model parameters, rather than an image (together with a routine that can generate the image from the model parameters). So, I'm tempted to wait until the next round of KCWI scattered light removal to incorporate this.

pypeit/core/parse.py
kbwestfall: Yeah, I'm constantly learning the the ins and outs of docstrings / sphinx / rst, as well!

kbwestfall:
On saving the scattered light image, I'm fine with putting this off, but I wanted to advocate for saving the full image.  In similar situations I also used to be of the opinion that it's better to save the model parameters and provide a script if people want to reconstruct the model.  But I've come around to the opinion that this presents a barrier (albeit small) to people being able to assess the results.  Disk space is cheap and, IMO, enabling people to quickly open the file with astropy and view it with matplotlib is worth the cost of a larger output file.

kbwestfall (approved):
Approving!

rcooke-ast:
Tests pass
image

jhennawi (approved):
Approved. Thanks!

Issue #1662, opened 2023-09-06
Separate illumflat is not made if pixelflat already exists (when `run_pypeit` is executed with the `-c` argument).


rcooke-ast:
`run_pypeit` with the `-c` option appears to ignore the BPM. This was tested with the `keck_kcwi/medium_bl` data.

rcooke-ast:
This is actually an issue to do with the pixelflat being generated first, and then a separate illumflat is trying to be generated and merged into the pixelflat.

rcooke-ast:
Also, in this case, the BPM doesn't work, because the dataset has multiple different BPM, but only 1 BPM is actually used (the one not relevant to the Science frame).

rcooke-ast:
Addressed by PR #1663

PR #1663, opened 2023-09-07
hotfix BPM

rcooke-ast:
If different calibrations use different amplifier settings, then the BPM needs to be specific to the calibration frame. This PR resets the BPM every time a new frame is being generated. It's a bit overkill, but it takes no time at all, and is "safer".

Addresses Issue #1662

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@
 - Hotfix for skysub regions GUI that used np.bool
 - Hotfix to stop pypeit_setup from crashing on data from lbt_luci1, lbt_luci2, magellan_fire,
   magellan_fire_long, p200_tspec, or vlt_sinfoni.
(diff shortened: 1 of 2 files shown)


profxj (approved):
Approved!

We are confident this will pass the DevSuite, right? :)

rcooke-ast:
We are... but I've already started a dev-suite run, just in case  :-)

kbwestfall (approved):
I think this is okay, but we should run tests.  I did a quick check of the code and I don't think there are any changes made to `msbpm` that need to propagate between functions (so a refresh is good).  This also ensures each bpm is specific to the relevant frame type (not always based on the frame to be calibrated).

rcooke-ast:
Tests pass... Can merge if tox doesn't timeout.

image

PR #1664, opened 2023-09-07
slitspatnum bug

jhennawi:
It looks like the functionality to reduce specific slits was never properly implemented. It apparently will fail on any instrument that does not have the spectrograph.list_detectors() method overloaded which is all instruments except gemini_gmos, keck_deimos, and keck_mosfire.

[rdx]
  slitspatnum = DET01:1557

The current PR attempts to fix this using the specific case of LBT mods where the failure was identified.  I need help with Keck hires, since the mosaic hooks are not there and I'm confused with how this function is supposed to behave since the list_detectors function is different for the no mosaic case for deimos and mosfire (I'm not sure why). 

See the dev-suite PR of the same name for a test example.  I'm handing off this PR to @kbwestfall and @debora-pe who I think are in a better position to understand whether my changes are correct. Also, please note that I think that list_detectors should maybe be overloaded for keck_hires which also has mosaics, but I'm a bit unclear on that.

It would also be a good idea to add tests.

### DIFF
--- a/pypeit/tests/test_spectrographs.py
+++ b/pypeit/tests/test_spectrographs.py
@@ -90,6 +90,16 @@ def test_list_detectors_mods():
         mosaics = mods.list_detectors(mosaic=True)
 
 
+def test_list_detectors_hires():
+    hires = load_spectrograph('keck_hires')
+    dets = hires.list_detectors()
+    assert dets.ndim == 1, 'HIRES has a 1D array of detectors'
+    assert dets.size == 3, 'HIRES has 1 detector'
+    mosaics = hires.list_detectors(mosaic=True)
+    assert mosaics.ndim == 1, 'Mosaics are listed as 1D arrays'
+    assert mosaics.size == 1, 'HIRES has 1 predefined mosaic'
+
+
 def test_configs():
 
     spec = load_spectrograph('keck_deimos')
--- a/pypeit/tests/test_spectrographs.py
+++ b/pypeit/tests/test_spectrographs.py
@@ -94,7 +94,7 @@ def test_list_detectors_hires():
     hires = load_spectrograph('keck_hires')
     dets = hires.list_detectors()
     assert dets.ndim == 1, 'HIRES has a 1D array of detectors'
-    assert dets.size == 3, 'HIRES has 1 detector'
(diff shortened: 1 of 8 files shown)


kbwestfall:
Hi @jhennawi .  I made some changes to the branch and now 
```
run_pypeit lbt_mods_mods1r_longslit.pypeit
```
finishes successfully.  I get a `spec1d_mods1r.20180112.0051-QSO_J1104+2134_MODS1R_20180112T113720.755.txt` file that looks like this:

```
| slit |                    name | spat_pixpos | spat_fracpos | box_width | opt_fwhm |  s2n | wv_rms |
| 1557 | SPAT1643-SLIT1557-DET01 |      1643.5 |        0.677 |      3.00 |    1.224 | 0.31 |  0.170 |
```

Is that what you're expecting?  If so, I can run the `mods_bug` dev-suite branch on this pypeit branch.

pypeit/tests/test_spectrographs.py
```diff
@@ -62,6 +62,44 @@ def test_select_detectors_mosaic():
...
+    assert dets.size == 1, 'MODS1R has 1 detector'
+    with pytest.raises(PypeItError):
+        mosaics = mods.list_detectors(mosaic=True)
+
+
+def test_list_detectors_hires():
+    hires = load_spectrograph('keck_hires')
+    dets = hires.list_detectors()
+    assert dets.ndim == 1, 'HIRES has a 1D array of detectors'
+    assert dets.size == 3, 'HIRES has 1 detector'
```
debora-pe: Did you want to say "HIRES has 3 detectors"?

debora-pe (approved):
Thank you @kbwestfall for taking care of this quickly and @jhennawi for starting this PR.
Just for the records,  `list_detectors` was first introduced to be used for slitmask design purposes (as the docstring says). This is why it is different for DEIMOS (we wanted to list red and blue detectors separately). I'm glad it can now be used for a broader purpose.

pypeit/tests/test_spectrographs.py
kbwestfall: Yes!  Good catch.  Fixed.

kbwestfall (approved):

PR #1669, opened 2023-09-11
Datacube core

rcooke-ast:
This looks like a large PR, but it's a relatively simple refactoring to make the CoAdd3D look a lot more like CoAdd2D. Here are the main changes:

* `IFU` reduction is replaced throughout the code by `SlicerIFU`, with a view to possibly support `FiberIFU` as another reduction mode in the (distant) future.
* `pypeit.core.datacube.py` has been refactored to `core` routines and code that uses `PypeIt`-specific objects.
* Unused routines from `pypeit.core.datacube.py` have been moved to `deprecated`

I should emphasise that there's virtually no new code here. It's just a reorganising of the code that makes datacubes so it's easier for others to read, improve, and it mostly just required writing _a lot_ more docstrings... joy!

I'm pretty sure that nothing is broken, but I'll run the tests just to be sure...


rcooke-ast:
Tests passing (at the moment)...
image

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+            :class:`CoAdd3D`: One of the subclasses with
+            :class:`CoAdd3D` as its base.
+        """
+
+        return next(c for c in cls.__subclasses__()
+                    if c.__name__ == (spectrograph.pypeline + 'CoAdd3D'))(
+                        spec2dfiles, opts, spectrograph=spectrograph, par=par, det=det, overwrite=overwrite,
+                        show=show, debug=debug)
+
+    def __init__(self, files, opts, spectrograph=None, par=None, det=None, overwrite=False,
```
jhennawi: Can you change this name to spec2dfiles for consistency with naming above and also for clarity.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+        if isinstance(spectrograph, str):
+            self.spec = load_spectrograph(spectrograph)
+            self.specname = spectrograph
+        else:
+            # Assume it's a Spectrograph instance
+            self.spec = spectrograph
+            self.specname = spectrograph.name
+
+        # Grab the parset, if not provided
+        if par is None:
```
jhennawi: self.par = self.spec.default_pypeit_par() if par is None else par

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+        wav_min = self.cubepar['wave_min'] if self.cubepar['wave_min'] is not None else np.min(all_wave)
+        wav_max = self.cubepar['wave_max'] if self.cubepar['wave_max'] is not None else np.max(all_wave)
+        if self.cubepar['wave_delta'] is not None:
+            dwave = self.cubepar['wave_delta']
+
+        return datacube.create_wcs(all_ra, all_dec, all_wave, dspat, dwave, ra_min=ra_min, ra_max=ra_max,
+                                   dec_min=dec_min, dec_max=dec_max, wave_min=wav_min, wave_max=wav_max,
+                                   reference=self.cubepar['reference_image'], collapse=collapse, equinox=equinox,
+                                   specname=specname)
+
```
jhennawi, line 552: It looks like someone has to pass in the standard cube here rather than an archived sensitivity function. I'm not sure I follow the logic there. In PypeIt sensitivity function generation is always a separate step, and sensitivity functions are passed as arguments to functions, not as filenames in parsets. This thus constitutes a significant departure from the model we use for fluxing. I also don't see why you would want to re-extract the standard and recompute sensitivity functions every time you want to coadd in 3d. Perhaps I'm missing some of the intricacies of the complex control flow here, but please elaborate. To be clear, the question I'm asking is why not do this once and pass in the sensitivity function. Also, you are using the UV sensitivity function algorithm here, but that is about to be deprecated. So you should not be writing new code that depends on code that will be deprecated. We are transitioning to IR sensitivity function everywhere.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+    def set_default_skysub(self):
+        """
+        Set the default mode to use for sky subtraction.
+        """
+        if self.cubepar['skysub_frame'] in [None, 'none', '', 'None']:
+            self.skysub_default = "none"
+            self.skyImgDef = np.array([0.0])  # Do not perform sky subtraction
+            self.skySclDef = np.array([0.0])  # Do not perform sky subtraction
+        elif self.cubepar['skysub_frame'].lower() == "image":
```
jhennawi: These .lower() instances are not necessary. We have functionality in the parsets to only tolerate specific options, and the codes fault when those options are not passed in.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+                image - Use the sky model derived from the science frame
+                none - Do not perform sky subtraction
+
+        Returns:
+            :obj:`str`_: A string that describes the sky subtration mode to be used (see opts_skysub description)
+            `numpy.ndarray`_: 2D image (same shape as science frame) containing the sky frame to be subtracted from the science frame
+            `numpy.ndarray`_: 2D image (same shape as science frame) containing the relative spectral scaling that has been applied to the returned sky frame
+        """
+        this_skysub = self.skysub_default
+        if self.skysub_default == "image":
```
jhennawi, line 700: This control flow would be cleaner, if skysub_default were a parameter in a parset. Is there a way to do that, or is the decision making too complex?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+            skyImg = spec2DObj.skymodel
+            skyScl = spec2DObj.scaleimg
+        else:
+            skyImg = self.skyImgDef.copy() * exptime
+            skyScl = self.skySclDef.copy()
+        # See if there's any changes from the default behaviour
+        if opts_skysub is not None:
```
jhennawi, line 707: Same comment here about .lower() and also about opts_skysub being a parameter in a parset.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+        details of this procedure, see the child routines.
+        """
+        msgs.bug("This routine should be overridden by child classes.")
+        msgs.error("Cannot proceed without coding the coadd routine.")
+        return
+
+
+class SlicerIFUCoAdd3D(CoAdd3D):
+    """
+    Child of CoAdd3D for SlicerIFU data reduction. For documentation, see CoAdd3d parent class above.
```
jhennawi, line 801: Add some more documentation here indicating now a SlicerIFU coadd3d differs from the parent one.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+
+        return datacube.create_wcs(all_ra, all_dec, all_wave, dspat, dwave, ra_min=ra_min, ra_max=ra_max,
+                                   dec_min=dec_min, dec_max=dec_max, wave_min=wav_min, wave_max=wav_max,
+                                   reference=self.cubepar['reference_image'], collapse=collapse, equinox=equinox,
+                                   specname=specname)
+
+    def make_sensfunc(self):
+        """
+        Generate the sensitivity function to be used for the flux calibration.
+        """
```
jhennawi, line 556: This should be in a core method, since it is core IFU sensfunc computation.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+            if flatimages.pixelflat_spec_illum is None:
+                # Calculate the relative scale
+                scale_model = flatfield.illum_profile_spectral(flatframe, waveimg, slits,
+                                                               slit_illum_ref_idx=self.flatpar['slit_illum_ref_idx'],
+                                                               model=None,
+                                                               skymask=None, trim=self.flatpar['slit_trim'],
+                                                               flexure=spat_flexure,
+                                                               smooth_npix=self.flatpar['slit_illum_smooth_npix'])
+            else:
+                msgs.info("Using relative spectral illumination from FlatImages")
```
jhennawi: Can this complex control flow below should be in a core method?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+            msgs.info("Adopting a wavelength sampling of {0:f} Angstrom".format(self._dwv))
+
+    def load(self):
+        """
+        This is the main function that loads in the data, and performs several frame-specific corrections.
+        If the user does not wish to align or combine the individual datacubes, then this routine will also
+        produce a spec3d file, which is a DataCube representation of a PypeIt spec2d frame for SlicerIFU data.
+        """
+        # Initialise variables
+        wave_ref = None
```
jhennawi: Loading from files should be a class method. But these frame specific transformations should probably be in a core method.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+
+    def run_align(self):
+        """
+        This routine aligns multiple cubes by using manual input offsets or by cross-correlating white light images.
+        """
+        # Grab cos(dec) for convenience
+        cosdec = np.cos(np.mean(self.all_dec) * np.pi / 180.0)
+
+        # Register spatial offsets between all frames
+        if self.opts['ra_offset'] is not None:
```
jhennawi: Likewise, all of what I'm seeing here is core functionality. Classes should wrap core methods, and reduce the amount of replicated code exploiting inheritance, and other niceties of object oriented programming. What I see here is a gigantic hugely complex class, without the management of variables (local vs global) or control flow afforded by core functions.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+                    # Apply the shift
+                    self.all_ra[self.all_idx == ff] += ra_shift
+                    self.all_dec[self.all_idx == ff] += dec_shift
+
+    def compute_weights(self):
+        """
+        Compute the relative weights to apply to pixels that are collected into the voxels of the output DataCubes
+        """
+        # Calculate the relative spectral weights of all pixels
+        if self.numfiles == 1:
```
jhennawi: This should also probably be moved to core. If the user cannot understand what comes in and what goes out, the control flow is impossible to comprehend.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+                                              self.all_tilts, self.all_slits, self.all_align, voxedge, all_idx=self.all_idx,
+                                              spec_subpixel=1, spat_subpixel=1, combine=True)
+            # Compute the weights
+            self.all_wghts = datacube.compute_weights(self.all_ra, self.all_dec, self.all_wave, self.all_sci, self.all_ivar, self.all_idx, wl_full[:, :, 0],
+                                                      self._dspat, self._dwv, relative_weights=self.cubepar['relative_weights'])
+
+    def coadd(self):
+        """
+        This is the main routine called to convert PypeIt spec2d files into PypeIt DataCube objects. It is specific
+        to the SlicerIFU data.
```
jhennawi, line 1268: Should this be a core method.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+                                           self.all_spatpos[ww], self.all_specpos[ww], self.all_spatid[ww], self.all_tilts[ff],
+                                           self.all_slits[ff], self.all_align[ff], vox_edges,
+                                           all_idx=self.all_idx[ww], overwrite=self.overwrite, blaze_wave=self.blaze_wave,
+                                           blaze_spec=self.blaze_spec,
+                                           fluxcal=self.fluxcal, sensfunc=sensfunc, specname=self.specname,
+                                           whitelight_range=wl_wvrng,
+                                           spec_subpixel=self.spec_subpixel, spat_subpixel=self.spat_subpixel)
+
+
+def generate_image_subpixel(image_wcs, all_ra, all_dec, all_wave, all_sci, all_ivar, all_wghts,
```
jhennawi: Should this go in core.datacube?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+            img, _, _ = subpixellate(image_wcs, all_ra[ww], all_dec[ww], all_wave[ww],
+                                     all_sci[ww], all_ivar[ww], all_wghts[ww], all_spatpos[ww],
+                                     all_specpos[ww], all_spatid[ww], tilts[fr], slits[fr], astrom_trans[fr], bins,
+                                     spec_subpixel=spec_subpixel, spat_subpixel=spat_subpixel)
+        all_wl_imgs[:, :, fr] = img[:, :, 0]
+    # Return the constructed white light images
+    return all_wl_imgs
+
+
+def generate_cube_subpixel(outfile, output_wcs, all_ra, all_dec, all_wave, all_sci, all_ivar, all_wghts,
```
jhennawi: Same comment, core.datacube?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1772 @@
...
+        img_hdu.writeto(out_whitelight, overwrite=overwrite)
+
+    # Write out the datacube
+    msgs.info("Saving datacube as: {0:s}".format(outfile))
+    final_cube = DataCube(flxcube.T, np.sqrt(varcube.T), bpmcube.T, specname, blaze_wave, blaze_spec,
+                          sensfunc=sensfunc, fluxed=fluxcal)
+    final_cube.to_file(outfile, hdr=hdr, overwrite=overwrite)
+
+
+def subpixellate(output_wcs, all_ra, all_dec, all_wave, all_sci, all_ivar, all_wghts, all_spatpos, all_specpos,
```
jhennawi: Same comment about all of these methods, which appear to be core methods.

pypeit/core/datacube.py
```diff
@@ -571,84 +349,84 @@ def extract_standard_spec(stdcube, subpixel=20, method='boxcar'):
         box_gpm = flxscl > 1/3  # Good pixels are those where at least one-third of the standard star flux is measured
```
jhennawi: Not sure I follow why you would every want to mask in this way. Boxcar is defined to be all the flux in the aperture.

pypeit/core/datacube.py
```diff
@@ -571,84 +349,84 @@ def extract_standard_spec(stdcube, subpixel=20, method='boxcar'):
...
-        waveimg = np.outer(wave.value, np.ones(nspat))
-        global_sky = np.zeros_like(box_sciimg)
-        # Find objects and then extract
-        sobj = findobj_skymask.objs_in_slit(box_sciimg, thismask, slit_left, slit_right)
-        skysub.local_skysub_extract(box_sciimg, box_sciivar, tilts, waveimg, global_sky, thismask, slit_left,
-                             slit_right, sobj, model_noise=False)
-        opt_flux, opt_var, opt_gpm = sobj.OPT_COUNTS[0,:], sobj.OPT_COUNTS_SIG[0,:]**2, sobj.OPT_MASK[0,:]
-        # Setup the return values
-        ret_flux, ret_var, ret_gpm = opt_flux, opt_var, opt_gpm
+    # elif method == 'gauss2d':
```
jhennawi: Remove?

jhennawi (changes requested):
Thanks for the hard work here @rcooke-ast. I'm happy to see the 3D coadding so well developed and am excited to give it a spin! My main comment here is that I dislike what looks like hundreds of lines of core functionality being performed in classes. This makes the control flow extremely hard to follow since it is unclear what the inputs and outputs are, and all the variables are basically global in the class. I appreciate that the coadd2d class is not the easiest to follow either, but that was counter-balanced against lots of exploitation of class inheritance (multslit vs echelle), whereas as I don't see much of that here in the coadd3d class. My opinion is that most of what you are doing in class methods, can be done in core methods that are wrapped. That would make the code base more flexible, for example if we wanted to use your code for JWST IFU coadds, but did not want to have all these fixed PypeIt object dependencies baked in. It would also make the control flow and variable scope much more clear. Please take a close look at your methods and move anything that is complex to core methods with well defined inputs variables and output variables.

pypeit/core/datacube.py
rcooke-ast: Gee... this was such a long time ago...

So, near the (spectral) edges there are masked pixels because each slice doesn't cover the same wavelength range. In this case, `flxscl` appears to be correcting for the flux that is missing (a value of 1 is no correction needed). You could also consider that the only good pixels are those with `flxscl = 1`, but it's been a while since I've looked at this.

pypeit/core/datacube.py
rcooke-ast: Yes, agreed... consider it gone... actually... I'll get rid of the `method`.

pypeit/coadd3d.py
rcooke-ast: Ah, yes, I meant to do this - good catch.

pypeit/coadd3d.py
rcooke-ast: Done

pypeit/coadd3d.py
rcooke-ast: OK - I'll move this to the parset

rcooke-ast: Right... the issue here is that it uses the parset, and then there's also a chance to override the default setting with `pypeit.inputfiles.Coadd3DFile()` for each individual frame. In the case of the `default` setting, we can remove the `.lower()`, and I've done that now, but I think the issue is that there is a conflict in the way we handle `inputfiles.py` vs the parset. In particular, the `inputfiles` approach doesn't currently do checks in the same way that the parset does... Ideally they should be unified, but they are not at the moment.

I agree that the control flow is not obvious, but it does work... probably we need to rethink this approach.

One possibility is to move the checks to `inputfiles.py`, and just do the two `validate` checks separately (once in the parset for the default values, and once again in inputfiles for the frame-specific values). The issue with this is that it's prone to bugs, since we'll need to make sure to change it in both places if we ever make updates.

pypeit/coadd3d.py
rcooke-ast: I don't think this should be a parset. In this case, `skysub_frame` is already in the parset, and sets the default usage. However, if a user wants to use a different skysub frame for different spec2d files, then we need to override the default behaviour.

pypeit/coadd3d.py
rcooke-ast: See the comments above. I haven't changed anything here until we discuss next steps.

pypeit/coadd3d.py
rcooke-ast: I think this should stay here, but some of the functions called in this section can go into core. I've really just used this to break up individual parts of the code

pypeit/coadd3d.py
rcooke-ast: This is easy to move to core, but it has PypeIt objects that need to be passed in. If you're OK with that, then I'm happy to move it to core, but I somehow feel that's going against the ethos of what core should represent.

pypeit/coadd3d.py
rcooke-ast: Same as above

pypeit/coadd3d.py
rcooke-ast: Same as above

pypeit/coadd3d.py
rcooke-ast: Sounds good - moved to core.

pypeit/coadd3d.py
rcooke-ast: Just to be clear, I'm not writing new code here, my only goal is to reorganise the code that was already written for DataCubes, but make the code look more similar to `CoAdd2D()`. The UV sensitivity function stuff here was written years ago, and I'm more than happy for it to be updated (in fact, I thought this was one of the reasons to restructure the code in the first place, because someone wanted to transition the datacube stuff to deal with the IR sensitivity stuff, so I needed to make it more logically structured and more documented!).

The question about making the sensitivity function in advance is fair, and we can perhaps generate a new script to deal with extracting standard star spectra from datacubes and making the sensitivity function in advance. In fact, given how long the IR algorithm takes, it's probably beneficial to do this. The only "corner case" that I'm aware of is when we need to make a small correction to the sensitivity function when the standard star and science frame uses a _slightly_ different grating angle. This can introduce a pseudo-linear (as a function of wavelength) sensitivity variation in the flux (due to the different grating angles). For example, imagine that you have a standard star taken on only one night with a certain grating angle. You take science observations with this exact same grating angle. These do not need a grating correction. But, if you take another observation that tried to get close to the same grating angle (but it's off slightly) then you'll either need to make a sensitivity function using a different standard star or you'll need to perform a grating correction.

We also need to make sure that the extinction correction is not applied to the sensitivity function (as it was previously). This needs to be done within the datacube routines.

doc/coadd3d.rst
```diff
@@ -155,13 +155,13 @@ Sky Subtraction
...
 derived from the science frame during the reduction. If you would like
-to turn off sky subtraction, set the following keyword argument:
+to turn off sky subtraction, set the following keyword argument (all lowercase):
 
 .. code-block:: ini
 
     [reduce]
         [[cube]]
-            skysub_frame = None
+            skysub_frame = none
```
kbwestfall, line 164: At least in terms of parsing the input parameters, we might want to handle this in the code.  I.e., in the `CubePar.validate` function, 
```
if self.data['skysub_frame'] == 'None':
    msgs.warn('skysub_frame should be all lower case.  Changing None to none.')
    self.data['skysub_frame'] = 'none'
```

Also, when looking at the validate function I noticed that you're not using the `options` functionality of `ParSet` for `method`.  You can restrict the options allowed for a given parameter so that it can only ever be one among a limited set of values.  E.g., see `frametype` in `FrameGroupPar`.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+        Utility function to compute the inverse variance cube
+        """
+        return utils.inverse(self.sig**2)
+
+    @property
+    def wcs(self):
+        """
+        Utility function to provide the world coordinate system of the datacube
+        """
+        return wcs.WCS(self.head0)
```
kbwestfall: I would recommend lazy loading both ivar and wcs, if possible.  That is, add `self._ivar` and `self._wcs` to the internals and then the ivar property method would be:
```
@property
def ivar(self):
    if self._ivar is None:
        self._ivar = utils.inverse(self.sig**2)
    return self._ivar
```
The only caveat is that, if you ever change `self.sig`/`self.head0`, you need to update `self._ivar`/`self._wcs`.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+
+
+class DARcorrection:
+    """
+    This class holds all of the functions needed to quickly compute the differential atmospheric refraction correction.
+    """
+    def __init__(self, hdr0, cosdec, spectrograph=None, wave_ref=4500.0):
```
kbwestfall: I'd rather this class not use a file header for its default construction.  I think the base constructor should just take all the parameters and then we can have a class method that instantiates from a header and spectrograph object.  E.g.:
```
class DARcorrection:
    def __init__(self, airmass, parangle, pressure, temperature, humidity, cosdec, wave_ref=4500., co2=400.):
        ...

    @classmethod
    def from_hdr(cls, hdr0, cosdec, spectrograph=None, wave_ref=4500., co2=400.):
        ...
        return cls(airmass, parangle, pressure, temperature, humidity, cosdec, wave_ref=wave_ref, co2=co2)
```
This will make the class more general and easier to test.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+
+    def __init__(self, spec2dfiles, opts, spectrograph=None, par=None, det=None, overwrite=False,
+                 show=False, debug=False):
+        """
+
+        Args:
+            spec2dfiles (:obj:`list`):
+                List of all spec2D files
+            opts (:obj:`dict`):
+                Options associated with each spec2d file
```
kbwestfall: I'd prefer that we expand out what is in this dictionary into a set of keyword parameters and/or items to add to the relevant parameter set.  Regardless, I'd like to better understand use of this dictionary is just a convenience, or if there is some need for it.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+            spectrograph (:obj:`str`, :class:`~pypeit.spectrographs.spectrograph.Spectrograph`, optional):
+                The name or instance of the spectrograph used to obtain the data.
+                If None, this is pulled from the file header.
+            par (:class:`~pypeit.par.pypeitpar.PypeItPar`, optional):
+                An instance of the parameter set.  If None, assumes that detector 1
+                is the one reduced and uses the default reduction parameters for the
+                spectrograph (see
+                :func:`~pypeit.spectrographs.spectrograph.Spectrograph.default_pypeit_par`
+                for the relevant spectrograph class).
```
kbwestfall, line 367: Is there a specific reason not to require this ParSet?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+        self.opts = opts
+        self.overwrite = overwrite
+
+        # Check on Spectrograph input
+        if spectrograph is None:
+            with fits.open(spec2dfiles[0]) as hdu:
+                spectrograph = hdu[0].header['PYP_SPEC']
+
+        if isinstance(spectrograph, str):
+            self.spec = load_spectrograph(spectrograph)
```
kbwestfall: Note that `load_spectrograph` can take a `Spectrograph` instance and simply return it.  So you can instead replace this if-else block with:
```
self.spec = load_spectrograph(spectrograph)
self.specname = self.spec.name
```
Although, you don't really need the last line.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+                msgs.info("Setting 'align' to True")
+            self.align = True
+        # TODO :: The default behaviour (combine=False, align=False) produces a datacube that uses the instrument WCS
+        #  It should be possible (and perhaps desirable) to do a spatial alignment (i.e. align=True), apply this to the
+        #  RA,Dec values of each pixel, and then use the instrument WCS to save the output (or, just adjust the crval).
+        #  At the moment, if the user wishes to spatially align the frames, a different WCS is generated.
+        # Check if fast-histogram exists
+        if histogramdd is None:
+            msgs.warn("Generating a datacube is faster if you install fast-histogram:"+msgs.newline()+
+                      "https://pypi.org/project/fast-histogram/")
```
kbwestfall: How many packages are there that are specific to IFU reductions?  I think we should collect them into an IFU option in `setup.cfg` so that people can do something like `pip install "pypeit[ifu]"`.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+            if self.cubepar['scale_corr'] == "image":
+                msgs.info("The default relative spectral illumination correction will use the science image")
+                self.scalecorr_default = "image"
+            else:
+                msgs.info("Loading default scale image for relative spectral illumination correction:" +
+                          msgs.newline() + self.cubepar['scale_corr'])
+                try:
+                    spec2DObj = spec2dobj.Spec2DObj.from_file(self.cubepar['scale_corr'], self.detname)
+                    self.relScaleImgDef = spec2DObj.scaleimg
+                    self.scalecorr_default = self.cubepar['scale_corr']
```
kbwestfall: I would limit this `try` to one line and then use an else block:
```
try:
    spec2DObj = spec2dobj.Spec2DObj.from_file(self.cubepar['scale_corr'], self.detname)
except:
    ...
else:
    self.relScaleImgDef = spec2DObj.scaleimg
    self.scalecorr_default = self.cubepar['scale_corr']
```
You might also want to collect and print the exception:
```
except Exception as e:
    msgs.warn(f'Loading spec2d file raised {type(e).__name__}: {str(e)}')
    ...
```

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+
+        Args:
+            spec2DObj (:class:`~pypeit.spec2dobj.Spec2DObj`_):
+                2D PypeIt spectra object.
+            opts_scalecorr (:obj:`str`, optional):
+                A string that describes what mode should be used for the sky subtraction. The
+                allowed values are:
+                default - Use the default value, as defined in self.set_default_scalecorr()
+                image - Use the relative scale that was derived from the science frame
+                none - Do not perform relative scale correction
```
kbwestfall: I'm guessing this doesn't format correctly.  And can this be `scalecorr` instead of `opts_scalecorr`?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+
+        Returns:
+            :obj:`str`_: A string that describes the scale correction mode to be used (see opts_scalecorr description)
+            `numpy.ndarray`_: 2D image (same shape as science frame) containing the relative spectral scaling to apply to the science frame
```
kbwestfall: This won't format correctly.  You might want to switch to numpy-style if this is the kind of return block you want.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+        """
+        this_scalecorr = self.scalecorr_default
+        relScaleImg = self.relScaleImgDef.copy()
+        if opts_scalecorr is not None:
+            if opts_scalecorr.lower() == 'default':
+                if self.scalecorr_default == "image":
+                    relScaleImg = spec2DObj.scaleimg
+                    this_scalecorr = "image"  # Use the current spec2d for the relative spectral illumination scaling
```
kbwestfall, line 621: I'm never sure of the rules, so you might want to check that this doesn't alter `self.scalecorr_default`.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+                this_scalecorr = "none"  # Don't do relative spectral illumination scaling
+            else:
+                # Load a user specified frame for sky subtraction
+                msgs.info("Loading the following frame for the relative spectral illumination correction:" +
+                          msgs.newline() + opts_scalecorr)
+                try:
+                    spec2DObj_scl = spec2dobj.Spec2DObj.from_file(opts_scalecorr, self.detname)
+                except:
+                    msgs.error(
+                        "Could not load skysub image from spec2d file:" + msgs.newline() + opts_scalecorr)
```
kbwestfall: Same comment about catching the exception and printing it.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+                      msgs.newline() + self.cubepar['skysub_frame'])
+            try:
+                spec2DObj = spec2dobj.Spec2DObj.from_file(self.cubepar['skysub_frame'], self.detname)
+                skysub_exptime = fits.open(self.cubepar['skysub_frame'])[0].header['EXPTIME']
+                self.skysub_default = self.cubepar['skysub_frame']
+                self.skyImgDef = spec2DObj.sciimg / skysub_exptime  # Sky counts/second
+                # self.skyImgDef = spec2DObj.skymodel/skysub_exptime  # Sky counts/second
+                self.skySclDef = spec2DObj.scaleimg
+            except:
+                msgs.error("Could not load skysub image from spec2d file:" + msgs.newline() + self.cubepar['skysub_frame'])
```
kbwestfall: Same comment about using `try-except-else`.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+                A string that describes what mode should be used for the sky subtraction. The
+                allowed values are:
+                default - Use the default value, as defined in self.set_default_skysub()
+                image - Use the sky model derived from the science frame
+                none - Do not perform sky subtraction
+
+        Returns:
+            :obj:`str`_: A string that describes the sky subtration mode to be used (see opts_skysub description)
+            `numpy.ndarray`_: 2D image (same shape as science frame) containing the sky frame to be subtracted from the science frame
+            `numpy.ndarray`_: 2D image (same shape as science frame) containing the relative spectral scaling that has been applied to the returned sky frame
```
kbwestfall: Same docstring comments as earlier.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+            elif opts_skysub.lower() == 'none':
+                skyImg = np.array([0.0])
+                skyScl = np.array([1.0])
+                this_skysub = "none"  # Don't do sky subtraction
+            else:
+                # Load a user specified frame for sky subtraction
+                msgs.info("Loading skysub frame:" + msgs.newline() + opts_skysub)
+                try:
+                    spec2DObj_sky = spec2dobj.Spec2DObj.from_file(opts_skysub, self.detname)
+                    skysub_exptime = fits.open(opts_skysub)[0].header['EXPTIME']
```
kbwestfall: I would move this outside the try block.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+        ref_shift_ra = self.ifu_ra[0] - self.ifu_ra
+        ref_shift_dec = self.ifu_dec[0] - self.ifu_dec
+        for ff in range(self.numfiles):
+            # Apply the shift
+            self.all_ra[self.all_idx == ff] += ref_shift_ra[ff] + self.opts['ra_offset'][ff] / 3600.0
+            self.all_dec[self.all_idx == ff] += ref_shift_dec[ff] + self.opts['dec_offset'][ff] / 3600.0
+            msgs.info("Spatial shift of cube #{0:d}:" + msgs.newline() +
+                      "RA, DEC (arcsec) = {1:+0.3f} E, {2:+0.3f} N".format(ff + 1,
+                                                                           self.opts['ra_offset'][ff],
+                                                                           self.opts['dec_offset'][ff]))
```
kbwestfall: Is it worth using astropy's SkyOffsetFrame here?  Requiring the user to include the cos(dec) factor feels awkward.   https://docs.astropy.org/en/stable/coordinates/matchsep.html

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+
+    def coadd(self):
+        """
+        Main entry routine to set the order of operations to coadd the data. For specific
+        details of this procedure, see the child routines.
+        """
+        msgs.bug("This routine should be overridden by child classes.")
+        msgs.error("Cannot proceed without coding the coadd routine.")
+        return
```
kbwestfall: Don't need the return.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+            locations = [0.0, 1.0]
+            traces = np.append(left[:, None, :], right[:, None, :], axis=1)
+        else:
+            locations = self.par['calibrations']['alignment']['locations']
+            traces = alignments.traces
+        # Generate an RA/DEC image
+        msgs.info("Generating RA/DEC image")
+        alignSplines = alignframe.AlignmentSplines(traces, locations, spec2DObj.tilts)
+        # Return the alignment splines
+        return alignSplines
```
kbwestfall: Skip the middle man: `return alignframe.AlignmentSplines(traces, locations, spec2DObj.tilts)`.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+            hdr0 = spec2DObj.head0
+            self.ifu_ra = np.append(self.ifu_ra, self.spec.compound_meta([hdr0], 'ra'))
+            self.ifu_dec = np.append(self.ifu_dec, self.spec.compound_meta([hdr0], 'dec'))
+
+            # Get the exposure time
+            # TODO :: Surely this should be retrieved from metadata... although it's coming from spec2d file?
+            exptime = hdr0['EXPTIME']
+
+            # Setup for PypeIt imports
+            msgs.reset(verbosity=2)
```
kbwestfall: Not sure why this is here and not in the calling script.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+            # Remove edges of the spectrum where the sky model is bad
+            sky_is_good = datacube.make_good_skymask(slitid_img_init, spec2DObj.tilts)
+
+            # Construct a good pixel mask
+            # TODO: This should use the mask function to figure out which elements are masked.
+            onslit_gpm = (slitid_img_init > 0) & (bpmmask.mask == 0) & sky_is_good
+
+            # Grab the WCS of this frame
+            frame_wcs = self.spec.get_wcs(spec2DObj.head0, slits, detector.platescale, wave0, dwv)
+            self.all_wcs.append(copy.deepcopy(frame_wcs))
```
kbwestfall: A couple of suggestions:

- I understand why it's convenient, but I would append the output of `self.spec.get_wcs` directly to `self.all_wcs` and then replace `frame_wcs` with `self.all_wcs[-1]` throughout the remainder.
- Barring that, I would use the `WCS.deepcopy()` convenience function instead of importing `copy.deepcopy` in case they diverge.  I generally distrust use of `deepcopy` on classes.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+            wave_sort = wave_ext[wvsrt]
+            dwav_sort = dwav_ext[wvsrt]
+            ra_sort = ra_ext[wvsrt]
+            dec_sort = dec_ext[wvsrt]
+            # Here's an array to get back to the original ordering
+            resrt = np.argsort(wvsrt)
+
+            # Compute the DAR correction
+            cosdec = np.cos(np.mean(dec_sort) * np.pi / 180.0)
+            darcorr = DARcorrection(spec2DObj.head0, cosdec, spectrograph=self.spec)
```
kbwestfall: If you follow the suggestion above, this would become:
```
darcorr = DARCorrection.from_hdr(spec2dObj.head0, cosdec, spectrograph=self.spec)
```

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+
+            # Perform extinction correction
+            msgs.info("Applying extinction correction")
+            longitude = self.spec.telescope['longitude']
+            latitude = self.spec.telescope['latitude']
+            airmass = spec2DObj.head0[self.spec.meta['airmass']['card']]
```
kbwestfall: Get this from `darcorr`?  Or use `self.spec.get_meta_value`?

pypeit/coadd3d.py
kbwestfall: Personally, I like this as it is.

pypeit/coadd3d.py
kbwestfall: Ask 3 pypeit developers "What is `core` for?" and be prepared for 5 different answers.  That being said, there are plenty of examples of high-level pypeit class objects being passed to functions in `core`.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+        outfile_resid = outfile.replace(".fits", "_resid.fits")
+        msgs.info("Saving residuals datacube as: {0:s}".format(outfile_resid))
+        hdu = fits.PrimaryHDU(residcube.T, header=hdr)
+        hdu.writeto(outfile_resid, overwrite=overwrite)
+    else:
+        flxcube, varcube, bpmcube = subpix
+
+    # Get wavelength of each pixel, and note that the WCS gives this in m, so convert to Angstroms (x 1E10)
+    nspec = flxcube.shape[2]
+    wave = 1.0E10 * output_wcs.spectral.wcs_pix2world(np.arange(nspec), 0)[0]  # The factor 1.0E10 convert to Angstroms
```
kbwestfall: Why is the wcs not constructed in angstroms? or nm? or micron?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+        # Get the output filename for the white light image
+        out_whitelight = datacube.get_output_whitelight_filename(outfile)
+        whitelight_img = datacube.make_whitelight_fromcube(flxcube, wave=wave, wavemin=whitelight_range[0], wavemax=whitelight_range[1])
+        msgs.info("Saving white light image as: {0:s}".format(out_whitelight))
+        img_hdu = fits.PrimaryHDU(whitelight_img.T, header=whitelight_wcs.to_header())
+        img_hdu.writeto(out_whitelight, overwrite=overwrite)
+
+    # Write out the datacube
+    msgs.info("Saving datacube as: {0:s}".format(outfile))
+    final_cube = DataCube(wave, flxcube.T, np.sqrt(varcube.T), bpmcube.T, specname, blaze_wave, blaze_spec,
```
kbwestfall: I think I've mentioned this before, but given my experience with MaNGA, I'm really wary of the costs of transposing things.  There is likely an adopted convention for the WCS that we should match.  Does this do that?  Are there other useful conventions from either from what the Keck KCWI DRP produced or what JWST is producing for their NIRSpec and MIRI datacubes?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+            flux should have mean=0 and std=1.
+
+    Returns:
+        :obj:`tuple`: Three or four `numpy.ndarray`_ objects containing (1) the
+        datacube generated from the subpixellated inputs, (2) the corresponding
+        variance cube, (3) the corresponding bad pixel mask cube, and (4) the
+        residual cube.  The latter is only returned if debug is True.
+    """
+    # Check for combinations of lists or not
+    if type(tilts) is list and type(slits) is list and type(astrom_trans) is list and type(all_dar) is list:
```
kbwestfall: I would use `isinstance` instead.  E.g.
```
if all([isinstance(l, list) for l in [tilts, slits, astrom_trans, all_dar]]):
```

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1803 @@
...
+                msgs.error("Missing required argument for combining frames: all_idx")
+            else:
+                all_idx = np.zeros(all_sci.size)
+        else:
+            tmp = np.unique(all_idx).size
+            if tmp != numframes:
+                msgs.warn("Indices in argument 'all_idx' does not match the number of frames expected.")
+        # Store in the following variables
+        _tilts, _slits, _astrom_trans, _all_dar = tilts, slits, astrom_trans, all_dar
+    elif type(tilts) is not list and type(slits) is not list and \
```
kbwestfall: `not isinstance(...)`

pypeit/core/datacube.py
```diff
@@ -695,6 +381,54 @@ def make_good_skymask(slitimg, tilts):
...
+
+    Args:
+        outfile (str):
+            The output filename used for the datacube.
+
+    Returns:
+        str: The output filename to use for the whitelight image.
+    """
+    out_wl_filename = os.path.splitext(outfile)[0] + "_whitelight.fits"
+    return out_wl_filename
```
kbwestfall: Skip the middle man.

pypeit/core/datacube.py
```diff
@@ -695,6 +381,54 @@ def make_good_skymask(slitimg, tilts):
...
+        # Check the output files don't exist
+        outfile = par_outfile if ".fits" in par_outfile else par_outfile + ".fits"
+    else:
+        if par_outfile == "":
+            outfile = fil.replace("spec2d_", "spec3d_")
+        else:
+            # Use the output filename as a prefix
+            outfile = os.path.splitext(par_outfile)[0] + "_{0:03d}.fits".format(idx)
+    # Return the outfile
+    return outfile
```
kbwestfall: return is your friend:
```
if combine:
    if par_outfile = '':
        par_outfile = 'datacube.fits'
    return par_outfile if '.fits' in par_outfile else f'{par_outfile}.fits'
if par_outfile == '':
    return fil.replace('spec2d_', 'spec3d_')
return os.path.splitext(par_outfile)[0] + f'_{idx:03}.fits'
```

pypeit/coadd3d.py
rcooke-ast: Done, I've moved this to `core.extract`, and will make some changes to other parts of the code as well, since this same function gets used quite a bit throughout PypeIt.

pypeit/coadd3d.py
rcooke-ast: I've made some additional changes, so that there is always something returned and I've moved some additional code to core. Also, many of the lines of code here are already calling core routines.

pypeit/coadd3d.py
rcooke-ast: Again, pretty much every line of code here _is_ calling a core routine (the core routines are in `datacube`). If you have some specific lines of code you'd like me to clear up, please be more specific. For now, I'm just going to return `all_wghts`.

pypeit/coadd3d.py
rcooke-ast: I disagree. This is effectively an `__init__` method (loading the data), so I've moved it there

pypeit/coadd3d.py
rcooke-ast: OK - more documentation added

rcooke-ast (commented):
@jhennawi's comments are addressed.

pypeit/core/datacube.py
```diff
@@ -660,6 +280,72 @@ def extract_standard_spec(stdcube, subpixel=20, method='boxcar'):
...
+                                 nresln=senspar['UVIS']['nresln'],
+                                 resolution=senspar['UVIS']['resolution'],
+                                 trans_thresh=senspar['UVIS']['trans_thresh'],
+                                 polyorder=senspar['polyorder'],
+                                 polycorrect=senspar['UVIS']['polycorrect'],
+                                 polyfunc=senspar['UVIS']['polyfunc'])
+    wgd = np.where(zeropoint_fit_gpm)
+    sens = np.power(10.0, -0.4 * (zeropoint_fit[wgd] - flux_calib.ZP_UNIT_CONST)) / np.square(wave[wgd])
+    flux_spline = interp1d(wave[wgd], sens, kind='linear', bounds_error=False, fill_value="extrapolate")
+    return flux_spline
```
kbwestfall: Skip the middle man.

pypeit/core/datacube.py
```diff
@@ -829,120 +563,61 @@ def load_imageWCS(filename, ext=0):
...
-    # Determine number of files
-    numfiles = np.unique(all_idx).size
-
-    if whitelightWCS is None:
-        # Generate a 2D WCS to register all frames
-        coord_min = [np.min(all_ra), np.min(all_dec), np.min(all_wave)]
-        coord_dlt = [dspat, dspat, np.max(all_wave) - np.min(all_wave)]
-        whitelightWCS = generate_WCS(coord_min, coord_dlt)
+        `numpy.ndarray`_: A new set of RA values that have been aligned
+        `numpy.ndarray`_: A new set of Dec values that has been aligned
```
kbwestfall: This won't format correctly.

pypeit/core/extract.py
```diff
@@ -451,6 +451,51 @@ def extract_boxcar(sciimg, ivar, mask, waveimg, skyimg, spec, fwhmimg=None, base
...
+            The frame to use to extract a spectrum. Shape should be the same as waveimg
+        gpm (`numpy.ndarray`_, optional):
+            A boolean array indicating the pixels to include in the histogram (True = include)
+        bins (`numpy.ndarray`_, int, optional):
+            Either a 1D array indicating the bin edges to be used for the histogram,
+            or an integer that specifies the number of bin edges to generate
+
+    Returns:
+        `numpy.ndarray`_: The wavelength at the centre of each histogram bin
+        `numpy.ndarray`_: The spectrum at each pixel of the returned wavelength array
```
kbwestfall: Formatting.

pypeit/core/extract.py
```diff
@@ -451,6 +451,51 @@ def extract_boxcar(sciimg, ivar, mask, waveimg, skyimg, spec, fwhmimg=None, base
...
+    """
+    # Check the inputs
+    if waveimg.shape != frame.shape:
+        msgs.error("Wavelength image is not the same shape as the input frame")
+    # Check the GPM
+    _gpm = gpm if gpm is not None else waveimg > 0
+    if waveimg.shape != _gpm.shape:
+        msgs.error("Wavelength image is not the same shape as the GPM")
+    # Set the bins
+    if type(bins) is int:
```
kbwestfall: Use `isinstance`

pypeit/core/meta.py
```diff
@@ -143,22 +143,28 @@ def define_additional_meta(nlamps=20):
...
-                       'dithoff': dict(dtype=float, comment='Dither offset'),
+                       'posang': dict(dtype=float, comment='Position angle of the observation (degrees, positive is East from North)'),
+                       'ra_off': dict(dtype=float, comment='Dither offset in RA'),
+                       'dec_off': dict(dtype=float, comment='Dither offset in DEC'),
                        'echangle':dict(dtype=float, comment='Echelle angle'),
                        'filter1': dict(dtype=str, comment='First filter in optical path'),
                        'frameno': dict(dtype=str, comment='Frame number provided by instrument software'),
                        'hatch': dict(dtype=str, comment='Position of instrument hatch'),
-                       'humidity': dict(dtype=float, comment='Relative humidity (0 to 1) at observation time'),
+                       'humidity': dict(dtype=float, comment='Humidity at observation time (as a percentage, not a fraction)'),
```
kbwestfall, line 159: Just flagging this to make sure that we caught these unit changes throughout the rest of the code: humidity from a fraction to a percentage, pressure from bar to pascal.

pypeit/core/ref_index.py
```diff
@@ -0,0 +1,738 @@
+"""
```
kbwestfall: I worry about wholesale copying of code from another repository.  Credit is given, and it doesn't look like the repository is well maintained, so I'm not saying this is unreasonable.  But is there another well-maintained version of this?  astropy has some functionality for calculating DAR (e.g., https://docs.astropy.org/en/stable/api/astropy.coordinates.AltAz.html), but the documentation is not great.  I played around with it a long time ago, so I might be able to dig up some code so that we can compare results with this package.  What does the Keck KCWI pipeline use?

pypeit/find_objects.py
```diff
@@ -42,7 +42,7 @@ class FindObjects:
             Specifies object being reduced.  Should be 'science',
             'standard', or 'science_coadd2d'.
         wv_calib (:class:`~pypeit.wavecalib.WaveCalib`, optional):
-            This is only used for the IFU child when a joint sky subtraction
+            This is only used for the SlicerIFU child when a joint sky subtraction
```
kbwestfall: Give the class reference?
```
:class:`SlicerIFUFindObjects`
```

pypeit/find_objects.py
```diff
@@ -471,7 +471,7 @@ def get_platescale(self, slitord_id=None):
 
         Args:
             slitord_id (:obj:`int`, optional):
-                slit spat_id (MultiSlit, IFU) or ech_order (Echelle) value
+                slit spat_id (MultiSlit, SlicerIFU) or ech_order (Echelle) value
```
kbwestfall: Same comment here.

pypeit/par/pypeitpar.py
```diff
@@ -1591,9 +1588,20 @@ def from_dict(cls, cfg):
         return cls(**kwargs)
 
     def validate(self):
-        allowed_methods = ["subpixel", "NGP"]#, "resample"
+        # Check the method options
+        allowed_methods = ["subpixel", "ngp"]
```
kbwestfall: Move these to options for the `method` parameter.

pypeit/spectrographs/gemini_gnirs.py
```diff
@@ -135,21 +135,34 @@ def compound_meta(self, headarr, meta_key):
                 return 0.0
         elif meta_key == 'pressure':
             try:
-                return headarr[0]['PRESSURE'] * 0.001  # Must be in astropy.units.bar
+                return headarr[0]['PRESSUR2']  # Must be in astropy.units.pascal
```
kbwestfall: Is the change to `PRESSUR2` (with the `2` at the end) on purpose?

pypeit/spectrographs/gemini_gnirs.py
```diff
@@ -135,21 +135,34 @@ def compound_meta(self, headarr, meta_key):
...
             except KeyError:
                 msgs.warn("Pressure is not in header")
-                return 0.0
+                msgs.info("The default pressure will be assumed: 61.1 kPa")
```
kbwestfall: Here and a few places you have paired `warn` then `info` statements.  Can we combine them?

pypeit/spectrographs/gemini_gnirs.py
```diff
@@ -135,21 +135,34 @@ def compound_meta(self, headarr, meta_key):
...
+                # Humidity expressed as a percentage, not a fraction
                 return headarr[0]['HUMIDITY']
             except KeyError:
                 msgs.warn("Humidity is not in header")
+                msgs.info("The default relative humidity will be assumed: 20 %")
+                return 20.0  # van Kooten & Izett, arXiv:2208.11794
+        elif meta_key == 'parangle':
+            try:
+                # Humidity expressed as a percentage, not a fraction
+                msgs.work("Parallactic angle is not available for GNIRS - DAR correction may be incorrect")
```
kbwestfall: Personally, as an outward facing software package, I don't think we should have these "work in progress" statements.  I would change this to a warning.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -569,18 +598,20 @@ def get_wcs(self, hdr, slits, platescale, wave0, dwv, spatial_scale=None):
...
-        else:
-            rref = 0.
+        #if 'ROTPOSN' in hdr:
+        #    rpos = hdr['ROTPOSN']
+        #else:
+        #    rpos = 0.
+        #if 'ROTREFAN' in hdr:
+        #    rref = hdr['ROTREFAN']
+        #else:
+        #    rref = 0.
```
kbwestfall: remove commented code?

pypeit/tests/test_ref_index.py
```diff
@@ -0,0 +1,177 @@
+"""
```
kbwestfall: Nice!  ...  Although see my questions about using this package.

kbwestfall (changes requested):
Lots of mostly minor comments from me.

doc/coadd3d.rst
rcooke-ast: As discussed on Slack, I don’t think that we can use the options feature here, because users can specify an arbitrary spec2d filename. Also, the key confusion is that if a user provides “None” with a capital “N”, then this is interpreted as NoneType, which reverts to the default value. The only possibility is to use lowercase “none”, which isn’t captured as NoneType.

pypeit/core/extract.py
rcooke-ast: Done!

pypeit/coadd3d.py
rcooke-ast: I can see the benefit of not computing `ivar` repeatedly, but I worry that we'll then be carrying around `self._ivar` and `self.ivar`, which seems a little messy to me. I also suspect it's rare that the `ivar` will ever be used... and it's also pretty quick to generate `self.wcs`. Also, is there a point in having the `@property` and function in this case? Couldn't we just have `self.ivar = utils.inverse(self.sig**2)` when a DataCube gets loaded?

Overall, I'm tempted to leave things the way they are, but it might be that I'm not appreciating the full benefit of this, so if you think this really ought to be done, then should it be done in the `from_file()` routine?

pypeit/coadd3d.py
rcooke-ast: That's a good idea - I've made the relevant updates!

doc/coadd3d.rst
rcooke-ast: As per slack, I've fixed the `method` options.

pypeit/coadd3d.py
rcooke-ast: Good point - there isn't, so I've made that a required input.

pypeit/coadd3d.py
rcooke-ast: We can expand it out, and I've done so.

pypeit/coadd3d.py
rcooke-ast: Cool - that's much cleaner - done!

pypeit/coadd3d.py
rcooke-ast: I think it's just this one, and it's a very small package (but requires some c extensions)... Maybe we should just add it as a dependency for PypeIt in general? Actually, I think this could be important for CoAdd2D (and pretty much any histogram in PypeIt that uses equal bin sizes). Another possibility is to just grab the code and put it directly in PypeIt. Let me know what you prefer.

I'm leaning towards making it a required dependency.

pypeit/coadd3d.py
rcooke-ast: You're right :-) I've updated the docstring, and the argument

pypeit/coadd3d.py
rcooke-ast: That's much cleaner - done!

pypeit/coadd3d.py
rcooke-ast: Checked, and it's fine (`scalecorr_default` remains the same when `this_scalecorr` changes).

pypeit/coadd3d.py
rcooke-ast: Haha... I often just copy a docstring that's already formatted in PypeIt, so if I'm getting it wrong, it's likely because I've copied some other function that's getting it wrong!

On the developer documentation it says that we prefer the "Google" style docstrings, so I've been preferentially trying to do this instead of the numpy ones, but I do find the return stuff of the Google format to be a little vague. Nevertheless, I think I've fixed the formatting here!

pypeit/coadd3d.py
rcooke-ast: Done!

pypeit/coadd3d.py
rcooke-ast: Done!

pypeit/coadd3d.py
rcooke-ast: Fixed!

pypeit/coadd3d.py
rcooke-ast: But in this case both are to do with file loading, so isn't it preferred in this case to keep both in the `try` statement?

pypeit/coadd3d.py
rcooke-ast: Indeed  👍

pypeit/coadd3d.py
rcooke-ast: Good catch - done.

pypeit/coadd3d.py
rcooke-ast: Yep - now get this from above (whent he DAR correction is being called.

pypeit/coadd3d.py
rcooke-ast: I now follow your other suggestion, to pass each individual parameter in separately.

pypeit/coadd3d.py
rcooke-ast: I'm also not sure why this is here... the same thing is in the calling script already (and takes advantage of the verbosity that's actually fed in by the user)... Good catch - I'm removing this line.

pypeit/coadd3d.py
rcooke-ast: I think that's spot on!  In this case, given that this functionality is going to be used exclusively by `CoAdd3D()`, I think it's best to keep it in this routine (otherwise in `pypeit.core.datacube` we'll be importing `pypeit.coad3d.DataCube()`, and in `pypeit.coadd3d` we'll be importing `pypeit.core.datacube`). So, let's keep it here...

pypeit/coadd3d.py
rcooke-ast: Hmmm... it is, everywhere it's set with `CUNIT=Angstrom`, but it somehow defaults back to `units.m`... Instead of putting 1.0E10 in several places throughout the code without checking units, I've now updated this so it looks at the units first.

pypeit/coadd3d.py
rcooke-ast: Good suggestion - much cleaner

pypeit/coadd3d.py
rcooke-ast: I've opted for something similar to your suggestion above.

pypeit/core/datacube.py
rcooke-ast: Done!

pypeit/core/datacube.py
rcooke-ast: Much cleaner - thanks!

pypeit/core/datacube.py
rcooke-ast: Done!

pypeit/core/datacube.py
rcooke-ast: Fixed.

pypeit/core/extract.py
rcooke-ast: Fixed

pypeit/core/meta.py
rcooke-ast: Yes - we've got it all, although PR #1647 uses different units for pressure (mbar). I'm not sure if it matters whether or not we are consistent, but it probably makes sense.

I've changed all units here so that it's consistent with PR #1647, and everything in PypeIt is consistent now, I believe. I've also checked that the results are unchanged from what they were before.

pypeit/core/ref_index.py
rcooke-ast: I feel the same... we can always add it as another dependency, and I think that would be fine. I checked in PR #1679 that this routine is working well for the purposes of our pipeline. The official Keck KCWI DRP also use `ref_index`. I was previously using the astropy version but it's very slow for our purposes. This repository is effectively just implementing the information on [this website](https://emtoolbox.nist.gov/Wavelength/Documentation.asp).

pypeit/find_objects.py
rcooke-ast: OK - done!

pypeit/find_objects.py
rcooke-ast: Done (I've also updated the MultiSlit and Echelle, as well.)

pypeit/par/pypeitpar.py
rcooke-ast: Good idea - done! There is also a bug here, which is now fixed.

pypeit/spectrographs/gemini_gnirs.py
rcooke-ast: Thanks for checking - yes, this is correct. The header stores the pressure in two different units.

pypeit/spectrographs/gemini_gnirs.py
rcooke-ast: OK - I'll just throw the warning message.

pypeit/spectrographs/gemini_gnirs.py
rcooke-ast: Done

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Yes, good call.

pypeit/tests/test_ref_index.py
rcooke-ast: Sounds good - let's discuss this on the slack to decide the best approach.

pypeit/coadd3d.py
rcooke-ast: I'm not sure what JWST does. I just tried to look through the KCWI DRP code and they appear to generate cubes differently to us... they don't transpose, afaik. I share your concern though about transposing. One thing we could do (and is perhaps the most sensible option) is to generate the cubes in transposed form, that way, we wouldn't need to transpose at the end, we'd just need to reverse the bin order on the histograms. But, I'd need to do a bunch of tests to make sure I didn't mess anything up. If it's OK for now, I'll raise an issue (#1699) on this, and come back to this with a later (dedicated) PR...

rcooke-ast (commented):
Addressed PR comments

pypeit/coadd3d.py
jhennawi: If the code base isn't changing, maybe we should just grab it to avoid annoyances in installation.

pypeit/coadd3d.py
jhennawi: I'm fine with leaving this as for now @rcooke-ast but can you at least make an issue for these changes. It does not make sense to me to have sensitivity function computation as part of a coadd algorithm.

pypeit/coadd3d.py
jhennawi: Okay as a compromise, please add the following to the documentation to make the control flow more clear. Add a list of all of the class attributes that are assigned by this function, i.e. all the self.X assignments. I agree that return values would be combersome and dilly, but it is needs to be clear what attributes this method assigns withoiut having to read through 500 lines of code.

pypeit/coadd3d.py
jhennawi: I still think  that there is no reason to have this method in this class since it does not specifically depend on complicated objects or specific methods for this class. The only class method you are calling is create_wcs which itself wraps a core method. From what I see here,  you could easily make this function work with lists of standard python inputs i.e. lists and arrays.  That would involve a longer set of argument lists, but would also be more transparent and more generalizable. For example, suppose I want to compute weights for a set of cubes for  fiber IFU or a JWST IFU, but for whatever reason cannot perform all the operations in this class.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1853 @@
...
+            specstr = ", ".join(["{0:.6f}".format(ss) for ss in self._specscale])
+            msgs.info("Wavelength samplings of all input frames:" + msgs.newline() + specstr)
+
+        # If the user has not specified the spatial scale, then set it appropriately now to the largest spatial scale
+        if self._dspat is None:
+            self._dspat = np.max(self._spatscale)
+            msgs.info("Adopting a square pixel spatial scale of {0:f} arcsec".format(3600.0 * self._dspat))
+        # If the user has not specified the spectral sampling, then set it now to the largest value
+        if self._dwv is None:
+            self._dwv = np.max(self._specscale)
```
jhennawi: Can you return these values instead of assigning to self. For a small number of return values that makes the control flow more transparent.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1853 @@
...
+        sensfunc = None
+        if self.flux_spline is not None:
+            # Get wavelength of each pixel
+            numwav = vox_edges[2].size - 1
+            wcs_scale = (1.0 * cube_wcs.spectral.wcs.cunit[0]).to(units.Angstrom).value  # Ensures the WCS is in Angstroms
+            senswave = wcs_scale * cube_wcs.spectral.wcs_pix2world(np.arange(numwav), 0)[0]
+            sensfunc = self.flux_spline(senswave)
+
+        # Generate a datacube
+        outfile = datacube.get_output_filename("", self.cubepar['output_filename'], True, -1)
```
jhennawi: My advice is to create the this outfile name int he __init__ method or before coadd is called. And then have self.coadd(outfile) be the calling sequence, to make it clear that this coadds and writes to disk.

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1853 @@
...
+        # No need to continue if we are not combining nor aligning frames
+        if not self.combine and not self.align:
+            return
+
+        # If the user is aligning or combining, the spatial scale of the output cubes needs to be consistent.
+        # Set the spatial and spectral scales of the output datacube
+        self.set_voxel_sampling()
+
+        # Align the frames
+        if self.align:
```
jhennawi, line 1307: Is this a bug, run_align now returns ra and dec?

pypeit/coadd3d.py
```diff
@@ -0,0 +1,1853 @@
...
+          Note, there are several algorithms used to combine multiple frames. Refer to the subpixellate() routine for
+          more details about the combination options.
+        """
+        # No need to continue if we are not combining nor aligning frames
+        if not self.combine and not self.align:
+            return
+
+        # If the user is aligning or combining, the spatial scale of the output cubes needs to be consistent.
+        # Set the spatial and spectral scales of the output datacube
+        self.set_voxel_sampling()
```
jhennawi: See my comments above, set_voxel_sampling should return something.  As a general rule,  control flows like 

self.do_x()
self.do_y()
self.do_z()

Are not only inscrutable, but often  do not manage variable scope and are prone to things changing accidentally changing behind the scenes, like if x changes variables that are then required for y. I know that everything is an attribute of the object, but actually that should not really be the case. The fewer global variables the better.

pypeit/coadd3d.py
jhennawi: I'm fine with this method remaining in the class, but am pushing back on self.compute_weights which I think should be in core. Please also see my comment about return values to make the control flow more clear.

pypeit/coadd3d.py
jhennawi: I'm fine with this being here, as it is not a class method and avoids circular imports.

pypeit/coadd3d.py
jhennawi: Fine.

jhennawi (changes requested):
Thanks for the major refactor @rcooke-ast. I have only a few relatively minor comments. Only one place self.compute_weights, where I think it makes more sense for a method to be in core. Many thanks!

pypeit/coadd3d.py
rcooke-ast: It appears that the code has still a few live issues, a pull request, and multiple branches. I'm tempted to make this a dependency. I have now done so.

pypeit/coadd3d.py
rcooke-ast: OK - that sounds like a good idea. It will make sense to address this in a separate PR, anyway.

pypeit/coadd3d.py
rcooke-ast: Good catch - I agree 👍  Actually, I decided to move this entire section of code to core.

pypeit/coadd3d.py
rcooke-ast: Sounds good - If I understand correctly, I've added this information to the docstring.

pypeit/coadd3d.py
rcooke-ast: Yep - good catch - as mentioned above, I've moved this function to core.

pypeit/coadd3d.py
rcooke-ast: Yes, it is a bug - great catch! Fixed...

pypeit/coadd3d.py
rcooke-ast: This is a good idea, but there is also a method here that writes out separate datacubes (when `align=True` and `combine=False`). I've just realised that this is confusing, given that the routine is called "coadd"... So I've renamed the routine to `run()`, which I think is more consistent with the rest of PypeIt anyway.

It has made me realise though that this routine is more about making datacubes, and not necessarily "coadding"... Coadding is just one possible outcome. So I wonder if this might be confusing, and we should name it something else (perhaps `spec3dobj` is more apt)?

pypeit/coadd3d.py
rcooke-ast: I do agree, and I started to rework this, but then realised that this routine calls `generate_image_subpixel`, which is located at the bottom of this `coadd3d.py` file (outside of this class). It required a big change throughout a lot of the code (and avoid circular imports), and I've now moved considerably more code over to core (for better or worse).

pypeit/coadd3d.py
rcooke-ast: Good point - done!

pypeit/coadd3d.py
rcooke-ast: This was a painful change, but I think it's all there now.

rcooke-ast (commented):
OK, I think that addresses all comments of @jhennawi and @kbwestfall. I'll run the dev suite again to make sure I haven't broken anything in the process!

rcooke-ast:
Tests passing...
image

pypeit/coadd3d.py
kbwestfall: In trying to respond to this, I actually learned some new things about the `@property` decorator!  Anyway, the root of my comment was to try to limit the number of times we compute `ivar` or build `wcs`.  To me, this is the risk with `property` decorators because, even though they look like you're simply accessing something stored in memory when you see `self.ivar` elsewhere in the code, you're actually performing a function and allocating new memory.  If it's possible, yes, I would suggest just allocating them once, but this isn't necessarily straight-forward...

pypeit/coadd3d.py
kbwestfall: Yeah, I think I recently removed the statement that we prefer Google-style (I might be the only one!).  You should feel free to use either.

pypeit/coadd3d.py
kbwestfall: Ah, amazingly, I didn't realize we were reading the file again.  Is the header (or simply just the exposure time) not kept as part of the `Spec2DObj`?  Can we add it there?  I really dislike when we open files multiple times.

pypeit/coadd3d.py
kbwestfall: Thoughts on this?

pypeit/coadd3d.py
kbwestfall: Sounds good.

kbwestfall (approved):
I had a couple of lingering comments/responses, but I'm happy to approve now.

pypeit/coadd3d.py
rcooke-ast: Yes, that is a good point - it's included in the header 🤦‍♂️ so I've done that now... thanks for spotting that! Fixed throughout.

pypeit/coadd3d.py
rcooke-ast: Sorry - I somehow missed this during the first pass!

This offset is usually something that the user sets at the telescope (i.e. it's usually a dither or sky offset - Like moving the telescope "5 arcsec East and 2 arcsec North", which already includes the cos(dec) factor). I think this is more practical than specifying two separate RA/Dec combinations (which is what appears to be needed by the SkyOffsetFrame).

Perhaps I need to clarify the docs on this?

pypeit/coadd3d.py
rcooke-ast: Ahh, I think I see. No one will actually no about `self._ivar`, it will just be a convenience variable so that when the `@property` ivar is called, it doesn't (necessarily) compute the `1/sig**2`. I think this is a reasonable compromise. I've implemented the change you have suggested, and I initialise `self._ivar=None`. Same for the WCS

rcooke-ast (commented):
Thanks again @kbwestfall - all comments addressed... I think!

jhennawi (approved):
Thanks for all the hard work here @rcooke-ast

PR #1673, opened 2023-09-15
Stop collate1d test download

badpandabear:
Fix issues found when CI tests fail to download the archived sensitivity functions:

- `test_collate_1d::test_flux` was only intended to test `collate_1d.py`'s logic, not fluxing logic. But it still would try to download the sensitivity function. 
- `collate_1d.py::flux()` did not properly skip a file if there was an error downloading the sensitivity function.

Verified unit tests passed and any collate_1d related devsuite tests passed. If requested I can do a full dev-suite run.


profxj (approved):
thx for this

kbwestfall (approved):
Thanks!

PR #1679, opened 2023-09-21
Bugfix for differential atmospheric refraction correction

rcooke-ast:
I have downloaded and reduced >90 standard stars observed with KCWI spanning the last 6 years to check the DAR correction implemented in PypeIt. This covers a large range of position angles, airmasses, and atmospheric conditions. The plots below show to offsets in arcsec to the RA (red) and dec (blue). I have horizontally shifted the RA values by -90 degrees so the corrections appear to line up. The correction is evaluated at 3700A, and is computed relative to 4500A.

This plot does not have the DAR correction implemented:
image

This plot shows the results with the new DAR correction algorithm, which is both faster and more accurate than the previously implemented DAR correction. This plot is shown on the same y-axis scale as the one above.
image

There is still some work to be done for observations that don't use position angle (i.e. "stationary" mode), and both GNIRS and GTC/OSIRIS+/MAAT do not include the parallactic angle in the header information (KCWI does)... so we will need to calculate that manually based on other information in the header...


pypeit/find_objects.py
```diff
@@ -1244,6 +1244,11 @@ def global_skysub(self, skymask=None, update_crmask=True, trim_edg=(0,0),
         correction using the sky spectrum, if requested. See Reduce.global_skysub()
         for parameter definitions.
         """
+        # Skip the sky subtraction, if requested by the user
```
jhennawi: It is not ideal for this to be hidden here. To make the control flow more transparent, can you move this into the routine that calls global_skysub in the main control flow.

jhennawi (changes requested):
Hi @rcooke-ast minor changed requested here. Thanks for the hard work!

pypeit/find_objects.py
rcooke-ast: Yep, sounds good - done!

jhennawi (approved):
Approved.

PR #1685, opened 2023-09-27
Remove use of `pkg_resources`

kbwestfall:
As titled.  `pkg_resources` is being deprecated, so this changes the relevant functions to use `importlib` instead.  Changes mostly touch the scripts used for automated doc construction, but it also touches the remote file download/caching system.  I'll run the dev-suite just to be sure.


tbowers7 (approved):
Thanks for getting on top of the deprecated `pkg_resources` API.  Looks good to me.

profxj (approved):
nice work and much appreciated.

doc/conf.py
```diff
@@ -12,21 +12,20 @@
...
 
 # Get configuration information from setup.cfg
-from configparser import ConfigParser
+pypeit_root = resources.files("pypeit").parent
 conf = ConfigParser()
-
-conf.read([os.path.join(os.path.dirname(__file__), '..', 'setup.cfg')])
-setup_cfg = dict(conf.items('metadata'))
+conf.read([pypeit_root / "setup.cfg"])
+setup_cfg = dict(conf.items("metadata"))
```
tbowers7: @kbwestfall : Be sure this change is compatible with RTD... I used the same syntax on a package of my own (GH pages rather than RTD) and needed to revert to the `os.path...` version because `resources.files()` pulls the installed package location (somewhere in `site-packages`) rather than the source code location (which includes the desired `setup.cfg` file).

tbowers7: I am reverting this problematic change that I introduced.

tbowers7:
@kbwestfall : The reversions in 4d89d3ead8f72b0995ce60b54f59cf53189e364b take these files back to before my #1687 PR.

kbwestfall:
Running the dev-suite again...

doc/conf.py
```diff
@@ -18,7 +18,7 @@
 
 # Get configuration information from setup.cfg
 conf = ConfigParser()
-conf.read([pathlib.Path(__file__).resolve.parents[1] / 'setup.cfg'])
+conf.read([pathlib.Path(__file__).resolve().parents[1] / 'setup.cfg'])
```
tbowers7, line 21: Whoops, sorry.  That was me.

doc/scripts/write_script_help.py
```diff
@@ -27,7 +27,7 @@ def write_help(script_cls, opath, width=80):
 
     pypeit_root = resources.files('pypeit').parent
     path = pypeit_root / 'doc' / 'help'
-    if path.is_dir():
+    if not path.is_dir():
```
tbowers7, line 30: Whoops, sorry.  That was me.

PR #1687, opened 2023-09-28
Going all-in on `pathlib`

tbowers7:
In the `doc/scripts` directory, replace all possible `os` calls with their equivalent `pathlib` call.  Also, changed string formatting to f-strings where applicable.

In `pypeit.data.utils`, greatly simplified the Path class in addition to shifting from `pkg_resources` to `importlib`.


doc/conf.py
```diff
@@ -12,13 +12,14 @@
...
-conf = ConfigParser()
+from importlib import metadata, resources
 
-import os
-conf.read([os.path.join(os.path.dirname(__file__), '..', 'setup.cfg')])
-setup_cfg = dict(conf.items('metadata'))
+# Get configuration information from setup.cfg
+pypeit_root = resources.files("pypeit").parent
+conf = ConfigParser()
+conf.read([pypeit_root / "setup.cfg"])
```
kbwestfall, line 21: Nice catch on this!

kbwestfall (approved):
Looks good to me!  Thanks for digging in.

doc/conf.py
tbowers7: Turns out this was wrong.  The `resources.files()` looks in the installed package directory (if running in CI, is somewhere in a `site-packages`) rather than the source code directory.  For editable installations (_e.g._, `pip install -e .`), these amount to the same thing, which is why I didn't catch it initially.

Issue #1686, opened 2023-09-28
Inconsistency with cutoff date for Keck LRIS red spectrographs


amasegian:
The PypeIt source code identifies 2021-04-15 as the date that the Keck LRIS red spectrograph switched to `keck_lris_red_mark4`. I was (un)lucky enough to be reducing a spectrum taken on this exact date with the old spectrograph (`keck_lris_red`). During calibrations, I got the following error:
```
keck_lris.py 1182 get_detector_par() - This is the new detector.  Use keck_lris_red_mark4
```
But of course, running `pypeit_setup` with `-s keck_lris_red_mark4` produces this error:
```
[ERROR]   :: This is not the correct spectrograph. Use keck_lris_red instead.
```
which is expected since the data were taken with `keck_lris_red`, NOT `keck_lris_mark4`.

I traced the error back to line 1168 of `keck_lris.py`:
```
t2021_upgrade = time.Time("2021-04-15", format='isot')
```
This date should probably be updated to accommodate the fact that it is possible to have data taken with the old spectrograph on 2021-04-15 (and potentially later).

For reference, here's an archive containing one of each type of file in the dataset I was reducing: [data.zip](https://github.com/pypeit/PypeIt/files/12744653/data.zip)

debora-pe:
@amasegian this issue was resolved. If you are able to run PypeIt on the the `develop` branch, you will find the fix there, otherwise it would be available in the next release.

debora-pe:
resolved in #1689. Closing this issue.

PR #1689, opened 2023-09-29
Issue #1686

debora-pe:
This resolve issue #1686 
An error message triggered by an incorrect date was causing a crash. Now the date is corrected.


profxj (approved):
nice fix

rcooke-ast (approved):
Good - thanks!

Issue #1684, opened 2023-09-27
Improved scattered light model


rcooke-ast:
At the moment a very crude model of the scattered light is implemented, and this is only used for KCWI reductions. Ideally, we should use the inter-order regions (for `Echelle` reductions) and the interslice region for `SlicerIFU` reductions. This enhancement would require that we pass the `Slits` calibration files into `rawimage.py` (because the scattered light should be removed before the flatfielding frames), and this is not currently well setup for this.

The current simplistic model for scattered light subtraction (at least for KCWI) was implemented in PR #1661. I assign myself to look into a better model for KCWI. Part of this effort should involve using the Slits information.

kbwestfall:
Partially addressed by #1704 .  @rcooke-ast , can you comment on what's left to do?

rcooke-ast:
Yes, this is now "fully" addressed. KCRM is not implemented, but KCWI and ESI are both implemented. So, this can be closed.

PR #1704, opened 2023-10-15
Scattered light model

rcooke-ast:
This PR introduces a "proper" scattered light model using the inter-slit pixels (with some padding to reject light near the slit edges). This has been primarily tested with KCWI data (some tweaks are still being implemented), but this PR contains the main base to model the scattered light.


rcooke-ast:
Partially addresses issue #1684

pypeit/calibrations.py
```diff
@@ -537,6 +542,95 @@ def get_bpm(self, frame=None):
...
 
+    def get_scattlight(self):
+        """
+        Load or generate the scattered light model.
+
+        Returns:
+            :class:`~pypeit.scattlight.ScatteredLight`: The processed calibration image including the model.
+        """
+        # Initialise to None
+        self.msscattlight = None
```
kbwestfall: Just noting that I'm a little wary of this because it's not what we do in the other `get_*` functions for other internals.  Having said that, I'm fine with leaving it as is, if you prefer.

pypeit/calibrations.py
```diff
@@ -537,6 +542,95 @@ def get_bpm(self, frame=None):
...
+                                                      binning=scattlightImage.detector.binning,
+                                                      pad=self.par['scattlight']['pad'],
+                                                      scattlight_raw=scattlightImage.image,
+                                                      scattlight_model=model,
+                                                      scattlight_param=modelpar)
+
+        # TODO :: Should we go back and recalculate the slit edges once the scattered light is known?
+
+        # Show the result if requested
+        if self.show:
```
kbwestfall: `if self.show and self.msscattlight is not None:` ?

pypeit/calibrations.py
```diff
@@ -321,8 +324,8 @@ def get_arc(self):
         self.msarc = buildimage.buildimage_fromlist(self.spectrograph, self.det,
                                                     self.par['arcframe'], raw_files,
                                                     bias=self.msbias, bpm=self.msbpm,
-                                                    dark=self.msdark, calib_dir=self.calib_dir,
-                                                    setup=setup, calib_id=calib_id)
+                                                    dark=self.msdark, scattlight=self.msscattlight,
```
kbwestfall: My first principles question is do we need/want to remove scattered light when building the arc frame?  The more practical question is, given the calibration order, isn't `msscattlight` always going to be None here?  Same goes for `get_tilts` and `get_align` below.

pypeit/core/framematch.py
```diff
@@ -29,9 +29,10 @@ def __init__(self):
                      ('pinhole', 'Pinhole observation used for tracing slit centers'),
                    ('pixelflat', 'Flat-field exposure used for pixel-to-pixel response'),
                    ('illumflat', 'Flat-field exposure used for illumination flat'),
-                   ('lampoffflats', 'Flat-field exposure with lamps off used to remove '
-                                   'persistence from lamp on flat exposures and/or thermal emission '
-                                   'from the telescope and dome'),
+                ('lampoffflats', 'Flat-field exposure with lamps off used to remove '
+                                 'persistence from lamp on flat exposures and/or thermal emission '
+                                 'from the telescope and dome'),
+                  ('scattlight', 'Frame (usually similar to pixelflat) used to determine scattered light model'),
```
kbwestfall: "usually similar to a pixelflat" is confusing to me.  Do you mean "usually a flat-field exposure"?

pypeit/images/rawimage.py
```diff
@@ -512,6 +517,8 @@ def process(self, par, bpm=None, flatimages=None, bias=None, slits=None, dark=No
             msgs.error('No bias available for bias subtraction!')
         if self.par['use_darkimage'] and dark is None:
             msgs.error('No dark available for dark subtraction!')
+        if self.par['subtract_scattlight'] and scattlight is None:
+            msgs.error('Scattered light subtraction requested, but scattered light model provided.')
```
kbwestfall: "... scattered light model *not* provided."

pypeit/images/rawimage.py
```diff
@@ -1125,10 +1132,25 @@ def subtract_scattlight(self):
...
         for ii in range(self.nimg):
-            binning = self.detector[0].binning
-            scatt_img = self.spectrograph.scattered_light(self.image[ii, ...], binning)
+            # Calculate a model specific for this frame
+            spatbin = parse.parse_binning(self.detector[0]['binning'])[1]
+            # TODO :: Need to pass in the fitstbl here... currently this is a dummy value
+            tbl = dict(binning=self.detector[0]['binning'], dispname='BL')
+            pad = scattlight.pad // spatbin
+            offslitmask = slits.slit_img(pad=pad, initial=True, flexure=None) == -1
+            scatt_img, _, success = self.spectrograph.scattered_light(self.image[ii, ...], offslitmask, tbl)
```
kbwestfall: I think there are larger scale things that we can do to enable passing the full instrument configuration to `scattered_light`, but for now I would suggest the following.

 - Instead of passing `tbl` to `scattered_light`, just pass `binning` and `dispname`.
 - You already have binning, but you'll have to get the dispname from the file (outside the loop): `dispname = self.spectrograph.get_meta_value(self.spectrograph.get_headarr(self.file), 'dispname')`

The latter seems particularly important to do now because otherwise the code is going to think all data are "BL" data, right?

pypeit/calibrations.py
```diff
@@ -537,6 +542,95 @@ def get_bpm(self, frame=None):
...
+            return self.msscattlight
+
+        # Now generate the DataModel
+        self.msscattlight = scattlight.ScatteredLight(PYP_SPEC=self.spectrograph.name,
+                                                      pypeline=self.spectrograph.pypeline,
+                                                      detname=scattlightImage.detector.name,
+                                                      nspec=scattlightImage.shape[0], nspat=scattlightImage.shape[1],
+                                                      binning=scattlightImage.detector.binning,
+                                                      pad=self.par['scattlight']['pad'],
+                                                      scattlight_raw=scattlightImage.image,
```
kbwestfall, line 617: Similar to the way the `EdgeTraceSet` object has the full `TraceImage` object in its datamodel, is it worth doing the same here? I.e., keeping the full `ScatteredLightImage` object in the `ScatteredLight` object?

pypeit/scattlight.py
```diff
@@ -0,0 +1,169 @@
...
+import numpy as np
+
+from pypeit import msgs
+from pypeit import datamodel
+from pypeit import calibframe
+from pypeit.display import display
+from pypeit.spectrographs.util import load_spectrograph
+
+
+class ScatteredLight(calibframe.CalibFrame):
```
kbwestfall, line 22: Add this `doc/scripts/build_datacontainer_datamodels.py`?  I think you need that to get the `class_datamodel_scatteredlight.rst` file.

pypeit/scattlight.py
```diff
@@ -0,0 +1,169 @@
...
+
+    # Define the data model
+    datamodel = {'PYP_SPEC': dict(otype=str, descr='PypeIt spectrograph name'),
+                 'pypeline': dict(otype=str, descr='PypeIt pypeline name'),
+                 'detname': dict(otype=str, descr='Identifier for detector or mosaic'),
+                 'nspec': dict(otype=int, descr='Number of pixels in the image spectral direction.'),
+                 'nspat': dict(otype=int, descr='Number of pixels in the image spatial direction.'),
+                 'binning': dict(otype=str, descr='Binning in PypeIt orientation (not the original)'),
+                 'pad': dict(otype=int, descr='Integer number of pixels to mask beyond the slit edges.'),
+                 'scattlight_raw': dict(otype=np.ndarray, atype=np.floating, descr='Processed, combined scattered light image'),
```
kbwestfall: See my comment about making this the full `ScatteredLightImage` data container, instead of only the image.

pypeit/scattlight.py
```diff
@@ -0,0 +1,169 @@
...
+            A model of the expected scattered light in the input image. Shape is (nspec, nspat).
+        """
+        msgs.info("Generating a scattered light image")
+        if self.scattlight_param is None:
+            msgs.warn("No scattered light parameters are available")
+            return np.zeros_like(image)
+        # Load the spectrograph
+        spec = load_spectrograph(self.PYP_SPEC)
+        # Return the model of the scattered light
+        return spec.scattered_light_model(self.scattlight_param, image)
```
kbwestfall: Does this need to be spectrograph-dependent?  I.e., the KCWI `scattered_light_model` function never uses `self`.  It does require 9 parameters, and that is specified in the comments as something required for KCWI, but this seems like it could be flexible.

pypeit/scattlight.py
```diff
@@ -0,0 +1,169 @@
...
+        modmax = np.max(_model)
+        image_list = zip([_data, _model, _resid],
+                             ['data', 'model', 'data-model'],
+                             [(0.0, modmax), (0.0, modmax), (-modmax/2,modmax/2)])
+
+        # Display frames
+        show_scattered_light(image_list, slits=slits, wcs_match=wcs_match)
+
+
+def show_scattered_light(image_list, slits=None, wcs_match=True):
```
kbwestfall: Put this in `pypeit/display/display.py`?

pypeit/scripts/chk_scattlight.py
```diff
@@ -0,0 +1,72 @@
+"""
```
kbwestfall, line 1: Nice! Add this to `pypeit/scripts/__init__.py`.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -350,7 +351,7 @@ def check_frame_type(self, ftype, fitstbl, exprng=None):
                     & self.lamps(fitstbl, 'off') & (fitstbl['hatch'] == 'Open')
         if ftype == 'bias':
             return good_exp & (fitstbl['idname'] == 'BIAS')
-        if ftype == 'pixelflat':
+        if ftype in ['pixelflat', 'scattlight']:
```
kbwestfall: What's the argument for using the internal continuum lamp exposures as opposed to exposures that (nominally) have the same illumination pattern as the sky?  (An associated comment is a reminder to update `doc/spectrographs/keck_kcwi.rst` with the details of the new scattered light modeling approach.)

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1031,46 +1032,164 @@ def get_rawimage(self, raw_file, det):
...
+            but the default value ('gaussian') is known to provide the best fit.
+
+        Returns
+        -------
+        model : `numpy.ndarray`_
+            Model of the scattered light for the input
         """
-        Calculate a model of the scattered light of the input frame.
+        # For clarity, unpack the parameters
+        assert param.size == 9  # For KCWI, the scattered light model requires 9 model parameters
```
kbwestfall: I'd prefer a `msgs.error` here.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1031,46 +1032,164 @@ def get_rawimage(self, raw_file, det):
         # Return
         return detpar, raw_img, hdu, exptime, rawdatasec_img, oscansec_img
 
-    def scattered_light(self, frame, binning):
+    def scattered_light_model(self, param, img, kernel='gaussian'):
```
kbwestfall: I would encourage us to treat the scattered light modeling more generically, something like what we do with `core.fitting.PypeItFit` or `core.wavecal.wv_fitting.WaveFit`.  So I would put much of this in `scattlight.py` and only keep the KCWI-specific stuff here (which I think is only the constraint that the model must have 9 parameters).

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1031,46 +1032,164 @@ def get_rawimage(self, raw_file, det):
...
+        spat, spec = np.meshgrid(spatvec, specvec / (specvec.size - 1))
+        # Convolve the input image (note: most of the time is spent here)
+        # oaconvolve is the fastest option when the kernel is much smaller dimensions than the image
+        # scale_img = (term0 + term1 * spec + term2*spec**2 + term3*spec**3) * signal.fftconvolve(img, kernel, mode='same')
+        scale_img = (term0 + term1 * spec + term2 * spec ** 2 + term3 * spec ** 3) * signal.oaconvolve(img, kernel,
+                                                                                                       mode='same')
+        spl = interpolate.RectBivariateSpline(specvec, spatvec, scale_img, kx=1, ky=1)
+        return spl(zoom * (specvec + shft_spec), zoom * (spatvec + shft_spat))
+
+    def scattered_light(self, frame, offslitmask, tbl, detpad=300):
```
kbwestfall: Similar comment: I would generalize and move much/most of this to `scattlight.py`.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1031,46 +1032,164 @@ def get_rawimage(self, raw_file, det):
...
-        # (2) The right half of the detector (using the right and middle scattered light spectrum)
-        # We then ensure that the model is continuous and smooth at the middle of the detector.
-        medl = np.median(scattlightl / scattlight)
-        medr = np.median(scattlightr / scattlight)
-        spatimg = np.meshgrid(np.arange(frame.shape[1]), np.arange(frame.shape[0]))[0]
-        scatt_img = np.outer(scattlight, np.ones(frame.shape[1]))
-        scatt_img[spatimg < ym] *= 1 + (medl - 1) * (spatimg[spatimg < ym] / ym - 1) ** 2
-        scatt_img[spatimg >= ym] *= 1 + (medr - 1) * (spatimg[spatimg >= ym] / (4095//spatbin - ym) - ym / (4095//spatbin - ym)) ** 2
-        return scatt_img
+        def resid(param, wpix, img):
```
kbwestfall: Nested (Inner) functions are a pet peeve of mine, in the sense that I'd rather we not do this, but I realize this is largely a style preference.

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -1031,46 +1032,164 @@ def get_rawimage(self, raw_file, det):
...
+        tbl : :class:`~pypeit.metadata.PypeItMetaData`_
+            One row of the fitstbl PypeItMetaData that contains metadata about the file being used to
+            optimize the scattered light model parameters.
+        detpad : :obj:`int`_, optional
+            Number of pixels to pad to each of the detector edges to reduce edge effects.
 
         Returns
         -------
         scatt_img : `numpy.ndarray`_
             A 2D image of the scattered light determined from the input frame
```
kbwestfall: Update to describe the 3 objects returned.

pypeit/spectrographs/spectrograph.py
```diff
@@ -1891,21 +1891,50 @@ def calc_pattern_freq(self, frame, rawdatasec_img, oscansec_img, hdu):
...
-            Binning of the frame (e.g. '2x1' refers to a binning of 2 in the spectral
-            direction, and a binning of 1 in the spatial direction). For the supported
-            formats, refer to :func:`~pypeit.core.parse.parse_binning`.
+        offslitmask : `numpy.ndarray`_
+            A boolean mask indicating the pixels that are on/off the slit (True = off the slit)
+        tbl : :class:`~pypeit.metadata.PypeItMetaData`_
+            One row of the fitstbl PypeItMetaData that contains metadata about the file being used to
+            optimize the scattered light model parameters.
+        detpad : :obj:`int`_, optional
+            Number of pixels to pad to each of the detector edges to reduce edge effects.
```
kbwestfall: Return needs to be updated so that it returns 3 objects.

pypeit/images/rawimage.py
```diff
@@ -1125,10 +1132,25 @@ def subtract_scattlight(self):
...
+            # If failure, revert back to the Scattered Light calibration frame model parameters
+            if not success:
+                msgs.warn("Scattered light model failed - using predefined model parameters")
+                # Use predefined model parameters
+                scatt_img = self.spectrograph.scattered_light_model(scattlight.scattlight_param, self.image[ii, ...])
```
kbwestfall: My understanding of the current flow is:
 - If scattered-light frames are provided, fit a scattered light model to them.  If they are not, scattered light is never subtracted.
 - For the science frames, the code will try to independently model the scattered light in them (only if scattered light frames are provided).  If that fails, use the scattered light model from the scattered-light frames.

Is this right?  If not, can you describe the flow?

If so, can we enable the scattered-light model to be subtracted from the science data regardless of whether or not you provide scattered-light frames?

Also, if I understand the source of the scattered light for KCWI, at least, (i.e., its intensity is proportional to the light flux through the system), then is it appropriate to subtract the model from the science frames given the significant difference in flux?

kbwestfall (changes requested):
Thanks, @rcooke-ast !  I've got lots of comments for you.  The most significant ones (I think) are:

- Update the docs, and more specifically KCWI docs with a description of the new approach
- Add some tests (will there be an associated dev-suite PR?)
- I'm confused about the control flow
- It feels like we can make the scattered-light modeling code more general

I'm fine if you want to push off the latter two points to future PRs.

profxj:
Head's up that I am going to branch off this and try it on ESI.

We can then compare to the old IDL code (by having @jhennawi run it)

pypeit/calibrations.py
rcooke-ast: Sure, I see no issue removing this, and if it's actually needed later, then we can add it later.

pypeit/calibrations.py
rcooke-ast: It's initialised to `None` in the `__init__` method, so I've removed this from here.

pypeit/calibrations.py
rcooke-ast: Done (but I don't think `self.msscattlight` can be None here). I've merged this with the `if` statement below, so it's a bit clearer.

pypeit/core/framematch.py
rcooke-ast: I was going to clarify that this is `ideally an internal flatfield frame`, but I'm not sure if this is true... I think we just want it to have lots of counts, so I've mentioned this instead!

pypeit/calibrations.py
rcooke-ast: Yep, good idea  👍   added.

pypeit/images/rawimage.py
rcooke-ast: Good catch - updated!

pypeit/images/rawimage.py
rcooke-ast: Yes, that's right. I'm happy to implement just that for now. Looking to other instruments, they may require different metadata to decide what the scattered light model parameters should be. `binning` should be generically needed, though. It's not ideal that we need to reload the file again, but not a big deal.

pypeit/scattlight.py
rcooke-ast: Good call, and well-spotted  👍

pypeit/scattlight.py
rcooke-ast: Done!

pypeit/scattlight.py
rcooke-ast: While it doesn't use anything from `self`, it could be spectrograph dependent (e.g. the number of model parameters, or some spectrographs might require the scattered light to be flipped relative to the science data, etc.). So, the model is (likely) spectrograph dependent.

pypeit/scattlight.py
rcooke-ast: Good call. I didn't know about this (I was just copying the `flatfield.py`, which also does this. I've moved this.

pypeit/scripts/chk_scattlight.py
rcooke-ast: Done!

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: OK - I've updated the documentation.  I agree with this (see my comment above, as well)...  I don't think it needs to be an internal flatfield, it just needs lots of counts to get good counts on the scattered light model. I might try a couple of other frames (e.g. standard star, or Arc frame). I'll update this with the best results.

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Agreed...

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: I tend to agree... it would be good to have a default model that can be used for all spectrographs, but I think this will require spectrograph-specific coding to do it well.

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: See comment above. Perhaps we can discuss this further on Slack, but the reason I've implemented it like this is because I suspect the scattered light model requires spectrograph specific code.

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: Done!

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: I also hesitate about this, but it's not that much different to the lambda function (I think), and this is a bit clearer, because we can add a docstring. I'm not really sure how to do this otherwise (i.e. call a spectrograph specific model).

pypeit/spectrographs/spectrograph.py
rcooke-ast: Indeed! Done.

rcooke-ast (commented):
Thanks @kbwestfall for the thorough review again! There's still one more comment that I'm thinking about, but I'll post the responses to everything else now.

pypeit/images/rawimage.py
rcooke-ast: Your understanding of the control flow is correct. But, I'm still deciding what is the best approach. I think it's best to general the model parameters and always use those, unless the user explicitly wants to calculate the scattered light model from each individual frame. I will change the flow such that it's like that, and document it here so the flow is clearer.

> If so, can we enable the scattered-light model to be subtracted from the science data regardless of whether or not you provide scattered-light frames?

Yes, that is possible, and I think that may even be better in some cases... still testing...

> Also, if I understand the source of the scattered light for KCWI, at least, (i.e., its intensity is proportional to the light flux through the system), then is it appropriate to subtract the model from the science frames given the significant difference in flux?

That is correct. This is precisely what is currently being done. For each setup there's a set of model parameters that:
(1) take the light distribution of the current frame (e.g. a science frame)
(2) based on this light distribution on the detector (and the pre-defined model parameters), calculate the scattered light image
(3) subtract this scattered light image from the current frame.

kbwestfall (approved):
Thanks for addressing my comments!  I'm approving now, but I have two lingering requests for follow-up work.
 - Be as verbose as you're willing to be in the description of how the scattered light is modeled and subtracted in the docs.  The added text in `keck_kcwi.rst` is appreciated, but users will still not really understand what's happening without digging into the code.
 - I could be wrong, but it still feels like you could create general, abstracted functions that handle generating and fitting the scattered light models that live in `scattlight.py` and accept arguments that specify the spectrograph specific behavior.  The spectrograph functions then are mostly wrappers for the general functions.

pypeit/core/scattlight.py
```diff
@@ -0,0 +1,308 @@
...
+        Model used to calculate the scattered light. This function is used to
+        generate a model of the scattered light, based on a set of model parameters
+        that have been optimized using self.scattered_light().
+
+    Returns
+    -------
+    model : `numpy.ndarray`_
+        Model of the scattered light for the input
+    """
+    # Extract the parameters into more conveniently named variables
```
jhennawi, line 54: Is there a reason why you don't just use the simpler: 

sigmx_g, sigmy_g, sigmx_l, sigmy_l, shft_spec, shft_spat, zoom_spec, zoom_spat, constant, kern_angle, kern_scale, polyterms_spat, polyterms_spec = param

pypeit/core/scattlight.py
```diff
@@ -0,0 +1,308 @@
...
+def scattered_light(frame, bpm, offslitmask, x0, bounds, detpad=300, debug=False):
+    """ Calculate a model of the scattered light of the input frame.
+
+    Parameters
+    ----------
+    frame : `numpy.ndarray`_
+        Raw 2D data frame to be used to compute the scattered light.
+    bpm : `numpy.ndarray`_
+        2D boolean array indicating the bad pixels (True=bad)
+    offslitmask : `numpy.ndarray`_
```
jhennawi, line 125: Can you specify image shapes for frame and then state that bpm and offslitmask are the same size.

pypeit/core/scattlight.py
```diff
@@ -0,0 +1,308 @@
...
+    good_mask = offslitmask.copy()
+    for ii in _mask_regions:
+        if ii == 0:  # All pixels to the left of the first slit
+            indx = spat[None, :] < centrace[:, ii, None]
+        elif ii == nslit:  # All pixels to the right of the last slit
+            indx = spat[None, :] > centrace[:, ii-1, None]
+        else:  # Everything else in between
+            indx = (spat[None, :] > centrace[:, ii-1, None]) \
+                   & (spat[None, :] < centrace[:, ii, None])
+        # Exclude these pixels
```
jhennawi, line 249: Since offslitmask is boolean, I think the logic would be clearer if you did something like 

good_mask = offslitmask &  np.logical_not(indx)

I would also name indx to indicate what it means, i.e. bad_pixels or something like that.

pypeit/core/scattlight.py
```diff
@@ -0,0 +1,308 @@
...
+    for yy in range(nspec):
+        ext = frame[yy, :]
+        gd = np.where(offslitmask[yy, :] & gpm[yy, :])
+        coeff = np.polyfit(xspat[gd], ext[gd], polyord)
+        model[yy, :] = np.polyval(coeff, xspat)
+    # Median filter in the spectral direction to smooth out irregularities in the fine correction
+    model_med = ndimage.median_filter(model, size=(50, 1))  # Median filter to get rid of CRs
+    scatt_light_fine = ndimage.gaussian_filter(model_med, sigma=10)  # Gaussian filter to smooth median filter
+    if debug:
+        embed()
```
jhennawi: Remove the embed.

pypeit/core/scattlight.py
```diff
@@ -0,0 +1,308 @@
...
+    # Return the mask of good inter-slit pixels
+    return good_mask
+
+
+def fine_correction(frame, bpm, offslitmask, polyord=2, debug=False):
+    """ Calculate a fine correction to the residual scattered light of the input frame.
+
+    Parameters
+    ----------
+    frame : `numpy.ndarray`_
```
jhennawi, line 260: Reminder, image shapes.

pypeit/images/rawimage.py
```diff
@@ -1112,24 +1120,143 @@ def subtract_continuum(self, force=False):
...
+            # Get a copy of the best-fitting model parameters
+            this_modpar = msscattlight.scattlight_param.copy()
+            this_modpar[8] = 0.0  # This is the zero-level of the scattlight frame. The zero-level is determined by the finecorr
+            # Apply the requested method for the scattered light
+            do_finecorr = self.par["scattlight"]["finecorr"]
+            if self.par["scattlight"]["method"] == "model":
+                # Use predefined model parameters
+                scatt_img = scattlight.scattered_light_model(this_modpar, _img)
+                if debug:
+                    embed()
```
jhennawi: remove embed.

pypeit/images/rawimage.py
```diff
@@ -1112,24 +1120,143 @@ def subtract_continuum(self, force=False):
...
+                    # import astropy.io.fits as fits
```
jhennawi: Remove?

pypeit/images/rawimage.py
```diff
@@ -1112,24 +1120,143 @@ def subtract_continuum(self, force=False):
...
+                    strprint = f"x0 = np.array([{tmp[0]*specbin} / specbin, {tmp[1]*spatbin} / spatbin,  # Gaussian kernel widths\n" + \
+                               f"               {tmp[2]*specbin} / specbin, {tmp[3]*spatbin} / spatbin,  # Lorentzian kernel widths\n" + \
+                               f"               {tmp[4]*specbin} / specbin, {tmp[5]*spatbin} / spatbin,  # pixel offsets\n" + \
+                               f"               {tmp[6]}, {tmp[7]},  # Zoom factor (spec, spat)\n" + \
+                               f"               {tmp[8]},  # constant flux offset\n" + \
+                               f"               {tmp[9]},  # kernel angle\n" + \
+                               f"               {tmp[10]},  # Relative kernel scale (>1 means the kernel is more Gaussian, >0 but <1 makes the profile more lorentzian)\n" + \
+                               f"               {tmp[11]}, {tmp[12]},  # Polynomial terms (coefficients of \"spat\" and \"spat*spec\")\n" + \
+                               f"               {tmp[13]}, {tmp[14]}, {tmp[15]}])  # Polynomial terms (coefficients of spec**index)\n"
+                    print(strprint)
```
jhennawi, line 1189: This printing should be via msgs right?

pypeit/par/pypeitpar.py
```diff
@@ -990,7 +994,99 @@ def from_dict(cls, cfg):
...
+
+        defaults['finecorr_order'] = 2
+        dtypes['finecorr_order'] = int
+        descr['finecorr_order'] = 'Polynomial order to use for the fine correction to the scattered light ' \
+                                  'subtraction. It should be a low value.'
+
+        defaults['finecorr_mask'] = None
+        dtypes['finecorr_mask'] = [int, list]
+        descr['finecorr_mask'] = 'A list containing the inter-slit regions that the user wishes to mask during ' \
+                                 'the fine correction to the scattered light. Each integer corresponds to an ' \
```
jhennawi, line 1070: Just mention that the defaults mean no additional masking.

jhennawi (approved):
Hi @rcooke-ast excellent work on this PR. The code looks great. I have minor comments throughout that I hope you can look at, but I'm approving anyway. Only one substantive question which is why do you opt to hard-code the parameters in the spectrograph file versus fitting them from flats every time? Is it that the fitting is too time consuming? Also, I'm not sure I followed how things actually work. Do you fit for a template and then determine its amplitude from flats to account for the fact that the scattered light depends on the total amount of light? Sorry if these questions are naive.

pypeit/core/scattlight.py
rcooke-ast: Not really, I just didn't want a very long line of code, so I put these into sensible groups. Also, `polyterms_spat` is two parameters, while `polyterms_spec` can have any number of free parameters, so they have to be separated.

pypeit/core/scattlight.py
rcooke-ast: Good idea - done!

pypeit/core/scattlight.py
rcooke-ast: Good idea - I've changed the logic to make it clearer, as you suggest, and changed the name.

pypeit/core/scattlight.py
rcooke-ast: Yep, good call 👍

pypeit/core/scattlight.py
rcooke-ast: Removed

pypeit/images/rawimage.py
rcooke-ast: Removed

pypeit/images/rawimage.py
rcooke-ast: Removed

pypeit/images/rawimage.py
rcooke-ast: This is in the debug statement. It's only used for developers that want a pretty printout of the best-fit parameters to include in the spectrograph file. I've added a more detailed comment now to make this a little more obvious!

pypeit/par/pypeitpar.py
rcooke-ast: Done.

rcooke-ast (commented):
Thanks @jhennawi for the comments! Let me respond to your other queries:

> Only one substantive question which is why do you opt to hard-code the parameters in the spectrograph file versus fitting them from flats every time? Is it that the fitting is too time consuming?

The hard-coded parameters in the spectrograph files just provide the initial guesses for the fitting. The scattered light model parameters are always derived using the user's data (unless they request to use the archived model parameters). It's a little time-consuming, so starting somewhere near the expected solution speeds things up and makes sure the solution doesn't get stuck in model space far from the expected "best" solution.

> Do you fit for a template and then determine its amplitude from flats to account for the fact that the scattered light depends on the total amount of light?

The model currently assumes that the scattered light is a reprocessed version of the detector image (this seems to work well for both ESI and KCWI). So, with this approach, the model parameters are assumed to be the same for _all_ frames, the only thing that changes is the image of the detector (i.e. the detector image + any fixed [pre-calculated] model parameters determine the total scattered light). This way, we can derive the model parameters using a `ScatteredLight` calibration frame, and then apply these model parameters (without any adjustment) to any other frame.

I've also updated the `calibrations/sacttlight.rst` docs to include all of these details, and hopefully that makes things a little clearer.

I hope that makes sense!

rcooke-ast:
All comments addressed. Running the dev suite again to make sure nothing is broken 🤞

rcooke-ast:
Tests pass... merging
image

PR #1705, opened 2023-10-16
Hotfix : pypeit_identify echelle

rcooke-ast:
A user identified a bug using `pypeit_identify` with echelle data. Here is the fix.


kbwestfall (approved):
Great, thanks!

profxj (approved):
thx for the quick fix

PR #1708, opened 2023-10-19
Allow a way to circumvent monotonic wavelength requirement in Spectrum1D

kbwestfall:
As titled.

This only affects the specutils interface.  Tests pass.  No need for a dev-suite run.


pypeit/specutils/pypeit_loaders.py
```diff
@@ -35,6 +37,69 @@
...
+    """
+    Force the spectrum to have a monotonically increasing wavelength vector.
+
+    Parameters
+    ----------
+    wave : `numpy.ndarray`_
+        Spectrum wavelengths
+    flux : `numpy.ndarray`_
+        Spectrum flux
+    ivar : `numpy.ndarray`_
```
rcooke-ast, line 51: should we make it an optional argument, then? `ivar=None`

pypeit/specutils/pypeit_loaders.py
```diff
@@ -35,6 +37,69 @@
...
+    msgs.warn('Wavelengths are not monotonically increasing!  Strict was set to False, so '
+              'measurements after a negative step in wavelength are removed from the constructed '
+              'spectrum.  BEWARE that this is likely the result of an error in the data '
+              'reduction!')
+
+    # NOTE: This is the brute force approach.  If this becomes something that we
+    # want to be more acceptable, we should consider instead fitting a low-order
+    # polynomial to the pixel vs. wavelength function and rejecting strong
+    # outliers.
+    _wave = wave.copy()
```
rcooke-ast: I think all of the `wave` below this line should be changed to `_wave`

EDIT: Actually it is probably fine, since you pass in and set `wave` in `pypeit_onespec_loader()`... but I still flag this just in case

pypeit/tests/test_specutils.py
```diff
@@ -134,3 +136,32 @@ def test_spec1d_io():
 
     ofile.unlink()
 
+
+@specutils_required
```
rcooke-ast, line 140: Nice!

rcooke-ast (approved):
Looks good to me, just some minor comments!

pypeit/specutils/pypeit_loaders.py
tbowers7: I think if this function were moved to somewhere in `core` to be used outside of this loader, then making this argument optional is fine.  Otherwise, this is just a helper function for use within this module and it doesn't matter since `ivar` in some form will be passed in all the time.

pypeit/specutils/pypeit_loaders.py
tbowers7: I think I agree with @rcooke-ast 's first comment, else `_wave` doesn't do anything here.

tbowers7 (approved):
Looks good.  Thanks for addressing this edge case.  I'll insist on the variable name change, but am approving now.

pypeit/specutils/pypeit_loaders.py
kbwestfall: Although @rcooke-ast 's suggestion is perfectly reasonable, I agree with @tbowers7 : this is really just a helper function that isn't intended to be used outside this module.  To signal this is a "private" function, I added a leading underscore to the function name.

pypeit/specutils/pypeit_loaders.py
kbwestfall: Good catch, and apologies.  This was a leftover line I used during testing.  Removing the line fixes the issue.

PR #1709, opened 2023-10-20
Add sensible error message for `specutils` loaders

tbowers7:
When users try to ingest a `spec1d` file into a Spectrum1D object, the current
scheme ends badly with an obscure error message like:

```
OSError: Could not identify column containing the wavelength, frequency or energy
```

Because PypeIt `spec1d` files may have multiple extensions, by convention we
load these into SpectrumList objects in specutils, even if there is only a
single spectrum.  This commit catches the case when a user tries to load a
`spec1d` file into a Spectrum1D object.  Previously, such an attempt ended up
with specutils loading in the `spec1d` file as a tabular FITS format, then
didn't know what to do about the spectral axis (leading to the above obscure
error).

This commit specifically includes a loader for `spec1d` files into Spectrum1D
objects, but emits a sensible error message in this case rather than trying to
actually load the file.

As it happens, in order to properly do this, I needed to make an adjustment in
`specutils` itself.  That PR (https://github.com/astropy/specutils/pull/1068), was
finally included in `specutils` version 1.12 (released 2023-10-17).

The `specutils` optional dependency has been updated to require this version.


profxj (approved):
Thanks for this!

Just check that you have directed this PR where you intended.

PR #1713, opened 2023-10-26
Script that allows interactive editing of the slit/order edges

kbwestfall:
As titled.  See changes to `doc/calibrations/slit_tracing.rst` for a description of the script.

I'm directing to the `hires_trace_orders` branch because I branched from there.  Will redirect once that PR clears.

I should add some tests, and it would be useful for one or more volunteers to take this for a test drive.


doc/calibrations/slit_tracing.rst
```diff
@@ -89,6 +89,66 @@ additional output that can be used to diagnose the parameterized fits to the
...
+   :width: 50%
+
+The trace image is shown in grayscale.  The left and right traces are green and
+magenta, respectively.  The orange horizontal line shows the "reference row"
+used when locating new traces.  Only the vertical line of the cursor (not shown)
+moves; the horizontal line always stays at the reference row.  Note that the
+vertical line marking the cursor may not appear if the typical matplotlib zoom
+or pan buttons are active.
+
+Pressing the ``?`` key anywhere in the window will print the key bindings:
```
rcooke-ast: "key bindings to the terminal window."

pypeit/core/gui/edge_inspector.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        Delete the trace nearest the cursor spatial (x) position.
+
+        Args:
+            pos (:obj:`list`):
+                List with x and y position of the cursor at the time of the
+                window event.
+        """
+        # Find the nearest trace
+        i = self.nearest_trace(pos[0])
```
rcooke-ast, line 267: could we just pass in the zeroth element of `pos`, since pos[1] is never used?

pypeit/core/gui/edge_inspector.py
```diff
@@ -0,0 +1,321 @@
...
+        color = self._trace_color(side)
+        self.trace_plot += [self.image_plot.axes.plot(new_trace, self.spec_pix,
+                                                      color=color, lw=2)[0]]
+        # Draw the plot
+        pyplot.draw()
+        # Report
+        msgs.info(f'Added {"right" if side > 0 else "left"} trace passing through '
+                  f'({spat:.1f}, {self.reference_row:.2f}).')
+
+    def add_left(self, pos):
```
rcooke-ast, line 313: same here, and below for `add_right`. Can we just pass in pos[0] as the argument?

pypeit/core/gui/gui_util.py
```diff
@@ -0,0 +1,257 @@
...
+            self.disconnect('?')
+        self.register('?', self.print_help, descr='Print this list of key bindings')
+
+    def print_help(self, pos):
+        """
+        Print the help dialog.
+
+        Args:
+            pos (list):
+                Location where the help event was triggered.  This is ignored.
```
rcooke-ast: I'm sure there's a good reason why `pos` is an argument (and then is ignored), but I flag it anyway!

pypeit/core/gui/gui_util.py
```diff
@@ -0,0 +1,257 @@
...
+                         valstep=valstep, orientation=orientation, track_color=track_color,
+                         handle_style=handle_style, **kwargs)
+        self.label.set_position((0.5, 0.8))
+        self.label.set_verticalalignment('bottom')
+        self.label.set_horizontalalignment('center')
+        self.label.set_weight('bold')
+        # "Removes" the labels showing the current range
+        self.valtext.set_visible(False)
+
+    def update_range(self, rng, label=None):
```
rcooke-ast, line 197: should we add these args into the docstring?

pypeit/core/gui/gui_util.py
```diff
@@ -0,0 +1,257 @@
...
+        Adjust the plotted range.
+
+        Args:
+            val (:obj:`list`):
+                New range to plot
+        """
+        self.image_plot.set_clim(*val)
+        pyplot.draw()
+
+    def next_image(self, *args, **kwargs):
```
rcooke-ast: Also here. Why do we need to include the args/kwargs in some places?

pypeit/scripts/edge_inspector.py
```diff
@@ -0,0 +1,46 @@
...
+        edges = edgetrace.EdgeTraceSet.from_file(trace_file)
+        # Inspector object
+        pointer = edge_inspector.EdgeInspectorGUI(edges)
+        # Run.  Ends when window is closed
+        pyplot.show()
+        # Close (restore pyplot rc defaults)
+        pointer.close()
+
+        # Write the updated edges and slits
+        edges.to_file(file_path=trace_file)
```
rcooke-ast, line 42: Should this be `pointer.edges.to_file`? (and similarly on line 44)?

rcooke-ast (commented):
This is really impressive @kbwestfall!  The structure of the GUI is really clear, and much better organised than what I previously implemented in `identify`, `skysub_regions`, and `object_find`. Eventually, we should aim to migrate these over to the same structure that you've setup to make all GUI codes a bit easier to manage in the long run.

I have mostly minor comments, but I'm a little confused why there are some unused arguments being passed to some functions (including a few `*args` and `**kwargs` as well). In some cases, there doesn't appear to be any inheritance, and I can't understand why they are needed... I'm certain that I'm about to learn something new in python that I didn't know before!

Without major comments, I'm approving - thanks again!  👍

rcooke-ast (approved):
Now actually approving...

doc/calibrations/slit_tracing.rst
kbwestfall: Done!

pypeit/core/gui/edge_inspector.py
```diff
@@ -0,0 +1,321 @@
...
+            self.trace_plot += [self.image_plot.axes.plot(self.trace_cen[:,i], self.spec_pix,
+                                                          color=color, lw=2)[0]]
+
+    def update_traces(self, *args, **kwargs):
+        """
+        Update the underlying trace data and edges object.
+
+        Changes to the traces are kept by the internals until we're ready to
+        "update" the edges object.  This performs the update and replots the
+        trace data.
```
kbwestfall, line 150: I added some explanation here for why this function has *args and **kwargs.  I originally only used `self` in the arguments, but the code barfed because the Button was passing something to the function.  Adding the placeholders that are always ignored is really just a lazy way out.  I removed **kwargs because it wasn't necessary.

pypeit/core/gui/edge_inspector.py
kbwestfall: Good question!  The issue is that all the functions connected to pointer events have to have the same calling sequence, which is set by `_set_event` in `gui_util.Pointer`.  I added a few doc-string sentences to clarify this.

pypeit/core/gui/edge_inspector.py
kbwestfall: Same response.

pypeit/core/gui/gui_util.py
kbwestfall: Yep, same response as to why we can't just pass pos[0] to some of the `EdgeInspectorGUI` functions.

pypeit/core/gui/gui_util.py
kbwestfall: Thanks for catching this!  Added.

pypeit/core/gui/gui_util.py
kbwestfall: Same as `update_traces` in `EdgeInspectorGUI`.  I updated the calling sequence (we only need *args), and added the same note to the doc string.

pypeit/scripts/edge_inspector.py
kbwestfall: They're the same object.  I.e., `EdgeInspectorGUI` doesn't copy `edges`, it edits it directly.  I added a comment in its doc string indicating this.

kbwestfall (commented):
Thanks @rcooke-ast !  I've written a few of these now and stackoverflow is key!

On the argument lists, as you'll see in my responses, some of these were pure laziness and others were restrictions to the required calling sequence for cursor events.

debora-pe (approved):
Solid! Thanks @kbwestfall!

PR #1715, opened 2023-10-30
Hotfix: use mask in specutils loader

kbwestfall:
As titled.  The new check on a monotonically increasing wavelength vector caused problems for some spectrographs that have wavelengths set to 0.  This is avoided by using the mask.

### DIFF
--- a/pypeit/specutils/pypeit_loaders.py
+++ b/pypeit/specutils/pypeit_loaders.py
@@ -197,6 +197,9 @@ def pypeit_spec1d_loader(filename, extract=None, fluxed=True, strict=True, **kwa
                                         else sobj.get_opt_ext(fluxed=_cal)
         if not np.all(_gpm):
             msgs.warn(f'Ignoring {np.sum(np.logical_not(_gpm))} masked pixels.')
+        if not np.any(_gpm):
+            msgs.warn(f'Spectrum {sobj.NAME} is fully masked and will be ignored!')
+            continue
         _wave, _flux, _ivar = _enforce_monotonic_wavelengths(_wave[_gpm], _flux[_gpm], _ivar[_gpm],
                                                              strict=strict)
         flux_unit = astropy.units.Unit("1e-17 erg/(s cm^2 Angstrom)" if _cal else "electron")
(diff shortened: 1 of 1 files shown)


pypeit/specutils/pypeit_loaders.py
```diff
@@ -195,7 +195,10 @@ def pypeit_spec1d_loader(filename, extract=None, fluxed=True, strict=True, **kwa
         _ext, _cal = sobj.best_ext_match(extract=extract, fluxed=fluxed)
         _wave, _flux, _ivar, _gpm = sobj.get_box_ext(fluxed=_cal) if _ext == 'BOX' \
                                         else sobj.get_opt_ext(fluxed=_cal)
-        _wave, _flux, _ivar = _enforce_monotonic_wavelengths(_wave, _flux, _ivar, strict=strict)
+        if not np.all(_gpm):
+            msgs.warn(f'Ignoring {np.sum(np.logical_not(_gpm))} masked pixels.')
+        _wave, _flux, _ivar = _enforce_monotonic_wavelengths(_wave[_gpm], _flux[_gpm], _ivar[_gpm],
+                                                             strict=strict)
```
tbowers7, line 204: What happens if there are no good pixels (_i.e._, all are zero for some reason)?  Will `_enforce_monotonic_wavelengths()` as written choke on empty input arrays?  Maybe add an edge-case check to bypass the enforcement in this case?

tbowers7 (changes requested):

pypeit/specutils/pypeit_loaders.py
kbwestfall: Good catch!  What do you think of my fix:
```
if not np.any(_gpm):
    msgs.warn(f'Spectrum {sobj.NAME} is fully masked and will be ignored!')
    continue
```

tbowers7: That looks good.

tbowers7 (approved):
Nice fix.  Approving.

rcooke-ast (approved):
Looks good to me!

PR #1716, opened 2023-10-30
Hotfix: Improve error message when `reid_arxiv` file missing

tbowers7:
As titled.

From the User's slack (https://pypeit-users.slack.com/archives/CST2Q4QPN/p1694827540917799), this fix deals with the case when a `reid_arxiv` file doesn't exist.  It manifests itself as a `None` type filename being passed to the template loader.  This PR adds a check for `None` before presenting an error message.


pypeit/core/wavecal/waveio.py
```diff
@@ -73,6 +73,9 @@ def load_template(arxiv_file, det, wvrng=None):
         binning of the template arc spectrum
 
     """
+    if not isinstance(arxiv_file, (str, pathlib.Path)) or arxiv_file is None or arxiv_file == "":
```
rcooke-ast: Should we be also doing an `os.path.exists()` call here? Also note the inconsequential typo `non-existent`

rcooke-ast (approved):
All looks good to me - thanks!

One minor comment...

pypeit/core/wavecal/waveio.py
kbwestfall: FWIW, I would recommend putting all of this error handling into `data.get_reid_arxiv_filepath` so that the first two lines of this function are:
```
calibfile, fmt = data.get_reid_arxiv_filepath(arxiv_file)
tbl = astropy.table.Table.read(calibfile, format=fmt)
```
As for this `if` statement, I think `isinstance` will catch the None case, so it can be
```
if not isinstance(arxiv_file, (str, pathlib.Path)) or arxiv_file == '':
```

kbwestfall (approved):
A minor recommendation, but I'm approving!

kbwestfall (approved):
Thanks!

PR #1712, opened 2023-10-25
Update LDT/DeVeny + "quality of life" changes

tbowers7:
Primarily an update to LDT/DeVeny:
* Default parameter changes (mainly edge tracing and wavelength solution orders)
* Instrument-Specific arc line lists
* Updated wavelength template for one grating

A series of small "quality of life" improvements:
* Allow `pypeit_chk_wavecalib` to accept lists of input files.
* Allow restricted wavelength range for spectral flexure cross-correlation (this required touching a number of files in order to plumb the user-specifiable parameter all the way down to the function where it is used). This feature is useful when the end(s) of the wavelength calibration are not well constrained because of arc line placement and the user doesn't want those regions to affect the flexure correction.
* Update verbosity assignment in `pypeit_show_2dspec`.
* Doc/param fix to align with code for polynomial overscan fitting; update `np.polyfit()` -> `np.polynomial.Chebyshev.fit()`.

Main repo unit tests pass, will request cloud run of DevSuite following review.
```
================= test session starts ================
platform darwin -- Python 3.11.6, pytest-7.4.2, pluggy-1.3.0
rootdir: /Users/tbowers/d1/codes/PypeIt
configfile: setup.cfg
testpaths: pypeit/tests
plugins: doctestplus-1.0.0, astropy-0.11.0, hypothesis-6.88.1, remotedata-0.4.1, asdf-3.0.0, cov-4.1.0, filter-subpackage-0.1.2, astropy-header-0.2.2, mock-3.12.0, arraydiff-0.5.0
collected 240 items
...
==== 240 passed, 67 warnings in 408.87s (0:06:48) ====
```

Companion DevSuite PR (https://github.com/pypeit/PypeIt-development-suite/pull/295) adds wavelength templates.

PR #1718, opened 2023-11-02
HotFix: SkyRegions GUI

rcooke-ast:
This hotfix ensures that the calib key/dir/ID are all stored in the SkyRegions calibrations file. Several errors and bits of code cleanup are also implemented.


pypeit/core/gui/skysub_regions.py
```diff
@@ -571,22 +571,41 @@ def operations(self, key, axisID):
...
-                msgs.info(f"Saving regions to: {outfil}")
-                self._overwrite = True
-            msskyreg = buildimage.SkyRegions(image=inmask.astype(float), PYP_SPEC=self.spectrograph)
-            msskyreg.to_file(file_path=outfil)
+            if np.all(np.logical_not(inmask)):
+                msgs.warn("Sky regions are empty - master calibration frame will not be generated")
+            else:
+                # Build the master Sky Regions calibration frame
+                msskyreg = buildimage.SkyRegions(image=inmask.astype(float), PYP_SPEC=self.spectrograph)
+        return msskyreg
```
kbwestfall: I think this is a bit clearer:
```
if not self._use_updates:
    return None

# Generate the mask
inmask = skysub.generate_mask(self.pypeline, self._skyreg, self.slits, self.slits_left,
                              self.slits_right)
if np.all(np.logical_not(inmask)):
    msgs.warn("Sky regions are empty - master calibration frame will not be generated")
    return None

# Build the master Sky Regions calibration frame
return buildimage.SkyRegions(image=inmask.astype(float), PYP_SPEC=self.spectrograph)

```

kbwestfall (approved):
One minor suggestion, but it looks good otherwise.  This touches a core function, so should we run tests?

pypeit/core/gui/skysub_regions.py
rcooke-ast: Agreed - I've made that change!

>

rcooke-ast:
> This touches a core function, so should we run tests?

Yes, we probably should - running now...

rcooke-ast:
Two dev suite fails, but these are unrelated to this PR. One is fixed by #1710, and the other is fixed by #1712.

image

pypeit/core/gui/skysub_regions.py
```diff
@@ -571,22 +571,41 @@ def operations(self, key, axisID):
         self.replot()
 
     def get_result(self):
-        """Save a mask containing the skysub regions, and print information
+        """Generate a master calibration file containing a mask of the skysub regions, and print information
```
profxj: please remove `master`

pypeit/core/gui/skysub_regions.py
```diff
@@ -571,22 +571,41 @@ def operations(self, key, axisID):
...
+        if not self._use_updates:
+            return None
+
+        # Generate the mask
+        inmask = skysub.generate_mask(self.pypeline, self._skyreg, self.slits, self.slits_left, self.slits_right)
+        if np.all(np.logical_not(inmask)):
+            msgs.warn("Sky regions are empty - master calibration frame will not be generated")
+            return None
+
+        # Build the master Sky Regions calibration frame
```
profxj: same

profxj (approved):
all good, modulo the `master` bit
  time to scrub that from our past  :)

pypeit/core/gui/skysub_regions.py
rcooke-ast: Done (and I've also removed some other instances of this, as well)!

pypeit/core/gui/skysub_regions.py
rcooke-ast: Done!

PR #1727, opened 2023-11-20
Added Gr4 support

rcooke-ast:
Added support for GR4, based on Users Slack request.

image


profxj (approved):
thx!  Is there a DevSuite test on the way?

rcooke-ast:
You bet there is! I knew you'd ask - I was just awaiting permission from the user.

rcooke-ast:
Dev-suite test added, and all passes (the failures are because I only ran NTT/EFOSC2):
image

kbwestfall (approved):
Nice!

PR #1728, opened 2023-11-21
Sanity check Keck ESI MJD values

badpandabear:
Older Keck ESI files seem to have an invalid MJD value. This PR detects this and uses the DATE-OBS and UT keywords instead.


profxj (approved):
thx for this

kbwestfall (approved):
Looks great!

PR #1731, opened 2023-12-04
LDT/DeVeny: Adjust source for 'target' metadata item

tbowers7:
For long target lists, some observers do not type the individual object names into the LOUI, but rather rely upon the SCITARG keyword that comes from the TCS via the ObserverTargetList tool.  This results in the OBJECT/OBJNAME keywords being set to UNKNOWN.  When PypeIt tries to read in the metadata, this leaves all object frames with the same, unhelpful name.

Solution: Add a `compound_meta` case for 'target' that uses the SCITARG FITS keyword for OBJECT frames if the OBJNAME is "UNKNOWN" or blank. This means that only science frames could have the 'target' metadata pulled from SCITARG, leaving all calibration frames as before.

PypeIt repo unit tests pass, and LDT/DeVeny DevSuite tests pass (no changes elsewhere in the code, so no general DevSuite run):
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  242 passed, 69 warnings in 480.89s (0:08:00) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 13/13 TESTS  ---
Testing Started at 2023-12-04T10:22:37.830855
Testing Completed at 2023-12-04T10:46:47.808049
Total Time: 0:24:09.977194
```

### DIFF
--- a/doc/releases/1.14.1dev.rst
+++ b/doc/releases/1.14.1dev.rst
@@ -49,8 +49,8 @@ Instrument-specific Updates
 - Keck/KCWI and Keck/KCRM: Turned on polynomial correction for sky subtraction.
 - We now support the reduction of VLT/FORS2 data taken in MOS mode.
 - Fixed fluxing file format in the ``Keck-DEIMOS HOWTO`` documentation.
-- Various updates to default parameters for LDT/DeVeny, plus updated wavelength
-  templates and instrument-specific line lists.
+- LDT/DeVeny: default parameter & wavelength template updates, add instrument-
+  specific line lists, use SCITARG for target name if none entered in the LOUI.
 - Added support for NTT/EFOSC2 Gr#4
 
 Script Changes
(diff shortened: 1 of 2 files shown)


kbwestfall (approved):
Looks good to me!  We should consider adding a useful test to the dev-suite.

tbowers7:
> Looks good to me! We should consider adding a useful test to the dev-suite.

Yes, I've been pondering where/how to do that.  Probably a VET test?

kbwestfall:
> > Looks good to me! We should consider adding a useful test to the dev-suite.
> 
> Yes, I've been pondering where/how to do that. Probably a VET test?

I think it can be a unit test that reads a raw file, right?  I.e., I think of VET tests as tests of PypeIt output, not input.

profxj (approved):
thx thx

your call on whether to add a DevSuite test on this

tbowers7:
> thx thx
> 
> your call on whether to add a DevSuite test on this

I've tested this with various data sets I have on hand and it works great.  Will probably add this DevSuite test for long-term stability when I dig in to adding our NIR spectrograph after the first of the year, since I will need to add DevSuite tests for it.

PR #1732, opened 2023-12-05
PypeIt setup GUI changes that affect code shared with the rest of PypeIt

badpandabear:
These changes were made to help the new pypeit_setup_gui do the following:

- Re-open pypeit files and get the same contents that were written.
- Save and open invalid pypeit files that the user may not have completed editing
- Preserve commented out files and comments in the parameters section.
- Monitor the log to provide progress dialog.


pypeit/inputfiles.py
```diff
@@ -91,84 +108,126 @@ def readlines(ifile:str):
 
         # Read the input lines and replace special characters
         with open(ifile, 'r') as f:
-            lines = np.array([l.replace('\t', ' ').replace('\n', ' ').strip() for l in f.readlines()])
-        # Remove empty or fully commented lines
-        lines = lines[np.array([ len(l) > 0 and l[0] != '#' for l in lines ])]
-        # Remove appended comments and return
-        return np.array([ l.split('#')[0] for l in lines ])
+            lines = np.array([l.replace('\t', ' ').rstrip()  for l in f.readlines()])
+        return lines
```
kbwestfall: FWIW, you can skip the middle man:
```
with open(ifile, 'r') as f:
    return np.array([...
```

pypeit/inputfiles.py
```diff
@@ -255,29 +312,38 @@ def _read_data_file_table(lines):
         Args:
             lines (:obj:`list`):
                 List of lines *within the data* block read from the input file.
-        
+            preserve_comments (bool,Optional):
```
kbwestfall: Not optional, right?

pypeit/inputfiles.py
```diff
@@ -391,18 +463,26 @@ def path_and_files(self, key:str, skip_blank=False, check_exists=True):
             if skip_blank and row[key].strip() in ['', 'none', 'None']:
                 continue
 
+            # Skip commented out entries
+            if row[key].strip().startswith("#"):
+                if not include_commented_out:
+                    continue
+                name = row[key].strip("# ")
```
kbwestfall, line 487: Might be worth adding an in-line comment here that the space after the hash is important.  I.e., `"# "` vs. `"#"` is an important difference.

pypeit/pypmsgs.py
```diff
@@ -135,11 +137,16 @@ def _initialize_log_file(self, log=None):
...
         """
-        if log is None:
+
+        if log is not None:
+            if isinstance(log, io.IOBase):
+                self._log = log
+            else:
+                # Initialize the log
+                self._log = open(log, 'w')
+        else:
```
kbwestfall: I think the code is cleaner when minimizing nesting.  So I would instead:
```
if log is None:
    return

self._log = log if isinstance(log, io.IOBase) else open(log, 'w')
```
but that's just my preference.  No need to change if you don't agree.

pypeit/pypmsgs.py
```diff
@@ -183,15 +195,12 @@ def error(self, msg, cls='PypeItError'):
         premsg = '\n'+self._start + self._white_RD + '[ERROR]   ::' + self._end + ' '
         self._print(premsg, msg)
 
-        # Close log file
+        # Close QA plots
         # TODO: This no longer "closes" the QA plots
-        self.close()
+        close_qa(self.pypeit_file, self.qa_path)
```
kbwestfall, line 195: Can we delete the TODO?  Also, is there any worry about `self._log` not being closed?

kbwestfall (approved):
Looks good!  I had some very minor comments.

pypeit/inputfiles.py
```diff
@@ -91,84 +108,126 @@ def readlines(ifile:str):
...
-        elif s >= 0 and e > 0:
+        elif setup_start >= 0 and setup_end > 0:
             setup_found = True
 
         # Proceed
         if setup_found:
-            setups, sdict = cls._parse_setup_lines(lines[s:e])
-            is_config[s-1:e+1] = False
+            setup_lines = lines[setup_start:setup_end]
+            if preserve_comments:
```
debora-pe, line 184: Should this be `if not preserve_comments`?

debora-pe (approved):
@badpandabear this looks good! I only had one question.

pypeit/inputfiles.py
badpandabear: When preserving comments I kept running into parsing problems and backwards compatibility problems with the pypeit files in the dev-suite. So I only support them in the configuration section and the data section.  This code removes them from the setup section if they weren't removed earlier.  I added comments to make this clearer

pypeit/pypmsgs.py
badpandabear: I left the TODO because I didn't know what it does. But looking at it it appears to do something so I removed the comment

pypeit/pypmsgs.py
badpandabear: I think you're right. The trinary operator in C/C++ has made me too averse for using this construct in python

badpandabear:
Dev suite resutls:

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  242 passed, 70 warnings in 393.81s (0:06:33) ---
--- PYTEST UNIT TESTS PASSED  133 passed, 171 warnings in 973.27s (0:16:13) ---
--- PYTEST VET TESTS PASSED  60 passed, 70 warnings in 1605.82s (0:26:45) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 235/235 TESTS  ---
Testing Started at 2023-12-06T22:49:55.290644
Testing Completed at 2023-12-07T08:26:56.493246
Total Time: 9:37:01.202602

```

PR #1734, opened 2023-12-05
Hotfix arc/tilt frame-typing for KCWI

rcooke-ast:
This hotfix ensures that only FeAr frames are typed as `arc`. ThAr frames are preferred for the `tilt`, but if none of these are available, the FeAr frames will be used.


kbwestfall (approved):
Looks good!

Do we need to update the documentation?  I.e., we should be clear that pypeit will *only* use FeAr lamps for wavelength calibration and use either ThAr or FeAr lamps for tilts, where preference is given to ThAr lamps.

rcooke-ast:
> Do we need to update the documentation? I.e., we should be clear that pypeit will _only_ use FeAr lamps for wavelength calibration and use either ThAr or FeAr lamps for tilts, where preference is given to ThAr lamps.

Good idea - I've just realised that we already mention this in the docs, but I've added a little more text.

profxj (approved):
thx thx

PR #1748, opened 2023-12-18
Increase time limit for GTC-OSIRIS standards

Raindogjones:
Increased exposure time limit for GTC-OSIRIS standards to account for longer standards with high-res grisms. Without this change some standards are mis-typed as science,standard and lead to an error running pypeit_setup where they are associated with two different setups.


rcooke-ast (approved):
This looks just fine to me!  Thanks for the update.

profxj (approved):
you bet

PR #1749, opened 2023-12-19
Correct lookup of Subaru Telescope location

ejeschke:
Simply fixes the name for the lookup of Subaru Telescope


ejeschke:
@profxj , for your quick merge, hopefully. Tried to push a commit to the `subaru_focas` branch on github PypeIt home, but got a 403 (forbidden). It's ok, I prefer PR way of working if you are OK with that.

ejeschke:
> but got a 403 (forbidden).

ah, hadn't accepted the github invite... did that...

profxj (approved):
thx!

PR #1750, opened 2023-12-20
FIX sky_spec FITS file

debora-pe:
A PypeIt user pointed out that the archival sky spectrum `pypeit/data/sky_spec/sky_LRISr_600_7500_5460_7950.fits` file had the FITS extensions inverted and, therefore, `data.load_sky_spectrum(sky_file)` was not able to load the file and giving an error.

This PR fixes this.


profxj (approved):
good catch and fix

kbwestfall (approved):

PR #1751, opened 2023-12-21
Make show_2dspec more robust against bitmask changes

kbwestfall:
This addresses the `show_2dspec` issue Joe had with this branch.


jhennawi (approved):
@kbwestfall I'm going to approve this now because I'm observing and need this functionality. But we should still have the discussion with the broader group as we discussed. Sorry I missed the last meeting.

jhennawi (approved):
Thanks @kbwestfall

PR #1754, opened 2024-01-11
Fix setup gui type annotations to be Python 3.9 compatible.

badpandabear:
The setup GUI was using type annotation stuff from python 3.10, but the docs are built under 3.9. This makes the annotations python 3.9 compatible.


rcooke-ast (approved):
Looks good to me!

profxj (approved):
all good

PR #1757, opened 2024-01-18
Bitmask flagging update

kbwestfall:
Fixes what I consider to be a cumbersome construction by introducing an `and_not` keyword when finding flipped bits.


jhennawi (changes requested):
Hi @kbwestfall I'm pretty sure I got this syntax from elsewhere in calibrations, i.e. I was just copying things from other slit_bpms. I don't disagree that this syntax is cumbersome and something along the lines of your change may well be needed. However, the documentation of the new and_not keyword argument is not clear to me. This line I did not quite follow:  "If any of these bits are flagged, the element in the returned array is as if it was not flagged, even if other flags not in this list *are* flagged.  I.e., the returned value can only be True if one of these "and_not" bits are *not* flagged". This is the line  Maybe you can try to clarify it or provide an example of the usage. 

I also find that it is typically better to keep the logic explicit in the calling routine unless we are doing this regularly (which I believe we are in calibrations.py, but currently and_not is only being used in one location).  Furthermore, you appear to be achieving what I was doing in one line of code with ten lines of code. Your functionality is likely more general, but it is this added complexity needed? If so seems like it should be used in more than one place. Please let me know your thoughts?

kbwestfall:
Hi @jhennawi .  Yes, good point about the documentation not being clear.  I've tried to improve it and (more importantly) added a bunch of examples in the docs.  I've also propagated this change to other similar calls that were made throughout the code (I expected they were there; I was just lazy and only replaced the new line).  Even here, I've only replaced ones that use the `exclude_from_reducing` object (because it was easy to search for), so there may be more, similar calls that I missed.  For the lines I replace, please pay particular attention to the change in `find_objects.py`.  The previous code was trying to exclude BOXSLIT flags, but include others.  I'm not sure the command I replaced this with is going to produce *exactly* the same result, but I'm also not sure that the previous code was actually achieving what it was meant to given the description in the inline comments.

kbwestfall:
FWIW, tests pass.  Note, I used the dev-suite `develop` branch for this, so it missed the [HIRES vet test fix](https://github.com/pypeit/PypeIt-development-suite/pull/306).

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  244 passed, 56 warnings in 591.60s (0:09:51) ---
--- PYTEST UNIT TESTS PASSED  133 passed, 171 warnings in 1060.41s (0:17:40) ---
--- PYTEST VET TESTS FAILED  1 failed, 60 passed, 70 warnings in 5862.06s (1:37:42) ---
Testing Started at 2024-01-26T18:12:45.544170
Testing Completed at 2024-01-27T15:26:29.847681
Total Time: 21:13:44.303511
```

jhennawi (approved):
Thanks @kbwestfall. Waiting on @debora-pe to check the boxslit/short slit implementation.

pypeit/find_objects.py
```diff
@@ -166,8 +166,19 @@ def __init__(self, sciImg, slits, spectrograph, par, objtype, wv_calib=None, wav
         # Internal bpm mask
         # We want to keep the 'BOXSLIT', which has bpm=2. But we don't want to keep 'BOXSLIT'
         # with other bad flag (for which bpm>2)
-        self.reduce_bpm = (self.slits.mask > 2) & (np.invert(self.slits.bitmask.flagged(
-                        self.slits.mask, flag=self.slits.bitmask.exclude_for_reducing)))
+        # TODO: To my mind, we should never be using the value of the bit to
```
debora-pe, line 169: @kbwestfall I tested this and it looks OK.

Her is the test case, in case you want to add it to the tests:

```
>>> import numpy
>>> from pypeit.slittrace import SlitTraceBitMask
>>> bm = SlitTraceBitMask()

>>> v = numpy.array([10,0,2])

>>> print([bm.flagged_bits(_v) for _v in v])
[['BOXSLIT', 'BADWVCALIB'], [], ['BOXSLIT']]

>>> desired_bpm = (v > 2) & (np.invert(bm.flagged(v, flag=bm.exclude_for_reducing)))
>>> desired_bpm
array([ True, False, False])

>>> new_bpm = bm.flagged(v, exclude='BOXSLIT', and_not=bm.exclude_for_reducing)
>>> new_bpm
array([ True, False, False])

>>> np.all(new_bpm == desired_bpm)
True
```

debora-pe (approved):

pypeit/find_objects.py
kbwestfall: Thanks!  I added this test to `test_bitmask.py`.

PR #1759, opened 2024-01-18
GMOS detector list fix

jhod0:
Per discussion with @kbwestfall and @profxj on slack, this fixes a tiny bug in `list_detectors` for GMOS. This issue popped up because I was reducing a single detector at a time, instead of the full 3-detector mosaic.

Sorry for all the tiny changes, my editor is set to automatically clean up trailing whitespace and it seems that there was a lot of that in this file. The real changes is a tiny one-line thing


pypeit/spectrographs/gemini_gmos.py
```diff
@@ -565,13 +565,13 @@ def list_detectors(self, mosaic=False):
             return np.array([self.get_det_name(_det) for _det in self.allowed_mosaics])
 
         return np.array([detector_container.DetectorContainer.get_name(i+1)
-                             for i in range(self.ndet)]).reshape(2,-1)
+                             for i in range(self.ndet)]).reshape(3,1)
```
jhod0: This is the actual change. Instead of 2 rows of detectors in the spectral direction, like DEIMOS, GMOS has 3 detectors covering the whole dispersed spectrum. This should return

```
array([['DET01'],
       ['DET02'],
       ['DET03']], dtype='<U5')
```

kbwestfall (approved):
Thanks!

debora-pe (changes requested):
@jhod0 Thank you for doing this, and spotting this bug.
Actually, I realized that this is not the needed change here, the only thing we need to do is to remove `list_detectors` from `pypeit/spectrographs/gemini_gmos.py`. The parent method in `pypeit/spectrographs/spectrograph.py` is already properly taking care of this, but because we forgot to remove this in  `gemini_gmos.py`, that was getting overwritten.

kbwestfall:
> @jhod0 Thank you for doing this, and spotting this bug. Actually, I realized that this is not the needed change here, the only thing we need to do is to remove `list_detectors` from `pypeit/spectrographs/gemini_gmos.py`. The parent method in `pypeit/spectrographs/spectrograph.py` is already properly taking care of this, but because we forgot to remove this in `gemini_gmos.py`, that was getting overwritten.

@debora-pe I had the same thought, but I was not sure if the shape of the mosaic array needed to follow the (spec,spat) convention.  For DEIMOS, the array is specifically (2,4) because there are two chips spectrally and four chips spatially.  The GMOS chips are primarily spectral (if I remember right), so the shape should be (3,1) if we need to follow that convention.  I didn't dig far enough to understand if the convention actually mattered in this case...

debora-pe:
> > @jhod0 Thank you for doing this, and spotting this bug. Actually, I realized that this is not the needed change here, the only thing we need to do is to remove `list_detectors` from `pypeit/spectrographs/gemini_gmos.py`. The parent method in `pypeit/spectrographs/spectrograph.py` is already properly taking care of this, but because we forgot to remove this in `gemini_gmos.py`, that was getting overwritten.
> 
> @debora-pe I had the same thought, but I was not sure if the shape of the mosaic array needed to follow the (spec,spat) convention. For DEIMOS, the array is specifically (2,4) because there are two chips spectrally and four chips spatially. The GMOS chips are primarily spectral (if I remember right), so the shape should be (3,1) if we need to follow that convention. I didn't dig far enough to understand if the convention actually mattered in this case...

@kbwestfall This `list_detector` method always creates a lot of confusion! I see your point and it makes sense. The problem is that this method was originally created for DEIMOS for the purpose of dividing the blue and red detector so that `pypeit.slittrace.average_maskdef_offset` could computes the offsets separately for red and blue detectors (side note, since we mostly use mosaic for DEIMOS, this is not crucial anymore but it's good to keep it). So, `average_maskdef_offset` can take a 1D array of detectors (in that case it does not make the differentiation between red and blue) or a 2D array of detectors where the red and blue detector are separated. 
In your reasoning, it would make sense to have the shape be (3,1) so that  `average_maskdef_offset` would compute the offsets for each detector in the spectral direction, but `average_maskdef_offset` doesn't know how to deal with a third (or fourth, fifth etc) detectors. It knows "red" and "blue", but not "green", so having the shape (3,1) does not help, and possibly harms the reduction if using the slitmask definition. We could make changes to `average_maskdef_offset`, but considering that we generally use the mosaic, it's probably not worth it to spend that time.

kbwestfall:
Thanks for digging in on this @debora-pe !

@jhod0 , I agree with @debora-pe .  I think you should try to simply remove the `list_detectors` method from the `gemini_gmos` class, and make sure that works with your data.  Thanks!

jhod0:
I have removed `list_detectors` from the GMOS class and am re-running a reduction that previously completed. I'll update you all when it completes. Thanks for the quick support!

jhod0:
The reduction seems to have run to completion with the same result as before, so it looks like removing the function worked. Thanks @debora-pe @kbwestfall !

debora-pe (approved):
Thanks!

PR #1760, opened 2024-01-19
Propagate toggling of datamodel version checking

kbwestfall:
Toggling on and off version checking of the datamodels (using `chk_version`) was pretty limited.  This is my attempt to try to fully propagate it through the code.  This essentially means that I tried to properly set the `chk_version` keyword anytime we read a pypeit-produced file.  This includes adding `try_old` to many scripts, and I added a `chk_version` parameter to the `ReduxPar` parameter set (the same group that defines the spectrograph); because of the latter, I deprecated similar parameters in `Collate1DPar` and `CoAdd1DPar`.

I also edited the `SlitTraceSet` datamodel because of an inconsistency in how it checks/stores the bits.  I'm now using the built-in function that writes the bits to a fits header, and the class no longer includes `slitbitm` in the datamodel.

I need to update the docs and run tests, but the code is ready for review.


pypeit/core/telluric.py
```diff
@@ -755,7 +760,8 @@ def general_spec_reader(specfile, ret_flam=False):
         spect_dict = spec.spect_meta
         head = spec.head0
     else:
-        sobjs = specobjs.SpecObjs.from_fitsfile(specfile, chk_version=False)
+        sobjs = specobjs.SpecObjs.from_fitsfile(specfile, chk_version=chk_version)
+        # TODO: What bug?  Is it fixed now?  How can we test if it's fixed?
```
rcooke-ast, line 764: I suspect we want to throw a reasonable user error here, and maybe make a note that this needs to be looked into.

pypeit/core/wavecal/templates.py
```diff
@@ -266,6 +266,7 @@ def pypeit_arcspec(in_file, slit, binspec, binning=None):
         #                   minx=iwv_calib['fmin'], maxx=iwv_calib['fmax'])
         flux = np.array(iwv_calib['spec']).flatten()
     elif '.fits' in in_file:
+        # TODO: This should not default to chk_version=False ...
```
rcooke-ast, line 269: Should we force this to `True`?

pypeit/par/pypeitpar.py
```diff
@@ -1321,12 +1321,16 @@ def __init__(self, ex_value=None, flux_value=None, nmaskedge=None,
         dtypes['coaddfile'] = str
         descr['coaddfile'] = 'Output filename'
 
-
-        # Version checking
-        defaults['chk_version'] = True
-        dtypes['chk_version'] = bool
-        descr['chk_version'] = 'If True enforce strict PypeIt version checking to ensure that spec1d*.fits files were created' \
-                               'with the current version of PypeIt'
+        # NOTE: Deprecated in favor of the parameter in ReduxPar!
```
rcooke-ast: Should we just delete this? I don't mind keeping it around for a while for developers, but I think a quick search of this file will find where `chk_version` is now kept.

pypeit/par/pypeitpar.py
```diff
@@ -5244,9 +5261,10 @@ def __init__(self, tolerance=None, dry_run=None, ignore_flux=None, flux=None, ma
         descr['refframe'] = 'Perform reference frame correction prior to coadding. ' \
                          'Options are: {0}'.format(', '.join(options['refframe']))
 
-        defaults['chk_version'] = False
-        dtypes['chk_version'] = bool
-        descr['chk_version'] = "Whether to check the data model versions of spec1d files and sensfunc files."
+        # NOTE: Deprecated in favor of the parameter in ReduxPar!
```
rcooke-ast: Same here?

pypeit/scripts/collate_1d.py
```diff
@@ -124,7 +124,9 @@ def find_slits_to_exclude(spec2d_files, par):
     exclude_map = dict()
     for spec2d_file in spec2d_files:
 
-        allspec2d = AllSpec2DObj.from_fits(spec2d_file, chk_version=par['collate1d']['chk_version'])
+        allspec2d = AllSpec2DObj.from_fits(spec2d_file,
+                                           chk_version=par['rdx']['chk_version'])
+#                                           chk_version=par['collate1d']['chk_version'])
```
rcooke-ast: remove?

pypeit/scripts/collate_1d.py
```diff
@@ -247,7 +249,9 @@ def read_spec1d_files(par, spec1d_files, failure_msgs):
     good_spec1d_files = []
     for spec1d_file in spec1d_files:
         try:
-            sobjs = SpecObjs.from_fitsfile(spec1d_file, chk_version = par['collate1d']['chk_version'])
+            sobjs = SpecObjs.from_fitsfile(spec1d_file,
+                                           chk_version=par['rdx']['chk_version'])
+#                                           chk_version=par['collate1d']['chk_version'])
```
rcooke-ast: remove?

pypeit/scripts/collate_1d.py
```diff
@@ -301,7 +305,9 @@ def flux(par, spectrograph, spec1d_files, failed_fluxing_msgs):
         # Flux calibrate the spec1d file
         try:
             msgs.info(f"Running flux calibrate on {spec1d_file}")
-            FxCalib = fluxcalibrate.flux_calibrate([spec1d_file], [sens_file], par=par['fluxcalib'], chk_version=par['collate1d']['chk_version'])
+            FxCalib = fluxcalibrate.flux_calibrate([spec1d_file], [sens_file], par=par['fluxcalib'],
+                                                   chk_version=par['rdx']['chk_version'])
+#                                                   chk_version=par['collate1d']['chk_version'])
```
rcooke-ast: remove?

pypeit/scripts/collate_1d.py
```diff
@@ -610,7 +616,8 @@ def build_parameters(args):
         params['collate1d']['refframe'] = args.refframe
 
     if args.chk_version is True:
-        params['collate1d']['chk_version'] = True
+#        params['collate1d']['chk_version'] = True
```
rcooke-ast: remove?

pypeit/scripts/collate_1d.py
```diff
@@ -425,8 +431,8 @@ def coadd(par, coaddfile, source):
     # Set destination file for coadding
     par['coadd1d']['coaddfile'] = coaddfile
     
-    # Whether to be forgiving of data model versions
-    par['coadd1d']['chk_version'] = par['collate1d']['chk_version']
+#    # Whether to be forgiving of data model versions
+#    par['coadd1d']['chk_version'] = par['collate1d']['chk_version']
```
rcooke-ast: This is being lost in the call to `coadd1d.CoAdd1D.get_instance` on line 458 below (although, I'm not really sure it's even needed...)

also, remove these commented lines?

pypeit/scripts/collate_1d.py
```diff
@@ -664,12 +671,16 @@ class Collate1D(scriptbase.ScriptBase):
...
         # A blank Colate1DPar to avoid duplicating the help text.
-        blank_par = pypeitpar.Collate1DPar()
+        blank_pypar = pypeitpar.PypeItPar()
+        blank_par = blank_pypar['collate1d']
 
         parser = super().get_parser(description='Flux/Coadd multiple 1d spectra from multiple '
                                                 'nights and prepare a directory for the KOA.',
                                     width=width, formatter=scriptbase.SmartFormatter)
 
+#                                 'F|  chk_version           If true, spec1ds and archival sensfuncs must match the currently\n'
```
rcooke-ast: remove?

pypeit/scripts/skysub_regions.py
```diff
@@ -55,7 +57,8 @@ def main(args):
             detname = DetectorContainer.get_name(det)
 
         # Load it up
-        spec2DObj = spec2dobj.Spec2DObj.from_file(args.file, detname, chk_version=True)
+        spec2DObj = spec2dobj.Spec2DObj.from_file(args.file, detname,
+                                                  chk_version=(not args.try_old))
```
rcooke-ast: In quite a few places, there's `(not args.try_old)`. Would it be cleaner to define a variable `chk_version=not args.try_old`, and then replace all instances with `chk_version=chk_version`? Probably just a difference of style...

pypeit/slittrace.py
```diff
@@ -30,6 +30,7 @@ class SlitTraceBitMask(BitMask):
     Mask bits used during slit tracing.
     """
     version = '1.0.1'
+    # TODO: Need a unique bit prefix?
```
rcooke-ast: Just flagging this   :-)

pypeit/wavecalib.py
```diff
@@ -401,6 +401,7 @@ def wave_diagnostics(self, print_diag=False):
         diag['IDs_Wave_cov(%)'] = lines_cov
         diag['IDs_Wave_cov(%)'].format = '0.1f'
         # FWHM
+        # TODO: Why not me*a*sured_fwhm?
```
rcooke-ast: yeah, this looks like a typo. It seems to have been propagated through several bits of the code and docs (e.g. `pypeit.scripts.chk_wavecalib`)

rcooke-ast (approved):
This is a great unification of `chk_version` and `try_old`, and I personally think this makes a lot more sense, and the code is much cleaner. There's only the (slight) disadvantage that one can't disable `chk_version` for just one datamodel. However, I think if you're in the territory of disabling this, then you're probably happy to disable all datamodel checking simultaneously  :-)

Anyway, this all looks good to me, and I'm happy to approve with minor comments, but please confirm tests are passing (I'm sure you will!).

pypeit/core/telluric.py
kbwestfall: I need some help with this.  The TODO comment was definitely not the best way to communicate, but I want to understand what bug has been identified in DataContainer, if it's still a bug, and a useful test.  I'd rather deal with the root issue instead of improving the error message.

pypeit/core/wavecal/templates.py
kbwestfall: I hesitate because I assume there was some reason why this was set to False.  Another issue I need to understand better before taking any action.

pypeit/par/pypeitpar.py
kbwestfall: Removed.

pypeit/par/pypeitpar.py
kbwestfall: Removed.

pypeit/scripts/collate_1d.py
kbwestfall: Done.

pypeit/scripts/collate_1d.py
kbwestfall: Done.

pypeit/scripts/collate_1d.py
kbwestfall: Done.

pypeit/scripts/collate_1d.py
kbwestfall: Done.

pypeit/scripts/collate_1d.py
kbwestfall: Done.

pypeit/scripts/collate_1d.py
kbwestfall: Done.

pypeit/slittrace.py
kbwestfall: Thanks!  I expanded the comment (i.e., how to do it and saying that it's not currently necessary).  Given that it's not necessary, I'm okay with deleting the comment altogether, if you prefer.

pypeit/wavecalib.py
kbwestfall: I've changed all the instances I could find to `measured_fwhm`.

pypeit/scripts/skysub_regions.py
kbwestfall: Good recommendation.  All the scripts using `try_old` now define `chk_version = not args.try_old` somewhere near the top of the script.

kbwestfall (commented):
Thanks for the review!

kbwestfall:
I need to fix a couple of vet tests and re-run the dev-suite.

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  242 passed, 66 warnings in 443.54s (0:07:23) ---
--- PYTEST UNIT TESTS PASSED  133 passed, 171 warnings in 1065.64s (0:17:45) ---
--- PYTEST VET TESTS FAILED  2 failed, 59 passed, 69 warnings in 1668.25s (0:27:48) ---
Failed tests:
    keck_kcrm/medium_rh3 pypeit
Skipped tests:
Testing Started at 2024-01-24T16:57:13.931812
Testing Completed at 2024-01-25T03:45:57.467014
Total Time: 10:48:43.535202
```

kbwestfall:
Tests pass

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  242 passed, 70 warnings in 527.47s (0:08:47) ---
--- PYTEST UNIT TESTS PASSED  133 passed, 171 warnings in 966.50s (0:16:06) ---
--- PYTEST VET TESTS PASSED  61 passed, 70 warnings in 1775.89s (0:29:35) ---
Failed tests:
    keck_kcrm/medium_rh3 pypeit
Skipped tests:
Testing Started at 2024-01-25T20:49:36.044969
Testing Completed at 2024-01-26T14:48:21.489263
Total Time: 17:58:45.444294
```

pypeit/wavecalib.py
```diff
@@ -182,28 +182,28 @@ def _parse(cls, hdu, **kwargs):
                 else:
                     # TODO -- Replace the following with WaveFit._parse() and pass that back!!
                     iwavefit = wv_fitting.WaveFit.from_hdu(ihdu)# , chk_version=False)
-                    parsed_hdus += ihdu.name
+                    parsed_hdus += [ihdu.name]
```
debora-pe: `wv_fitting.WaveFit.from_hdu` also needs a `chk_version` otherwise will crash if for example `pypeit_chk_wavecalib` is run.
Did you leave it out on purpose?

pypeit/scripts/show_2dspec.py
```diff
@@ -137,7 +141,7 @@ def main(args):
         # Try to read the Spec2DObj using the current datamodel, but allowing
         # for the datamodel version to be different
         try:
-            spec2DObj = spec2dobj.Spec2DObj.from_file(args.file, detname, chk_version=False)
+            spec2DObj = spec2dobj.Spec2DObj.from_file(args.file, detname, chk_version=chk_version)
```
debora-pe, line 144: I tried to run this with an older version of PypeIt and if I don't use the `--try_old` flag it still tries to show the 2D spectra. but they are a bit off. If I run it with the `--try_old` flag, everything looks good.
I think the problem is the `try/except` code here and the one below after `if spec2DObj is None:`
Should we remove it since now we have the `--try_old` flag?

pypeit/flatfield.py
```diff
@@ -414,7 +419,7 @@ def show(self, frametype='all', slits=None, wcs_match=True):
             slits_file = slittrace.SlitTraceSet.construct_file_name(self.calib_key,
                                                                     calib_dir=self.calib_dir)
             try:
-                slits = slittrace.SlitTraceSet.from_file(slits_file)
+                slits = slittrace.SlitTraceSet.from_file(slits_file, chk_version=chk_version)
```
debora-pe, line 422: Same here as `pypeit_show_2dspec`, should we remove `try/except` and the code below `if slits is not None:`?

pypeit/coadd3d.py
```diff
@@ -578,7 +588,9 @@ def set_default_scalecorr(self):
                 msgs.info("Loading default scale image for relative spectral illumination correction:" +
                           msgs.newline() + self.cubepar['scale_corr'])
                 try:
-                    spec2DObj = spec2dobj.Spec2DObj.from_file(self.cubepar['scale_corr'], self.detname)
+                    spec2DObj = spec2dobj.Spec2DObj.from_file(self.cubepar['scale_corr'],
```
debora-pe, line 591: Same here (and a few times more below) as `pypeit_show_2dspec`? Can we remove `try/except` and only rely on setting  the `chk_version` parameter. Maybe @rcooke-ast can respond to this.

pypeit/scripts/chk_scattlight.py
```diff
@@ -47,18 +51,19 @@ def main(args):
...
+        slits = slittrace.SlitTraceSet.from_file(args.slits, chk_version=chk_version)
 
         # Load the alternate file if requested
         display_frame = None  # The default is to display the frame used to calculate the scattered light model
         if args.spec2d is not None:
             msgs.error("displaying the spec2d scattered light is not currently supported")
             try:
                 # TODO :: the spec2d file may have already had the scattered light removed, so this is not correct. This script only works when the scattered light is turned off for the spec2d file
-                spec2D = spec2dobj.Spec2DObj.from_file(args.spec2d, detname, chk_version=False)
+                spec2D = spec2dobj.Spec2DObj.from_file(args.spec2d, detname,
```
debora-pe, line 65: here it seems you can remove `try/except` and modify `display_frame` below?
See the other comments about this.

pypeit/calibrations.py
```diff
@@ -153,6 +158,7 @@ def __init__(self, fitstbl, par, spectrograph, caldir, qadir=None,
 
         # Calibrations
         self.reuse_calibs = reuse_calibs
+        self.chk_version = chk_version
```
debora-pe, line 161: To be able to use chk_version for `run_pypeit` should we pass `par['rdx']['chk_version']` to `pypeit.calib_all()` and `pypeit.calib_one()`?

debora-pe (changes requested):
Hi @kbwestfall. I have an old Dev Suite reduction (v1.13.1.dev) so I tested a bit these changes. I have a few comments.

pypeit/coadd3d.py
rcooke-ast: Given that `from_file` calls `io.fits_open` which checks if the file exists (and includes a few more try/except statements), I think this is a good point. It might be a good thing to fault, anyway. So, perhaps this code can just be changed to:

```
spec2DObj = spec2dobj.Spec2DObj.from_file(self.cubepar['scale_corr'],
                                          self.detname,
                                          chk_version=self.chk_version)
self.relScaleImgDef = spec2DObj.scaleimg
self.scalecorr_default = self.cubepar['scale_corr']
```

and the equivalent for the three other instances in `coadd3d.py`:
(`get_current_scalecorr`, `get_current_skysub`, `set_current_skysub`)

pypeit/calibrations.py
kbwestfall: Wowza!  Thanks for catching that!  I don't need to pass it to these functions, they already have access to `self.par['rdx']['chk_version']`, so I now use that whenever a `Calibrations` object is instantiated.

pypeit/scripts/show_2dspec.py
kbwestfall: Thanks for the prompt to think this through more deeply.  I think we should leave it as is for the following reason.

The `chk_version` approach is very basic.  It just checks if the version numbers are different, and raises a `PypeItDataModelError` if they aren't.  But the datamodel code may throw a `PypeItDataModelError` for other reasons.  So it's possible for their to be cases when someone tries `chk_version=False`, but this try block still catches an exception.  In that case, we still want the code to try to pull the necessary data from the spec2d file and show it.  The warning should hopefully prompt the user to appreciate that, if the results don't look right, it could be because the datamodel changed and they should re-reduce the data.

What do you think?

pypeit/coadd3d.py
kbwestfall: For similar reasons as in `show_spec2d.py`, I'm not sure we want to remove the try block entirely.  But I think it would be good to specify that the except block is specific to datamodel errors.  I.e., does it make sense to replace `Exception` with `PypeItDataModelError`?

pypeit/flatfield.py
kbwestfall: Same basic response.  I've changed `PypeItError` to `PypeItDataModelError`.

pypeit/scripts/chk_scattlight.py
kbwestfall: Same basic response.

pypeit/wavecalib.py
kbwestfall: Good call!  This prompted me to basically rewrite how we read in `WaveCalib` objects.

kbwestfall (commented):
Thanks for the commens, @debora-pe !  This prompted me to rewrite how we read in `WaveCalib` files (addressing a long-standing TODO).  As for the questions about the try/except blocks, I think we should keep them, and I explain why in the comments.  But please let me know what you think, and it feels like it's worth a longer discussion, either over Slack or at the next telecon.

pypeit/coadd3d.py
rcooke-ast: Yes, I think that would be equally fine here. However, would we need to keep in the `Exception`, as well? For example:

```
try:
    BLAH
except PypeItDataModelError as e:
    BLAH
except Exception as e:
    BLAH
else:
    BLAH
```

kbwestfall: If possible, I'd rather we not use the base Exception class, unless we know that we should handle all other exceptions in the same way.  I.e., try/except blocks are easier to maintain/debug when they focus on specific known failure modes that we know how to handle.

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  243 passed, 70 warnings in 567.31s (0:09:27) ---
--- PYTEST UNIT TESTS PASSED  133 passed, 171 warnings in 1033.73s (0:17:13) ---
--- PYTEST VET TESTS PASSED  61 passed, 70 warnings in 1872.43s (0:31:12) ---
Failed tests:
    keck_kcrm/medium_rh3 pypeit
Skipped tests:
Testing Started at 2024-01-31T00:47:34.895983
Testing Completed at 2024-01-31T21:43:03.974158
Total Time: 20:55:29.078175
```

pypeit/scripts/show_2dspec.py
debora-pe: Hi @kbwestfall, in general I'm OK to keep try/except and, as you said, it makes sense for when  `chk_version=False`. What I am most worried is when `chk_version=True`, which is the default. If users run `pypeit_show_2dspec` without `--try_old` (as they are used to) with an older version of spec2d, the script still shows something (because of the try/except and the code below that try to pull the necessary data from the spec2d file and show it) and I'm assuming users will not read the warning and will be OK with what they get. The problem is that the code after `if spec2DObj is None:` is not often updated, so it may not show the correct frames (I tried it with a v1.13.1 spec2d and the SCIIMG is not right).
That said, I think the ideal situation would be that if the script is run with `chk_version=True`, the script faults and directs the user to use `chk_version=False`. If the script is run with `chk_version=False`, try/except caches any `PypeItDataModelError` and allows the script to show something anyway using the code after `if spec2DObj is None:`.

debora-pe (approved):
Hi @kbwestfall I am OK with all the changes. Thank you!
I only have still a concern about `try/except` in `show_2dspec`. See my comment. If you could  implement what I explain there would be great. If not, it's fine. I'm approving anyway.

pypeit/scripts/show_2dspec.py
kbwestfall: Ah, I get it now.  Yes, I like your suggested ideal situation.  I've pushed changes that should implement this.  Can you give it a test run?

debora-pe: Thanks @kbwestfall! This looks great!

debora-pe (approved):
Approve again!

PR #1761, opened 2024-01-25
Vetting edge traces resulting from peak finding in rectified and collapsed image

kbwestfall:
One of the steps in edge finding is, after the PCA of the edge trace form has been constructed, the trace image is rectified and collapsed and used to re-identify the trace locations (this is in `EdgeTraceSet.peak_refine`).  Once the locations are found, the edges are retraced and the modeled by a polynomial.  Particularly for edges near the edges of the detector or others that don't have well-detected edges over the full range of the detector along the dispersion direction, the polynomial fit is poorly constrained and goes wonky.  Here's an example from the `keck_hires/J0100+2802_N255Hr_RED_C2_ECH_-0.84_XD_1.29_1x3` dataset in the dev-suite:

Figure_1

The left most trace is poorly constrained and causes havoc later in the code flow.  This PR tries to identify and remove such traces.  This is something I had in mind as something that should be done for a while, but this HIRES dataset provided a reason to implement it.

The code now matches the traces found by the `core.trace.peak_trace` algorithm to the original set of traces used to define the PCA.  It then calculates the RMS between the old and new traces and removes any that are above the tolerance set by the new parameter `trace_rms_tol`.  This will not catch *new* traces identified by `core.trace.peak_trace` that are poorly constrained, but it does catch cases like the one above.

Test pass; I'll post the dev-suite results when they finish.


pypeit/edgetrace.py
```diff
@@ -3572,6 +3584,43 @@ def peak_refine(self, rebuild_pca=False, debug=False):
...
+            # Determine the RMS difference between the input and output traces.
+            # This allows us to compare traces that had already been identified
+            # to their new measurements resulting from peak_trace, and remove
+            # them if they are too discrepant from their original form.  This is
+            # largely meant to find and remove poorly constrained traces, where
+            # the polynomial fit goes wonky.
+            diff = fit - self.edge_fit.T[gpm][peak_indx].T
+            rms = np.sqrt(np.mean((diff - np.mean(diff, axis=0)[None,:])**2, axis=0))
+
+            # TODO: Add a report to the screen or a QA plot?
```
rcooke-ast: I think it might be good to print the rms of each order to screen.

pypeit/par/pypeitpar.py
```diff
@@ -3325,6 +3325,13 @@ def __init__(self, filt_iter=None, sobel_mode=None, edge_thresh=None, sobel_enha
                                 'image (see `trace_median_frac`), values in the median-filtered ' \
                                 'image *below* this threshold are masked in the refitting of ' \
                                 'the edge trace data.  If None, no masking applied.'
+        
+        dtypes['trace_rms_tol'] = [int, float]
+        descr['trace_rms_tol'] = 'After retracing edges using peaks detected in the rectified ' \
+                                 'and collapsed image, the RMS difference between the original ' \
+                                 'and refit traces are calculated.  This sets the upper limit ' \
```
rcooke-ast: It's probably obvious, but maybe just be explicitly clear that the units is pixels

pypeit/edgetrace.py
```diff
@@ -3572,6 +3584,43 @@ def peak_refine(self, rebuild_pca=False, debug=False):
...
+
+            # Select traces below the RMS tolerance or that were newly
+            # identified by peak_trace.  I.e., this will *not* catch newly
+            # identified traces found by peak_trace that are also poorly
+            # constrained!
+            indx = (rms < self.par['trace_rms_tol']) | (peak_indx == -1)
+            if not all(indx):
```
rcooke-ast: `if not np.all(indx):` ?

rcooke-ast (commented):
Thanks @kbwestfall for the fix. I have added a few minor comments inline, and raise a few more general things to think about here:

(1) It would be good to see an image of what the final result is (i.e. you show an image of it not working, it would be good to see what it looks like when you run this new branch on the data)

(2) I wonder if removing the slit is the right approach. I would think the PCA extrapolation might give a better result in some cases than the refitting approach, so is it possible to revert back to using the PCA for the additional edge by default, and then give the user an option to delete the additional edge? I worry about adding yet another parameter (!) to cover this option of either including the (unfitted) PCA edge or removing the slit entirely.

(3) Maybe add something to the `release_notes`?

pypeit/edgetrace.py
kbwestfall: Done!  Here's an example:
```
[INFO]    :: edgetrace.py 3610 peak_refine() - ------------------------------
[INFO]    :: edgetrace.py 3611 peak_refine() - Matched spatial locations and RMS difference along spectral direction
[INFO]    :: edgetrace.py 3612 peak_refine() -       OLD      NEW      RMS
[INFO]    :: edgetrace.py 3613 peak_refine() -  -------- -------- --------
[INFO]    :: edgetrace.py 3617 peak_refine() -      13.6     18.6   25.434
[INFO]    :: edgetrace.py 3617 peak_refine() -      37.2     37.2    0.015
[INFO]    :: edgetrace.py 3617 peak_refine() -      74.2     74.2    0.041
[INFO]    :: edgetrace.py 3617 peak_refine() -     112.0    112.0    0.062
[INFO]    :: edgetrace.py 3617 peak_refine() -     150.6    150.6    0.091
[INFO]    :: edgetrace.py 3617 peak_refine() -     232.4    232.4    0.088
[INFO]    :: edgetrace.py 3617 peak_refine() -     274.1    274.1    0.078
[INFO]    :: edgetrace.py 3617 peak_refine() -     317.4    317.3    0.035
[INFO]    :: edgetrace.py 3617 peak_refine() -     362.2    362.2    0.075
[INFO]    :: edgetrace.py 3617 peak_refine() -     408.5    408.4    0.103
[INFO]    :: edgetrace.py 3617 peak_refine() -     456.2    456.2    0.132
[INFO]    :: edgetrace.py 3617 peak_refine() -     505.4    505.4    0.160
[INFO]    :: edgetrace.py 3617 peak_refine() -     556.3    556.3    0.193
[INFO]    :: edgetrace.py 3617 peak_refine() -     609.0    609.0    0.223
[INFO]    :: edgetrace.py 3617 peak_refine() -     663.5    663.4    0.223
[INFO]    :: edgetrace.py 3617 peak_refine() -     720.1    720.1    0.225
[INFO]    :: edgetrace.py 3617 peak_refine() -     778.6    778.6    0.175
[INFO]    :: edgetrace.py 3617 peak_refine() -     839.3    839.3    0.124
[INFO]    :: edgetrace.py 3617 peak_refine() -     902.4    902.4    0.076
[INFO]    :: edgetrace.py 3617 peak_refine() -     967.9    967.9    0.023
[INFO]    :: edgetrace.py 3617 peak_refine() -    1036.1   1036.1    0.051
[INFO]    :: edgetrace.py 3617 peak_refine() -    1107.2   1107.2    0.126
[INFO]    :: edgetrace.py 3617 peak_refine() -    1181.5   1181.4    0.208
[INFO]    :: edgetrace.py 3617 peak_refine() -    1259.1   1259.0    0.299
[INFO]    :: edgetrace.py 3617 peak_refine() -    1340.3   1340.3    0.319
[INFO]    :: edgetrace.py 3617 peak_refine() -    1426.3   1426.3    0.483
[INFO]    :: edgetrace.py 3617 peak_refine() -    1515.8   1515.7    0.367
[INFO]    :: edgetrace.py 3617 peak_refine() -    1609.9   1609.9    0.237
[INFO]    :: edgetrace.py 3617 peak_refine() -    1709.3   1709.3    0.090
[INFO]    :: edgetrace.py 3617 peak_refine() -    1814.3   1814.3    0.088
[INFO]    :: edgetrace.py 3617 peak_refine() -    1925.8   1925.8    0.277
[INFO]    :: edgetrace.py 3617 peak_refine() -      42.3     42.3    0.084
[INFO]    :: edgetrace.py 3617 peak_refine() -      78.1     78.1    0.075
[INFO]    :: edgetrace.py 3617 peak_refine() -     114.8    114.8    0.047
[INFO]    :: edgetrace.py 3617 peak_refine() -     152.8    152.8    0.020
[INFO]    :: edgetrace.py 3617 peak_refine() -     230.4    230.4    0.132
[INFO]    :: edgetrace.py 3617 peak_refine() -     272.0    272.1    0.120
[INFO]    :: edgetrace.py 3617 peak_refine() -     314.5    314.5    0.047
[INFO]    :: edgetrace.py 3617 peak_refine() -     357.9    357.9    0.048
[INFO]    :: edgetrace.py 3617 peak_refine() -     402.6    402.7    0.071
[INFO]    :: edgetrace.py 3617 peak_refine() -     448.8    448.8    0.096
[INFO]    :: edgetrace.py 3617 peak_refine() -     496.4    496.4    0.120
[INFO]    :: edgetrace.py 3617 peak_refine() -     545.6    545.6    0.146
[INFO]    :: edgetrace.py 3617 peak_refine() -     596.4    596.4    0.174
[INFO]    :: edgetrace.py 3617 peak_refine() -     648.9    648.9    0.160
[INFO]    :: edgetrace.py 3617 peak_refine() -     703.6    703.7    1.011
[INFO]    :: edgetrace.py 3617 peak_refine() -     759.9    759.9    0.252
[INFO]    :: edgetrace.py 3617 peak_refine() -     818.2    818.2    0.206
[INFO]    :: edgetrace.py 3617 peak_refine() -     878.8    878.8    0.156
[INFO]    :: edgetrace.py 3617 peak_refine() -     941.7    941.7    0.091
[INFO]    :: edgetrace.py 3617 peak_refine() -    1007.1   1007.2    0.011
[INFO]    :: edgetrace.py 3617 peak_refine() -    1075.3   1075.4    0.043
[INFO]    :: edgetrace.py 3617 peak_refine() -    1146.4   1146.4    0.128
[INFO]    :: edgetrace.py 3617 peak_refine() -    1220.6   1220.7    0.209
[INFO]    :: edgetrace.py 3617 peak_refine() -    1298.3   1298.3    0.299
[INFO]    :: edgetrace.py 3617 peak_refine() -    1465.7   1465.7    0.471
[INFO]    :: edgetrace.py 3617 peak_refine() -    1555.3   1555.3    0.354
[INFO]    :: edgetrace.py 3617 peak_refine() -    1649.6   1649.6    0.227
[INFO]    :: edgetrace.py 3617 peak_refine() -    1749.1   1749.1    0.088
[INFO]    :: edgetrace.py 3617 peak_refine() -    1854.3   1854.4    0.099
[INFO]    :: edgetrace.py 3617 peak_refine() -    1965.9   1965.9    0.289
[INFO]    :: edgetrace.py 3628 peak_refine() - Removing 1 trace(s) due to large RMS difference with previous trace locations.
```

pypeit/edgetrace.py
kbwestfall: `all` actually works, but, yes, good call.

pypeit/par/pypeitpar.py
kbwestfall: Good catch.  Added!

kbwestfall (commented):
Thanks for the quick review!

> (1) It would be good to see an image of what the final result is (i.e. you show an image of it not working, it would be good to see what it looks like when you run this new branch on the data)

Good point.  Here's an image right after `peak_refine` (showing the removed trace):

Figure_2

And here's after it does the left-right syncing:

Figure_3

`overlap` is set to True for this dataset, which means it identifies abnormally short slits as overlap regions, swaps the left-right designation for these "slits", and then inserts lefts and rights as necessary (just as in a normal sync).

> (2) I wonder if removing the slit is the right approach. I would think the PCA extrapolation might give a better result in some cases than the refitting approach, so is it possible to revert back to using the PCA for the additional edge by default, and then give the user an option to delete the additional edge? I worry about adding yet another parameter (!) to cover this option of either including the (unfitted) PCA edge or removing the slit entirely.

Good question.  A few points of clarification:  First, this new check doesn't remove a full slit (i.e., both left and right edges), just a bad edge trace.  In this example dataset, it's only removing the left edge, which is then replaced by the PCA during the syncing.  It's only the case when only one edge of a slit/order is detected (and this RMS tolerance is violated) that you would lose an entire slit/order.

Second, the refitting of the traces once they've been identified in the rectified and collapsed image is deeply ingrained in the `core.trace.peak_trace` function.  We could alter that function, but I think the idea is that a direct fit guided by (initialized by) the location of the PCA typically provides a better description of the edges.  All this is to say, if we want to allow your suggestion, we probably also want to enable users to continue using the current algorithm.  So, I'm not sure there's a way around the need for a new parameter, which either specifies this tolerance to identify traces to remove or toggles a different tracing mode.

Third, I just wanted to re-emphasize that this RMS rejection only works for traces that existed both before and after running `peak_trace`.  The point being, if the trace is dramatically different between the two tracing methods, it's likely a poor trace to begin with.

Let me know what you think.

> (3) Maybe add something to the release_notes?

Thanks for the prompt to do this.  Done!

pypeit/edgetrace.py
```diff
@@ -2542,7 +2544,7 @@ def check_synced(self, rebuild_pca=False):
                 msgs.info('Found {0} slit(s) with gaps below {1} arcsec ({2:.2f} pixels).'.format(
                             np.sum(indx), self.par['minimum_slit_gap'], gap_atol))
                 rmtrace = np.concatenate(([False],np.repeat(indx,2),[False]))
-                self.remove_traces(rmtrace, rebuild_pca=rebuild_pca)
+                self.remove_traces(rmtrace, rebuild_pca=_rebuild_pca)
```
debora-pe, line 2547: I take this occasion to mention that I found a typo in line 2514. The variable defined there should be `dlength_atol` and not `length_atol`, which is defined 2 lines below that.

debora-pe (approved):
Thanks @kbwestfall. This looks great. I checked all the HIRES datasets in the Dev Suite and `trace_rms_tol = 5.` works great for all.
Just a clarification, the dataset `keck_hires/J0100+2802_N255Hr_RED_C2_ECH_-0.84_XD_1.29_1x3` is in the Dev Suite Google Drive, but it's not yet added to the Dev Suite runs. I'm planning to add it at certain point as part of the ADAP work.

rcooke-ast:
> Let me know what you think.

All sounds good to me - thanks for the detailed response!

rcooke-ast (approved):

pypeit/edgetrace.py
kbwestfall: Thanks for finding that!! Definitely a bug.  Fixed.

kbwestfall:
I had to restart the tests for this branch.  Something caused them to hang...

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  243 passed, 66 warnings in 393.73s (0:06:33) ---
--- PYTEST UNIT TESTS PASSED  133 passed, 171 warnings in 1010.43s (0:16:50) ---
--- PYTEST VET TESTS PASSED  61 passed, 71 warnings in 1555.49s (0:25:55) ---
Failed tests:
    keck_kcrm/medium_rh3 pypeit
Skipped tests:
Testing Started at 2024-01-29T16:27:37.613667
Testing Completed at 2024-01-30T03:00:24.389698
Total Time: 10:32:46.776031
```

PR #1764, opened 2024-01-31
HOTFIX : spec2d hotfix

rcooke-ast:
Makes sure that the BPM (and no other channels) get cleared when displayed with the `pypeit_show_2dspec` script.


kbwestfall (approved):
I'm not sure why the change for the use of `clear` works, but I confirmed it fixes the bug.  I added some new functionality and updated the docs.  @rcooke-ast , please take it for a test drive!

rcooke-ast:
This all looks good to me @kbwestfall - thanks for the updates! I was also thinking of adding a script, that you supply the spec2d file, and a number, and it tells you what the masked bit is. That way, folks can call:

`pypeit_show_2dspec spec2d_file.fits --showmask`

_User inspects the BPM in the ginga window, identifies a masked pixel (let's say in this case the BPM value is 24), and then on a command line does:_

```
pypeit_print_bpm spec2d_file.fits 24
>> The specified bad pixel contains the following masked values:
>> Cosmic Ray, Saturation, Bad Slit
```

What do you think?

kbwestfall:
Hi @rcooke-ast .  Perhaps in the longer term, it would be interesting to work with @ejeschke to develop a ginga plugin that essentially allows the user to hover their mouse above a mask pixel value and have ginga display the associated flag names in a dialog box somewhere.  I think your script is a fine stop-gap solution.  Though, why would you need to provide the fits image?  I think you should be able to just do `pypeit_print_bpm 24`, right?

rcooke-ast:
> Hi @rcooke-ast . Perhaps in the longer term, it would be interesting to work with @ejeschke to develop a ginga plugin that essentially allows the user to hover their mouse above a mask pixel value and have ginga display the associated flag names in a dialog box somewhere. I think your script is a fine stop-gap solution. Though, why would you need to provide the fits image? I think you should be able to just do `pypeit_print_bpm 24`, right?

That's a great idea - I've already implemented a script that does that (I'll push my changes momentarily)... the spec2d file is only needed if the imagebitmask changes for different version of PypeIt... not critical...

kbwestfall:
> the spec2d file is only needed if the imagebitmask changes for different version of PypeIt... not critical...

Ah, right.  And this is how things get messy...  FWIW, I would allow users to pass the file as an option in the script, and I would then signal the script to set `chk_version` to False whenever the user provides the file.

rcooke-ast:
Here is what the output looks like in a terminal (`pypeit_print_bpm 23`):

image

kbwestfall:
Nice.  I would actually give both the flag name and the description in the output.  I.e.,:
```
 * BPM: Component of the instrument-specific bad pixel mask
 * CR: Cosmic ray detected
```
etc.

rcooke-ast:
Sounds good... I've included that in as well  :-)

image

debora-pe (approved):
Looks good. Thanks!

PR #1769, opened 2024-02-13
Reduce package distribution size

kbwestfall:
The primary goal of this PR is to reduce the size of the package distribution installed using pip.  To do so, I 

 1. refactored @tbowers7 cache system, largely to make the code more general,
 2. added more data files to the list being controlled by the cache
 3. removed the doc/ directory from the distribution

For item 1, I added the `pypeitdata.py` and `cache.py` files, which are refactored pieces of what was `data/utils.py`, and then I moved some of the `load_*` functions from `data/utils.py` into `io.py`.  With this refactor, I've removed the `pypeit.data` module and replaced it with the `pypeit.dataPaths` instance; see the changes to the main `__init__.py` file.  If this is accepted, I'd like to make the `data/` directory exclusively for data files; i.e., it should not have any code in it.

After some testing, this reduces the size of the tar file from ~145MB for version 1.15.0 to about 12MB.  The big ticket items were the test data files, some of the standards, and doc figures directory.

There's no hurry on this one, but I wanted to issue this so that everyone can start to comment.


tbowers7:
@kbwestfall .  I'm starting to dig into this one.  Could you provide a rationale for the change from `pathlib.Path.resolve()` to `pathlib.Path.absolute()`?  What I'm seeing in the docs suggests that the former resolves symlinks and normalizes paths (removing, _e.g._, `/../`) whereas the latter does not.  Perhaps in the end it doesn't matter for our use case, but I'm curious.

kbwestfall:
> @kbwestfall . I'm starting to dig into this one. Could you provide a rationale for the change from `pathlib.Path.resolve()` to `pathlib.Path.absolute()`? What I'm seeing in the docs suggests that the former resolves symlinks and normalizes paths (removing, _e.g._, `/../`) whereas the latter does not. Perhaps in the end it doesn't matter for our use case, but I'm curious.

Yes, thanks for bringing this up.  The underlying reason is that I didn't realize (embarrassingly) that `resolve` actually resolves symlink paths.  Basically anytime I've used `resolve` in my own coding up until about a week ago, I actually really should have just been using `absolute`.  I think, unless we have a good reason to do otherwise on a case-by-case basis, we should be generally using `absolute` instead of `resolve`. 

The more important reason for this PR is that I "moved" all the data that used to be in the `tests/files` directory into the cache system (and move the directory to `data/tests`).  There are a number of tests where we need the name of the file.  When the file is downloaded into the cache system, its name is always `contents`.  By adding a symlink to the file in the pkg directory structure and using `absolute`, I can keep the same access system and also keep the test the same because the name of the symlink is identical to the original filename.  Does that make sense?

pypeit/cache.py
```diff
@@ -0,0 +1,354 @@
...
+from importlib import resources
+import pathlib
+import urllib.error
+from urllib.parse import urljoin
+
+from IPython import embed
+
+import astropy.utils.data
+import github
+from linetools.spectra import xspectrum1d
```
tbowers7: `xspectrum1d` not needed in this module.

pypeit/cache.py
```diff
@@ -0,0 +1,354 @@
...
+
+.. code-block:: python
+
+    import astropy.config.paths
+    print(astropy.config.paths.get_cache_dir('pypeit'))
+
+.. note::
+
+    If the hostname URL for the telluric atmospheric grids on S3 changes, the
+    only place that needs to change is the file ``s3_url.txt``.
```
tbowers7: Include the full path to `s3_url.txt`, since it is no longer in the same directory as this module.

pypeit/cache.py
```diff
@@ -0,0 +1,354 @@
...
+import requests
+
+try:
+    from pygit2 import Repository
+except:
+    Repository = None
+
+# NOTE: To avoid circular imports, avoid (if possible) importing anything from
+# pypeit!  Objects created or available in pypeit/__init__.py are the
+# exceptions, for now.
```
tbowers7: I assume you mean avoid importing anything from `pypeit` _in this module only_.

pypeit/cache.py
```diff
@@ -0,0 +1,354 @@
...
+from IPython import embed
+
+import astropy.utils.data
+import github
+from linetools.spectra import xspectrum1d
+import requests
+
+try:
+    from pygit2 import Repository
+except:
```
tbowers7: `except ImportError:`?  (PEP 8)

pypeit/cache.py
```diff
@@ -0,0 +1,354 @@
...
+    else:
+        remote_url, sources = _build_remote_url(filename, filetype, remote_host=remote_host)
+
+    if remote_host == "s3_cloud" and not install_script:
+        # Display a warning that this may take a while, and the user
+        #   may wish to download using the `pypeit_install_telluric` script
+        # TODO: Is this only true for telluric files?
+        msgs.warn(f"You may wish to download {filename}{msgs.newline()}"
+                  f"independently from your reduction by using the{msgs.newline()}"
+                  "`pypeit_install_telluric` script.")
```
tbowers7: The telgrid files were the monsters (measured in GBs), whereas anything else is generally small (handfulls of MB).  With the shift toward the TelPCA scheme, even these monsters are becoming a thing of the past and this warning may become obsolete at some point.  I originally added this warning to alert a user that they may have to wait a bit for their reduction to run if the monster file was being downloaded on-the-fly.  AFAIK, the only things specifically hosted at `s3_cloud` rather than GitHub are the telluric files.  If this changes, then the warning should also change.

pypeit/cache.py
```diff
@@ -0,0 +1,354 @@
...
+
+    # Get the file from cache, if available, or download from the remote server
+    try:
+        cache_fn = astropy.utils.data.download_file(
+            remote_url,
+            sources=sources,
+            timeout=10,
```
tbowers7, line 208: It may make sense at some point to make the `timeout` adjustable, but I haven't heard complaints about this.

pypeit/pypeitdata.py
```diff
@@ -0,0 +1,368 @@
...
+    thar = dataPaths.nist.get_file_path('ThAr_vacuum.ascii')
+
+All of the attributes of :class:`~pypeit.pypeitdata.PypeItDataPaths` are
+:class:`~pypeit.pypeitdata.PypeItDataPath` objects, such that the code above is
+effectively equivalent to:
+
+.. code-block:: python
+
+    from pypeit.pypeitdata import PypeItDataPath
+    thar = PypeItDataPath('arc_lines/NIST').get_file_path('ThAr_vacuum.ascii')
```
tbowers7, line 41: Nice!

pypeit/pypeitdata.py
```diff
@@ -0,0 +1,368 @@
...
+    """
+
+    _data = resources.files('pypeit') / 'data'
+    """
+    Package-level data path.  This is defined without checking that it exists!
+    """
+
+    def __init__(self, subdirs):
+        self.data = self.check_isdir(self._data)
+        self.path = self.check_isdir(self.data / subdirs)
```
tbowers7: What if `subdirs` is a string containing POSIX-style forward slashes (_i.e._, what happens on Windows)?  For instance, see `pypeit/core/convert_DEIMOSsavfiles.py`, line 35.

pypeit/core/convert_DEIMOSsavfiles.py
```diff
@@ -32,7 +32,7 @@ def sav_to_fits(savfile):
 
     io.write_to_fits(
         sav,
-        data.Paths.static_calibs / 'keck_deimos', f'{savfile_name}.fits',
+        dataPaths.static_calibs.get_file_path(f'keck_deimos/{savfile_name}.fits'),
```
tbowers7, line 35: How does the POIX-style forward slash here play with Windows and the `PypeItDataPath.get_file_path()` function?

pypeit/core/flux_calib.py
```diff
@@ -306,7 +322,7 @@ def stellar_model(V, sptype):
     logg_sol = np.log10(6.67259e-8) + np.log10(1.989e33) - 2.0 * np.log10(6.96e10)
 
     # Load Schmidt-Kaler (1982) table
-    sk82_file = data.Paths.standards / 'kurucz93' / 'schmidt-kaler_table.txt'
+    sk82_file = dataPaths.standards.get_file_path('kurucz93/schmidt-kaler_table.txt')
```
tbowers7, line 317: Same Windows-compatible question about POSIX-style forward slash.

pypeit/core/flux_calib.py
```diff
@@ -351,7 +367,7 @@ def stellar_model(V, sptype):
     indg = np.argmin(np.abs(loggk - logg))
 
     # Grab Kurucz filename
-    std_file = data.Paths.standards / 'kurucz93' / 'kp00' / f'kp00_{int(Tk[indT])}.fits.gz'
+    std_file = dataPaths.standards.get_file_path(f'kurucz93/kp00/kp00_{int(Tk[indT])}.fits.gz')
```
tbowers7, line 362: Forward slash in filename versus `__truediv__` forward slash in the red code.

pypeit/core/flux_calib.py
```diff
@@ -462,7 +478,8 @@ def load_extinction_data(longitude, latitude, extinctfilepar,
         # Observation coordinates
         obs_coord = coordinates.SkyCoord(longitude, latitude, frame='gcrs', unit=units.deg)
         # Read list
-        extinct_summ = data.Paths.extinction / 'README'
+        # TODO: We should give this file a different name!
+        extinct_summ = dataPaths.extinction.get_file_path('README')
```
tbowers7: Agreed.

pypeit/core/wavecal/templates.py
```diff
@@ -1,4 +1,10 @@
-""" Module to generate templates for the PypeIt full_template wavelength calibration routine
+"""
+Module to generate templates for the PypeIt full_template wavelength calibration routine
+
+This function uses direct access to the ``path`` attribute in the relevant
+pypeit data paths.  This should *not* be replicated in other parts of the code.
+This ``templates.py`` script is a developer only script, and it really shouldn't
+be distributed as part of the code base.
```
tbowers7, line 7: Should we move this to the DevSuit as part of this PR?

pypeit/core/wavecal/waveio.py
```diff
@@ -107,7 +108,9 @@ def load_reid_arxiv(arxiv_file):
     """
     # This function allows users to specify their own `reid_arxiv`, in
     #   particular, the output from `pypeit_identify`.
-    calibfile, arxiv_fmt = data.get_reid_arxiv_filepath(arxiv_file)
+    # NOTE: I use to_pkg='symlink' here so that the extension of the file is correct.
+    calibfile, arxiv_fmt = dataPaths.reid_arxiv.get_file_path(arxiv_file, return_format=True,
+                                                              to_pkg='symlink')
```
tbowers7: What happens with the `to_pkg='symlink'` option when the user's installation of PypeIt is read-only for some reason (_e.g._, running on a server or cluster, maybe)?

This was something I had thought about when crafting the cache system in the first place, so that things downloaded would be completely independent of the installed package.

pypeit/pypeitdata.py
```diff
@@ -0,0 +1,368 @@
...
+            format.  Any ``'.gz'`` extensions are stripped.
+        """
+        _f = f.with_suffix('') if f.suffix == '.gz' else f
+        return _f.suffix.replace('.','').lower()
+
+    # TODO: Should to_pkg default to 'symlink'?
+    def get_file_path(self, data_file, to_pkg=None, return_format=False):
+        """
+        Return the path to a file.
+
```
tbowers7: I really like how this generalizes the various `get_filepath` scripts I had originally!

pypeit/spectrographs/keck_deimos.py
```diff
@@ -1230,13 +1231,13 @@ def get_amapbmap(self, filename):
...
 
-        mp_dir = data.Paths.static_calibs / 'keck_deimos'
-
         if slider in [3,4]:
-            self.amap = fits.getdata(mp_dir / f'amap.s{slider}.2003mar04.fits')
-            self.bmap = fits.getdata(mp_dir / f'bmap.s{slider}.2003mar04.fits')
+            self.amap = fits.getdata(dataPaths.static_calibs.get_file_path(
+                                        f'keck_deimos/amap.s{slider}.2003mar04.fits'))
+            self.bmap = fits.getdata(dataPaths.static_calibs.get_file_path(
+                                        f'keck_deimos/bmap.s{slider}.2003mar04.fits'))
```
tbowers7, line 1238: String `/` vs `__truediv__`

pypeit/spectrographs/keck_lris.py
```diff
@@ -1527,19 +1527,15 @@ def get_detector_par(self, det, hdu=None):
             amp_mode = hdu[0].header['AMPMODE']
             msgs.info("AMPMODE = {:s}".format(amp_mode))
             # Load up translation dict
-            ampmode_translate_file = (
-                data.Paths.data / 'spectrographs' / 'keck_lris_red_mark4' / 'dict_for_ampmode.json'
-            )
+            ampmode_translate_file = dataPaths.spectrographs.get_file_path(
+                    'keck_lris_red_mark4/dict_for_ampmode.json')
```
tbowers7, line 1532: String `/` vs `__truediv__`

pypeit/spectrographs/keck_lris.py
```diff
@@ -1527,19 +1527,15 @@ def get_detector_par(self, det, hdu=None):
...
-            header_file = (
-                data.Paths.data /
-                'spectrographs' /
-                'keck_lris_red_mark4' /
-                f'header{ampmode_translate_dict[amp_mode]}_{swap_binning.replace(",","_")}.fits'
-            )
+            _amp = ampmode_translate_dict[amp_mode]
+            swap_binning = f"{binning[-1]}_{binning[0]}" # LRIS convention is oppopsite ours
+            header_file = dataPaths.spectrographs.get_file_path(
+                    f'keck_lris_red_mark4/header{_amp}_{swap_binning}.fits')
```
tbowers7, line 1539: String `/` vs `__truediv__`

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -505,7 +508,8 @@ def bpm(self, filename, det, shape=None, msbias=None):
                 filt_bpm = bpm_data_pypeit[1:y_len,1:x_len]>100.
                 y_bpm, x_bpm = np.where(filt_bpm)
                 bpm_loc = np.array([y_bpm,x_bpm]).T
-                np.savetxt(bpm_dir / 'BP_MAP_RP_NIR.dat', bpm_loc, fmt=['%d','%d'])
+                # NOTE: This directly access the path, but we shouldn't be doing that...
+                np.savetxt(vlt_sc.path / 'BP_MAP_RP_NIR.dat', bpm_loc, fmt=['%d','%d'])
```
tbowers7: Would it make sense to add a method to `PypeItDataPath` that returns a filename within the package structure (or cache -- see other comments about read-only package installations) for these kinds of saves?

tbowers7 (changes requested):
This is a really nice improvement for the data and caching system that brings it in line with the rest of the PypeIt infrastructure.

There are some comments throughout about POSIX-style string `/` path dividers versus the `pathlib` style `__truediv__` that should be considered/addressed.  Also, symlink creation should be considered in light of possible read-only PypeIt installs (where the cache should be used exclusively).

I have not reviewed or commented on the tests yet, and will await the remainder of the DevSuite work before doing so.

pypeit/cache.py
kbwestfall: Good point.  I generalized the warning statement a bit.

pypeit/cache.py
kbwestfall: Good question.  I added a TODO comment.

pypeit/core/convert_DEIMOSsavfiles.py
kbwestfall: Good question, and we should make sure we warn any windows users to keep an eye out for problems possibly associated with this.  But I expect it should be fine.  The windows CI tests pass, and here's an example (at least one that I can do on my mac):
```
>>> import pathlib
>>> p = pathlib.PureWindowsPath('test/this')
>>> p
PureWindowsPath('test/this')
>>> str(p)
'test\\this'
>>> str(p / 'and' / 'that')
'test\\this\\and\\that'
```

pypeit/core/flux_calib.py
kbwestfall: See prev response.

pypeit/core/flux_calib.py
kbwestfall: Sorry, I don't understand what you mean.  Can you reword/restate the issue?

pypeit/core/flux_calib.py
kbwestfall: Punting on this for now.

pypeit/core/wavecal/templates.py
kbwestfall: Let's discuss at the next pypeit meeting.

pypeit/pypeitdata.py
kbwestfall: Same response.

pypeit/spectrographs/keck_deimos.py
kbwestfall: Same response.

pypeit/spectrographs/keck_lris.py
kbwestfall: Same response.

pypeit/spectrographs/keck_lris.py
kbwestfall: Same response.

pypeit/spectrographs/vlt_xshooter.py
kbwestfall: IMO, the code should never be writing files to `pypeit/data`; it should only be accessing them.  I'd really like to get rid of these lines of code, if agreeable to the team.  In that sense, I don't think we want such a function, even if it's only used by developers.

The only exception to this is the tests ... and we might want to change the code so that output from the tests are always written to some tmp directory.

pypeit/core/wavecal/waveio.py
kbwestfall: Good question.  The tests seem to run okay in the cloud, but I agree this may be a problem for system-level installations of pypeit.  I think I've now limited the use of this to the test functions, meaning this should not affect a general user.

kbwestfall (commented):
Thanks for the initial review @tbowers7 .  I think I've addressed most of your comments, and I'm now re-running the dev-suite.  Since the last time you looked at this, I've finished updating the relevant scripts and I added a new script that allows users to view/remove files from the cache.  I also made a lot of edits to the installation docs to reflect the changes.

Thanks for bringing up the Windows path and symlink issues.  I think I've addressed both, but let me know what you think.

kbwestfall:
Test "pass".  All of the failures are in NIRSPEC, presumably because of recent changes to the raw data available.  I'll re-run after the NIRSPEC branch is merged.
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  257 passed, 56 warnings in 399.69s (0:06:39) ---
--- PYTEST UNIT TESTS FAILED  3 failed, 130 passed, 171 warnings in 937.68s (0:15:37) ---
--- PYTEST VET TESTS PASSED  61 passed, 73 warnings in 5528.28s (1:32:08) ---
Failed tests:
Skipped tests:
Testing Started at 2024-03-07T21:33:16.697132
Testing Completed at 2024-03-08T10:08:43.978332
Total Time: 12:35:27.281200
```

pypeit/core/flux_calib.py
tbowers7: It's the same issue -- just another way of saying "same Windows-compatible question about POSIX-style forward slash".  In your `dataPaths` class, you defined the `__truediv__` method to join path segments just like `pathlib.Path` does.  This was my shorthand for warning about the Windows issue.  But, apparently it's less of an issue than I thought.

doc/scripts/build_cache_data_tbl.py
```diff
@@ -0,0 +1,40 @@
...
+from importlib import resources
+
+import numpy as np
+
+from pypeit.pypeitdata import PypeItDataPaths
+from pypeit.utils import string_table
+
+from IPython import embed
+
+def write_table(ofile):
```
rcooke-ast, line 14: Possibly not worth adding a docstring... but could be beneficial

doc/scripts/make_example_files.py
```diff
@@ -19,78 +21,125 @@
 
 #-----------------------------------------------------------------------------
 
+def obscure_path(paths):
```
rcooke-ast, line 25: Glad that you've implemented this!

pypeit/core/wavecal/waveio.py
```diff
@@ -261,10 +271,10 @@ def load_tree(polygon=4, numsearch=20):
     # TODO: Can we save these as fits files instead?
     # TODO: Please don't use imports within functions
     import pickle
-    filename = data.get_linelist_filepath(f'ThAr_patterns_poly{polygon}_search{numsearch}.kdtree')
-    fileindx = data.get_linelist_filepath(
-        f'ThAr_patterns_poly{polygon}_search{numsearch}.index.npy'
-    )
+    filename = dataPaths.linelist.get_file_path(
```
rcooke-ast, line 291: I have no idea how old this code is, but I would bet that it's not used...

pypeit/io.py
```diff
@@ -36,6 +32,12 @@
 import pypeit
 import time
 
+from linetools.spectra import xspectrum1d
```
rcooke-ast, line 37: I had a feeling there was a generally feeling to remove the xspectrum1d dependency... If that still stands, perhaps add a TODO above this line to this effect?

pypeit/scripts/cache_github_data.py
```diff
@@ -1,99 +1,141 @@
...
+        msgs.info('Searching github repository ... done.')
+        
+        # Determine which files should be in the cache (or in the repo)
+        # TODO: This is currently broken because not all spectrograph-dependent
+        # files are named after their *exact* pypeit spectrograph names.  E.g.,
+        # the reid_arxiv file for gemini_flamingos2 is
+        # ``Flamingos2_HK_HK.fits``, not ``gemini_flamingos2_HK_HK.fits``.  It's
+        # a major effort to rename all those files.  The other (better?) option
+        # would be to put all these reid_arxiv files into subdirectories named
+        # after each spectrograph.
```
rcooke-ast, line 108: I suspect both might be better, but I'm a folder enthusiast... In the example you give, I would do:
`gemini_flamingos2/gemini_flamingos2_HK_HK.fits`
but, I agree that this would be tedious...

pypeit/scripts/install_ql_calibs.py
```diff
@@ -5,8 +5,10 @@
 .. include:: ../include/links.rst
 """
 
+# TODO: Do people actually use this script?  The directory on the Google drive
```
rcooke-ast, line 8: Maybe pose this question on the Slack. I have certainly never used it, and the QL stuff was changed recently, so maybe it's not needed?

pypeit/spectrographs/vlt_xshooter.py
rcooke-ast: Maybe check with @jhennawi. My thought is to just remove them... the lines to generate the BPM are not going to take long, and I agree that we shouldn't be writing this to disk.

rcooke-ast (approved):
This is impressive @kbwestfall - job very well done! I only have a few minor points, but the new approach is a lot better than before. I also have one general question... If a user updates their version of PypeIt, will this require them to re-download all data files when needed (or, only if those data files have been updated)? It just wasn't clear to me how any updated (i.e. changed from one release to the next release) data files are being managed. Thanks again!

I have no major issues with this PR, so I'm approving.

doc/installing.rst
```diff
@@ -203,109 +226,192 @@ please `Submit an issue`_.
...
+
+.. code-block:: console
+
+    % pypeit_clean_cache -l
+           HOST               BRANCH               SUBDIR FILE
+         github                 dirs                tests gemini_gnirs_32_1_spat_fit.npz
+         github                 dirs            sensfuncs keck_deimos_600ZD_sensfunc.fits
+       s3_cloud                  ...   telluric/atm_grids TellPCA_3000_26000_R10000.fits
+         github                 dirs                tests solution_arrays.npz
+
```
tbowers7: This thought feels unfinished.  Also, in the final version, perhaps use an example with a versioned (_e.g._, `1.15.1`) branch or tag.

doc/installing.rst
```diff
@@ -203,109 +226,192 @@ please `Submit an issue`_.
...
+spectrograph.  (Unlike previous versions) This script does *not* download any
+files hosted in ``s3_cloud`` (see the table :ref:`above<data_dir>`); instead,
+use the scripts below.
 
+.. note::
+
+    Beware of rate limits imposed by GitHub.  If you run into this, try setting
+    up an access token and export it as the ``GITHUB_TOKEN`` environmental
+    variable; see
+    `here <https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api>`__.
```
tbowers7, line 334: I didn't know this was a thing.  Thanks for knowing about it and adding it to the docs!

doc/scripts/build_cache_data_tbl.py
tbowers7: I second the desire for a docstring.

doc/scripts/make_example_files.py
tbowers7: Nice!

doc/scripts/make_example_files.py
```diff
@@ -200,14 +259,15 @@ def make_meta_examples():
 
 if __name__ == '__main__':
     t = time.perf_counter()
+    tag, date = git_most_recent_tag()
```
tbowers7, line 224: Very nice!

pypeit/cache.py
```diff
@@ -223,40 +310,95 @@ def write_file_to_cache(filename: str, cachename: str, filetype: str, remote_hos
...
+
+
+def parse_cache_url(url):
+    """
+    Parse a URL from the cache into its relevant components.
+
+    Args:
+        url (:obj:`str`):
+            URL of a file in the pypeit cache. A valid cache URL must include
+            either ``'github'`` or ``'s3.cloud'`` in its address.
```
tbowers7, line 370: This is great!

pypeit/pypeitdata.py
```diff
@@ -302,67 +328,95 @@ class PypeItDataPaths:
...
-        self.skisim = PypeItDataPath('skisim')
-        self.filters = PypeItDataPath('filters')
-        self.sensfunc = PypeItDataPath('sensfuncs')
-
-        # Other
-        self.sky_spec = PypeItDataPath('sky_spec')
-        self.static_calibs = PypeItDataPath('static_calibs')
-        self.spectrographs = PypeItDataPath('spectrographs')
+        for key, a in PypeItDataPaths.defined_paths.items():
+            setattr(self, key, PypeItDataPath(a['path'], remote_host=a['host']))
```
tbowers7, line 373: Clever.

tbowers7 (approved):
Looks great.  Thanks for addressing the Windows path questions.

pypeit/pypeitdata.py
```diff
@@ -0,0 +1,422 @@
...
+import shutil
+
+from IPython import embed
+
+from pypeit import msgs
+from pypeit import cache
+
+# NOTE: A better approach may be to subclass from Path.  I briefly tried that,
+# but quickly realized it was going to be more complicated than I'd hoped.  This
+# is a clean and sufficient solution for now.
```
tbowers7: FYI -- It looks like Python 3.12 has introduced "The [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) class now supports subclassing".  You may want to add a TODO here that when PypeIt supports `python>=3.12` you can try the subclassing again.

doc/scripts/build_cache_data_tbl.py
kbwestfall: Done!

pypeit/core/wavecal/templates.py
kbwestfall: Leaving it for now...

pypeit/core/wavecal/waveio.py
kbwestfall: In infinite time (ha!), we would go through the code and move functions that aren't used into `deprecated`.

pypeit/io.py
kbwestfall: Added!

pypeit/scripts/cache_github_data.py
kbwestfall: Leaving as a TODO for now.

pypeit/scripts/install_ql_calibs.py
kbwestfall: Added  #1786

doc/installing.rst
kbwestfall: Thanks for catching this!  I completed the thought.

pypeit/pypeitdata.py
kbwestfall: Added!

pypeit/spectrographs/vlt_xshooter.py
kbwestfall: I went ahead and commented this out.  We can remove completely in a future PR.

kbwestfall:
There is a test failure, but it's because of a recent change to the files in the relevant Keck/HIRES dataset.  The vet-test failure is related.  So I think this one is ready to go.

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  256 passed, 56 warnings in 431.72s (0:07:11) ---
--- PYTEST UNIT TESTS PASSED  148 passed, 167 warnings in 1038.48s (0:17:18) ---
--- PYTEST VET TESTS FAILED  1 failed, 60 passed, 96 warnings in 5677.00s (1:34:37) ---
Failed tests:
    keck_hires/J1723+2243_W241_RED_C5_ECH_-0.15_XD_0.90_2x2 pypeit
Skipped tests:
Testing Started at 2024-06-12T01:28:58.711750
Testing Completed at 2024-06-12T14:16:18.163584
Total Time: 12:47:19.451834
```

Issue #1701, opened 2023-10-08
Make a script that will calculate the sensitivity function for DataCubes.


rcooke-ast:
At the moment the sensitivity function is derived "on the fly" while datacubes are being made. It makes a lot more sense to do this in advance.

Also be aware of the following corner case that was mention in PR #1669:

`The question about making the sensitivity function in advance is fair, and we can perhaps generate a new script to deal with extracting standard star spectra from datacubes and making the sensitivity function in advance. In fact, given how long the IR algorithm takes, it's probably beneficial to do this. The only "corner case" that I'm aware of is when we need to make a small correction to the sensitivity function when the standard star and science frame uses a slightly different grating angle. This can introduce a pseudo-linear (as a function of wavelength) sensitivity variation in the flux (due to the different grating angles). For example, imagine that you have a standard star taken on only one night with a certain grating angle. You take science observations with this exact same grating angle. These do not need a grating correction. But, if you take another observation that tried to get close to the same grating angle (but it's off slightly) then you'll either need to make a sensitivity function using a different standard star or you'll need to perform a grating correction.`

The obvious way around this is to have a new column in the `spec2d` section of the `.coadd3d` file that allows the user to apply any pre-prepared sensitivity function to any again `spec2d` file

rcooke-ast:
Addressed by #1770

PR #1770, opened 2024-02-15
Nonlinearity correction + datacube spec1d extractions + datacube fluxing improvements

rcooke-ast:
This PR contains multiple updates (mostly in the context of KCWI/KCRM reductions).

**Non-linearity correction**
Based on commissioning data, I have calculated the non-linearity of KCWI (KCRM data are not currently available). I've also prepared a short document to describe the correction. See the following [non-linearity report (pdf)](https://github.com/pypeit/PypeIt/files/14295468/Linearity_Report.pdf) for further details.

**Datacube spec1d extractions**
Introduced a new script that performs point source extractions from datacubes: `pypeit_extract_datacube`. This script produces `spec1d` files including boxcar + optimal extractions that can then be used for flux calibration or science analysis (if observing point sources).

**Flux calibration**
The flux calibration of datacubes has been moved to outside of the datacube generation. This allows the datacube generation to make use of the `pypeit_sensfunc` script. Note, there is a "bug" with the current approach to generating sensitivity functions. PypeIt only uses only one spec1d file (of a standard star) to generate the sensitivity function. The datacube module allows you to combine multiple standard star spec2d files into a single datacube. However, the extinction correction is performed on the extracted datacube - this means that the extinction correction cannot be performed on individual standard star spec2d files.

**extended a test for datacube and fluxing (+ docs)**
I've redeveloped a test that tests the entire datacube creation process + added new docs. There is an associated
[dev-suite PR as well](https://github.com/pypeit/PypeIt-development-suite/pull/311).

Part of this PR addresses Issue #1701.

Tests pass:
image


doc/coadd3d.rst
```diff
@@ -96,10 +118,50 @@ Then run the script:
...
+    The datacube is created by running the following command:
+
+    .. code-block:: console
+
+        pypeit_coadd_datacube StandardStarName.coadd3d -o
+
+#. Step 2 - Extract the 1D spectra from the datacube. This is done by running the following command,
+    assuming that the output datacube from the previous step was called ``StandardStarName.fits``.
+    The ``pypeit_extract_datacube`` script will produce an output file called
+    ``StandardStarName_spec1d.fits``:
```
kbwestfall: Is this difference from the naming convention of the other spec1d files in pypeit meaningful?  I.e., why not `spec1d_StandardStarName.fits`?  Same question for the sensitivity function file.

doc/coadd3d.rst
```diff
@@ -114,41 +176,50 @@ into many subpixels, and assigns each subpixel to a voxel of the datacube. Flux
 but voxels are correlated, and the error spectrum does not account for covariance between
 adjacent voxels. The subpixellation scale can be separately set in the spatial and spectral
 direction on the 2D detector. If you would like to change the subpixellation factors from
-the default values (5), you can set the ``spec_subpixel`` and  ``spat_subpixel`` keywords
-as follows:
+the default values (5), you can optionally set (one or all of) the ``spec_subpixel``,
+``spat_subpixel``, and ``slice_subpixel`` parameters as follows:
```
kbwestfall, line 188: Can you define `slice_subpixel`?  It's unclear to me what this is.

doc/coadd3d.rst
```diff
@@ -196,16 +268,27 @@ but the grating angle of the two nights were slightly different).
...
+    Science/spec2d_scienceframe_01.fits  |  Calibrations/Flat_A_0_DET01.fits
+    Science/spec2d_scienceframe_02.fits  |  Calibrations/Flat_B_1_DET01.fits
+    spec2d end
 
+If all spec2d files were reduced with the same Flat calibration file,
+then you do not need to specify the grating correction file. Also, if you
+generate a sensitivity function for each spec2d file, then you do not need
+to specify the grating correction file. The grating correction file is only
+needed if you have one sensitivity function for all spec2d files, even though
+the spec2d files were acquired with different grating angles.
```
kbwestfall, line 340: Presumably these are for small changes in grating angle, so maybe it doesn't matter, but how does the code treat the difference in wavelength range between the two (or more) setups?  Is the sensitivity function extrapolated or are the data truncated to the common wavelength range (or something else)?

doc/releases/1.15.1dev.rst
```diff
@@ -19,6 +19,13 @@ Instrument-specific Updates
 Script Changes
 --------------
 
+- A new script allows spec1d files to be extracted from datacubes. This
+  script is called `pypeit_extract_datacube`, and can be used to extract
+  1D spectra from a datacube.
```
kbwestfall: Note is a bit repetitious?

pypeit/coadd3d.py
```diff
@@ -226,9 +233,41 @@ def wcs(self):
             not be accessed directly, and you should only call self.wcs
         """
         if self._wcs is None:
-            self._wcs = wcs.WCS(self.head0)
+            if self.headwcs is None:
+                msgs.error('No WCS information stored in the DataCube')
+            else:
+                self._wcs = wcs.WCS(self.headwcs)
```
kbwestfall: Sorry if I've asked this before, but do you need to lazy-load the wcs?  Can we instead, e.g., set `self.wcs = wcs.WCS(hdu[1].header)` in the `from_file` function?  As far as I can tell `headwcs` is only used here.

pypeit/coadd3d.py
```diff
@@ -226,9 +233,41 @@ def wcs(self):
...
         return self._wcs
 
+    def extract_spec(self, parset, outname=None, overwrite=False):
+        """
+        Extract a spectrum from the datacube
+
+        Parameters
+        ----------
+        parset : dict
+            A dictionary containing the 'Reduce' PypeItPar parameters.
```
kbwestfall: This is an instance of `ReducePar`, right?  They act similarly, but I would be specific with the type.

pypeit/coadd3d.py
```diff
@@ -226,9 +233,41 @@ def wcs(self):
...
+        outname : str, optional
+            Name of the output file
+        overwrite : bool, optional
+            Overwrite any existing files
+        """
+        # Extract the spectrum
+        # TODO :: should we set the boxcar_width argument to None so the automated algorithm is used by default? At the moment, the user needs to set the boxcar_width to None in the parset to use the automated algorithm
```
kbwestfall: Is this possible?  An annoyance with `ParSet`s are that setting a value to `None` can invoke the default when the `ParSet` is instantiated.  Is the automated approach robust?  If so, I agree that changing the default value for `boxcar_width` to `None` makes sense.

pypeit/coadd3d.py
```diff
@@ -226,9 +233,41 @@ def wcs(self):
...
+        fwhm = parset['findobj']['find_fwhm'] if parset['extraction']['use_user_fwhm'] else None
+
+        # Datacube's are counts/second, so set the exposure time to 1
+        exptime = 1.0
+        # TODO :: Avoid transposing these large cubes
+        sobjs = datacube.extract_standard_spec(self.wave, self.flux.T, self.ivar.T, self.bpm.T, self.wcs,
```
kbwestfall: Is there anything in this function that's specific to treatment of a standard star?  I.e., why not `extract_point_source`?

pypeit/coadd3d.py
```diff
@@ -378,9 +417,15 @@ def __init__(self, spec2dfiles, par, skysub_frame=None, scale_corr=None, ra_offs
...
                 If not None, this should be a list of frames to use for the sky subtraction of each individual
                 entry of spec2dfiles. It should be the same length as spec2dfiles.
+            sensfile (:obj:`list`, optional):
+                If not None, this should be a list of frames to use for the sensitivity function of each individual
+                entry of spec2dfiles. It should be the same length as spec2dfiles.
             scale_corr (:obj:`list`, optional):
                 If not None, this should be a list of relative scale correction options. It should be the
                 same length as spec2dfiles.
+            grating_corr (:obj:`list`, optional):
+                If not None, this should be a list of relative grating tilt correction options. It should be the
```
kbwestfall: Are these always `Flat` calibration files?  If so, does it make sense to be specific about this in the description?

pypeit/coadd3d.py
```diff
@@ -1101,47 +1144,67 @@ def load(self):
...
+            # TODO :: Need to make a note somewhere that the extinction correction cannot currently be done
+            #         in the datacube because the sensitivity function algorithms correct the standard star
+            #         for extinction when generating the sensitivity function. Including the extinction correction
+            #         in the datacube would result in a double correction of the standard star for extinction.
+            #         This could be wrong when combining multiple standard star exposures if the airmass of the
+            #         standard star exposures is significantly different. For now, to stay consistent with the
+            #         current pipeline, the extinction correction is done in the sensitivity function algorithms,
+            #         with the caveat that the standard star exposures are assumed to have similar airmasses.
+            #         For now, comment out the extinction correction, and reapply this later when the sensitivity
+            #         function algorithms are unified.
```
kbwestfall, line 1143: It feels like we should make it a priority to address this.  But, for this PR, should we elevate the comments you make in the PR description to the docs?  E.g., can you comment on this in coadd3d.rst (or another place that's relevant)?

pypeit/core/datacube.py
```diff
@@ -98,15 +98,19 @@ def fitGaussian2D(image, norm=False):
     y = np.linspace(0, image.shape[1] - 1, image.shape[1])
     xx, yy = np.meshgrid(x, y, indexing='ij')
     # Setup the fitting params
-    idx_max = [image.shape[0]/2, image.shape[1]/2]  # Just use the centre of the image as the best guess
-    #idx_max = np.unravel_index(np.argmax(image), image.shape)
+    med_filt_image = signal.medfilt2d(image, kernel_size=3)
+    # idx_max = [image.shape[0]/2, image.shape[1]/2]  # Just use the centre of the image as the best guess
```
kbwestfall: remove?  Or add some comments explaining the choice?

pypeit/core/datacube.py
```diff
@@ -187,76 +191,119 @@ def correct_grating_shift(wave_eval, wave_curr, spl_curr, wave_ref, spl_ref, ord
...
+    bpmcube : `numpy.ndarray`_
+        Datacube of the bad pixel mask
+    wcscube : `astropy.wcs.WCS`_
+        WCS of the datacube
+    exptime : float
+        Exposure time listed in the header of the datacube
+    subpixel : int, optional
         Number of pixels to subpixelate spectrum when creating mask
+    boxcar_width : float, optional
+        Width of the boxcar (in arcseconds) to use for the extraction
```
kbwestfall: What does the boxcar look like in 2D?  Is it a circular or square aperture?

pypeit/core/datacube.py
```diff
@@ -187,76 +191,119 @@ def correct_grating_shift(wave_eval, wave_curr, spl_curr, wave_ref, spl_ref, ord
...
     smask -= mask
 
-    # Subtract the residual sky
+    msgs.info("Subtracting the residual sky")
+    # Subtract the residual sky from the datacube
     skymask = np.logical_not(bpmcube) * smask
-    skycube = flxcube * skymask
+    skycube = _flxcube * skymask
     skyspec = skycube.sum(0).sum(0)
     nrmsky = skymask.sum(0).sum(0)
```
kbwestfall: Does `skycube.sum(axis=(0,1))` produce the same result?  IMO, it's important to include the name for keyword arguments because the order of the keywords can potentially change.  I.e., I would instead write `skycube.sum(axis=0).sum(axis=0)`.

pypeit/core/datacube.py
```diff
@@ -187,76 +191,119 @@ def correct_grating_shift(wave_eval, wave_curr, spl_curr, wave_ref, spl_ref, ord
...
+    arcsecSQ = (wcscube.wcs.cdelt[0] * wcscube.wcs.cunit[0].to(units.arcsec)) * \
+               (wcscube.wcs.cdelt[1] * wcscube.wcs.cunit[1].to(units.arcsec))
+    if fluxed:
+        # The datacube is flux calibrated, in units of 10^-17 erg/s/cm**2/Ang/arcsec**2
+        # Scale the flux and ivar cubes to be in units of erg/s/cm**2/Ang
+        unitscale = arcsecSQ
+    else:
+        # Scale the flux and ivar cubes to be in units of counts
+        deltawave = wcscube.wcs.cdelt[2]*wcscube.wcs.cunit[2].to(units.Angstrom)
+        unitscale = exptime * deltawave * arcsecSQ
```
kbwestfall, line 250: It would be useful to have some more detail here.  Am I right that the input datacube when not flux calibrated is in units of counts / s / Ang?  Why do we need to multiply by exposure time and wavelength pixel scale for uncalibrated cubes, but not for calibrated cubes?

pypeit/core/datacube.py
```diff
@@ -187,76 +191,119 @@ def correct_grating_shift(wave_eval, wave_curr, spl_curr, wave_ref, spl_ref, ord
...
     xx, yy = np.meshgrid(x, y, indexing='ij')
 
     # Generate a mask
-    newshape = (flxcube.shape[0] * subpixel, flxcube.shape[1] * subpixel)
+    msgs.info("Generating an object mask")
+    newshape = (numxx * subpixel, numyy * subpixel)
     mask = np.zeros(newshape)
-    nsig = 4  # 4 sigma should be far enough... Note: percentage enclosed for 2D Gaussian = 1-np.exp(-0.5 * nsig**2)
-    ww = np.where((np.sqrt((xx - popt[1]) ** 2 + (yy - popt[2]) ** 2) < nsig * wid))
+    ww = np.where((np.sqrt((xx - popt[1]) ** 2 + (yy - popt[2]) ** 2) < wid))
```
kbwestfall, line 282: I recommend we be a bit careful with the nomenclature.  If I'm reading this right, you're extracting a circular aperture around the star.  I think that's the right approach, but it's different from a boxcar extraction of a slit spectrum.  As a user, I would want to know the detailed shape of the extraction aperture, so we should be clear about this in the docs.  But maybe by "boxcar" we're really referencing that the extraction is unweighted, not the detailed shape of the extraction aperture?

pypeit/core/datacube.py
```diff
@@ -268,87 +315,101 @@ def extract_standard_spec(stdcube, subpixel=20):
...
+    sobj.BOX_WAVE = wave.astype(float)
+    if fluxed:
+        sobj.BOX_FLAM = box_flux
+        sobj.BOX_FLAM_SIG = np.sqrt(box_var)
+        sobj.BOX_FLAM_IVAR = utils.inverse(box_var)
+    else:
+        sobj.BOX_COUNTS = box_flux
+        sobj.BOX_COUNTS_SIG = np.sqrt(box_var)
+        sobj.BOX_COUNTS_IVAR = utils.inverse(box_var)
+        sobj.BOX_COUNTS_SKY = skyspec  # This is not the real sky, it is the residual sky. The datacube is presumed to be sky subtracted
```
kbwestfall, line 334: Is this articulated in the docs somewhere?

pypeit/core/datacube.py
```diff
@@ -268,87 +315,101 @@ def extract_standard_spec(stdcube, subpixel=20):
...
+    ivarcube2d = _ivarcube[flxslice].T
+    gpmcube2d = np.logical_not(bpmcube[flxslice].T)
+    waveimg = wave.reshape((numwave,1)).repeat(numspat, axis=1)
+    skyimg = np.zeros((numwave, numspat))  # Note, the residual sky has already been subtracted off _flxcube
+    oprof = objprof.reshape((1,numspat)).repeat(numwave, axis=0)
+    thismask = np.ones_like(flxcube2d, dtype=bool)
+
+    # Now do the optimal extraction
+    extract.extract_optimal(flxcube2d, ivarcube2d, gpmcube2d, waveimg, skyimg, thismask, oprof,
+                            sobj, min_frac_use=0.05, fwhmimg=None, base_var=None, count_scale=None, noise_floor=None)
```
kbwestfall, line 392: It's difficult for me to follow this without some direct interaction with the code.  But stepping back a bit I'm wondering, is the 3D->2D conversion necessary or a convenience?

pypeit/core/flux_calib.py
```diff
@@ -1357,7 +1369,7 @@ def zeropoint_to_throughput(wave, zeropoint, eff_aperture):
     """
 
     eff_aperture_m2 = eff_aperture*units.m**2
-    S_lam_units = 1e-17*units.erg/units.cm**2
+    S_lam_units = PYPEIT_FLUX_SCALE*units.erg/units.cm**2
```
kbwestfall, line 1404: Good catch!

pypeit/core/procimg.py
```diff
@@ -1372,3 +1372,52 @@ def variance_model(base, counts=None, count_scale=None, noise_floor=None):
...
+    -------
+    corr_counts :
+        Array with the corrected counts.
+    """
+    msgs.info('Applying a non-linearity correction to the counts.')
+    # Check the input
+    if counts.shape != ampimage.shape:
+        msgs.error('Counts and amplifier image have different shapes.')
+    if isinstance(nonlinearity_coeffs, list):
+        nonlinearity_coeffs = np.array(nonlinearity_coeffs)
```
kbwestfall: This potentially alters `nonlinearity_coeffs`.  Instead I would replace the above two lines with:
```
_nonlinearity_coeffs = np.asarray(nonlinearity_coeffs)
```
I.e., if you do this, you don't have to check if it's a list, and it won't result in a copy if it's an array.

pypeit/core/procimg.py
```diff
@@ -1372,3 +1372,52 @@ def variance_model(base, counts=None, count_scale=None, noise_floor=None):
...
+    # Setup the output array
+    corr_counts = counts.copy()
+    unqamp = np.unique(ampimage)
+    for uu in range(unqamp.size):
+        thisamp = unqamp[uu]
+        indx = (ampimage == thisamp)
+        corr_counts[indx] = counts[indx] * (1. + nonlinearity_coeffs[thisamp]*counts[indx])
+    # Apply the correction
+    return corr_counts
```
kbwestfall, line 1440: How does this correction propagate to the Poisson counting statistics?  Specifically, is it correct to apply Poisson statistics to the *corrected* counts?  Or do you need to apply Poisson stats to the uncorrected counts and the propagate the error to the corrected counts?  The corrections are very small, so maybe it doesn't matter.

pypeit/data/standards/blackbody/blackbody_info.txt
```diff
@@ -7,20 +7,20 @@
...
+J1045p0157.fits  J1045p0157  10:45:23.866  +01:57:21.96  19.09  DC     8956  0.666
+J1117p4059.fits  J1117p4059  11:17:20.801  +40:59:54.67  18.08  DC    10992  0.844
+J1147p1713.fits  J1147p1713  11:47:22.608  +17:13:25.21  18.65  DC     9962  0.669
+J1245p4238.fits  J1245p4238  12:45:35.626  +42:38:24.58  17.14  DC     9912  2.817
+J1255p1924.fits  J1255p1924  12:55:07.082  +19:24:59.00  18.53  DC     8882  1.157
+J1343p2706.fits  J1343p2706  13:43:05.302  +27:06:23.98  18.93  DC    10678  0.427
+J1417p4941.fits  J1417p4941  14:17:24.329  +49:41:27.85  17.25  DC    10461  2.192
+J1518p0028.fits  J1518p0028  15:18:59.717  +00:28:39.58  19.44  DC     9072  0.458
+J1617p1813.fits  J1617p1813  16:17:04.078  +18:13:11.96  18.78  DC     8568  0.996
+J2302m0030.fits  J2302m0030  23:02:40.032  -00:30:21.60  17.80  DC    10478  1.241
```
kbwestfall, line 26: What does this DB -> DC change mean?

pypeit/images/rawimage.py
```diff
@@ -653,6 +680,8 @@ def process(self, par, bpm=None, scattlight=None, flatimages=None, bias=None, sl
         flat_bpm = self.flatfield(flatimages, slits=slits, debug=debug) if self.use_flat else None
 
         # Calculate the inverse variance
+        # TODO :: I think the IVAR should technically be calculated before the flat-fielding.
+        #         This is because the flat-fielding is done in counts, and the IVAR is in counts^2.
```
kbwestfall: I'm not sure I follow.  The flat-fielding is a scale factor that is applied to the counts in the image.  `build_ivar` incorporates this scale factor in the IVAR calculation.  I.e., the IVAR is in ("corrected counts")^2, not counts^2.  Happy to discuss more though.

pypeit/scripts/coadd_datacube.py
```diff
@@ -32,6 +25,15 @@ def get_parser(cls, width=None):
...
     @staticmethod
     def main(args):
+        import time
+
+        from pypeit import msgs
+        from pypeit import par
+        from pypeit import inputfiles
+        from pypeit import utils
+        from pypeit.coadd3d import CoAdd3D
+        from pypeit.spectrographs.util import load_spectrograph
```
kbwestfall, line 35: Good catch.

pypeit/specobj.py
```diff
@@ -486,6 +486,7 @@ def update_flex_shift(self, shift, flex_type='local'):
         self.FLEX_SHIFT_TOTAL += shift
 
     # TODO This should be a wrapper calling a core algorithm.
+    #      It is, right? Can this TODO be removed?
```
kbwestfall: Yeah, I agree.  This effectively wraps `flux_calib.get_sensfunc_factor`, so I think you can remove the TODO.

pypeit/specobj.py
```diff
@@ -502,8 +503,8 @@ def apply_flux_calib(self, wave_zp, zeropoint, exptime, tellmodel=None, extinct_
             exptime (float):
                 Exposure time
             tellmodel (?):
-                Telluric correction
-            extinct_correct (?):
+                Telluric correction. Note: This is deprecated and will be removed in a future version.
```
kbwestfall, line 504: I followed the code a bit but not far enough to understand how this is deprecated.  If providing tellmodel here (and ultimately how it's used in `get_sensfunc_factor`) produces an incorrect result, we should deprecate it now.

pypeit/specobjs.py
```diff
@@ -238,14 +238,15 @@ def unpack_object(self, ret_flam=False, extract_type='OPT'):
         ech_orders = np.zeros(norddet, dtype=int)
 
         # TODO make the extraction that is desired OPT vs BOX an optional input variable.
+        #   This TODO is already so, right?
```
kbwestfall: Agreed.  Another good catch!

pypeit/specobjs.py
```diff
@@ -720,16 +721,23 @@ def write_to_fits(self, subheader, outfile, overwrite=True, update_det=None,
...
               If provided, do not clobber the existing file but only update
               the indicated slits.  Useful for re-running on a subset of slits
-            update_det (int or list, optional):
-              If provided, do not clobber the existing file but only update
-              the indicated detectors.  Useful for re-running on a subset of detectors
-
+            history (:obj:`str`, optional):
+                String to be added to the header HISTORY keyword.
+            debug (:obj:`bool`, optional):
+                If True, run in debug mode.
```
kbwestfall, line 737: Thanks for improving the docs!

kbwestfall (approved):
Lots of nit-picks from me, and a few pointed questions about nomenclature and procedure.  But I'm approving now, and leaving the final word to @debora-pe.

Please make sure to update the docs.

kbwestfall:
Also, I feel like I'm missing some context in your report on the non-linearity correction.  Can you provide more detail?  E.g., how are you constructing the alpha histograms in the 3rd figure?  Sorry if I'm being dense.

doc/coadd3d.rst
rcooke-ast: Good point - I've switched these around.

doc/coadd3d.rst
rcooke-ast: I've added some additional text to hopefully make this a little clearer (without resorting to the parameter definitions)

doc/coadd3d.rst
rcooke-ast: I've clarified this in the docs. The ideal situation is that the user doesn't need to perform this correction now that separate sensitivity functions can be applied to individual spec2d files. But, yes, it is extrapolated.

doc/releases/1.15.1dev.rst
rcooke-ast: Indeed... I've cleaned this up

pypeit/core/procimg.py
rcooke-ast: Yeah, good catch  👍  I've implemented this

pypeit/data/standards/blackbody/blackbody_info.txt
rcooke-ast: It has no affect at all. I originally just copied across the DB label, but these are all a subclass of the DC white dwarfs, not DB white dwarfs.

pypeit/specobj.py
rcooke-ast: I just added this to the docstring, because it's in the docs of `flux_calib.get_sensfunc_factor()` which is the only call here in this function, so I'm just propagating it to these docs. I'll add this as an Issue for now, as I'm not sure who deprecated it, and if this has been carried forward through the code.

pypeit/specobjs.py
rcooke-ast: Thank GitHub Copilot  😉

pypeit/images/rawimage.py
rcooke-ast: Yes, you are quite right -- I hadn't spotted that. I've removed the TODO.

pypeit/coadd3d.py
rcooke-ast: Yes, good point - added!

pypeit/coadd3d.py
rcooke-ast: I think this is just done for consistency with the `ivar` function above. It's not an intensive operation, so I've moved this to the `from_file` as you suggest.

pypeit/coadd3d.py
rcooke-ast: Good call - updated.

pypeit/coadd3d.py
rcooke-ast: Quite right - this is not possible. The automated algorithm works by fitting a 2D Gaussian. I can't see why this would fail. Nevertheless, I've added this as a parameter for the script

pypeit/coadd3d.py
rcooke-ast: Yes, they should only be Flat calibration files - updated.

pypeit/coadd3d.py
rcooke-ast: I'll add this as an issue, for now.

pypeit/core/datacube.py
rcooke-ast: Added a comment

pypeit/core/datacube.py
rcooke-ast: Ahh, good point - it's a circle. I've added this to the description, and changed "width" to "radius"

pypeit/core/datacube.py
rcooke-ast: When cubes have not been flux-calibrated, they are in units counts/s/Ang/arcsec^2. `pypeit_sensfunc` expects the input to be counts, which is why the unitscale is different here. Also, the spec1d files are in units `counts`

When the datacubes are fluxed, we want to extract a spectrum that has units of flux density.

pypeit/core/datacube.py
rcooke-ast: Yes, that's correct. I've changed everything from `boxcar_width` -> `boxcar_radius`, and noted that it's a circular aperture. I've added some notes to the docs, as well.

pypeit/core/datacube.py
rcooke-ast: Good point... I never knew that this was possible with some numpy functions! I've made the suggested change.

pypeit/core/datacube.py
rcooke-ast: It is now  :-)  thanks for the suggestion.

pypeit/core/datacube.py
rcooke-ast: It's necessary in the sense that the `extract.extract_optimal()` currently works on 2D frames only. However, it need not be, but this would require a more significant refactor.

rcooke-ast (commented):
Thanks @kbwestfall - I've addressed all comments, except for how to deal with errors associated with the non-linearity correction. It is indeed small, but I will have a think about the most appropriate approach (i.e. to include this in the `count_scale`... or not).

rcooke-ast:
> Also, I feel like I'm missing some context in your report on the non-linearity correction. Can you provide more detail? E.g., how are you constructing the alpha histograms in the 3rd figure? Sorry if I'm being dense.

Sorry - I just saw this additional comment. I honestly thought no-one would read the report, and it was more for my record!

I basically calculate the alpha value of each detector pixel. Then, create a distribution of alpha values for each amplifier. For most science applications, the correction will be at most 0.1%, so it's really not that important for KCWI.

pypeit/core/procimg.py
rcooke-ast: I've been thinking about this, and I think it depends on what causes the non-linearity. I can convince myself both ways... If we think of this as an "efficiency" correction, it should be treated just like we treat the flatfielding. However, this method is correcting to the "linear" count rate, so it should be the error on the true count rate. My current feeling is that the most appropriate approach is to treat this in the same way that we treat the flatfield correction. I doubt that this will ever really affect any result... but probably good to get it "right". Any thoughts?

doc/coadd3d.rst
kbwestfall: I'm confused about what you mean by the "slice direction" and, specifically, the difference between `spat_subpixel` and `slice_subpixel`.  E.g., is
```
spat_subpixel = 5
slice_subpixel = 5
```
equivalent to
```
spat_subpixel = 25
slice_subpixel = 1
```
?

pypeit/coadd3d.py
kbwestfall: Are there specific things that a user needs to be aware of, or is effectively everything handled correctly behind the scenes?  I.e., adding an issue is definitely useful, but I'm also worried about users not understanding some aspect of the output.

kbwestfall:
> Sorry - I just saw this additional comment. I honestly thought no-one would read the report, and it was more for my record!
> 
> I basically calculate the alpha value of each detector pixel. Then, create a distribution of alpha values for each amplifier. For most science applications, the correction will be at most 0.1%, so it's really not that important for KCWI.

The code always applies this correction, right?  If so, I feel like a description of the correction should be included in the user-visible docs at a level that is easily found, like in the KCWI pages and/or the basic image processing pages.  And it would be great to be able to reference or include the detailed justification for the correction, maybe as a developer doc building from your internal report.  Following up on that, how do you calculate alpha for each pixel?  I.e., how do you know the true counts?  Sorry to push on this.  Some of it is out of just simple interest, but I'm also thinking of someone using pypeit with KCWI and wanting to understand where this correction came from and why it's being applied.

doc/coadd3d.rst
rcooke-ast: Ahh, I see... the slice direction is perpendicular to the spat direction. I've added some additional text for context. `spat_subpixel` and `slice_subpixel` are needed to subdivide the two spatial directions of the field-of-view.

pypeit/coadd3d.py
rcooke-ast: Ahh... no, sorry, there is nothing here that the user needs to know. It's primarily a developer note only. However, it might be useful for users that are trying to combine lots of standard star observations at different airmass, so I'll add a temporary note to the documentation.

rcooke-ast:
> The code always applies this correction, right? If so, I feel like a description of the correction should be included in the user-visible docs at a level that is easily found, like in the KCWI pages and/or the basic image processing pages. And it would be great to be able to reference or include the detailed justification for the correction, maybe as a developer doc building from your internal report. Following up on that, how do you calculate alpha for each pixel? I.e., how do you know the true counts? Sorry to push on this. Some of it is out of just simple interest, but I'm also thinking of someone using pypeit with KCWI and wanting to understand where this correction came from and why it's being applied.

Yes, that is correct, and you are right - there should be some docs to describe this in a little more detail. I've added a new page in the documentation that describes this correction (including some math equations that I hope render correctly!).

pypeit/coadd3d.py
```diff
@@ -1101,47 +1125,67 @@ def load(self):
...
             # Correct for sensitivity as a function of grating angle
             # (this assumes the spectrum of the flatfield lamp has the same shape for all setups)
             gratcorr_sort = 1.0
-            if self.cubepar['grating_corr']:
-                # Load the flatfield file
-                key = flatfield.FlatImages.calib_type.upper()
-                if key not in spec2DObj.calibs:
-                    msgs.error('Processed flat calibration file not recorded by spec2d file!')
-                flatfile = os.path.join(spec2DObj.calibs['DIR'], spec2DObj.calibs[key])
+            if self.grating_corr[ff] is not None:
```
debora-pe, line 1158: Not sure if I missed something, but it seems this will crash if `self.grating_corr` is `None` (i.e., `self.grating_corr[ff]` does not exist). Right?

pypeit/core/datacube.py
```diff
@@ -187,76 +190,119 @@ def correct_grating_shift(wave_eval, wave_curr, spl_curr, wave_ref, spl_ref, ord
...
+    wl_img = make_whitelight_fromcube(_flxcube, wavemin=whitelight_range[0], wavemax=whitelight_range[1])
+    popt, pcov, model = fitGaussian2D(wl_img, norm=True)
+    if boxcar_radius is None:
+        nsig = 4  # 4 sigma should be far enough... Note: percentage enclosed for 2D Gaussian = 1-np.exp(-0.5 * nsig**2)
+        wid = nsig * max(popt[3], popt[4])
+    else:
+        # Set the user-defined radius
+        wid = boxcar_radius / np.sqrt(arcsecSQ)
+    # Set the radius of the extraction boxcar for the sky determination
+    msgs.info("Using a boxcar width of {:0.2f} arcsec".format(wid*np.sqrt(arcsecSQ)))
```
debora-pe: This is boxcar radius, right?

pypeit/core/datacube.py
```diff
@@ -267,88 +313,102 @@ def extract_standard_spec(stdcube, subpixel=20):
...
+    ivarcube2d = _ivarcube[flxslice].T
+    gpmcube2d = np.logical_not(bpmcube[flxslice].T)
+    waveimg = wave.reshape((numwave,1)).repeat(numspat, axis=1)
+    skyimg = np.zeros((numwave, numspat))  # Note, the residual sky has already been subtracted off _flxcube
+    oprof = objprof.reshape((1,numspat)).repeat(numwave, axis=0)
+    thismask = np.ones_like(flxcube2d, dtype=bool)
+
+    # Now do the optimal extraction
+    extract.extract_optimal(flxcube2d, ivarcube2d, gpmcube2d, waveimg, skyimg, thismask, oprof,
+                            sobj, min_frac_use=0.05, fwhmimg=None, base_var=None, count_scale=None, noise_floor=None)
```
debora-pe, line 392: Just so you know, I have made a change to `extract.extract_optimal()` (which is now in `develop`) so that the first argument is not `sciimg` anymore but it's `imgminsky` (= `sciimg - skyimg`). I did that so that I can pass to `extract.extract_optimal()` a real sky (and not the residual sky) for dithered observations (A-B frame) and have an actual 1D extracted sky spectrum. I have made the same change also in `extract.extract_boxcar()`

pypeit/core/datacube.py
```diff
@@ -267,88 +313,102 @@ def extract_standard_spec(stdcube, subpixel=20):
...
+
+    # TODO :: The optimal extraction may suffer from residual DAR correction issues. This is because the
+    #      :: object profile assumes that the white light image represents the true spatial profile of the
+    #      :: object. One possibility is to fit a (linear?) model to the ratio of box/optimal extraction
+    #      :: and then apply this model to the optimal extraction. This is a bit of a fudge.
+    # Note that extract.extract_optimal() stores the optimal extraction in the
+    # sobj.OPT_COUNTS, sobj.OPT_COUNTS_SIG, and sobj.OPT_COUNTS_IVAR attributes.
+    # We need to store the fluxed extraction into the FLAM attributes (a slight fudge).
+    if fluxed:
```
debora-pe, line 401: It would be good to have a more consistent structure here for sobj. For both boxcar and optimal extraction, if the datacube is not fluxed the resulting spec1d would be filled as normal, i.e., `OPT_COUNTS*` and `BOX_COUNTS*` are set while `OPT_FLAM*` and `BOX_FLAM` are `None`. However, if the datacube is fluxed, `BOX_FLAM` would be set but `BOX_COUNTS*` is `None`, and `OPT_COUNTS*` and `OPT_FLAM*` would have the same value. Can you try to make all of these more consistent with each other and with the standard spec1d output?

debora-pe (approved):
@rcooke-ast I have only a few comments, but I'm approving. Just take a look at my latest PR (#1771) that was just merged, to make sure it's not affecting anything here.

pypeit/coadd3d.py
rcooke-ast: Good catch  👍  I've fixed this in the init method.

pypeit/core/datacube.py
rcooke-ast: Correct - I've updated this!

pypeit/core/datacube.py
rcooke-ast: Thanks for the heads-up. I think this is OK, because the `skyimg` I'm passing in is just an array of zeros.

pypeit/core/datacube.py
rcooke-ast: I agree, it would be good to have a consistent approach, but I can't see a way to do this sensibly. I'll have a think about it some more and report back.

rcooke-ast (commented):
Thanks @debora-pe - much appreciated! I've addressed all but one of your comments. I need to think some more about the `spec1d` files.

pypeit/core/procimg.py
```diff
@@ -1372,3 +1372,51 @@ def variance_model(base, counts=None, count_scale=None, noise_floor=None):
     return var
 
 
+def nonlinear_counts(counts, ampimage, nonlinearity_coeffs):
+    r"""
+    Apply a nonlinearity correction to the provided counts.
+
+    The nonlinearity correction is applied to the provided counts using the
```
VascoSch92:
```suggestion
    The nonlinearity correction is applied to the provided ``counts`` using the
```

pypeit/images/rawimage.py
```diff
@@ -304,6 +305,27 @@ def build_ivar(self):
                                      noise_floor=self.par['noise_floor'])
         return utils.inverse(var)
 
+    def correct_nonlinear(self):
+        """
+        Apply a non-linear correction to the image.
+
+        This is a simple wrapper for :func:`~pypeit.core.procimg.nonlinear_counts`.
+
```
VascoSch92:
```suggestion
```

kbwestfall:
@rcooke-ast Please check my conflict resolution in the following:
```
CONFLICT (content): Merge conflict in pypeit/core/flux_calib.py
CONFLICT (content): Merge conflict in pypeit/par/pypeitpar.py
CONFLICT (content): Merge conflict in pypeit/sensfunc.py
```

rcooke-ast:
Tests pass...
image

PR #1775, opened 2024-02-21
Deep copy for DetectorContainer

kbwestfall:
In discussing and thinking through https://github.com/pypeit/PypeIt/pull/1771, I noticed a lingering TODO in `pypeitimage.py` that notes we should be copying the `DetectorContainer` object in the `PypeItImage` returned by the subtraction function.  This PR does that.

@debora-pe, I'm directing this into your PR instead of `develop`.


debora-pe (approved):
Thanks!

profxj (approved):
thx thx

PR #1784, opened 2024-03-13
Coadd1d hotfix

debora-pe:
This allows coadd1d to be run if only one spec1d is provided.


kbwestfall (approved):
Quick and easy!

Is there anything that should be changed in the docs that indicate this?  I.e., how do we tell the user what happens when there's only one spectrum.

Also, add something in the change log.

debora-pe:
> Quick and easy!
> 
> Is there anything that should be changed in the docs that indicate this? I.e., how do we tell the user what happens when there's only one spectrum.
> 
> Also, add something in the change log.

@kbwestfall Added to the change log.
This is fixing a bug that was introduced recently, therefore the documentation is currently (after we fix this) correct.

freddavies:
Looks good to me!

freddavies (approved):

debora-pe:
merging

PR #1785, opened 2024-03-18
Deprecate support for python 3.9

kbwestfall:
This deprecates support for python 3.9, adds support for 3.12, and does a similar bump for the numpy version.  This currently includes no code changes.  It only updates the relevant config and testing files.


tbowers7 (approved):
Looks good.

Are we aware of anything in the code that was waiting for a base version of 3.10 to fix/update?
Also, will there be any `PypeIt` issues arising from NumPy 1.22 going away?

tbowers7:
FYI -- `distutils` has been removed from Python 3.12.

"[distutils](https://docs.python.org/3.10/library/distutils.html#module-distutils) is deprecated with removal planned for Python 3.12. See the [What’s New](https://docs.python.org/3.10/whatsnew/3.10.html#distutils-deprecated) entry for more information."

`distutils` only shows up in `pypeit/bspline/setup_package.py`, but this should be dealt with.

-----

Also, it seems that 3.12 has deprecated `datetime.datetime.utcnow()`.  Maybe don't need to deal with immediately, but we'll be getting these warnings until we do:

```
DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC). (pypmsgs.py:410)
```
But... `datetime.UTC` was not introduced until 3.11.  Maybe leave as a TODO until we set `python>=3.11`?

kbwestfall:
Heads up that I started a dev-suite run in a python 3.12 container.

@profxj , as is typical when we change the CI tests, you will likely need to do a merge override on this one.

tepickering:
> FYI -- `distutils` has been removed from Python 3.12.
> 
> "[distutils](https://docs.python.org/3.10/library/distutils.html#module-distutils) is deprecated with removal planned for Python 3.12. See the [What’s New](https://docs.python.org/3.10/whatsnew/3.10.html#distutils-deprecated) entry for more information."
> 
> `distutils` only shows up in `pypeit/bspline/setup_package.py`, but this should be dealt with.
> 
> Also, it seems that 3.12 has deprecated `datetime.datetime.utcnow()`. Maybe don't need to deal with immediately, but we'll be getting these warnings until we do:
> 
> ```
> DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC). (pypmsgs.py:410)
> ```
> 
> But... `datetime.UTC` was not introduced until 3.11. Maybe leave as a TODO until we set `python>=3.11`?

the `utcnow` warnings have been vexing in many places and as you note there's no real way to fix it for good without requiring >= 3.11. i would actually advocate for doing that since there is a pretty noticeable speed-up between 3.10 and 3.11. easily 10's of minutes of wall clock time for, e.g., binospec MOS data.

tepickering (approved):
looks good, but i'd argue that all the added conditionals required to support 3.10 might make it worth setting 3.11 as the requirement instead. unless there's a really strong argument for keeping support for 3.10 around...

kbwestfall:
> looks good, but i'd argue that all the added conditionals required to support 3.10 might make it worth setting 3.11 as the requirement instead. unless there's a really strong argument for keeping support for 3.10 around...

I don't think there's a strong argument to keep it around, but I also don't feel like there's a strong argument to remove support for it yet.  I'm fine with continuing to support 3.10 until 3.13 releases later this year.

PR #1789, opened 2024-03-21
Fix Reduction failiures trying to deep copy pypeit.images.mosaic.Mosaic

badpandabear:
PR  #1775  added a copy method to DetectorContainer and modified PypeItImage.sub() to use it when copying detectors. But mosaic images use a Mosaic object instead of a DetectorContainer, which has no copy method. Resulting in a `AttributeError: copy is not an attribute of Mosaic!` error message.

To fix this I decided to switch to using copy.deepcopy() instead of using  a copy method, to avoid having to create a new copy method for each DataContainer type object we create.

### DIFF
--- a/pypeit/images/detector_container.py
+++ b/pypeit/images/detector_container.py
@@ -228,14 +228,3 @@ class DetectorContainer(datamodel.DataContainer):
             gain = 1.
         return self.saturation * self.nonlinear * gain
     
-    def copy(self):
-        """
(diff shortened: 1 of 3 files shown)


badpandabear:
There will also be a dev-suite PR to add a test case that catches this.

badpandabear:
Based on discussion in slack, I decided to close this request and open a new one that solves the problem with a different change, adding a copy method to Mosaic.

PR #1790, opened 2024-03-22
Fix Reduction failiures trying to deep copy pypeit.images.mosaic.Mosaic (take 2)

badpandabear:
PR https://github.com/pypeit/PypeIt/pull/1775 added a copy method to DetectorContainer and modified PypeItImage.sub() to use it when copying detectors. But mosaic images use a Mosaic object instead of a DetectorContainer, which has no copy method. Resulting in a AttributeError: copy is not an attribute of Mosaic! error message.

To fix this I added a copy() method to the Mosaic class.

There will be related PR in the devsuite to test for this situation in the future.


kbwestfall (approved):
Looks great.  Thanks!

debora-pe (approved):
Thanks @badpandabear!

PR #1792, opened 2024-03-23
A fix for the Magellan/FIRE long setup.

rcooke-ast:
The FIRE long wavelength solution in the dev-suite is incorrect (it uses ThAr, when only Ne and Ar are present), and doesn't fit most of the lines across the detector. Here is the old version straight out of the dev-suite:

image

And here is the new solution:

image

Given that the FWHM ~12 pixels, I think this is a much better solution.


debora-pe (approved):
This looks good to me. I am worried, though, that the fit does not look that good still, but I imagine it's very hard to get a good fit for such a large wavelength range (and such a low resolution). I trust what you did @rcooke-ast . I'm approving.

kbwestfall (approved):
I agree with @debora-pe , and am approving, as well.

rcooke-ast:
I agree that the solution _looks_ bad, but the FWHM of the lines are 12 pixels, so the rms (relative to the FWHM) is actually pretty good. If nothing else, this is an improvement over the solution that is currently in the dev-suite, so hopefully users have less problems with this solution.

PR #1755, opened 2024-01-12
merging in the NIRSPEC spectrograph Pypeline to develop

Adolfo1519:
This pull request adds the NIRSPEC Pypeline, developed by Greg Doppmann and @Adolfo1519, to the `develop` branch. The merge will include a tutorial for using the new NIRSPEC Pypeline, as well as some updates to `pypeit_identify` and the `full_template` method of wavelength calibration.

PR #1794, opened 2024-03-29
Hotfix: Dependency version consolidation

kbwestfall:
Fixes some versioning issues.


tbowers7:
A comment more than a review: what other dependency versions do we need to check?

For instance, may want to bump `scipy` to `>=1.9` to allow Python3.11  (https://docs.scipy.org/doc/scipy/dev/toolchain.html), but our `numpy` dependence should already grab `scipy>=1.8`.

environment.yml
```diff
@@ -2,7 +2,7 @@ name: pypeit
 channels:
   - defaults
 dependencies:
-  - python>=3.9,<3.12
+  - python>=3.10,<3.13
```
VascoSch92, line 5: I don't know how much is rilevant but give a look here: https://dev.to/hugovk/the-python-3-1-problem-85g

Have you test the changes?

setup.cfg
```diff
@@ -34,7 +34,7 @@ setup_requires = setuptools_scm
 include_package_data = True
 install_requires =
     numpy>=1.23
-    astropy>=4.3
+    astropy>=6.0
```
rcooke-ast, line 37: I am fine with this change, of course, but I will note that `setup.cfg` changes here conflict with those made in PR #1755. I think one of these PRs need to revert the changes to this file.

setup.cfg
```diff
@@ -34,7 +34,7 @@ setup_requires = setuptools_scm
...
     extension-helpers>=0.1
     packaging>=0.19
     scipy>=1.7
```
rcooke-ast, line 40: Can we also update the ginga requirement to >=5.0.1, given the recent work by @ejeschke?

rcooke-ast (approved):
Approving but please consider the changes made in PR#1755 to see if this is a conflict. Thanks!

environment.yml
badpandabear: I played around with this using yaml.safe_load and it parses "python>=3.10,<3.13" as a single string, so it should work. But it wouldn't be a bad idea to wrap them in quotes for safety anyway.

badpandabear (approved):
Looks good, but I second @rcooke-ast 's question about using ginga 5.0.1

kbwestfall:
I mistakenly merged `develop` into this when I made the recent update, so I just redirected this PR.  After #1755 merges, we'll want to quickly move to a new released version anyway...

setup.cfg
kbwestfall: Done!

setup.cfg
kbwestfall: Thanks for pointing this out.  I'm working on consolidating this...

environment.yml
kbwestfall: Thanks for pointing this out, @VascoSch92 , but, given @badpandabear 's comment, I think it's safe to leave it as is.

PR #1795, opened 2024-04-05
Hotfix: Coadd sharex

rcooke-ast:
The top and bottom panels were panning independently. This fix ensures they are tied together.


profxj (approved):
thx!

debora-pe (approved):
Thanks!

PR #1796, opened 2024-04-05
Tweak edges using a gradient method

rcooke-ast:
This PR adds additional functionality to the tweak-edges algorithm. The current implementation is prone to error when there's vignetting (see below image for an example).

The top panel is the slit profile (black), and some vertical lines that show the initial, tweaked (threshold - already implemented in PypeIt), and tweaked (gradient - new method in this PR) determinations of the slit edges. The tweaked (gradient) is a new method that I’ve implementing that is based on the strongest gradients of the slit profile (see bottom panel)… the smoothed profile is in purple, which is what the tweaked locations is based on. The key problem with the old approach is vignetting.

image

I will soon be submitting another PR that refactors the spatial flexure to be (optionally) slit-dependent. This is the start of that effort, and I won't develop further on this until PR #1755 clears.


pypeit/core/flat.py
```diff
@@ -505,10 +470,98 @@ def poly_map(rawimg, rawivar, waveimg, slitmask, slitmask_trim, modelimg, deg=3,
...
+
+    # Median slit width
+    slitwidth = np.median(right - left)
+
+    # Calculate the gradient of the normalized flat profile
+    grad_norm_flat = np.gradient(norm_flat)
+    # Smooth with a Gaussian kernel
+    sig_res = norm_flat.size / slitwidth
+    gauss_kernel = convolution.Gaussian1DKernel(sig_res)
+    grad_norm_flat_smooth = convolution.convolve(grad_norm_flat, gauss_kernel, boundary='extend')
```
kbwestfall: Is there an advantage to using `astropy.convolution` as opposed to `scipy.ndimage.gaussian_filter1d`?

pypeit/core/flat.py
```diff
@@ -505,10 +470,98 @@ def poly_map(rawimg, rawivar, waveimg, slitmask, slitmask_trim, modelimg, deg=3,
...
+    if len(right) != nspec:
+        msgs.error('Input left and right traces must have the same length!')
+
+    # Median slit width
+    slitwidth = np.median(right - left)
+
+    # Calculate the gradient of the normalized flat profile
+    grad_norm_flat = np.gradient(norm_flat)
+    # Smooth with a Gaussian kernel
+    sig_res = norm_flat.size / slitwidth
```
kbwestfall, line 527: Can you explain the motivation/derivation of this?

pypeit/core/flat.py
```diff
@@ -505,10 +470,98 @@ def poly_map(rawimg, rawivar, waveimg, slitmask, slitmask_trim, modelimg, deg=3,
...
+    left_shift = spat_coo[np.argmax(grad_norm_flat_smooth)]
+    right_shift = spat_coo[np.argmin(grad_norm_flat_smooth)]-1.0
+    msgs.info('Tweaking left slit boundary by {0:.1f}%'.format(100 * left_shift) +
+              ' ({0:.2f} pixels)'.format(left_shift * slitwidth))
+    msgs.info('Tweaking right slit boundary by {0:.1f}%'.format(100 * right_shift) +
+                ' ({0:.2f} pixels)'.format(right_shift * slitwidth))
+
+    # Calculate the tweak for the left edge
+    new_left = np.copy(left) + left_shift * slitwidth
+    new_right = np.copy(right) + right_shift * slitwidth
```
kbwestfall: Do you need to copy the `left` and `right` first?

pypeit/find_objects.py
```diff
@@ -986,7 +986,7 @@ def initialize_slits(self, slits, initial=True):
                 Use the initial definition of the slits. If False,
                 tweaked slits are used.
         """
-        super().initialize_slits(slits, initial=True)
+        super().initialize_slits(slits)
```
kbwestfall: If initial is always going to be false, I would add this to the docstring.  I.e., the user needs to know that if they set `initial=True` the code is going to ignore it.  But, if you think there may be a reason to set `initial=True`, you might consider removing this method so that the class always uses the base class method exactly.

pypeit/flatfield.py
```diff
@@ -1943,7 +2017,6 @@ def illum_profile_spectral(rawimg, waveimg, slits, slit_illum_ref_idx=0, smooth_
         modelimg_copy /= relscl_model
         if max(abs(1/minv), abs(maxv)) < 1.005:  # Relative accuracy of 0.5% is sufficient
             break
-    debug = False
```
kbwestfall, line 1946: Good catch!

pypeit/slittrace.py
```diff
@@ -479,7 +479,7 @@ def get_slitlengths(self, initial=False, median=False):
         slitlen = right - left
         return np.median(slitlen, axis=1) if median else slitlen
 
-    def get_radec_image(self, wcs, alignSplines, tilts, slit_compute=None, slice_offset=None, initial=True, flexure=None):
+    def get_radec_image(self, wcs, alignSplines, tilts, slit_compute=None, slice_offset=None, initial=False, flexure=None):
```
kbwestfall, line 482: Have you followed how this change in the default affects other calls to `get_radec_image`?

pypeit/utils.py
```diff
@@ -1380,6 +1380,38 @@ def find_nearest(array, values):
     return idxs
 
 
+def linear_interpolate(x1, y1, x2, y2, x):
```
kbwestfall, line 1383: Thanks for moving this function!

kbwestfall (approved):
Looks good to me, so I'm approving now.  As far as I can tell, you mostly remove the `initial=True` flags for the code relevant to the SlicerIFU only.  Is that true?  There was at least one case where I wanted to make sure that we understand changes to the default behavior.

pypeit/core/flat.py
rcooke-ast: Nope - I've removed the copy -  thanks!

pypeit/find_objects.py
rcooke-ast: Agreed - I've removed the method.

pypeit/slittrace.py
rcooke-ast: Yes, there are two calls in the code, both only affect datacube generation, and this change to the default value is the right choice in both instances.

pypeit/core/flat.py
rcooke-ast: Sure thing - I've added a comment.

pypeit/core/flat.py
rcooke-ast: I did test both, and one was quicker... I've just tested it again, since I'm surprised I didn't choose the scipy one... I've now used the scipy one (2x faster) and this gives the same result to within about machine precision.

rcooke-ast (commented):
Thanks for the review, @kbwestfall - all addressed!

debora-pe (approved):
This looks good for me in general. I'm approving.
But I have a question. The fact that you removed `initial=True` to all the code relevant to the SlicerIFU, does it mean that the parameter `tweak_slits` cannot be `False` for SlicerIFU? If it's correct, is that clear to the users?
Similarly, `spatial_fit_finecorr()` also have `initial=False`. Does it mean that it is always run only when `tweak_slits=True`? Is that clear to the users?

rcooke-ast:
Thanks for the feedback, @debora-pe!

> The fact that you removed initial=True to all the code relevant to the SlicerIFU, does it mean that the parameter tweak_slits cannot be False for SlicerIFU?

Just to confirm, `tweak_slits` can be `False` for `SlicerIFU`. These latest changes basically just make the code identical for `MultiSlit` and `SlicerIFU` when it comes to using either the initial or tweaked edges. The main difference is that `MultiSlit` and `SlicerIFU` can now use an alternative method to define the tweaked edges. Because the way `slits.select_edges()` works, the default will be that the tweaked edges are used, and if those are not available (or if the initial edges are specifically requested) then the initial edges are used.

> Similarly, spatial_fit_finecorr() also have initial=False. Does it mean that it is always run only when tweak_slits=True?

Like the above response, the `spatial_fit_finecorr()` function uses `onslit_tweak`, but this variable is defined by using the tweaked edges by default, and if those are not available, then the initial ones are used.

So, the full functionality is still available to all users - thanks for checking 👍

debora-pe:
Thank you @rcooke-ast for the explanation!

PR #1798, opened 2024-04-12
GMOS SOUTH CCD upgrade and some other fixes

debora-pe:
As titled

### DIFF
--- a/pypeit/manual_extract.py
+++ b/pypeit/manual_extract.py
@@ -85,7 +85,7 @@ class ManualExtractionObj(datamodel.DataContainer):
         m_es = inp.split(';')
         for m_e in m_es:
             parse = m_e.split(':')
-            # det_strip will be a list of a single number (no mosaic) or 2 numbers (mosaic)
+            # det_strip will be a list of a single number (no mosaic) or >= 2 numbers (mosaic)
             det_strip = [int(d) for d in parse[0].strip('()').split(',')]
             # check if it's negative object (i.e., if the det number is negative)
             if np.all([item < 0 for item in det_strip]):
@@ -93,7 +93,7 @@ class ManualExtractionObj(datamodel.DataContainer):
                 det_strip = [item * -1 for item in det_strip]
             else:
                 idict['neg'] += [False]
-            if len(det_strip) == 2 and tuple(det_strip) in spectrograph.allowed_mosaics:
+            if len(det_strip) >= 2 and tuple(det_strip) in spectrograph.allowed_mosaics:
                 # we use detname, which is a string (e.g., 'DET01', 'MSC01')
                 idict['detname'] += [spectrograph.get_det_name(tuple(det_strip))]
             elif len(det_strip) == 1:
--- a/pypeit/spectrographs/gemini_gmos.py
+++ b/pypeit/spectrographs/gemini_gmos.py
@@ -842,9 +842,6 @@ class GeminiGMOSSHamSpectrograph(GeminiGMOSSpectrograph):
             ronoise         = np.atleast_1d([3.75]*4),  # Slow Readout,	Low Gain
             )
 
-        # Detector ID (it is used to identify the correct mosaic geometry)
-        self.detid = 'BI5-36-4k-2,BI11-33-4k-1,BI12-34-4k-1'
-
         # account for the CCD upgrade happened on 2023-12-14
         if hdu is not None:
             # date upgrade
@@ -863,14 +860,54 @@ class GeminiGMOSSHamSpectrograph(GeminiGMOSSpectrograph):
                 detector_dict3['ronoise'] = np.atleast_1d([3.50]*4)
                 # TODO: we may need to update the saturation values
 
-                # Update the detector ID
-                self.detid = 'BI11-41-4k-2,BI13-19-4k-3,BI12-34-4k-1'
-                msgs.info(f'Using the detector parameters for GMOS-S Hamamatsu after the upgrade on '
-                          f'{t_upgrade.iso.split(" ")[0]}')
-
         detectors = [detector_dict1, detector_dict2, detector_dict3]
         # Return
         return detector_container.DetectorContainer(**detectors[det-1])
+
+    def get_mosaic_par(self, mosaic, hdu=None, msc_order=0):
+        """
+        Return the hard-coded parameters needed to construct detector mosaics
+        from unbinned images.
+
+        The parameters expect the images to be trimmed and oriented to follow
+        the PypeIt shape convention of ``(nspec,nspat)``.  For returned
+        lists, the length of the list is the same as the number of detectors in
+        the mosaic, and they are ordered by the detector number.
+
+        Args:
+            mosaic (:obj:`tuple`):
+                Tuple of detector numbers used to construct the mosaic.  Must be
+                one among the list of possible mosaics as hard-coded by the
+                :func:`allowed_mosaics` function.
+            hdu (`astropy.io.fits.HDUList`_, optional):
+                The open fits file with the raw image of interest.  If not
+                provided, frame-dependent detector parameters are set to a
+                default.  BEWARE: If ``hdu`` is not provided, the binning is
+                assumed to be `1,1`, which will cause faults if applied to
+                binned images!
+            msc_order (:obj:`int`, optional):
+                Order of the interpolation used to construct the mosaic.
+
+        Returns:
+            :class:`~pypeit.images.mosaic.Mosaic`: Object with the mosaic *and*
+            detector parameters."""
+
+        # get self.detid
+        # Detector ID (it is used to identify the correct mosaic geometry)
+        if hdu is not None:
+            # account for the CCD upgrade happened on 2023-12-14
+            # date upgrade
+            t_upgrade = time.Time("2023-12-14", format='isot')
+            obs_date = time.Time(self.get_meta_value(self.get_headarr(hdu), 'mjd'), format='mjd')
+
+            if obs_date >= t_upgrade:
+                self.detid = 'BI11-41-4k-2,BI13-19-4k-3,BI12-34-4k-1'
+                msgs.info(f'Using the detector parameters for GMOS-S Hamamatsu after the upgrade on '
+                          f'{t_upgrade.iso.split(" ")[0]}')
+            else:
+                self.detid = 'BI5-36-4k-2,BI11-33-4k-1,BI12-34-4k-1'
+
+        return super().get_mosaic_par(mosaic, hdu=hdu, msc_order=msc_order)
 
     @classmethod
     def default_pypeit_par(cls):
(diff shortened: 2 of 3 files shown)


pypeit/spectrographs/gemini_gmos.py
```diff
@@ -145,14 +154,20 @@ def compound_meta(self, headarr, meta_key):
...
                 binning = headarr[0].get('BINNING')
             if binning is None:
                 msgs.error('Binning not found')
             return binning
+        elif meta_key == 'mjd':
+            obsepoch = headarr[0].get('OBSEPOCH')
+            if obsepoch is not None:
+                return time.Time(obsepoch, format='jyear').mjd
+            else:
+                return None
```
VascoSch92: you should update the docstring as well, because you are return None instead of a Metadata value read from the header(s)

debora-pe:
Tests pass. The failures are unrelated and waiting for another PR to clear.
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  249 passed, 64 warnings in 343.40s (0:05:43) ---
--- PYTEST UNIT TESTS FAILED  3 failed, 131 passed, 168 warnings in 1679.39s (0:27:59) ---
--- PYTEST VET TESTS PASSED  61 passed, 97 warnings in 5166.77s (1:26:06) ---
Failed tests:
Skipped tests:
Testing Started at 2024-04-12T17:55:53.048547
Testing Completed at 2024-04-13T07:12:08.495308
Total Time: 13:16:15.446761
```

jhod0:
I have been able to successfully run this version of PypeIt on GMOS-South data taken in January!

The resulting spectra seem to be perfectly reasonable, here is a red galaxy targeted by our observations.

Thank you Debora!

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -818,9 +835,33 @@ def get_detector_par(self, det, hdu=None):
...
             mincounts       = -1e10,
             numamplifiers   = 4,
-            gain            = np.atleast_1d([1.83]*4),
-            ronoise         = np.atleast_1d([3.98]*4),
+            gain            = np.atleast_1d([1.74]*4),  # Slow Readout,	Low Gain
+            ronoise         = np.atleast_1d([3.75]*4),  # Slow Readout,	Low Gain
             )
+
+        # account for the CCD upgrade happened on 2023-12-14
+        if hdu is not None:
```
rcooke-ast, line 846: Should this perhaps be defined before the detector dictionaries? (i.e. define parameters `gain1`, `gain2`, etc...). I think it's fine to update, but might be cleaner if we set variables near the top of this function and use those variables in the dictionary.

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -818,9 +835,33 @@ def get_detector_par(self, det, hdu=None):
...
+            # date upgrade
+            t_upgrade = time.Time("2023-12-14", format='isot')
+            obs_date = time.Time(self.get_meta_value(self.get_headarr(hdu), 'mjd'), format='mjd')
```
rcooke-ast, line 849: Could this be `None` sometimes?

pypeit/spectrographs/gemini_gmos.py
rcooke-ast: Agreed. I also wonder if it might be worthwhile throwing an error, or, throw a warning and then assume that the observations have ben taken since "2023-12-14".

rcooke-ast (approved):
Just minor comments from me. I'd encourage you to add some recent data to the dev-suite (and I presume there are old data just prior to the detector upgrade already in the dev suite)? Just to make sure that no bugs have crept in for either change.

pypeit/manual_extract.py
```diff
@@ -85,15 +85,15 @@ def by_fitstbl_input(cls, frame: str, inp: str, spectrograph):
         m_es = inp.split(';')
         for m_e in m_es:
             parse = m_e.split(':')
-            # det_strip will be a list of a single number (no mosaic) or 2 numbers (mosaic)
+            # det_strip will be a list of a single number (no mosaic) or >= 2 numbers (mosaic)
```
VascoSch92, line 88: If len(det_strip) == 1 the it is not a mosaic, otherwise it is

Just for the comment. In this way it is explicit

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -145,14 +154,20 @@ def compound_meta(self, headarr, meta_key):
             # binning in the raw frames
             ccdsum = headarr[1].get('CCDSUM')
             if ccdsum is not None:
-                binspatial, binspec = parse.parse_binning(ccdsum)
+                binspec, binspatial = parse.parse_binning(ccdsum)
```
kbwestfall, line 160: This is an excellent catch, but I just want to check that we're sure this is right.  I.e., the ordering of CCDSUM is native to the detector, whereas I think `binspec` and `binspatial` and the ordering in the `binning2string` function is native to pypeit.  I looked through the detector properties and all the various flips are false, so I'm guessing this is right, but I wanted to check we're sure.  Do we have some, e.g., 2x1 binned data that we can use to confirm the ordering.

pypeit/spectrographs/gemini_gmos.py
kbwestfall: We likely do not want it to throw an error because this would cause `pypeit_setup` to fault, which we want to avoid.  Throwing a warning makes sense, and we can add that the return can be `None`.

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -738,7 +755,7 @@ class GeminiGMOSSHamSpectrograph(GeminiGMOSSpectrograph):
     telescope = telescopes.GeminiSTelescopePar()
     supported = True
     comment = 'Hamamatsu detector (R400, B600, R831); see :doc:`gemini_gmos`'
-    detid = 'BI5-36-4k-2,BI11-33-4k-1,BI12-34-4k-1'
+    detid = 'BI5-36-4k-2,BI11-33-4k-1,BI12-34-4k-1'  # this is changed in get_detector_par if t_obs >= t_upgrade
```
kbwestfall: There's a difficult call here.  These "class attributes" are meant to be the same for any instance of the class.  If we need it to change, we should change its scope (i.e. remove this and instead define `self.detid = None` in the `__init__` function).  I think that's what I would recommend doing because these GMOS classes are the only ones that use this class attribute.  The other approach (which seems to be more effort than it's worth) is to define different spectrograph classes, one before and one after the upgrade.  That's cleaner, but requires more code and will require users to use the correct spectrograph name, e.g., `gemini_gmos_south_ham_old` instead of `gemini_gmos_south_ham`.

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -906,7 +947,7 @@ def bpm(self, filename, det, shape=None, msbias=None):
             _bpm_img[i,badr,:] = 1
             # Bad amp as of January 28, 2022
             # https://gemini.edu/sciops/instruments/gmos/GMOS-S_badamp5_ops_3.pdf
-            if obs_epoch > 2022.07:
+            if (obs_epoch > 2022.07) & (obs_epoch < time.Time("2023-12-14", format='isot').jyear):
```
kbwestfall: Are these single values or arrays?  If they're single values, I would instead use:

```
if obs_epoch > 2022.07 and obs_epoch < time.Time("2023-12-14", format='isot').jyear:
```
or
```
if 2022.07 < obs_epoch < time.Time("2023-12-14", format='isot').jyear:
```

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -945,10 +986,10 @@ def config_specific_par(self, scifile, inp_par=None):
             par['calibrations']['wavelengths']['method'] = 'reidentify'
 
         # The bad amp needs a larger follow_span for slit edge tracing
-        obs_epoch = self.get_meta_value(scifile, 'mjd')
+        obs_epoch = time.Time(self.get_meta_value(scifile, 'mjd'), format='mjd').jyear
         binning = self.get_meta_value(scifile, 'binning')
         bin_spec, bin_spat= parse.parse_binning(binning)
-        if obs_epoch > 2022.07:
+        if (obs_epoch > 2022.07) & (obs_epoch < time.Time("2023-12-14", format='isot').jyear):
```
kbwestfall: Same as above: `if 2022.07 < obs_epoch < time.Time("2023-12-14", format='isot').jyear:`

pypeit/spectrographs/gemini_gmos.py
```diff
@@ -511,7 +526,9 @@ def get_mosaic_par(self, mosaic, hdu=None, msc_order=0):
         for i, d in enumerate([det-1 for det in mosaic]):
             msc_sft[i] = shift[d]
             msc_rot[i] = rotation[d]
-            msc_tfm[i] = build_image_mosaic_transform(shape, msc_sft[i], msc_rot[i], binning)
+            # binning is here in the PypeIt convention of (binspec, binspat), but the mosaic tranformations
+            # occur in the raw data frame, which has spatial in y and spectral in x
+            msc_tfm[i] = build_image_mosaic_transform(shape, msc_sft[i], msc_rot[i], tuple(reversed(binning)))
```
kbwestfall, line 535: Based on this, it looks like you've tested the ordering.  So my concerns above are likely wrong, but I felt like I should ask.

kbwestfall (changes requested):
Looks good.  My main reason for clicking "request changes" is the `detid` issue.  All of the other comments are minor.

pypeit/spectrographs/gemini_gmos.py
rcooke-ast: Indeed, and I agree, but returning None could cause an error on line 846 (and several other places in this file where a possible None value is not checked)... I think?

kbwestfall: Ah, yes, you're right.  @debora-pe , you'll need to catch the case @rcooke-ast pointed out.  There are other places in the code that can catch a None mjd (e.g., starting at line 255 in `metadata.py`), but maybe not everywhere.

pypeit/spectrographs/gemini_gmos.py
debora-pe: Yes, I spent a lot of time on this. At certain point I thought I was going crazy. I have tested this on several combination of binning.

pypeit/spectrographs/gemini_gmos.py
debora-pe: I decided to use a dummy `mjd` (using today as date) and throwing a warning. What do you think?

pypeit/spectrographs/gemini_gmos.py
debora-pe: Yes, reasonable wondering.

pypeit/spectrographs/gemini_gmos.py
debora-pe: I knew I would catch your attention here. I did not like it either, but I was lazy to look for another solution. Thank you for the suggestion. Let me know if it looks OK.

pypeit/spectrographs/gemini_gmos.py
debora-pe: See my comment above

pypeit/spectrographs/gemini_gmos.py
debora-pe: I do actually prefer to leave it like this. Other spectrographs have the same flow. Also, the code below is only run if `hdu` is provided, which would make the code full of `if...else` if I define the parameters before the dictionary.

pypeit/spectrographs/gemini_gmos.py
debora-pe: Thanks, done.

pypeit/spectrographs/gemini_gmos.py
debora-pe: Done.

debora-pe:
@rcooke-ast I agree that it would be good to have data in the Dev Suite, but the CCD update is so knew that there are no available public data. A PypeIt user gave me a couple of frame (no science), but that's everything I have.
And yes, we do have older data in the Dev Suite.

pypeit/spectrographs/gemini_gmos.py
debora-pe: I turned out that `self.detid` is already needed in `get_mosaic_par`, which is run just before `get_detector_par`, so I had to update `get_mosaic_par` for the child classes. Let me know what you think.

kbwestfall: Looks good!

kbwestfall (approved):

debora-pe:
Tests still good. Except the unrelated failures.
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  249 passed, 64 warnings in 346.13s (0:05:46) ---
--- PYTEST UNIT TESTS FAILED  3 failed, 131 passed, 168 warnings in 1660.01s (0:27:40) ---
--- PYTEST VET TESTS PASSED  61 passed, 98 warnings in 5207.49s (1:26:47) ---
Failed tests:
Skipped tests:
Testing Started at 2024-04-17T19:09:40.292404
Testing Completed at 2024-04-18T08:17:22.885371
Total Time: 13:07:42.592967
```

debora-pe:
Merging this now.

PR #1800, opened 2024-04-23
Enable spec_samp_fact for wave_method=iref in coadds

freddavies:
Specifically designed for spectrally undersampled instruments with _interesting_ dispersion patterns (not linear or log-linear) like JWST NIRSpec prism. Before, the spec_samp_fact option was completely ignored -- now, `wvutils.get_wave_grid` linearly interpolates within the bounds of the reference wave grid.

Here is example outcome of 2D coadding JWST prism data:

Screenshot 2024-04-23 at 10 27 37


profxj (approved):
Looks good to me.

Is there a way to implement a test?

kbwestfall (approved):

PR #1802, opened 2024-05-06
Auto-generate tables for the wavelength calibration documentation

### LINKED ISSUE #1787: Out-of-date tables in wavelength-calibration docs
The tables here:

https://pypeit.readthedocs.io/en/release/calibrations/wave_calib.html#pypeit-included-line-lists

and here:

https://pypeit.readthedocs.io/en/release/calibrations/wave_calib.html#full-template-dev

are well out of date.  We should remove the table, point the reader to the relevant GitHub link, and/or auto-generate the tables so that they don't get so outdated.

kbwestfall:
As titled.

This is largely a documentation PR.  There are some changes to how `datetime` is imported because I realized that `datetime.datetime.UTC` has never existed (but `datetime.UTC` does), meaning the try blocks used to import `__UTC__` were always faulting (not just for python `<=3.10`).

As a team, we need to fill in the new `pypeit/data/arc_lines/reid_arxiv/summary.txt` file!


pypeit/data/arc_lines/reid_arxiv/summary.txt
tbowers7, line 1: It's going to be a bear of a job filling this in the first time.  Can't we pull the instrument and disperser from the files when building this table in the first place?

I like the checking logic added to ``build_wvcal_tbl_rst.py`` to ensure a 1:1 match between the directory listing and this file, but I worry that the maintenance of this monstrosity will fall to @kbwestfall when he preps the docs for each release.

While it may be a bit of a headache to do, could we add the necessary metadata to each ``reid_arxiv`` file (by hand, if necessary, but automatically for any new ones) so that this table can be generated automagically?

doc/scripts/build_wvcal_tbl_rst.py
```diff
@@ -0,0 +1,121 @@
...
+    ofile = output_root / 'linelist_table.rst'
+    lines = string_table(data_table, delimeter='rst')
+    with open(ofile, 'w') as f:
+        f.write(lines)
+    print(f'Wrote: {ofile}')
+
+
+def write_reid_arxiv_table(output_root):
+
+    # TODO: Pull wavelength range (and resolution?) from files
```
tbowers7, line 65: See comment about this on the ``summary.txt`` file.

doc/scripts/build_wvcal_tbl_rst.py
```diff
@@ -0,0 +1,121 @@
...
+    indx = numpy.isin(names, tbl['File'])
+    if not numpy.all(indx):
+        raise ValueError('Files found in reid_arxiv directory that are *not* in the summary.txt '
+                         f'file.  Include summary entries for the following (or delete them): '
+                         f'{names[numpy.logical_not(indx)]}')
+    indx = numpy.isin(tbl['File'], names)
+    if not numpy.all(indx):
+        raise ValueError('Files found in the summary.txt file that are *not* in the reid_arxiv '
+                         'directory.  Remove summary entries for the following: '
+                         f'{tbl["File"][numpy.logical_not(indx)]}')
```
tbowers7, line 86: Nice check!

tbowers7 (approved):
Nice and very needed PR.  Am approving, but think about this issue:

My main concern is over maintaining the summary table of information about the ``reid_arxiv`` files.  I would suggest adding code in the main repo that includes this information in the ``reid_arxiv`` files when created, and otherwise hand-adding the same to all the existing files.  In this way, no one would need to maintain the ``summary.txt`` file, and the ``.rst`` table would be automagically generated each time with the files in the directory.

profxj (approved):
Looks well worth having.

Merge in when you have the time.

kbwestfall:
Closing for now...

PR #1803, opened 2024-05-14
Enable general odd-even bias subtraction, + minor MODS updates

freddavies:
A short PR which updates the `odd_even` overscan method to work when it is aligned with the spatial axis of the detector instead of the spectral axis (or in the parlance of `procimg.py`, when the `compress_axis` is 0). This is the case for LBT/MODS, and I have updated its default bias correction to use the overscan as also used by their bespoke reduction pipeline modsCCDRed (anecdotally, the bias varies during the night). I have kept the code in a similar format as the pre-existing `odd_even` method but could be convinced to clean it up a bit.

Here is the center (showing all 4 amplifiers) of the standard star skysub frame in the dev suite for MODS2R  with the original bias subtraction:
bias

and the new version with odd-even overscan subtraction (with the same contrast):
overscan

(Some of the improvement is due to a difference in detected junk objects)

Note that the main ambiguity of this particular overscan geometry is which odd/even rows in the overscan correspond to the odd/even rows in the data -- for whatever reason, due to how the sectors are defined in PypeIt, in LBT/MODS this depends on which set of amplifiers you are looking at. So I've added a few lines which check this empirically from the pixel values in the datasec, which should work generally.

Finally, I've updated the red detectors of MODS to use the IR algorithm for telluric corrections, to be consistent with the other red-optical spectrographs.


pypeit/spectrographs/lbt_mods.py
```diff
@@ -40,6 +40,9 @@ def default_pypeit_par(cls):
             all of PypeIt methods.
         """
         par = super().default_pypeit_par()
+        
+        turn_on = dict(use_biasimage=False, use_overscan=True, overscan_method='odd_even')
+        par.reset_all_processimages_par(**turn_on)
```
kbwestfall: I don't know how we got into the habit of this construction, but we don't need to construct the dictionary first.  I.e., these two lines are identical to:
```
par.reset_all_processimages_par(use_biasimage=False, use_overscan=True, overscan_method='odd_even')
```

pypeit/core/procimg.py
```diff
@@ -697,14 +697,34 @@ def subtract_overscan(rawframe, datasec_img, oscansec_img, method='savgol', para
...
+                aligned = np.sign(np.median(odd-even)) == np.sign(np.median(odd_data-even_data))
+                if aligned:
+                    # if so, then subtract overscan
+                    no_overscan[data_slice][1::2,:] = (no_overscan[data_slice][1::2,:].T-odd[:,None]).T
+                    no_overscan[data_slice][0::2,:] = (no_overscan[data_slice][0::2,:].T-even[:,None]).T
+                else:
+                    # otherwise there has been an index mismatch somewhere -- swap odd/even rows
+                    tmp = np.copy(odd)
+                    odd = even
+                    even = tmp
```
kbwestfall: There's a nifty python way of avoiding the copying.  You should be able to perform the swap this way:
```
odd, even = even, odd
```

kbwestfall (changes requested):
Thanks, @freddavies !  The proof is in the pudding :)

I have a few requests, though, including a couple having to do with the original implementation of the odd_even method.

- `osvar` is not being defined properly (and it wasn't in the old odd_even method either).
- at the end of the `elif method.lower() == 'odd_even'` block, I would add a `continue` so that it skips the last few lines of the for loop, meaning you don't need to define `ossub` as a bunch of zeros.
- On the alignment calculation, does this determination require that much of the data-section of the image effectively have no flux?
- Unless I'm missing something, I think you can simplify the last 9 lines to:
```
if not aligned:
    even, odd = odd, even
no_overscan[data_slice][1::2,:] -= odd[None,:]
no_overscan[data_slice][0::2,:] -= even[None,:]
```

At the risk of micro-managing your PR, here's how I would re-write the code within the `odd_even` method block:
```
            if compress_axis == 1:
                _overscan = overscan
                _no_overscan = no_overscan[data_slice]
            else:
                _overscan = overscan.T
                _no_overscan = no_overscan[data_slice].T
            odd = np.median(_overscan[:,1::2], axis=1)
            even = np.median(_overscan[:,0::2], axis=1)
            odd_data = np.median(_no_overscan[:,1::2], axis=1)
            even_data = np.median(_no_overscan[:,0::2], axis=1)
            aligned = np.sign(np.median(odd-even)) == np.sign(np.median(odd_data-even_data))
            if not aligned:
                odd, even = even, odd
            ossub = np.zeros_like(osfit)
            ossub[0::2] = even
            ossub[1::2] = odd
            if var is not None:
                _osvar = osvar if compress_axis == 1 else osvar.T
                odd_var = np.sum(_osvar[:,1::2], axis=1) / _osvar[:,1::2].size**2
                even_var = np.sum(_osvar[:,0::2], axis=1) / _osvar[:,0::2].size**2
                if not aligned:
                    odd_var, even_var = even_var, odd_var
                osvar = np.zeros_like(osfit)
                osvar[0::2] = np.pi/2 * even_var
                osvar[1::2] = np.pi/2 * odd_var
```

This deals with the variance and creates the `ossub` and `osvar` vectors that can be used with the last few lines in the for loop.  The big caveat is that I haven't tested this code.  Up to you how much you do or do not use of the above.

freddavies:
Thanks for the comments, @kbwestfall ! I'll be implementing your suggestions shortly (I didn't mean to make you do the work!!) but I thought I would quickly respond to one point,

> On the alignment calculation, does this determination require that much of the data-section of the image effectively have no flux?

It looks like it does at first glance, but in practice it also works on amplifiers that are full of flux. The reason why it still works is that it is effectively taking the difference between neighboring pixels, so in the absence of bias striping `np.median(odd_data-even_data)` contains the median *slope* of the flux in counts per (spectral) pixel. As long as the odd-even effect is prominent (which is why you would use this method in the first place) I think it's very unlikely for this slope to come close to the true bias offset.

profxj (approved):
I'm approving as I have nothing additional to add on to what Kyle already requested.

pypeit/spectrographs/lbt_mods.py
freddavies: Indeed, I thought this was a bit cumbersome, but that's what I saw in several other spectrograph files...I've changed it to hopefully start a new trend.

pypeit/core/procimg.py
freddavies: This feels like cheating, somehow, but is nevertheless definitely much neater.

freddavies:
@kbwestfall I've updated the code to be closer to your suggestions with tweaks here and there to make it work. In the end the odd/even is not compatible with `ossub` if `compress_axis == 0` (it would need to be 2D instead of 1D) so I have re-introduced the subtraction into the main code and added  the `continue` at the end.

The `lbt_mods` setups in the dev suite run through with no crashes and look bias-cleaned to me.

pypeit/core/procimg.py
```diff
@@ -695,16 +695,37 @@ def subtract_overscan(rawframe, datasec_img, oscansec_img, method='savgol', para
...
+            no_overscan[data_slice] = _no_overscan if compress_axis == 1 else _no_overscan.T
+            if var is not None:
+                _osvar = var[os_slice] if compress_axis == 1 else var[os_slice].T
+                odd_var = np.sum(_osvar[:,1::2],axis=1)/_osvar[:,1::2].size**2
+                even_var = np.sum(_osvar[:,0::2],axis=1)/_osvar[:,0::2].size**2
+                if not aligned:
+                    odd_var, even_var = even_var, odd_var
+                _osvar[:,1::2] = np.pi/2 * odd_var[:,None]
+                _osvar[:,0::2] = np.pi/2 * even_var[:,None]
+                osvar = _osvar if compress_axis == 1 else _osvar.T
```
kbwestfall: Since you're using the `continue` approach, I think we want to replace this with
```
_var[data_slice] = _osvar if compress_axis == 1 else _osvar.T
```
to make sure that the correct `_var` is returned.

kbwestfall (changes requested):
Thanks @freddavies !  One more (hopefully) quick correction to the variance calculation.

pypeit/core/procimg.py
freddavies: Ack, you're right! Fixed.

freddavies: Turns out that didn't actually work, since _osvar has the shape of the overscan region and not the data region, whoops! I've modified it again accordingly.

freddavies:
Before merging, I wanted to mention one thing: the change in the default telluric behavior may cause trouble, as the sensitivity function is highly non-trivial due to the broad wavelength coverage of the long MODS detector. This means that a good sensitivity function can only be derived if one first divides out the flatfield using the `flatfile` SensFuncPar, but making that the default is not really possible (currently).

IR mode sensfunc QA (no flatfile): 
MODS_sensfunc_IR

UVIS mode sensfunc QA:
MODS_sensfunc_UVIS

I am conflicted, then, whether to make the IR algorithm the default one, since it will probably cause user headaches...

kbwestfall (approved):
Thanks for this!!  Copying from a post I sent to you on slack:

For the change to the defaults you made in #1803, I assume the issue about using the flatfile by default is because you need to know the correct flat frame ahead of time?  If so, it may be written to the header (or we can add it), meaning we may be able to add a `flatfile=default` option, or something similar, so that it grabs the correct file from the spec1d header.  If that seems like it will work, we probably should separate that out into a different PR.  In the meantime, can you add some comments on this issue in the `doc/spectrographs/lbt_mods.rst` file?

Approving now.

PR #1806, opened 2024-05-16
Bug fixes to Setup Gui 

badpandabear:
This includes bug fixes from:

- PR comments on the initial SetupGUI PR
- Improve Setup GUI `CTRL-C` signal handling to work with the default PypeIt handler
- Allow `pypeit_setup` to start the setup gui when it isn't given a spectrograph
- Fix horizontal/vertical truncation of of frame type editor
- Saving comments
- Viewing commented out files
- Some docstring fixes/updates

These correspond to v1 01 in the [Setup GUI feedback document](https://docs.google.com/document/d/1VTZYVMNoxmdxuWipPCPvb3CczdldfJnDgMjJ4OnFKhE)

There will be a corresponding dev-suite PR to add unit tests for the Setup GUI.

Devsuite Results:
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  249 passed, 55 warnings in 715.24s (0:11:55) ---
--- PYTEST UNIT TESTS FAILED  3 failed, 145 passed, 169 warnings in 1596.39s (0:26:36) ---
--- PYTEST VET TESTS PASSED  61 passed, 96 warnings in 9124.83s (2:32:04) ---
Failed tests:
Skipped tests:
Testing Started at 2024-05-14T01:10:14.787610
Testing Completed at 2024-05-14T23:31:40.445694
Total Time: 22:21:25.658084
```
All failures are due to NIRSPEC issues I expect to be fixed in the upcomming NIRSPEC PR


badpandabear:
I've updated this PR to include fixes for compatibility matplotlib 3.9.0 and re-run the dev-suite successfully.

debora-pe (approved):
Hi @badpandabear. Thank you for all these improvements.
I basically played around a bit with the gui and it looks great. I just have a couple of comments that you can decide when to address.
1. It's nice that the commented out files are now listed, but I noticed that if any file has a `None` frametype (which happen pretty often and those files are commented out), the script gives an error: 

> Failed to open file <class 'pypeit.pypmsgs.PypeItError'>: Improper frame type supplied!
> The following bit names are not recognized: None
> Check your PypeIt Reduction File

So it would be good to not have that failure.

2. I verified that comments in the parameters block are preserved, but they are not in the Data block (if anybody put a comment in between lines). Is this the expected behaviour? Maybe it makes sense, because it may be difficult to preserve those comments, but I wanted to make sure that this was the expected behaviour.
3.  I don't know if this is a problem with my machine, but when I click the plus sign to add a new tab, the main gui window becomes longer and goes outside of the screen and all the buttons at the bottom (e.g. Open, Clear, ..., Exit) go out of reach. I cannot resize the main gui window horizontally beforehand to prevent that and, once the window goes out of the screen, I cannot move the window around to be able to reach the buttons at the bottom. So, the only thing I can do is to force quit the script.  As said, I'm not sure this is a problem with my machine but I thought I should mention it.

badpandabear:
@debora-pe Thanks for checking things out! The first two comments you mention are problems caused by the way I use PypeItMetaData when loading a file. I plan to stop using PypeItMetaData for opening files but I think I'll wait until the next Setup GUI PR.
The third problem you mention is weird. I can't reproduce it on linux. You're on MacOS right?

debora-pe:
> @debora-pe Thanks for checking things out! The first two comments you mention are problems caused by the way I use PypeItMetaData when loading a file. I plan to stop using PypeItMetaData for opening files but I think I'll wait until the next Setup GUI PR. The third problem you mention is weird. I can't reproduce it on linux. You're on MacOS right?

@badpandabear yes, I have MacOS. So far I tested only on my laptop, I can try on my desktop, but it would probably be next week. Maybe @kbwestfall can check that too on his machine.

debora-pe:
> > @debora-pe Thanks for checking things out! The first two comments you mention are problems caused by the way I use PypeItMetaData when loading a file. I plan to stop using PypeItMetaData for opening files but I think I'll wait until the next Setup GUI PR. The third problem you mention is weird. I can't reproduce it on linux. You're on MacOS right?
> 
> @badpandabear yes, I have MacOS. So far I tested only on my laptop, I can try on my desktop, but it would probably be next week. Maybe @kbwestfall can check that too on his machine.

Hey @badpandabear. About this problem, I tested on my other machine (still MacOS) and actually there is notproblem. When I click the plus sign to add a new tab, the main gui window DOES NOT becomes longer, the size stays the same and I do not have problems to reach the buttons at the bottom. There is definitely a problem with my other machine, but I really don't know what it is.

pypeit/scripts/setup.py
```diff
@@ -79,25 +79,35 @@ def main(args):
...
-            raise IOError('spectrograph is a required argument.  Use the -s, --spectrograph '
-                          'command-line option.')
-
-        # Check that input spectrograph is supported
-        if args.spectrograph not in available_spectrographs:
-            raise ValueError(f'Instrument "{args.spectrograph}" unknown to PypeIt.\n'
-                             f'\tOptions are: {", ".join(available_spectrographs)}\n'
-                             '\tSelect an available instrument or consult the documentation '
-                             'on how to add a new instrument.')
+            if args.gui is False:
```
kbwestfall, line 82: or `if not args.gui` and do we need the `else` below, or could we change the above two lines to `if args.spectrograph is None and not args.gui:` and leave lines 85-90 as they were?

kbwestfall (approved):
One very minor comment from me (that you can choose to ignore).  Thanks for all the added doc strings!

PR #1807, opened 2024-05-21
Correct error message display for Collate1D

woodml:
Encountered an error where the ValueError did not print correctly as it was using a boolean array to index a list. Changed list to a numpy array in the effected line so that it could be indexed using a boolean array.

Example of the error:
``` 
  File "/Users/woodml/anaconda3/envs/pypeit/lib/python3.11/site-packages/pypeit/par/pypeitpar.py", line 5299, in from_dict
    raise ValueError('{0} not recognized key(s) for Collate1DPar.'.format(k[badkeys]))
                                                                          ~^^^^^^^^^
TypeError: only integer scalar arrays can be converted to a scalar index
```


badpandabear (approved):
Looks good. Thanks.

kbwestfall (approved):
Thanks!

PR #1808, opened 2024-05-22
Hotfix: fix FORS2 sensfunc crash

freddavies:
Pushing this to release since it is currently impossible to compute a sensfunc for FORS2.

The sensfunc script calls `config_specific_par`, with logic within it that uses `self.get_meta_value` to figure out the corresponding decker to set a slitedges parameter if it is a long-slit setup. This crashes pypeit_sensfunc because the decker is not read from the spec1d file header:

```
if 'lSlit' in self.get_meta_value(scifile, 'decker') or 'LSS' in self.get_meta_value(scifile, 'decker'):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```

The spec1d file does have a `DECKER` in the header, but the name does not match the corresponding `raw_header_cards`, and so no decker is loaded (printing a series of error messages). In lieu of actually solving this underlying problem, I've wrapped this check within a try/except structure to shove it under the rug; there is likely a more elegant/general solution....


pypeit/spectrographs/vlt_fors.py
```diff
@@ -336,9 +336,12 @@ def config_specific_par(self, scifile, inp_par=None):
...
 
-        if 'lSlit' in self.get_meta_value(scifile, 'decker') or 'LSS' in self.get_meta_value(scifile, 'decker'):
-            par['calibrations']['slitedges']['sync_predict'] = 'nearest'
-
+        try:
+            if 'lSlit' in self.get_meta_value(scifile, 'decker') or 'LSS' in self.get_meta_value(scifile, 'decker'):
+                par['calibrations']['slitedges']['sync_predict'] = 'nearest'
+        except:
+            # if the above fails, we are probably in pypeit_sensfunc
+            msgs.info('Decker not detected; if you are running pypeit_sensfunc, do not panic!')
```
kbwestfall: My recollection was that we always tried to include the original file header in the pypeit output products.  Is that not true for FORS2?  It could be that `get_meta_value` is looking at the header of the wrong extension for the `spec1d` files.

I would isolate the specific line that we know is causing the problem (it also saves us from having to open the file twice to read one value), so I would do the following instead:
```
try:
    decker = self.get_meta_value(scifile, 'decker')
except TypeError as e:
    msgs.warn('Decker not detected; if you are running pypeit_sensfunc, do not panic!')
else:
    if any([t in decker for t in ['lSlit', 'LSS']]):
        par['calibrations']['slitedges']['sync_predict'] = 'nearest'
```

kbwestfall (changes requested):
Thanks!  I'd suggest we isolate the line we know can fault, and provide a suggestion.  I'm fine with this hotfix, but we should add an issue to figure out why this fails for FORS2 and not other instruments...

freddavies:
You've convinced me to dig deeper!

I think the complication for FORS2 is that the header entry corresponding to the decker in the raw FITS files is different depending on whether it was a science exposure or a standard exposure, so the spec1d files for each would have to have a different format (which is probably not what we want?). As far as I can tell, `get_meta_value` calls `compound_meta`, which only checks for the handful of different possibilities from ESO. 

It occurs to me that one solution is to check for `'DECKER'` first! That seems a lot cleaner to me.

pypeit/spectrographs/vlt_fors.py
```diff
@@ -121,20 +121,23 @@ def compound_meta(self, headarr, meta_key):
...
-                        return headarr[0]['HIERARCH ESO SEQ SPEC TARG']
-                    except KeyError:
-                        return headarr[0]['HIERARCH ESO INS MOS CHECKSUM']
-            elif mode == 'IMG':
-                # This is for the bias frames
-                return None
-            else:
-                msgs.error(f"PypeIt does not currently support VLT/FORS2 '{mode}' data reduction.")
+            try:
+                return headarr[0]['DECKER']
```
kbwestfall, line 125: Sorry to continue to be nit-picky, but I can't help myself.  Personally, I'd rather we use if statements instead of these try/except blocks.  To keep the change simple, could we replace the two lines above with:
```
if 'DECKER' in headarr[0]:
    return headarr[0]['DECKER']
```
Then we can leave the rest of the code as it was before.

pypeit/spectrographs/vlt_fors.py
```diff
@@ -337,8 +340,7 @@ def config_specific_par(self, scifile, inp_par=None):
             #par['reduce']['skysub']['bspline_spacing'] = 0.6
 
         if 'lSlit' in self.get_meta_value(scifile, 'decker') or 'LSS' in self.get_meta_value(scifile, 'decker'):
-            par['calibrations']['slitedges']['sync_predict'] = 'nearest'
-
+                par['calibrations']['slitedges']['sync_predict'] = 'nearest'
```
kbwestfall: I think the added indentation here is a typo.  Can we add the edit from my previous suggestion so that we don't have to read the file twice?  I.e.
```
decker = self.get_meta_value(scifile, 'decker')
if 'lSlit' in decker or 'LSS' in decker:
    par['calibrations']['slitedges']['sync_predict'] = 'nearest'
```

kbwestfall (commented):
Sorry to continue to be nit-picky.  I can make these edits myself, if you'd prefer.  Thanks!

pypeit/spectrographs/vlt_fors.py
freddavies: That's what I get for trying to maintain a consistent style ;) no problem!

pypeit/spectrographs/vlt_fors.py
freddavies: Oops, that extra indent must have snuck in while I was reverting. Done.

kbwestfall (approved):
Thanks!

badpandabear (approved):

PR #1813, opened 2024-06-03
Combine refactor

jhennawi:
This short PR refactors the CombineImage class to work on lists of images rather than lists of files. The file processing is moved into the calling routine pypeit.images.buildimage.py which makes more sense since this already takes the file list as the argument. 

The advantages of this scheme is that it allows the class to work with JWST images for which the raw image reading routines do not work, because processing needs to be done to convert JWST outputs into PypeIt inputs. Furthermore, it produces a cleaner interface whereby CombineImage simply combines images but does not do file reading or image processing. 

@rcooke-ast can you please see my comments about the arc lamp checking that I believe you added to this routine on line 291 of combineimage.py?  My feeling is that checking specific things about the inputs to an image combining class should be performed outside of that class such that you pass in only the right images. The point is that others might want to check other things, and at some point the number of things we would check in this single purpose code would balloon. I think lamp exposure time checking should be done in calibrations.


pypeit/images/combineimage.py
```diff
@@ -16,28 +16,21 @@
...
+    Process and combine detector images. 
 
     Args:
-        spectrograph (:class:`~pypeit.spectrographs.spectrograph.Spectrograph`):
-            Spectrograph used to take the data.
-        det (:obj:`int`, :obj:`tuple`):
-            The 1-indexed detector number(s) to process.  If a tuple, it must
-            include detectors viable as a mosaic for the provided spectrograph;
-            see :func:`~pypeit.spectrographs.spectrograph.Spectrograph.allowed_mosaics`.
+            Parameters that dictate the processing of the images.
```
kbwestfall: I think this line is a left-over error from editing.

pypeit/images/combineimage.py
```diff
@@ -16,28 +16,21 @@
...
+        rawImages (:obj:`list`):
+            Either a single :class:`~pypeit.images.rawimage.RawImage` object or a list of one or more 
+            of these objects to be combined into a an image.
```
kbwestfall, line 27: Nice!  But if the type can be a list or RawImage, you should include the latter in the parentheses; i.e.
```
rawImages (:obj:`list`, :class:`~pypeit.images.rawimage.RawImage`):
```

pypeit/images/combineimage.py
```diff
@@ -133,112 +127,77 @@ def run(self, bias=None, scattlight=None, flatimages=None, ignore_saturation=Fal
...
             if ignore_saturation:  # Important for calibrations as we don't want replacement by 0
-                pypeitImage.update_mask('SATURATION', action='turn_off')
+                rawImage.update_mask('SATURATION', action='turn_off')
             # Get a simple boolean good-pixel mask for all the unmasked pixels
-            gpm_stack[kk] = pypeitImage.select_flag(invert=True)
+            gpm_stack[kk] = rawImage.select_flag(invert=True)
+            file_list.append(rawImage.filename)
 
+        # TODO JFH: This should not be here, but rather should be done in the Calibrations class. Putting in calibration specific
+        # lamp checking in an image combining class is not ideal.
```
kbwestfall: In principle, I agree.  A lamp check that is specific to arcs is out of place in a generic image-combining class.  We could push the lamp check into `Calibrations.get_arc`, but we need to be careful about the amount of code that would need to be repeated and minimize the number of times we need to open the raw files.

I would say that the exposure time check should stay here (and we should deal with combining images with different exposure times better...).

pypeit/images/combineimage.py
kbwestfall: Actually, I just realized that this is one or more `PypeItImage` objects.  I.e., this should be replaced with

```
rawImages (:obj:`list`, :class:`~pypeit.images.pypeitimage.PypeItImage):
```

The `process` method of `RawImage` returns a `PypeItImage`, not a `RawImage`.

pypeit/images/combineimage.py
```diff
@@ -46,24 +39,25 @@ class CombineImage:
...
-            A set of one or more images to process and combine.
+        rawImages (:obj:`list`):
+            A list of one or more :class:`~pypeit.images.rawimage.RawImage` objects 
+            to be combined.             
 
     """
-    def __init__(self, spectrograph, det, par, files):
-        self.spectrograph = spectrograph
-        self.det = det
+    def __init__(self, rawImages, spectrograph, par):
```
kbwestfall: Importantly, `spectrograph` is *not* an attribute of each `PypeItImage` (i.e. the actual type of every element of `rawImages`). But, if we move the lamp check out of this class, we only need the spectrograph name, which we can get from `PypeItImage.PYP_SPEC`.  If we do that, the instantiation arguments are only `rawImages` and `par`.

pypeit/images/combineimage.py
```diff
@@ -310,26 +269,26 @@ def run(self, bias=None, scattlight=None, flatimages=None, ignore_saturation=Fal
...
-        comb.files = self.files
-        comb.rawheadlist = pypeitImage.rawheadlist
-        comb.process_steps = pypeitImage.process_steps
+        comb.files = file_list
+        comb.rawheadlist = rawImage.rawheadlist
+        comb.process_steps = rawImage.process_steps
 
         # Build the base level mask
-        comb.build_mask(saturation='default', mincounts='default')
+        comb.build_mask(saturation='default' if not ignore_saturation else None, mincounts='default')
```
kbwestfall, line 262: I don't think this is necessary given how `ignore_saturation` is used in the lines above, but I also think this is a reasonable thing just to ensure that `ignore_saturation` is adhered to if the code above changes.

pypeit/images/rawimage.py
```diff
@@ -677,6 +677,7 @@ def process(self, par, bpm=None, scattlight=None, flatimages=None, bias=None, sl
                                               shot_noise=self.par['shot_noise'],
                                               bpm=_bpm.astype(bool))
 
+        pypeitImage.filename = self.filename
```
kbwestfall: You added `filename` to the argument list for `PypeItImage`.  Can you use that instead of adding this line?  I.e., the lines below are needed because `rawheadlist` and `process_steps` are not keyword arguments for `PypeItImage`.

pypeit/pypeit.py
```diff
@@ -789,6 +789,9 @@ def objfind_one(self, frames, det, bg_frames=None, std_outfile=None):
             bkg_redux_sciimg = sciImg
             # Build the background image
             bg_file_list = self.fitstbl.frame_paths(bg_frames)
+            # TODO I think we should create a separate self.par['bkgframe'] parameter set to hold the image
+            # processing parameters for the background frames.  This would allow the user to specify different
+            # parameters for the background frames than for the science frames.
```
kbwestfall, line 795: Let's talk more about this.

kbwestfall (approved):
Looks good to me.  I think the only requests I have are minor doc updates.  But we should talk more about moving the lamp checking code and whether we want to make that change in this PR.

debora-pe (approved):
It look good to me. But we need to run the tests!

pypeit/images/combineimage.py
rcooke-ast: I agree. In fact, since we are now passing in a `pypeitImage` (and not doing the reading/processing in this function as before), we should move the lamp checks to `pypeit/images/buildimage.buildimage_fromlist()`. This allows @jhennawi to remove the fudged function in `jwst_nirspec()` I think this makes the most sense.

I mostly agree with @kbwestfall that the exposure time stuff should stay here though, although I could be convinced either way, and that might be worth a separate discussion.

rcooke-ast (changes requested):
The PR looks good to me. I only have one request (+ running the tests as usual)

pypeit/images/combineimage.py
jhennawi: Fixed.

pypeit/images/combineimage.py
jhennawi: @rcooke-ast this arc lamp checking stuff was not added by me in this PR, but rather I think by you in a different context. Can you please confirm?  All I did was move that bit of code here from another of the image combine routines in order to be compatible with the refactor.

rcooke-ast: It's a simple change... I will submit a small PR into this so you can see what I mean. I just didn't want to monkey around with your PR.

rcooke-ast: OK @jhennawi - see PR #1822. I haven't tested this with any JWST data, so you might want to check it solves the issue you were having.

kbwestfall:
Lots of test failures:
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  256 passed, 6736 warnings in 508.76s (0:08:28) ---
--- PYTEST UNIT TESTS PASSED  148 passed, 2662 warnings in 916.21s (0:15:16) ---
--- PYTEST VET TESTS FAILED  9 failed, 52 passed, 92604 warnings in 697.22s (0:11:37) ---
--- PYPEIT DEVELOPMENT SUITE FAILED 13/233 TESTS  ---
Failed tests:
    keck_deimos/600ZD_M_6500 pypeit
    keck_deimos/600ZD_tilted pypeit
    keck_deimos/1200B_M_5200 pypeit
    keck_deimos/1200B_LVM_5200 pypeit
    keck_deimos/900ZD_LVM_5500 pypeit
    keck_deimos/1200G_M_5500 pypeit
    mdm_osmos/MDM4K pypeit
    keck_mosfire/long2pos1_H pypeit
    shane_kast_red/300_7500_Ne pypeit
    bok_bc/600 pypeit
    keck_lris_red/long_150_7500_d560 pypeit
    keck_kcwi/small_bh2_4200 pypeit
    mdm_modspec/Echelle pypeit
Skipped tests:
    keck_deimos/600ZD_M_6500 pypeit_flux
    keck_deimos/600ZD_M_6500 pypeit_ql
    keck_deimos/600ZD_M_6500 pypeit_ql
    keck_deimos/900ZD_LVM_5500 pypeit_sensfunc
    keck_deimos/900ZD_LVM_5500 pypeit_flux
    keck_mosfire/long2pos1_H pypeit_coadd_2dspec
Testing Started at 2024-07-17T16:02:18.335840
Testing Completed at 2024-07-18T08:39:00.901060
Total Time: 16:36:42.565220
```

Except for the known `bok_bc/600` failure, I think these all have to do with [this line](https://github.com/pypeit/PypeIt/blob/1827eb3040a3d2c24384781ce3d8fc817566e73b/pypeit/images/combineimage.py#L204) because `self.files` doesn't exist anymore.  Given that #1822 changes this block of code, I'm just rerunning the tests using that branch instead.  If they pass, we should just merge #1822 into this branch, consolidating the PRs.

rcooke-ast:
> Except for the known `bok_bc/600` failure, I think these all have to do with [this line](https://github.com/pypeit/PypeIt/blob/1827eb3040a3d2c24384781ce3d8fc817566e73b/pypeit/images/combineimage.py#L204) because `self.files` doesn't exist anymore. Given that #1822 changes this block of code, I'm just rerunning the tests using that branch instead. If they pass, we should just merge #1822 into this branch, consolidating the PRs.

That's fine with me, but note that @jhennawi hasn't responded to my comment on PR #1822, so I'm not sure if he approves of my proposed changes.

pypeit/images/combineimage.py
jhennawi: Okay I'm leaving it for now.

pypeit/images/combineimage.py
jhennawi: Great suggestion. This was also my hope, i.e. to remove the dependency on spectrograph initially. Now that the lamp check is out, I've removed it as a dependency.

pypeit/images/rawimage.py
jhennawi: Good call. Implemented.

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  256 passed, 6736 warnings in 514.92s (0:08:34) ---
--- PYTEST UNIT TESTS PASSED  148 passed, 2662 warnings in 925.14s (0:15:25) ---
--- PYTEST VET TESTS PASSED  61 passed, 106001 warnings in 5542.49s (1:32:22) ---
Failed tests:
    bok_bc/600 pypeit
Skipped tests:
Testing Started at 2024-07-19T15:34:33.772298
Testing Completed at 2024-07-20T10:06:24.521288
Total Time: 18:31:50.748990
```

rcooke-ast (approved):
Looks good - thanks!

PR #1815, opened 2024-06-04
New EdgeTracePar to truncate slits in the spectral direction

freddavies:
This PR adds a new parameter `trim_spec` to allow for slits to be truncated at both ends of the detector in the spectral direction. It essentially hijacks the `spec_min_max` functionality already used to truncate echelle orders. This can be useful for long-slit spectra with poorly behaved sensitivity functions, e.g. sharp breaks due to a filter or dichroic.

Currently applies the same trimming to all slits, although it would probably be straightforward to update it to allow a list of tuples or strings for slit-by-slit trimming of MOS modes in the future.


freddavies:
Ack, I notice some tests failing, will figure it out...

kbwestfall (approved):
Looks good!  Thanks for the extra effort of reformatting to keep the code lines short!

profxj (approved):
nice nice!

Issue #1823, opened 2024-06-30
run_pypeit -c fails


jhennawi:
See data and PypeIt file in the developer Slack to reproduce this bug. 

<img width="1562" alt="Screenshot 2024-06-29 at 17 43 47" src="https://github.com/pypeit/PypeIt/assets/8774563/d39ed6b8-326a-4914-bc1e-1aa51ae05af1">

kbwestfall:
See #1824.  This fixes the specific failure you found, but it might not fix larger issues associated with trying to only reduced calibrations (if they exist).

PR #1824, opened 2024-07-01
Calibration group hotfix

kbwestfall:
Addresses #1823

The logic used to identify frames with calibration groups assumed there would always be at least one frame that had a single (numeric) calibration group.  This fixes a failure caused when the pypeit file only contains frames that are set to `all` calibration groups.


profxj (approved):
boy we should refactor this someday :)

debora-pe (approved):
Looks good!

PR #1825, opened 2024-07-01
Allow for empty 2D wavecal solution in HDU extension of WaveCalib file

kbwestfall:
As titled.


profxj (approved):
do it!

debora-pe (approved):
Thanks @kbwestfall

PR #1830, opened 2024-07-19
UVES Popler support for coadding data

rcooke-ast:
I have added support to combine PypeIt data products with UVES_popler. The [UVES_popler changes](https://github.com/MTMurphy77/UVES_popler/pull/8) have now been merged, although there may still be a few minor issues that need to be sorted out. Michael Murphy has kindly joined the User's Slack and is happy to assist users with some support if they choose to use this package. I've also updated some of the docs.

Overall, if this is something that receives a lot of interest, we may want to consider redeveloping a GUI for such data combination in the future, possibly with python so users don't need to install CFITSIO and PGPLOT...


kbwestfall (commented):
This looks fine to me, but I think we need reviews from @debora-pe and @jhennawi .  Two requests from me:

 - Increment the datamodel version number for `SpecObj`.
 - I dislike our use of the term "blaze function" here.  The flat-field observations do not strictly provide the blaze function only (unless I'm missing something).  Maybe that's the dominant term, but there are other components (e.g., the detector QE, etc).  Why not just use "FLAT" instead?

debora-pe:
It seems a good idea to have the "blaze function" in the SpecObj, but we may want to consolidate this with what `pypeit_coadd_1dspec` is doing, which also uses the blaze. @jhennawi what do you think?

Also, @rcooke-ast can you add a quick example of how to use UVES_popler with PypeIt spec1ds?

rcooke-ast:
Thanks for the quick feedback @kbwestfall and @debora-pe! Responding to your queries:

> Increment the datamodel version number for SpecObj.

Done!

> I dislike our use of the term "blaze function" here. The flat-field observations do not strictly provide the blaze function only (unless I'm missing something). Maybe that's the dominant term, but there are other components (e.g., the detector QE, etc). Why not just use "FLAT" instead?

You're right, it's not strictly the blaze, but this is how UVES_popler reads in the PypeIt files. I have changed this to FLAT, and I'll submit a small PR to UVES_popler.

> Also, @rcooke-ast can you add a quick example of how to use UVES_popler with PypeIt spec1ds?

I've linked to the documentation, and I don't wish to reproduce too much here, but I've added some of the first simple steps to get things going. For the record, here is the documentation for your interest: 

[https://astronomy.swin.edu.au/~mmurphy/UVES_popler/](https://astronomy.swin.edu.au/~mmurphy/UVES_popler/)

Also, Michael Murphy has a video tutorial of how to interact with UVES_popler, and is going to share that with us.

doc/coadd1d.rst
```diff
@@ -303,3 +303,45 @@ and launches a GUI from the `linetools`_ package. e.g.:
...
+that were reduced by the ESO pipeline, and it has been recently modified
+to support the reduction of PypeIt longslit and echelle data. If you
+require some support on how to use UVES_popler with PypeIt data, please
+join the `PypeIt Users Slack <https://pypeit-users.slack.com>`__ and
+ask for help in the #coadd1d channel. All are welcome to join using
+`this invitation link <https://join.slack.com/t/pypeit-users/shared_invite/zt-1kc4rxhsj-vKU1JnUA~8PZE~tPlu~aTg>`__.
+
+For details on how to use the tool, please refer to the
+`UVES_popler documentation <https://astronomy.swin.edu.au/~mmurphy/UVES_popler/>`__.
+To get you started with reading in PypeIt :doc:`out_spec1D` files,
```
debora-pe, line 323: Thanks!

pypeit/specobjs.py
```diff
@@ -766,6 +766,9 @@ def write_to_fits(self, subheader, outfile, overwrite=True, update_det=None,
                         header[key.upper()] = line
             else:
                 header[key.upper()] = subheader[key]
+                # Also store the datetime in ISOT format
+                if key.upper() == 'MJD':
+                    header['DATETIME'] = (Time(subheader[key][0], format='mjd').isot, "Date and time of the observation in ISOT format")
```
debora-pe: I don't know if you noticed, but the CI tests are failing because of this. The error is:
```
                # Also store the datetime in ISOT format
                if key.upper() == 'MJD':
>                   header['DATETIME'] = (Time(subheader[key][0], format='mjd').isot, "Date and time of the observation in ISOT format")
E                   TypeError: 'float' object is not subscriptable
```

rcooke-ast: Thanks @debora-pe! This line of code is fine for the reduction but not the flux calibration step. So, I've implemented a fix, and this should work again now. It would probably be a good idea to run the dev-suite just to be sure, and I've just started this now.

rcooke-ast: Thanks @debora-pe! This line of code is fine for the reduction but not the flux calibration step. So, I've implemented a fix, and this should work again now. It would probably be a good idea to run the dev-suite just to be sure, and I've just started this now.

rcooke-ast:
Tests pass...

image

rcooke-ast:
OK - I've unified the approach to the blaze computation, and tests now pass. This also required a few bug fixes in the flatfield code which produces infinities in the flat model (even on the `develop` branch).
image

doc/coadd1d.rst
```diff
@@ -303,3 +303,45 @@ and launches a GUI from the `linetools`_ package. e.g.:
...
+.. image:: ../figures/uves_popler.png
+   :scale: 60%
+
+UVES_popler was originally written to coadd ESO/UVES echelle spectra
+that were reduced by the ESO pipeline, and it has been recently modified
+to support the reduction of PypeIt longslit and echelle data. If you
+require some support on how to use UVES_popler with PypeIt data, please
+join the `PypeIt Users Slack <https://pypeit-users.slack.com>`__ and
+ask for help in the #coadd1d channel. All are welcome to join using
+`this invitation link <https://join.slack.com/t/pypeit-users/shared_invite/zt-1kc4rxhsj-vKU1JnUA~8PZE~tPlu~aTg>`__.
```
kbwestfall: I'm a little uncomfortable with this.  A few thoughts:
- Specifically, I worry about volunteering to support use of someone else's software with our output data.  I.e., users should be going to the UVES_popler team first, and only coming to us if there's a problem with our datamodel that keeps them from using their code.
- If we keep this in, we should instead have a dedicated Slack channel for UVES_popler questions; i.e., the #coadd1d channel should be specifically for use of the pypeit 1D coadd code.
- Also, instead of repeating the link here, I would refer to the community section of the landing page (README.rst) so that we don't have to maintain the invitation link in multiple places.  I.e., ":ref:`community`" somewhere, instead of the link.

doc/releases/1.16.1dev.rst
```diff
@@ -17,6 +17,7 @@ Dependency Changes
 
 Functionality/Performance Improvements and Additions
 ----------------------------------------------------
+- Added support for data coaddition with the UVES_popler GUI tool.
```
kbwestfall: I would reword this:  "Adjusted spec1d datamodel to enable use with UVES_popler GUI tool"

pypeit/core/skysub.py
```diff
@@ -588,9 +588,12 @@ def local_skysub_extract(sciimg, sciivar, tilts, waveimg, global_sky, thismask,
         Global sky model produced by global_skysub without the background subtraction.
         If sciimg is an A-B image, then this is the global sky modeled from the A image, while `global_sky` is
         the global modeled from the A-B image. This is None if the sciimg is not an A-B image.
-    fwhmimg : `numpy.ndarray`_, None, optional:
+    fwhmimg : `numpy.ndarray`_, None, optional
```
kbwestfall, line 591: Good catch!

pypeit/core/telluric.py
```diff
@@ -189,6 +189,8 @@ def read_telluric_pca(filename, wave_min=None, wave_max=None, pad_frac=0.10):
 
     dwave, dloglam, resln_guess, pix_per_sigma = wvutils.get_sampling(wave_grid)
     tell_pad_pix = int(np.ceil(10.0 * pix_per_sigma))
+    if dloglam == 0.0:
+        msgs.error('The telluric PCA method has zero spacing in log wavelength. This is not supported.')
```
kbwestfall: Can we move these lines into the `wvutils.get_sampling` function?  The pro is that we don't have to repeat these lines so many time; the cons are that the error message might be less specific and there may be some application of `wvutils.get_sampling` where things can proceed, even if `dloglam==0` (although I don't know what that would be).

pypeit/core/telluric.py
```diff
@@ -2410,18 +2419,12 @@ def __init__(self, wave, flux, ivar, gpm, telgridfile, obj_params, init_obj_mode
         self.disp = disp or debug
         self.sensfunc = sensfunc
         self.debug = debug
-        self.log10_blaze_func_in_arr = None
 
         # 2) Reshape all spectra to be (nspec, norders)
-        if log10_blaze_function is not None:
+        self.wave_in_arr, self.flux_in_arr, self.ivar_in_arr, self.mask_in_arr, self.log10_blaze_func_in_arr, \
+            self.nspec_in, self.norders = utils.spec_atleast_2d(
+            wave, flux, ivar, gpm, log10_blaze_function=log10_blaze_function)
```
kbwestfall, line 2422: Nice.

pypeit/flatfield.py
```diff
@@ -1317,10 +1317,28 @@ def fit(self, spat_illum_only=False, doqa=True, debug=False):
...
             #  go bad?
 
+            # Check for infinities and NaNs
+            winfnan = np.where(np.logical_not(np.isfinite(self.flat_model[onslit_tweak])))
+            if winfnan[0].size != 0:
+                msgs.warn('There are {0:d} pixels with non-finite values in the flat-field model '
+                          'for slit {1:d}!'.format(winfnan[0].size, slit_spat) + msgs.newline() +
+                          'These model pixel values will be set to the raw pixel value.')
+                self.flat_model[np.where(onslit_tweak)[0][winfnan]] = rawflat[np.where(onslit_tweak)[0][winfnan]]
+                self.mspixelflat[np.where(onslit_tweak)[0][winfnan]] = 1.0
```
kbwestfall: I feel like we should be masking these instead, but it's likely a lot of work to propagate the masking through the code...

pypeit/flatfield.py
```diff
@@ -1495,7 +1522,11 @@ def spatial_fit_finecorr(self, normed, onslit_tweak, slit_idx, slit_spat, gpm, s
             illumflat_finecorr[this_slit] = fullfit.eval(xpos, ypos)
         else:
             msgs.warn(f"Fine correction to the spatial illumination failed for {slit_txt} {slit_ordid}")
-            return
+            return illumflat_finecorr
+
+        # If corrections are above the tolerance, then clip them to the level of the tolerance
+        illumflat_finecorr[np.where(illumflat_finecorr < 1-tolerance)] = 1-tolerance
+        illumflat_finecorr[np.where(illumflat_finecorr > 1+tolerance)] = 1+tolerance
```
kbwestfall: Can we use `np.clip`?

pypeit/par/pypeitpar.py
```diff
@@ -1949,7 +1949,7 @@ class SensFuncPar(ParSet):
     For a table with the current keywords, defaults, and descriptions,
     see :ref:`parameters`.
     """
-    def __init__(self, flatfile=None, extrap_blu=None, extrap_red=None, samp_fact=None, multi_spec_det=None, algorithm=None, UVIS=None,
+    def __init__(self, use_blaze=None, extrap_blu=None, extrap_red=None, samp_fact=None, multi_spec_det=None, algorithm=None, UVIS=None,
```
kbwestfall: Just logging my continued objection to the use of this term in this context...  Feel free to ignore.

pypeit/pypeit.py
```diff
@@ -1029,14 +1029,20 @@ def extract_one(self, frames, det, sciImg, bkg_redux_sciimg, objFind, initial_sk
 
         if not self.par['reduce']['extraction']['skip_extraction']:
             msgs.info(f"Extraction begins for {self.basename} on det={det}")
+            # set the flatimg, if it exists
+            try:
+                flatimg = self.caliBrate.flatimages.pixelflat_model
```
kbwestfall: Is there a specific case that you have where this faults?  I'm confused because `pixelflat_model` is a component of the datamodel for `pypeit.flatfield.FlatImages`, so I would have thought that the attribute should always exist.

pypeit/scripts/sensfunc.py
```diff
@@ -66,17 +66,6 @@ def get_parser(cls, width=None):
...
-                            help="R|Use a flat calibration file to compute the blaze function when generating "
-                                 "the sensitivity function.  This is helpful to account for small scale undulations "
-                                 "in the sensitivity function.  Note that it is not possible to set --flatfile and "
-                                 "simultaneously use a .sens file with the --sens_file option. If "
-                                 "you are using a .sens file, set the flatfile there via e.g.:\n\n"
-                                 "F|    [sensfunc]\n"
-                                 "F|         flatfile = Calibrations/Flat_A_0_DET01.fits\n\n"
-                                 "Where Flat_A_0_DET01.fits is the flat file in your "
-                                 "Calibrations directory\n")
-
```
kbwestfall, line 79: To maintain functionality, do we need to add a parameter that allows the user to not use the extracted flat-field spectrum during the sensitivity calculation?

pypeit/specobjs.py
```diff
@@ -791,6 +812,15 @@ def write_to_fits(self, subheader, outfile, overwrite=True, update_det=None,
...
                 header[key.upper()] = subheader[key]
+                # Also store the datetime in ISOT format
+                if key.upper() == 'MJD':
+                    if isinstance(subheader[key], (list, tuple)):
+                        mjdval = subheader[key][0]
+                    elif isinstance(subheader[key], float):
+                        mjdval = subheader[key]
+                    else:
+                        raise ValueError('Header card must be a float or a FITS header tuple')
+                    header['DATETIME'] = (Time(mjdval, format='mjd').isot, "Date and time of the observation in ISOT format")
```
kbwestfall, line 823: What if this is already in the header?  I.e., doesn't the header include most/all information from the raw images?  Would it make sense to instead add a new header keyword that's specific to the data reduction?  Maybe `RDXDATE` or `PYPDATE` or something similar?

pypeit/specobj.py
```diff
@@ -251,11 +255,11 @@ def __init__(self, PYPELINE, DET, OBJTYPE='unknown',
 
     @classmethod
     def from_arrays(cls, PYPELINE:str, wave:np.ndarray, counts:np.ndarray, ivar:np.ndarray,
-                    mode='OPT', DET='DET01', SLITID=0, **kwargs):
+                    flat:np.ndarray, mode='OPT', DET='DET01', SLITID=0, **kwargs):
```
kbwestfall: Given some of the associated changes, should `flat` be optional?

kbwestfall (approved):
A number of nit-picky comments from me, but hopefully they're all relatively easy to deal with.

doc/coadd1d.rst
rcooke-ast: Sounds good. It might be something worthwhile bringing up at the next PypeIt meeting to get opinions from others. I tend to agree with you, so I've removed this, but can reinstate it if it helps. I spoke to Michael Murphy about this, and there isn't a dedicated UVES_popler Slack channel.

doc/releases/1.16.1dev.rst
rcooke-ast: Yep, good call 👍 done. I also moved this to the Datamodel changes section.

pypeit/core/telluric.py
rcooke-ast: Good idea - done!

pypeit/flatfield.py
rcooke-ast: 🤦‍♂️ of course we can... done!

pypeit/flatfield.py
rcooke-ast: It probably is a good idea to mask, but I think these only go badly wrong in the cases when there is very little flux to actually determine the model. In the case of X-SHOOTER, not all orders fill the detector, so there's dead space with no science information, but still the model is being computed. Since we don't mask a few lines below this, I'm tempted to leave this as it is for now, and maybe have a dedicate PR for that in the future.

pypeit/par/pypeitpar.py
rcooke-ast: Fair point... I agree with you about this, and changed it throughout (and propagated these changes to the dev suite, as well).

pypeit/pypeit.py
rcooke-ast: It fails when a user chooses not to flatfield the data (this showed up as one of the failed dev suite tests).

pypeit/specobj.py
rcooke-ast: You're absolutely right - good catch 👍

pypeit/scripts/sensfunc.py
rcooke-ast: Since no default value is set (i.e. the default is `None`) then I think this is OK to remove. I suppose we could add in a `use_flat` boolean, but I'd rather encourage folks to put it in their `sens.par` for a record of what parameters they set.

pypeit/specobjs.py
rcooke-ast: This is the format required by `UVES_popler`, so we need to ensure that it's in the header. I don't think it matters if it gets overwritten, but it "needs" to be there if we don't want to modify the `UVES_popler` code.

rcooke-ast (commented):
Thanks @kbwestfall for the quick turnaround. I've addressed all comments, and made the appropriate changes to the dev suite, too. I'll await for any final comments before re-running the dev-suite to make sure I haven't broken anything in the process.

pypeit/pypeit.py
debora-pe: Yes, `self.caliBrate.flatimages` could be `None`. But, if it's not,  `pixelflat_model` exists (even if it's `None`). So 
`flatimg = self.caliBrate.flatimages.pixelflat_model if self.caliBrate.flatimages is not None else None` 
should work.

pypeit/scripts/sensfunc.py
debora-pe: I would agree with @kbwestfall . I am running `sensfunc` a lot these days and I do it always from command line (I do not create a sens configuration file). I would definitely prefer if the `--use_flat` boolean parameter could exists here.

doc/coadd1d.rst
kbwestfall: Thanks.  To be clear, I'm fine with there being a #uves_popler channel in the pypeit users slack where users can discuss that tool specifically.  I just want to keep it distinct from #coadd1d .

pypeit/pypeit.py
kbwestfall: Got it.  Is there a more direct check?  Something like ``if self.caliBrate.flatimages is None``?

doc/coadd1d.rst
rcooke-ast: Totally fine... but I think it's perhaps best that we don't explicitly list that they should receive help from the PypeIt Slack channel. If there's enough comments about it, then we can always add a #uver_popler channel in the Slack, but maybe best not to advertise it as such, since none of us are really able to address issues within UVES_popler. But if there's demand, I'm more than happy to add that in later, provided that Michael Murphy is happy to be the one who provides support

pypeit/pypeit.py
rcooke-ast: Hmm... good point - I'm not sure, but I'll try the direct check and run the dev-suite to see if it breaks  😃

doc/coadd1d.rst
kbwestfall: Sounds good to me!

pypeit/flatfield.py
```diff
@@ -1317,10 +1317,28 @@ def fit(self, spat_illum_only=False, doqa=True, debug=False):
             # Construct the pixel flat
             #trimmed_slitid_img_anew = self.slits.slit_img(pad=-trim, slitidx=slit_idx)
             #onslit_trimmed_anew = trimmed_slitid_img_anew == slit_spat
-            self.mspixelflat[onslit_tweak] = rawflat[onslit_tweak]/self.flat_model[onslit_tweak]
+            self.mspixelflat[onslit_tweak] = rawflat[onslit_tweak] * utils.inverse(self.flat_model[onslit_tweak])
             # TODO: Add some code here to treat the edges and places where fits
             #  go bad?
 
+            # Check for infinities and NaNs
```
debora-pe: This is really nit-picking...but you could add these checks for infinities and NaNs + unrealistically high or low values of the model before filling `self.mspixelflat[onslit_tweak]` in line 1320, so that you will not need to set 
`self.mspixelflat[np.where(onslit_tweak)[0][winfnan]] = 1.0`
and 
`self.mspixelflat[np.where(onslit_tweak)[0][whilo]] = 1.0`

pypeit/flatfield.py
```diff
@@ -1317,10 +1317,28 @@ def fit(self, spat_illum_only=False, doqa=True, debug=False):
...
+                msgs.warn('There are {0:d} pixels with non-finite values in the flat-field model '
+                          'for slit {1:d}!'.format(winfnan[0].size, slit_spat) + msgs.newline() +
+                          'These model pixel values will be set to the raw pixel value.')
+                self.flat_model[np.where(onslit_tweak)[0][winfnan]] = rawflat[np.where(onslit_tweak)[0][winfnan]]
+                self.mspixelflat[np.where(onslit_tweak)[0][winfnan]] = 1.0
+            # Check for unrealistically high or low values of the model
+            whilo = np.where((self.flat_model[onslit_tweak] >= nonlinear_counts) |
+                             (self.flat_model[onslit_tweak] <= 0.0))
+            if whilo[0].size != 0:
+                msgs.warn('There are {0:d} pixels with non-finite values in the flat-field model '
```
debora-pe: I think this warning should say something like `''There are {0:d} pixels with unrealistically high or low values in the flat-field model ..."`

pypeit/flatfield.py
debora-pe: This is unrelated, but I would like to ask you something because you understand the flat field code better than me. Here in the `fit()` method, there are several calls to `self.slits.slit_img()` and `self.slits.spatial_coordinate_image()` where the spatial flexure is a parameter, but it's never provided. While, for example, when `fit2illumflat()` is used, the spatial flexure is usually provided. 
What do you think about this? Should the spatial flexure always used?

debora-pe (approved):
Thanks @rcooke-ast for this and for fixing the bugs. I have a few comments (one kind of a philosophical question) and a request for the `sensfunc` script.
I am approving anyway.

pypeit/flatfield.py
rcooke-ast: Well-spotted - updated 👍

pypeit/scripts/sensfunc.py
rcooke-ast: OK, sounds good, and I've added it in!

pypeit/flatfield.py
rcooke-ast: Good idea. I've moved this code up a few lines.

Regarding your query about the spatial flexure, I think it's important when computing the spatial illumination, but for the pixel-flat it's less critical since a 2D fit is performed and the entire slit is normalised. @jhennawi may have a more well-informed idea, as I'm pretty sure he wrote this code.

This is quite intricate as well, since you always want to compute the spatial shift relative to the `trace` frame (which defines the slit edges), but the edges then get tweaked by the `illumflat`. So, I *think* that all of the spatial flexure corrections are done relative to the `illumflat`. Note that the spatial flexure correction is still somewhat primitive, but the correction is good enough for most applications.

rcooke-ast (commented):
Comments addressed - thanks @debora-pe!

pypeit/flatfield.py
debora-pe: Thanks for your answer @rcooke-ast. It is a bit more clear now and indeed I hadn't thought how the tweaking of the slits affect the spatial flexure.

pypeit/scripts/sensfunc.py
```diff
@@ -111,13 +110,14 @@ def main(args):
                        "    [sensfunc]\n"
                        "         algorithm = IR\n"
                        "\n")
-        if args.flatfile is not None and args.sens_file is not None:
-            msgs.error("It is not possible to set --flatfile and simultaneously use a .sens "
+
+        if not args.use_flat and args.sens_file is not None:
```
debora-pe: Thank you for adding this.
I think here you want `if args.use_flat and args.sens_file is not None:`
The idea is: if the `--use_flat` flag is not used, it means `args.use_flat` is `False` and in that case we use the default value for the instrument or the value in the `.sens` file. If the `--use_flat` flag  is used, `args.use_flat` is `True` and that would be the value used, which clashes with the `.sens` file (if provided), hence this error.

rcooke-ast: Indeed -- good catch! I'll fix this and restart my dev-suite run. Thanks for catching that!

rcooke-ast:
Tests pass...
image

rcooke-ast:
Tests still pass
image

rcooke-ast:
I've checked this branch works with `UVES_popler`, so I'm merging.

Issue #1634, opened 2023-07-17
Windows installation fails checking for openmp support


badpandabear:
Windows installation for PypeIt v 1.13 fails with a FileNotFoundError.  This was reported in the [users slack](https://pypeit-users.slack.com/archives/CNW4RHF98/p1689118863466359). 

This is caused by the recent code added to `pypeit/bspline/setup_packages.py` that looks to see if openmp is supported. This code attempts to compile sample code, and fails on windows if it can't find the compiler.

A work around is to set the "CC" environment variable to a dummy command that will fail without an exception. In the users slack I suggested the windows "help" command, which will return a failure with no side effects. This allows the installation to succeed but does not allow openmp support to work.

To fix this I believe we should attempt to use the compiler mechanisms built into setuptools so that the test for openmp support uses the same compiler as will be used to build the bspline extension. Ideally we would also use the extension_helpers code to add the openmp compiler arguments, but if that fails I think we should support the Visual studio "/openmp" syntax as well as the gcc "-fopenmp" syntax.  

The installation documentation should also be updated with instructions on what compiler to install on Windows prior to installing PypeIt.

PR #1831, opened 2024-07-19
Fix windows installation problem (Issue #1634)

badpandabear:
As described in issue [1634](https://github.com/pypeit/PypeIt/issues/1634) code added to support OpenMP was failing on windows because gcc was not installed. Through testing I found that installing gcc was not sufficient to deal with this because setup tools would always use the Microsoft Visual Studio compiler when building the extension.

It turns out the [extension_helpers](https://github.com/astropy/extension-helpers) package, already used by PypeIt, had similar code to test for and use OpenMP that worked on windows. So I updated `bspline/setup_package.py` and `pypeit_c_enabled` to use that code instead.  I also updated the installation docs to provide more information on how to install under Windows and how to enable OpenMP on Macs.  I also referenced Windows Subsystem for Linux as an alternative for running under Windows.


doc/installing.rst
```diff
@@ -522,7 +546,8 @@ GCC 12.x via ``homebrew``, you would get ``pypeit`` to use it by doing, for exam
 
 .. code-block:: console
 
-    CC=gcc-12 pip install pypeit
+    $ export CC=gcc-12
```
tepickering: should be gcc-14 as of now. `/opt/homebrew/bin/gcc` is more general option for homebrew installs.

pypeit/bspline/setup_package.py
```diff
@@ -55,16 +14,17 @@ def check_for_openmp():
...
+                          export_symbols=['bspline_model', 'solution_arrays',
+                                          'cholesky_band', 'cholesky_solve',
+                                          'intrv'])
+    
+    # extension_helpers will check for opnmp support by trying to build
+    # some test code with the appropriate flag. openmp provides a big performance boost, but some
+    # systems, notably apple's version of clang that xcode provides, don't support it out of the box.
+
+    add_openmp_flags_if_available(extension)
+    return [extension]
```
tepickering: looks like missing newline at end of file.

tepickering (approved):
i noted a couple changes to be done, most importantly update the version of `gcc` from 12 to 14. or just specify the homebrew gcc specifically. everything else looks good. thanks for the quick work!

doc/installing.rst
badpandabear: I went with the general /opt/homebrew/bin/gcc

pypeit/bspline/setup_package.py
badpandabear: Fixed

kbwestfall (approved):
Looks great!  Thanks for the detailed update to the windows installation instructions!

PR #1832, opened 2024-07-26
Propagate trace_npoly in ech_findobj + minor X-Shooter tweaks

freddavies:
Fixes a bug where the polynomial order for object trace fitting, `[reduce] [[findobj]] trace_npoly`, was not correctly propagated into `ech_objfind_ineach_order`, which lead to poor tracing in X-Shooter NIR. Fortunately, it is the only instrument that uses a higher order by default, so this bug probably did not affect many other reductions.

Additionally, I have updated a few parameters of the X-Shooter detectors to reflect the latest version of the manual, and have slightly reduced the order trimming of the NIR detector to keep some more pixels in the order overlap regions, based on visual inspection of bright object traces.


pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -444,6 +416,8 @@ def check_frame_type(self, ftype, fitstbl, exprng=None):
                                 | (fitstbl['target'] == 'LAMP,FLAT'))
                                & good_dark_seq)
         
+        # TODO: Figure out how to identify lampoffflats
```
kbwestfall: You could follow what @debora-pe did for MOSFIRE.

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -633,17 +578,17 @@ def get_detector_par(self, det, hdu=None):
...
             saturation      = 65535.,
             nonlinear       = 0.86,
             mincounts       = -1e10,
             numamplifiers   = 1,
-            gain            = np.atleast_1d(0.595), # FITS format is flipped: PrimaryHDU  (2106, 4000) w/respect to Python
-            ronoise         = np.atleast_1d(3.1), # raw unbinned images are (4000,2106) (spec, spat)
-            datasec=np.atleast_1d('[:,11:2058]'),  # pre and oscan are in the spatial direction
-            oscansec=np.atleast_1d('[:,2059:2106]'),
-        )
+            gain            = np.atleast_1d(0.64),   # Assumes high gain. TODO: grab this from the header
```
kbwestfall: There are a few instruments that pull this kind of information from the header.  The one that comes to mind is KCWI.

kbwestfall (changes requested):
Thanks @freddavies !  I may be overblowing this, but the changes to the hard-coded detector parameters feel significant.  We need to run the tests, but we should also log these changes in the `doc/releases/1.16.1dev.rst` doc.

I pointed to examples that have an approach to identifying lampoffflats (MOSFIRE) and how to pull detector parameters from file headers (KCWI).  Is that something you have time to implement in this PR, or can we find someone that does?

pypeit/spectrographs/vlt_xshooter.py
freddavies: I had forgotten that, actually, the lampoffflats were already being identified, but mistyped as dark frames (see the code block immediately above). So it was even easier than I was expecting!

pypeit/spectrographs/vlt_xshooter.py
freddavies: I noticed that the binning was already being grabbed via a single line just above, so I followed that example. Let me know if you think it should be done differently.

freddavies:
I have cleared out the TODOs that I left in the spectrograph file as requested, and have confirmed that the `NIR`, `VIS_1x1`, and `UVB_1x1` setups in the dev suite pass. I will run the rest of the setups now and update once they are done.

I have added one additional change to set `fwhm_fromlines = False` to be the default for NIR, as one of the orders seems to often be lost otherwise due to a bad wavelength solution (the dev suite pypeit file also does this). The reason for the poor wavelength solution is unclear, since the automatically measured FWHM looks correct, but the wavelength solution in that order seems to have consistently high RMS. I think investigating that underlying problem is outside the scope of this PR, however.

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -624,6 +571,10 @@ def get_detector_par(self, det, hdu=None):
         # Binning
         # TODO: Could this be detector dependent??
         binning = '1,1' if hdu is None else self.get_meta_value(self.get_headarr(hdu), 'binning')
+        
+        # Grab the gain and read noise from the header.
+        gain = None if hdu is None else hdu[0].header['HIERARCH ESO DET OUT1 CONAD']
```
debora-pe, line 577: Note sure how the code will react to a `gain=None` (also `ronoise = None`). Note that `hdu` is an optional parameter, so it can be None.
Can you just use a default value (maybe the one that was used before)?

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -915,6 +866,10 @@ def get_detector_par(self, det, hdu=None):
         """
         # Binning
         binning = '1,1' if hdu is None else self.get_meta_value(self.get_headarr(hdu), 'binning')
+        
+        # Grab the gain and read noise from the header.
+        gain = None if hdu is None else hdu[0].header['HIERARCH ESO DET OUT1 CONAD']
```
debora-pe, line 869: same here.

debora-pe (approved):
@freddavies thanks for fixing this bug. 
I am approving, but  please take into account my one comment.

pypeit/spectrographs/vlt_xshooter.py
freddavies: I borrowed the `None`s from from the KCWI implementation. I think the idea is that if there is no hdu, then no image is actually being processed (e.g. during sensfunc or coadd1d), so those values are irrelevant. But I can put in the nominal values if it might otherwise be necessary.

kbwestfall: Actually, the main reason these functions to allow `hdu=None` is so that [this table](https://pypeit.readthedocs.io/en/release/detectors.html#instrument-specific-data) can be built automatically.  Entries in the table set to None mean that the data are read from the files being processed.

pypeit/spectrographs/vlt_xshooter.py
```diff
@@ -569,12 +519,9 @@ def order_platescale(self, order_vec, binning=None):
...
         """
-        # TODO: Either assume a linear trend or measure this
-        # X-shooter manual says, but gives no exact numbers per order.
-        # NIR: 52.4 pixels (0.210"/pix) at order 11 to 59.9 pixels (0.184"/pix) at order 26.
-
-        # Right now I just assume a simple linear trend
-        plate_scale = 0.184 + (order_vec - 26)*(0.184-0.210)/(26 - 11)
+        # TODO: Figure out the order-dependence of the updated plate scale
+        # From the X-Shooter P113 manual, average over all orders. No order-dependent values given.
+        plate_scale = 0.245*np.ones_like(order_vec)
```
kbwestfall, line 524: Why doesn't this consider the binning?

kbwestfall (changes requested):
Hi @freddavies.  Thanks for adding the release comment, but I thought it would useful to have something more specific so that people don't have to try to look up the changes themselves by digging through the code.

I've started to do this, but in the process I realized that we may not be doing something correctly with the pixel scale.   There is a pixel scale defined for each detector, but there is also the `order_pixelscale` function.  Do these need to be consistent?  And in some cases the `order_pixelscale` function doesn't account for the binning.  Does that matter?

freddavies:
@kbwestfall Good point about the `order_platescale` function -- it appears to be used in object finding for echelle spectrographs instead of the value in the detector dict. I have now made them consistent with the new `plate_scale` numbers; while they no longer have any order-to-order variation, the typical values seem to be different enough from before that it is probably better to use the new ones.

pypeit/spectrographs/vlt_xshooter.py
freddavies: Oops, it does now! For VIS and UVB, at least. NIR is never binned.

kbwestfall (approved):

PR #1834, opened 2024-07-31
Adding support for the (decommissioned) spectrograph: AAT/UHRF

rcooke-ast:
There is an accompanying dev-suite PR. I will also add some documentation about this spectrograph. Tests currently pass.


pypeit/pypeitsetup.py
```diff
@@ -186,7 +186,20 @@ def from_file_root(cls, root, spectrograph, extension='.fits'):
         Returns:
             :class:`PypeitSetup`: The instance of the class.
         """
-        return cls.from_rawfiles(io.files_from_extension(root, extension=extension), spectrograph)
+        spec = load_spectrograph(spectrograph)
```
profxj: I hear that @kbwestfall has some issues/fixes for this

profxj (approved):
All good on my end.  I hear @kbwestfall will have one fix
to suggest/add.

pypeit/core/arc.py
```diff
@@ -1069,6 +1070,57 @@ def detect_lines(censpec, sigdetect=5.0, fwhm=4.0, fit_frac_fwhm=1.25, input_thr
...
+        # Find all pixels to the right of the peak that are above the noise level
+        ww = np.where((xarray >= pks[i]) & (specdiff < 0.0))[0]
+        # Find the first pixel to the right of the peak that is below the noise level
+        nmask = np.where(np.diff(ww) > 1)[0]
+        if nmask.size != 0 and nmask[0] > 5:
+            # Mask all pixels to the right of the peak
+            maxi = min(nmask[0], ww.size)
+            outbpm[pks[i]:ww[maxi]+2] = True
+    # Return the output bpm
+    return outbpm
```
kbwestfall: It would be great to have a test for this, but I'm fine with you punting on that for this PR.

pypeit/core/findobj_skymask.py
```diff
@@ -1762,9 +1762,16 @@ def objs_in_slit(image, ivar, thismask, slit_left, slit_righ,
...
+    # Check that the peak mask is not empty
+    if np.sum(np.logical_not(peak_bpm))/peak_bpm.size < 0.05:  # 95% masked!
+        msgs.warn('Peak masking was too aggressive. Attempting to sigma clip.')
+        smash_mask = np.logical_not(gpm_smash)
+    else:
+        smash_mask = np.logical_not(gpm_smash) | peak_bpm
     flux_smash_mean, flux_smash_med, flux_smash_std = astropy.stats.sigma_clipped_stats(
-        flux_smash, mask=np.logical_not(gpm_smash), sigma_lower=3.0, sigma_upper=3.0
-    )
+        flux_smash, mask=smash_mask, sigma_lower=3.0, sigma_upper=3.0)
```
kbwestfall: I haven't dug into the context of this change, but it feels important.  Can you give us more information about the problem that you've addressed with this code?  And, do you have any assessments of how this affects all the other instruments?

pypeit/spectrographs/keck_kcwi.py
```diff
@@ -860,7 +860,7 @@ def get_detector_par(self, det, hdu=None):
             # Some properties of the image
             binning = self.compound_meta(self.get_headarr(hdu), "binning")
             numamps = hdu[0].header['NVIDINP']
-            specflip = True if hdu[0].header['AMPID1'] == 2 else False
+            specflip = False if hdu[0].header['AMPMODE'] == 'ALL' else True
```
kbwestfall, line 863: I would comment on this in the change log.  Is this a general change for all KCWI taken, or does it reflect a change in header that needs an associated date `if..else` check?

kbwestfall (changes requested):
There are a couple of somewhat buried changes that I was hoping you could comment on.  I merged in `release` and `develop`.  Can we re-run tests?

pypeit/core/arc.py
rcooke-ast: Good point - I've added one in.

pypeit/spectrographs/keck_kcwi.py
rcooke-ast: I was finding a problem with one of the datasets I have. This is effectively the same thing, but it depends which 2 amplifiers are used. This is more conservative. I have added this to the change log.

pypeit/core/findobj_skymask.py
rcooke-ast: To be honest, I developed this many months ago, and I can't remember the motivation, or why it wasn't working in the first place. The code is basically finding peaks and masking around them, provided that pixel values continuously decrease away from the peak pixel. There needs to be a continuous decrease of 5+ pixels in order to activate this feature. I suspect this is only important when object profiles are excessively broad (which is what I was dealing with at the time).

Given that I can't recall what the motivation was, I am happy to revert this.

rcooke-ast (commented):
Comments addressed @kbwestfall - thanks!  I'll run tests

rcooke-ast:
Tests pass
image

kbwestfall (approved):

PR #1835, opened 2024-07-31
Issues with adding missing orders

kbwestfall:
This PR does two things:

 - Application to some HIRES datasets highlight that the left-right syncing procedure performed after adding missing orders has some conceptual issues.  This introduces a short-cut that we'll need to test.  Instead of redoing the full syncing procedure, the code now only checks that the edges are still synced after the missed orders are added.
 - When adding missed orders, the code will nominally add orders regardless of how much they overlap.  This adds a new parameters `max_overlap` that can be used to limit the added orders to only those that have less than this maximum overlap with adjacent orders.

There are also some random doc fixes.  I will run and post the dev-suite results.


kbwestfall:
The results of the run from a couple of weeks ago:
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  256 passed, 6736 warnings in 500.40s (0:08:20) ---
--- PYTEST UNIT TESTS PASSED  148 passed, 2662 warnings in 936.81s (0:15:36) ---
--- PYTEST VET TESTS FAILED  2 failed, 60 passed, 98674 warnings in 5529.67s (1:32:09) ---
--- PYPEIT DEVELOPMENT SUITE FAILED 2/239 TESTS  ---
Failed tests:
    bok_bc/600 pypeit
    keck_hires/J0306+1853_U074_RED_C2_ECH_0.72_XD_1.42_1x3 pypeit
Skipped tests:
Testing Started at 2024-07-31T18:15:57.941519
Testing Completed at 2024-08-01T12:14:56.105569
Total Time: 17:58:58.164050
```

Bok_bc/600 is now irrelevant.

I'm going to look into the HIRES failure today.

pypeit/edgetrace.py
```diff
@@ -5014,41 +5074,135 @@ def order_refine_free_format(self, reference_row, combined_order_tol=1.8, debug=
...
+        offset = _right[isrt][add_left.size:] - right
+        self.edge_fit[:,right_indx] += offset[None,:]
 
         # Get the adjusted traces to add.  Note this currently does *not* change
         # the original traces
-        return _left[isrt][:add_left.size], _right[isrt][:add_right.size], rmtraces
-        
+        good_order = good_order[isrt][:add_left.size]
+        add_left = _left[isrt][:add_left.size][good_order]
+        add_right = _right[isrt][:add_left.size][good_order]
```
debora-pe: Hey @kbwestfall it seems there is bug here (many ADAP datasets failed here). Should this line be:
`add_right = _right[isrt][:add_right.size][good_order]`

kbwestfall: Ah, yes.  Sorry about this!  I (hopefully) fixed this and pushed a new version.

pypeit/edgetrace.py
```diff
@@ -4855,13 +4863,17 @@ def order_refine(self, debug=False):
...
+            bracket = True
             add_left, add_right, rmtraces \
-                    = self.order_refine_free_format(reference_row, debug=debug)
+                    = self.order_refine_free_format(reference_row, bracket=bracket, debug=debug)
 
         if add_left is None or add_right is None:
             msgs.info('No additional orders found to add')
             return
-
+
```
debora-pe, line 4888: Hey @kbwestfall , I don't know if this is the best way, but by adding 
`self.clean_traces(force_flag='OFFDETECTOR', rebuild_pca=False, sync_mode='both', assume_synced=True)`
below, after `self.insert`, it fixed the problem.
I have noticed that in the case of added orders we need to force the flag `OFFDETECTOR`, because those orders have also the flag `ORDERINSERT` therefore 
`self.fully_masked_traces(flag=self.bitmask.bad_flags, exclude=self.bitmask.exclude_flags)`
alone will exclude those orders.

kbwestfall: The specific cause of the added orders was because the code always added the "bracketing" orders, but it only removed them if the code detected overlap.  I updated the code so that it now also removes those bracketing orders when there is no overlap, as it should have done to begin with!  Sorry for these bugs.

This means that we shouldn't need to run `clean_traces`.

pypeit/edgetrace.py
```diff
@@ -4953,7 +5012,7 @@ def order_refine_free_format(self, reference_row, combined_order_tol=1.8, debug=
                                      function='legendre', lower=3., upper=3., maxiter=5,
                                      sticky=True)
 
-        # Ideally, measured widths/gaps should be regected for one of the
+        # Ideally, measured widths/gaps should be rejected for one of the
```
debora-pe, line 5023: Hi @kbwestfall sorry if I keep bugging you with this. 
I was looking at the ADAP runs and all completed except one that failed (which is not bad!). But as I was diagnosing the reason of the failure, I noticed that, this particular dataset, for some reason, had a ghost orders (PypeIt detected an order in the gap) as the first order (the leftmost order). Since  `trace.find_missing_orders` starts the search for missing orders from the first order, the entire search is messed up because the first order is messed up.
So, I thought we could add a condition here (similar to the `combined_orders` below) to exclude this bad order before running `trace.find_missing_orders`. It seems that it could be beneficial in general for ghost orders. So it would be to do something like:
`ghost_orders = np.append(gap, gap[-1]) / gap_fit.eval(cen) < ghost_orders_tol`

`gap` and `gap_fit.eval(cen)` have different sizes, this is why I'm appending (but I'm not sure it's the right way to deal with it). And I'm not sure how much `ghost_orders_tol` should be, but maybe 0.6 or so.

What do you think? I can send you the data if you want to take a look.

kbwestfall: Hi @debora-pe.  Sorry for the slow response!  Your solution sounds good to me, but I'd like to have a look to see what's happening, as well.  Can you send me the data?

kbwestfall: I think this should now be fixed.  See the changes in [this commit](https://github.com/pypeit/PypeIt/pull/1835/commits/08a2798c8fc21ec22dfa61b11d23751cadf0e982).

pypeit/edgetrace.py
```diff
@@ -4996,123 +5079,328 @@ def order_refine_free_format(self, reference_row, combined_order_tol=1.8, debug=
...
-                    _right[i] -= ngap
+        if np.all(_left[1:] - _right[:-1] > 0):
+            if bracket:
+                add_left, add_right = self._handle_bracketing_orders(add_left, add_right)
+            # There is no overlap, so just return the orders to add
+            return add_left, add_right, rmtraces
+
+        # Used to remove orders that have too much overlap
+        nord = _left.size
+        good_order = np.ones(nord, dtype=bool)
```
debora-pe:
```suggestion
        ok_overlap = np.ones(nord, dtype=bool)
```

pypeit/edgetrace.py
```diff
@@ -4996,123 +5079,328 @@ def order_refine_free_format(self, reference_row, combined_order_tol=1.8, debug=
...
+
+        # Loop sequentially so that each pair is updated as the loop progresses
+        for i in range(1, nord):
+            # *Negative* of the gap; i.e., positives values means there's
+            # overlap
+            ngap = _right[i-1] - _left[i]
+            if ngap > 0:
+                if self.par['max_overlap'] is not None:
+                    good_order[i-1] = 2*ngap/(_right[i-1] - _left[i-1]) < self.par['max_overlap']
+                    good_order[i] = 2*ngap/(_right[i] - _left[i]) < self.par['max_overlap']
```
debora-pe:
```suggestion
                    ok_overlap[i-1] = 2*ngap/(_right[i-1] - _left[i-1]) < self.par['max_overlap']
                    ok_overlap[i] = 2*ngap/(_right[i] - _left[i]) < self.par['max_overlap']
```

pypeit/edgetrace.py
```diff
@@ -4996,123 +5079,328 @@ def order_refine_free_format(self, reference_row, combined_order_tol=1.8, debug=
...
 
         # Get the adjusted traces to add.  Note this currently does *not* change
         # the original traces
-        return _left[isrt][:add_left.size], _right[isrt][:add_right.size], rmtraces
-        
-    def order_refine_free_format_qa(self, cen, combined_orders, width, gap, width_fit, gap_fit,
-                                    order_cen, order_missing, ofile=None):
+        good_order = good_order[isrt][:nadd]
+        add_left = _left[isrt][:nadd][good_order]
+        add_right = _right[isrt][:nadd][good_order]
```
debora-pe:
```suggestion
        ok_overlap = ok_overlap[isrt][:nadd]
        add_left = _left[isrt][:nadd][ok_overlap]
        add_right = _right[isrt][:nadd][ok_overlap]
```

debora-pe (commented):
Hey @kbwestfall. I added some edit suggestions that fix the problem with the dataset that I shared with you on Slack, but I need you to check it to make sure I understood the code.

I also notice that the parameter  `order_outlier` is not helping much. In a couple of datasets makes things worse. In one case makes the reduction to fail and in another adds a missing order (I'm guessing one that was considered outlier) in a wrong location and fail later on.

There are still 2 more failures that I am trying to debug. They happen after the edge tracing (wavecalib and flat fielding) but I'm not sure they are caused by edge tracing or by recent changes to the code from other PRs.

kbwestfall:
> I added some edit suggestions that fix the problem 

Looks good!  Thanks for finding the fix, and sorry again for not catching these bugs.

> I also notice that the parameter `order_outlier` is not helping much. In a couple of datasets makes things worse. In one case makes the reduction to fail and in another adds a missing order (I'm guessing one that was considered outlier) in a wrong location and fail later on.

We can try setting the default back to `None` for this.  I tested doing this against the dataset that was causing the original problem and it works fine.  I.e., the other changes I made to the code catch the underlying issue for this dataset, so we may not need the `order_outlier` parameter in general.

pypeit/edgetrace.py
```diff
@@ -877,11 +877,19 @@ def auto_trace(self, bpm=None, debug=False, show_stages=False):
         if not self.is_empty and self.par['add_missed_orders']:
             # Refine the order traces
             self.order_refine(debug=debug)
-            # Check that the edges are still sinked (not overkill if orders are
-            # missed)
-            self.success = self.sync()
-            if not self.success:
-                return
+            # Check that the edges are still sinked
```
badpandabear: I think this should be "synced"

pypeit/edgetrace.py
```diff
@@ -877,11 +877,19 @@ def auto_trace(self, bpm=None, debug=False, show_stages=False):
...
+
+#           KBW: This was the old code that resynced.  The comment suggests
+#           this may be necessary if orders are missed, so I need to test if
+#           that's true.
+#            # Check that the edges are still sinked (not overkill if orders are
+#            # missed)
+#            self.success = self.sync(debug=True)
+#            if not self.success:
+#                return
+
```
badpandabear, line 894: Can the above commented out code be removed or is there still more testing needed?

badpandabear (approved):

pypeit/edgetrace.py
kbwestfall: Good catch!  Fixed.

pypeit/edgetrace.py
kbwestfall: I'd like to keep it around for a while.  I've improved the comment a bit to indicate this.

debora-pe (approved):
@kbwestfall I think we should just run the tests and then we can merge. There is one last favor...Can you please add some documentation about the the QA plot for the missing orders. With all the new improvements, it would be very useful to have something written that explain what all the different points in the plot are. 

Thank you so much for helping with this.

badpandabear:
Result of Oct 7 dev-suite run:
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  256 passed, 6840 warnings in 609.23s (0:10:09) ---
--- PYTEST UNIT TESTS PASSED  152 passed, 2670 warnings in 1092.98s (0:18:12) ---
--- PYTEST VET TESTS PASSED  62 passed, 106525 warnings in 6059.94s (1:40:59) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 247/247 TESTS  ---
Testing Started at 2024-10-07T22:02:52.633780
Testing Completed at 2024-10-08T17:54:37.289253

```

PR #1836, opened 2024-08-05
HOTFIX: pyqt6 version and ginga connection timeout

kbwestfall:
As titled.  Note this is going straight to `release`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ python_requires = >=3.10,<3.13
 setup_requires = setuptools_scm
 include_package_data = True
 install_requires =
-    numpy<2.0.0
+    numpy>=1.23,<2.0.0
     astropy>=6.0
     extension-helpers>=0.1
     packaging>=0.19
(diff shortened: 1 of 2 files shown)


jhennawi (approved):
Thanks!

jhennawi:
Will you also pull this into develop when approved?

kbwestfall:
> Will you also pull this into develop when approved?

Yes, but it will also require a PR.

setup.cfg
```diff
@@ -33,7 +33,7 @@ python_requires = >=3.10,<3.13
 setup_requires = setuptools_scm
 include_package_data = True
 install_requires =
-    numpy>=1.23
+    numpy<2.0.0
```
debora-pe: Quick question. If before this needed to be `>=1.23`, should it now still need to be `>=1.23` and `<2.0.0`?

debora-pe (approved):
One quick question

setup.cfg
kbwestfall: Yes!  Good catch!  Fixed.

PR #1837, opened 2024-08-09
Hotfix display

rcooke-ast:
Simple hotfix that deals with the case when `trc_name` is not set


profxj (approved):

kbwestfall (approved):
Look good!  Please add a change comment in the docs.

rcooke-ast:
Good call - done!

PR #1839, opened 2024-08-12
Refactor file discovery

kbwestfall:
Prompted by @rcooke-ast's changes to the file search functionality, I've done a couple of things:

- The pypeit code will now be *exclusive* (not *inclusive*) of compression signatures.  I.e., the file in the pypeit file, and the allowed extensions for each spectrograph must now be given explicitly and exactly match the file name on disk.  To me, it's bad practice to allow the file name to be different from what is given in the pypeit file.
- This means that I had to explicitly define the allowed extensions for each spectrograph.  This defaults to either `.fits` or `.fits.gz`, but is overwritten for a few spectrographs.

There's also an associated dev-suite PR that adds a unit test for this.  Both PRs are directed into the relevant AAT/UHRF PRs.


pypeit/spectrographs/gemini_gnirs.py
```diff
@@ -587,6 +587,7 @@ class GNIRSIFUSpectrograph(GeminiGNIRSSpectrograph):
     #   * Need to store a wavelength solution for different grating options (Note, the Holy Grail algorithm works pretty well, most of the time)
     name = 'gemini_gnirs_ifu'
     pypeline = 'SlicerIFU'
+    allowed_extensions = ['.fits', '.fits.bz2']
```
rcooke-ast: What happens is the user has a set of fits files in their `Raw/` directory, and also the exact same files but compressed?

pypeit/io.py
```diff
@@ -851,40 +851,48 @@ def create_symlink(filename, symlink_dir, relative_symlink=False, overwrite=Fals
...
+            if not _raw_path.is_dir():
+                msgs.error(f'{_raw_path} does not exist!')
+        ext = [extension] if isinstance(extension, str) else extension
+        files = numpy.concatenate([sorted(_raw_path.glob(f'{prefix}*{e}')) for e in ext])
+        return numpy.unique(files).tolist()
     
     if isinstance(raw_path, list):
-        return numpy.concatenate([files_from_extension(p, extension=extension) for p in raw_path]).tolist()
+        files = numpy.concatenate([files_from_extension(p, extension=extension) for p in raw_path])
+        return numpy.unique(files).tolist()
```
rcooke-ast, line 893: I think we possibly need to check if there are strings present in this list that are identical, except for the extension? I would be doubtful that anyone would carry around identical compressed and uncompressed files in the same directory, but then again, it's possible.

rcooke-ast (approved):
This is much cleaner, and a great way to search through potential extensions and paths 👍  Thanks for doing this!

pypeit/spectrographs/gemini_gnirs.py
kbwestfall: This is consistent with the previous behavior: the code would find both.  At least now, users can limit the search to one or the other using the `--extension` command-line argument.  I added some text in the setup doc about this.

pypeit/io.py
kbwestfall: I'd rather not over-complicate it.  I think it's enough to warn people that the code will potentially find both compressed and uncompressed files if they're in the same directory, which has always been true.

kbwestfall:
I've started a dev-suite run for this branch.

pypeit/io.py
rcooke-ast: OK - sounds good to me! 👍

rcooke-ast:
The tests are passing (except the HIRES failures), fixed in #1833.
image

PR #1840, opened 2024-08-13
Numpy Argsort fix

debora-pe:
(cherry picked from commit e6abfcd5e66d9777c680fd22165721c963ba6dbd)

This fixes how we use numpy.argsort. It currently uses the quicksort method, which is not good in many cases. these changes allows for a more stable method.

This changes were originally in PR #1820 , but since that PR is taking awhile to clear, I think it would be good to merge this as soon as possible.


kbwestfall (approved):
Nice!  Please add a comment to the release doc, and then we can merge asap!

debora-pe:
> Nice! Please add a comment to the release doc, and then we can merge asap!

Thanks @kbwestfall for the reminder. Just added.

badpandabear (approved):

debora-pe:
merging

PR #1841, opened 2024-08-16
Adding orderStacks as a default coadd.py output for Echelle spectrographs

Adolfo1519:
OrderStacks has been created to contain the order stacks output by coadd.py. Originally these were output as part of the onespec object when the -m flag was set, following the NIRSPEC merge. However, this has now been made into a default output for echelle spectrographs and the stacks are output in the original wavelength solution of each order (if it is monotonic).


pypeit/core/coadd.py
```diff
@@ -2726,6 +2728,13 @@ def ech_combspec(waves_arr_setup, fluxes_arr_setup, ivars_arr_setup, gpms_arr_se
     ivars_scale_arr_setup = utils.setup_list_to_arr_setup(ivars_scale_setup_list, norders, nexps)
     out_gpms_arr_setup = utils.setup_list_to_arr_setup(out_gpms_setup_list, norders, nexps)
 
+    #debugging for order stacks
+    #msgs.info(f'nexps = {nexps}')
+    #msgs.info(f'norders = {norders}')
+    #msgs.info(f'Shape of setups is {np.shape(waves_setup_list[0])}')
+    #msgs.info(f'median of first two orders of wave setup list is {np.median(waves_setup_list[0][0])} and {np.median(waves_setup_list[0][1])}')
+    
+#    msgs.info(f'Shape of setups is {np.shape(waves_setup_list)}')
```
kbwestfall: Remove?

pypeit/core/coadd.py
```diff
@@ -2738,9 +2747,32 @@ def ech_combspec(waves_arr_setup, fluxes_arr_setup, ivars_arr_setup, gpms_arr_se
...
+                #_, dloglam_data, _, _ = wvutils.get_sampling(wave_grid_ord)
+                #wave_grid_ord, wave_grid_mid_ord, _ = wvutils.wavegrid(np.min(wave_grid_ord), np.max(wave_grid_ord), dloglam_data,
+                #                    spec_samp_fact=spec_samp_fact, log10=True)
+                wave_grid_ord = np.linspace(np.min(wave_grid_ord), np.max(wave_grid_ord), len(wave_grid_ord))
+                wave_grid_diff_ord = np.diff(wave_grid_ord)
+                wave_grid_diff_ord = np.append(wave_grid_diff_ord, wave_grid_diff_ord[-1])
+                wave_grid_mid_ord = wave_grid_ord + wave_grid_diff_ord / 2.0
+                # removing the last bin since the midpoint now falls outside of wave_grid rightmost bin. This matches
+                # the convention in wavegrid above
+                wave_grid_mid_ord = wave_grid_mid_ord[:-1]
```
kbwestfall: Three things here:

1. I think you can trim this because there are a number of repeated lines.  E.g., I think you can start with:
```
wave_grid_ord = waves_setup_list[isetup][ind_start]
if np.any(np.diff(wave_grid_ord) < 0):
    msgs.warn(f'This order ({iord}) has a non-monotonic wavelength solution. Resampling.')
    wave_grid_ord = np.linspace(np.min(wave_grid_ord), np.max(wave_grid_ord), len(wave_grid_ord))
wave_grid_diff_ord = np.diff(wave_grid_ord)
...
```
That should limit the repeated lines.

2. I'm unsure whether you need the `append` line, given that you remove the last element later.

3. I would remove any commented code.

pypeit/core/flux_calib.py
```diff
@@ -29,6 +29,7 @@
 from pypeit.core import fitting
 from pypeit.core import wave
 from pypeit import dataPaths
+from pypeit import data
```
kbwestfall: Does this file need this module?  As far as I can tell, this should be removed.

pypeit/core/wavecal/autoid.py
```diff
@@ -1111,8 +1111,8 @@ def full_template(spec, lamps, par, ok_mask, det, binspectral, nsnippet=2, slit_
         lines_wav = template_dict['lines_wav'] 
         lines_fit_ord = template_dict['lines_fit_ord']
 
-    temp_wv = temp_wv_og
-    temp_spec = temp_spec_og
+    temp_wv = np.copy(temp_wv_og)
+    temp_spec = np.copy(temp_spec_og)
```
kbwestfall, line 1115: Good catch!

pypeit/core/wavecal/wvutils.py
```diff
@@ -281,6 +281,10 @@ def get_wave_grid(waves=None, gpms=None, wave_method='linear', iref=0, wave_grid
             wave_grid = np.power(10.0,newloglam)
 
         elif wave_method == 'iref': # Use the iref index wavelength array
+            msgs.info(f'iref for the list is set to {iref}')
+            msgs.info(f'The shape of the list is: {np.shape(waves)}')
+            msgs.info(f'shape of the first wave_grid in the list is: {np.shape(waves[iref])}')
+            msgs.info(f'The median wavelength in orders 0, 37, 74 is {np.median(waves[0])}, {np.median(waves[0])}, {np.median(waves[0])}')
```
kbwestfall: Is the last statement above specific to NIRSPEC?  I would either generalize or remove.

pypeit/sensfunc.py
```diff
@@ -801,6 +801,8 @@ def write_QA(self):
             wave_gpm = self.sens['SENS_FLUXED_STD_WAVE'][iorddet] > 1.0
             model_flux_sav[iorddet][wave_gpm] = model_interp_func(self.sens['SENS_FLUXED_STD_WAVE'][iorddet][wave_gpm])
 
+        self.sens['SENS_STD_MODEL_FLAM'] = model_flux_sav
+
```
kbwestfall, line 745: Good catch!

kbwestfall (changes requested):
Thanks!  I have some code clean-up requests.

pypeit/core/coadd.py
Adolfo1519: good suggestion, thanks! I've implemented it. The append is still necessary, I think, so that the two grids being added have the same length.

pypeit/core/coadd.py
Adolfo1519: Done!

pypeit/core/flux_calib.py
Adolfo1519: I don't think so. I left it to avoid breaking something in my previous PR.

pypeit/core/wavecal/autoid.py
Adolfo1519: Thanks!

pypeit/core/wavecal/wvutils.py
Adolfo1519: It must be... I am happy to remove it.

pypeit/sensfunc.py
Adolfo1519: Thanks!

debora-pe (commented):
Hey @Adolfo1519 , the changes here look OK. 
I am just confused about the description of the PR. It says "Adding orderStacks as a default coadd.py output for Echelle spectrographs", but I cannot tell which part of the code here is doing that. Was it done in a previous PR and this is an update to that?
Also, if the orderstack is now provided by default for echelle, what happened to the `-m` flag? Was it removed? 
Sorry if I missed something in previous PRs.

kbwestfall (approved):
Approving.

debora-pe (approved):

PR #1845, opened 2024-08-29
Move flexure diagnostic code to class methods

kbwestfall:
As titled.  This is directed into `order_sync`, #1835, because of a number of other doc fixes that I added to that branch.

This deals with a circular import that I was getting when compiling the docs.


debora-pe (approved):
Thank you for taking care of this and sorry for giving you more work!
I fixed a couple of things. All good to be merged.

profxj (approved):
thx thx

PR #1851, opened 2024-09-03
Enable Spectrograph base class to account for image sections that include on-chip binning

kbwestfall:
As titled.

This is to address the needs for P200/DBSP and LDT/DeVeny, without requiring each to have their own reader.

### DIFF
--- a/pypeit/spectrographs/spectrograph.py
+++ b/pypeit/spectrographs/spectrograph.py
@@ -1251,7 +1251,7 @@ class Spectrograph:
         # Always assume normal FITS header formatting
         one_indexed = True
         include_last = True
-        required_dim = 2
+        require_dim = 2
 
         # Read the image(s)
         raw_img = [None]*nimg
@@ -1285,8 +1285,7 @@ class Spectrograph:
                     if image_sections is not None:  # and image_sections[i] is not None:
                         # Convert the (FITS) data section from a string to a slice
                         datasec = parse.sec2slice(image_sections[j], one_indexed=one_indexed,
-                                                  include_end=include_last,
-                                                  require_dim=required_dim,
+                                                  include_end=include_last, require_dim=require_dim,
                                                   binning=binning_raw)
                         # Assign the amplifier
                         pix_img[datasec] = j+1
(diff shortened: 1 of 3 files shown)


pypeit/spectrographs/spectrograph.py
```diff
@@ -1254,26 +1269,20 @@ def get_rawimage(self, raw_file, det):
...
 
                 # Initialize the image (0 means no amplifier)
                 pix_img = np.zeros(raw_img[i].shape, dtype=int)
                 for j in range(detectors[i]['numamplifiers']):
 
                     if image_sections is not None:  # and image_sections[i] is not None:
-                        # Convert the data section from a string to a slice
+                        # Convert the (FITS) data section from a string to a slice
                         datasec = parse.sec2slice(image_sections[j], one_indexed=one_indexed,
                                                   include_end=include_last, require_dim=2,
```
finagle29: I think we want to have `require_dim=required_dim` here.

finagle29 (changes requested):
One small nitpick, but otherwise this LGTM!

I did a test reduction on a binned P200/DBSP dataset and it worked and looked good!

pypeit/spectrographs/spectrograph.py
kbwestfall: Thanks for the good catch!l Fixed.

pypeit/spectrographs/spectrograph.py
```diff
@@ -1225,15 +1234,26 @@ def get_rawimage(self, raw_file, det):
...
+            # a mosaic
+            if detectors[0]['specaxis'] == 1:
+                binning_raw = (',').join(binning.split(',')[::-1])
+            else:
+                binning_raw = binning
+
+        # Always assume normal FITS header formatting
+        one_indexed = True
+        include_last = True
+        require_dim = 2
```
tbowers7, line 1254: These are called out here, and passed later (unaltered) to `parse.sec2slice()`... why?

Are there any cases where these default values would not be correct?  If so, move these to be optional inputs to the method.  If not, hard-wire the values into the call to `parse.sec2slice()`.

tbowers7 (approved):
LGTM.  Approved, but one question about whether variables need to be called out.

Passes the sniff test vs 1.16.0 on LDT/DeVeny.

finagle29 (approved):

PR #1852, opened 2024-09-04
Specify `remote_host` in call to PypeItDataPath

tbowers7:
When caching from GitHub using the script `pypeit_cache_github_data`, an error was kicked complaining about "Remote host type None is not supported for package data caching."  This commit corrects the issue by specifying the `remote_host` in the instantiation of PypeItDataPath.

	modified:   pypeit/scripts/cache_github_data.py


kbwestfall (approved):
Thanks!

Underlying this issue is that I don't think we have a dev-suite test that tests this script.  I'll add an issue on this to the dev-suite repo.

finagle29 (approved):

PR #1855, opened 2024-09-11
Spatial flexure options

### LINKED ISSUE #1859: Manual spatial flexure correction for calibration frames
As title. The current implementation works only on science frames, and if a shift is provided for `arc`,`tilt`, etc frames, it would be ignored.

rcooke-ast:
This PR reworks a lot of the code that uses spatial flexure. The main change here is that instead of a single value, the spatial flexure is now stored for each slit edge. I haven't yet implemented the edge-by-edge spatial flexure adjustment. These updates are just to ensure that the code base is unaffected by the change from a single floating point value to a 2D array.

There is a matching PR in the dev-suite. I'm running tests now...


rcooke-ast:
OK - tests pass!
image

deprecated/find_objects.py
```diff
@@ -191,7 +191,7 @@ def illum_profile_spatial(self, skymask=None, trim_edg=(0, 0), debug=False):
     skymask_now = skymask if (skymask is not None) else np.ones_like(self.sciImg.image, dtype=bool)
     hist_trim = 0  # Trim the edges of the histogram to take into account edge effects
     gpm = self.sciImg.select_flag(invert=True)
-    slitid_img_init = self.slits.slit_img(pad=0, initial=True, flexure=self.spat_flexure_shift)
+    slitid_img_init = self.slits.slit_img(pad=0, initial=True, spat_flexure=self.spat_flexure_shift)
```
kbwestfall: My guess is this was because of a global find/replace, but should we be editing the deprecated code?

presentations/py/users.py
```diff
@@ -21,9 +21,9 @@ def set_fontsize(ax, fsz):
                  ax.get_xticklabels() + ax.get_yticklabels()):
         item.set_fontsize(fsz)
 
-user_dates = ["2021-03-11", "2022-04-29", "2022-11-07", "2022-12-06", "2023-06-08", "2023-06-29", "2023-07-11", "2023-09-03", "2023-10-13", "2023-12-01", "2023-12-15", "2024-02-22", "2024-03-21", "2024-04-09", "2024-05-02", "2024-05-19", "2024-06-06", "2024-06-10", "2024-08-20"]
+user_dates = ["2021-03-11", "2022-04-29", "2022-11-07", "2022-12-06", "2023-06-08", "2023-06-29", "2023-07-11", "2023-09-03", "2023-10-13", "2023-12-01", "2023-12-15", "2024-02-22", "2024-03-21", "2024-04-09", "2024-05-02", "2024-05-19", "2024-06-06", "2024-06-10", "2024-08-20", "2024-09-11"]
 user_dates = numpy.array([numpy.datetime64(date) for date in user_dates])
-user_number = numpy.array([125, 293, 390, 394, 477, 487, 506, 518, 531, 544, 551, 568, 579, 588, 596, 603, 616, 620, 643])
+user_number = numpy.array([125, 293, 390, 394, 477, 487, 506, 518, 531, 544, 551, 568, 579, 588, 596, 603, 616, 620, 643, 655])
```
kbwestfall: Thanks for continuing to track this!!

pypeit/core/flexure.py
```diff
@@ -55,14 +55,27 @@ def spat_flexure_shift(sciimg, slits, debug=False, maxlag = 20):
...
+            Science image
         slits (:class:`pypeit.slittrace.SlitTraceSet`):
+            Slits object
+        method (:obj:`str`, optional):
+            Method to use to calculate the spatial flexure shift. Options
+            are 'detector' (default), 'slit', and 'edge'. The 'detector'
+            method calculates the shift for all slits simultaneously, the
+            'slit' method calculates the shift for each slit independently,
+            and the 'edge' method calculates the shift for each slit edge
+            independently.
```
kbwestfall, line 55: I don't think we should add this argument if we don't actually use it yet.  Can we hold-off until it's actually implemented?

pypeit/core/flexure.py
```diff
@@ -1413,7 +1425,7 @@ def flexure_diagnostic(file, file_type='spec2d', flexure_type='spec', chk_versio
             failed.  If False, throw a warning only. Default is False.
 
     Returns:
-        :obj:`astropy.table.Table`, :obj:`float`: If file_type is 'spec2d' and
+        :obj:`astropy.table.Table`, :obj:`float`, None: If file_type is 'spec2d' and
```
kbwestfall: Technically, `None` isn't a type.  I would instead indicate in the description when None is returned.

pypeit/core/tracewave.py
```diff
@@ -851,7 +851,56 @@ def fit_tilts(trc_tilt_dict, thismask, slit_cen, spat_order=3, spec_order=4, max
...
+    Parameters
+    ----------
+    img_shape : :obj:`tuple`
+        Shape of the science image
+    slit_left : `numpy.ndarray`_
+        Left slit edge
+    slit_right : `numpy.ndarray`_
+        Right slit edge
+    thismask_science : `numpy.ndarray`_
+        Boolean mask for the science pixels in this slit (True = on slit)
```
kbwestfall, line 872: Given this, do you need `img_shape`?

pypeit/core/tracewave.py
```diff
@@ -863,31 +912,72 @@ def fit2tilts(shape, coeff2, func2d, spat_shift=None):
...
+    spec_eval : `numpy.ndarray`_, optional
+        1D array indicating how spectral pixel locations move across the
+        image. If spec_eval is provided, spat_eval must also be provided.
+    spat_shift : float, `numpy.ndarray`_, optional
         Spatial shift to be added to image pixels before evaluation
         If you are accounting for flexure, then you probably wish to
-        input -1*flexure_shift into this parameter.
+        input -1*flexure_shift into this parameter. Note that this
+        should either be a float or a 1D array with two elements
+        (and both of these elements must be equal).
```
kbwestfall: Is the fact that this can be a 2-element array a lead-in for what's to come?

pypeit/flatfield.py
```diff
@@ -1648,12 +1655,11 @@ def spectral_illumination(self, gpm=None, debug=False):
             gpm = self.rawflatimg.select_flag(flag='BPM', invert=True)
 
         # Obtain relative spectral illumination
+        spat_flexure = self.wavetilts.spat_flexure if self.wavetilts is not None else np.zeros((self.slits.nslits, 2))
```
kbwestfall, line 1666: Do we need to define this as an array of 0s when `wavetilts` is None, or can we set `spat_flexure = None`?

pypeit/images/pypeitimage.py
```diff
@@ -128,9 +128,12 @@ class PypeItImage(datamodel.DataContainer):
                  'exptime': dict(otype=(int, float), descr='Effective exposure time (s)'),
                  'noise_floor': dict(otype=float, descr='Noise floor included in variance'),
                  'shot_noise': dict(otype=bool, descr='Shot-noise included in variance'),
-                 'spat_flexure': dict(otype=float,
+                 'spat_flexure': dict(otype=np.ndarray, atype=np.floating,
```
kbwestfall, line 132: I think we need to update the datamodel version number.

pypeit/images/rawimage.py
```diff
@@ -667,8 +671,12 @@ def process(self, par, bpm=None, scattlight=None, flatimages=None, bias=None, sl
...
         # function checks that the slits exist if running the spatial flexure
         # correction, so no need to do it again here.
-        self.spat_flexure_shift = self.spatial_flexure_shift(slits, maxlag = self.par['spat_flexure_maxlag']) \
-                                    if self.par['spat_flexure_correct'] else None
+        self.spat_flexure_shift = self.spatial_flexure_shift(slits, method=self.par['spat_flexure_correct'],
+                                                                 maxlag=self.par['spat_flexure_maxlag']) \
+            if self.par['spat_flexure_correct'] != "none" else None
+
+        # self.spat_flexure_shift = self.spatial_flexure_shift(slits, maxlag=self.par['spat_flexure_maxlag']) \
+        #                             if self.par['spat_flexure_correct'] else None
```
kbwestfall: remove?

pypeit/images/rawimage.py
```diff
@@ -761,7 +769,7 @@ def _squeeze(self):
         return _det, self.image, self.ivar, self.datasec_img, self.det_img, self.rn2img, \
                 self.base_var, self.img_scale, self.bpm
 
-    def spatial_flexure_shift(self, slits, force=False, maxlag = 20):
+    def spatial_flexure_shift(self, slits, force=False, method="detector", maxlag=20):
```
kbwestfall: Same comment here about the method.

pypeit/metadata.py
```diff
@@ -1576,7 +1580,7 @@ def set_user_added_columns(self):
         if 'manual' not in self.keys():
             self['manual'] = ''
         if 'shift' not in self.keys():
-            self['shift'] = 0
+            self['shift'] = self.MASKED_VALUE
```
kbwestfall: I'm a little worried about this approach, but I don't have a better solution off the top of my head.

pypeit/par/pypeitpar.py
```diff
@@ -493,6 +499,13 @@ def valid_combine_methods():
         """
         return ['median', 'mean' ]
 
+    @staticmethod
+    def valid_spatial_flexure():
+        """
+        Return the valid methods for combining frames.
+        """
+        return ['none', 'detector', 'slit', 'edge']
```
kbwestfall: Same comment about the method.

pypeit/pypeit.py
```diff
@@ -595,7 +595,7 @@ def reduce_exposure(self, frames, bg_frames=None, std_outfile=None):
         # slitmask stuff
         if len(calibrated_det) > 0 and self.par['reduce']['slitmask']['assign_obj']:
             # get object positions from slitmask design and slitmask offsets for all the detectors
-            spat_flexure = np.array([ss.spat_flexure for ss in sciImg_list])
+            spat_flexure = [ss.spat_flexure for ss in sciImg_list]
```
kbwestfall: Does this need to be a list?

pypeit/pypeit.py
```diff
@@ -811,20 +811,32 @@ def objfind_one(self, frames, det, bg_frames=None, std_outfile=None):
             sciImg = bkg_redux_sciimg.sub(bgimg)
 
         # Flexure
-        spat_flexure = None
+        spat_flexure = np.zeros((self.caliBrate.slits.nslits, 2))  # No spatial flexure, unless we find it below
```
kbwestfall: Do we need to allocate the array?

kbwestfall (changes requested):
A few minor comments/requests from me.

deprecated/find_objects.py
rcooke-ast: Almost... it was a global find, but I manually replaced all instances. I figured that it's important to make it clear that this is a spatial flexure as opposed to a spectral flexure (even for those that look back at the code). Granted... it's not absolutely necessary in this case.

pypeit/core/flexure.py
rcooke-ast: This is a fair point, and I am happy to revert some changes, if needed. First let me check an alternative... if it's OK with you, can we postpone merging this PR into develop, until I direct another PR with the `slit` and `detector` into this current PR? I had intended to do it all together, but I think it's best to have a separate PR focus on the actual implementation of the new methods.

pypeit/core/flexure.py
rcooke-ast: Good call, I've updated the description 👍

pypeit/core/tracewave.py
rcooke-ast: Good catch - updated!

pypeit/core/tracewave.py
rcooke-ast: I realise now that this is a hack to get two different modes working. Instead of having two different types of inputs, I've consolidated this, since it only affected one other part of the code (in the flatfield).

pypeit/flatfield.py
rcooke-ast: I've updated this part of the code, when responding to your comment: https://github.com/pypeit/PypeIt/pull/1855/files#r1767740221

pypeit/images/pypeitimage.py
rcooke-ast: Indeed we do - done!

pypeit/images/rawimage.py
rcooke-ast: Yes, this can be removed - done!

pypeit/images/rawimage.py
rcooke-ast: Same response  :-)  Happy to remove it, but would prefer to direct a PR into this one before merging.

pypeit/metadata.py
rcooke-ast: I know... But we certainly can't keep the non-defined value to be 0. I' not sure if we can just set this to be `None`, instead. In the end, I picked a large enough value that shouldn't be realistic on any detector.

pypeit/par/pypeitpar.py
rcooke-ast: See previous responses.

pypeit/pypeit.py
rcooke-ast: So, this line is not to allocate the array, it's to set the default value. I could instead put it in the various if/else parts of the statement, but thought it was cleaner to set a `default` here.

pypeit/pypeit.py
rcooke-ast: While it doesn't have to be a list, I think it makes more sense for it to be a list. `ss.spat_flexure` is an array of shape `(nslits, 2,)`. We would otherwise be appending this onto a third axis, and the code this is used has lists as inputs. Overall, I find this a cleaner approach, but I'm always open to new suggestions/preferences!

rcooke-ast (commented):
Thanks for the review, @kbwestfall! I've made the changes to the PR, but I haven't changed the `method` parameter. I thought I'd check if it's OK to submit a separate PR (directed into this one) with the method changes, (and both need to be approved before either gets merged into develop).

pypeit/core/flexure.py
```diff
@@ -1442,11 +1454,23 @@ def flexure_diagnostic(file, file_type='spec2d', flexure_type='spec', chk_versio
                 return_flex = spec_flex
             # get and print the spatial flexure
             if flexure_type == 'spat':
-                spat_flex = allspec2D[det].sci_spat_flexure
-                # print the value
-                print(f'Spat shift: {spat_flex}')
+                spat_flexure = allspec2D[det].sci_spat_flexure
```
debora-pe: @rcooke-ast, just so you know, here there is a conflict with @kbwestfall's PR #1835
Kyle split and moved this method to `pypeit/spec2dobj.py` and `pypeit/specobjs.py`. You may want to pull those changes.

pypeit/flatfield.py
```diff
@@ -705,11 +710,12 @@ def build_waveimg(self):
         if self.wavetilts is None or self.wv_calib is None:
             msgs.error("Wavelength calib or tilts are not available.  Cannot generate wavelength image.")
         else:
-            flex = self.wavetilts.spat_flexure
-            slitmask = self.slits.slit_img(initial=True, flexure=flex)
-            tilts = self.wavetilts.fit2tiltimg(slitmask, flexure=flex)
+            spat_flexure = self.wavetilts.spat_flexure
+            left, right, msk = self.slits.select_edges(spat_flexure=spat_flexure)
```
debora-pe: Since `slit_img` below is generated with `initial=True`, should it be also here for `select_edges`?

pypeit/core/tracewave.py
```diff
@@ -851,7 +851,56 @@ def fit_tilts(trc_tilt_dict, thismask, slit_cen, spat_order=3, spec_order=4, max
...
+    # Evaluate the slit coordinates
+    _spec_eval = spec_img[thismask_science] / (nspec - 1)
+    flex_coord = (spat_img - slit_left[:, None]) / (slit_right - slit_left)[:, None]
+    # Linearly interpolate the spatial flexure over the slit coordinates
+    spat_eval = spat_img + _spat_flexure[0] + flex_coord * (_spat_flexure[1] - _spat_flexure[0])
+    _spat_eval = spat_eval[thismask_science] / (nspat - 1)
+    return _spec_eval, _spat_eval
+
+
+def fit2tilts(shape, coeff2, func2d, spec_eval=None, spat_eval=None):
```
debora-pe: It seems like that `shape` is only used if `spat_eval` and `spec_eval` are not provided. Should it be an optional parameter then?

pypeit/images/pypeitimage.py
```diff
@@ -782,9 +785,10 @@ def sub(self, other):
         # Spatial flexure
         spat_flexure = self.spat_flexure
         if other.spat_flexure is not None and spat_flexure is not None \
-                and other.spat_flexure != spat_flexure:
-            msgs.warn(f'Spatial flexure different for images being subtracted ({spat_flexure} '
-                      f'vs. {other.spat_flexure}).  Adopting {np.max(np.abs([spat_flexure, other.spat_flexure]))}.')
+                and np.array_equal(other.spat_flexure, spat_flexure):
+            msgs.warn(f'Spatial flexure different for images being subtracted. Adopting ' \
+                      f'the maximum spatial flexure of each individual edge.')
+            spat_flexure = np.maximum(spat_flexure, other.spat_flexure)
```
debora-pe: `np.maximum` here does not take into account the negative shifts. If an image has a shift of -1 and one of 0, the 0 would be used, which is not the maximum flexure.
`abs` should be use in the comparison, but the negative sign should be preserved when assigning `spat_flexure`

pypeit/par/pypeitpar.py
```diff
@@ -356,9 +356,15 @@ def __init__(self, trim=None, apply_gain=None, orient=None,
                                  '``slit_illum_relative=True`` in the ``flatfield`` parameter set!'
 
         # Flexure
-        defaults['spat_flexure_correct'] = False
-        dtypes['spat_flexure_correct'] = bool
-        descr['spat_flexure_correct'] = 'Correct slits, illumination flat, etc. for flexure'
+        defaults['spat_flexure_correct'] = 'none'
```
debora-pe: I must say that I am not a fan of the `'none'` parameter option. 
I wonder if we could change the name of the parameter to be `spat_flexure_corr_method` or something like that and use `'skip'` instead of `'none'`

pypeit/pypeit.py
```diff
@@ -811,20 +811,32 @@ def objfind_one(self, frames, det, bg_frames=None, std_outfile=None):
...
         manual_flexure = self.fitstbl[frames[0]]['shift']
-        if (self.objtype == 'science' and self.par['scienceframe']['process']['spat_flexure_correct']) or \
-                (self.objtype == 'standard' and self.par['calibrations']['standardframe']['process']['spat_flexure_correct']) or \
+        if (self.objtype == 'science' and self.par['scienceframe']['process']['spat_flexure_correct'] != "none") or \
+                (self.objtype == 'standard' and self.par['calibrations']['standardframe']['process']['spat_flexure_correct'] != "none") or \
                     manual_flexure:
-            if (manual_flexure or manual_flexure == 0) and not (np.issubdtype(self.fitstbl[frames[0]]["shift"], np.integer)):
-                msgs.info(f'Implementing manual flexure of {manual_flexure}')
-                spat_flexure = np.float64(manual_flexure)
+            if (manual_flexure != self.fitstbl.MASKED_VALUE) and np.issubdtype(self.fitstbl[frames[0]]["shift"], np.integer):
```
debora-pe: a `not` is missing. It should be:
`if (manual_flexure != self.fitstbl.MASKED_VALUE) and not np.issubdtype(self.fitstbl[frames[0]]["shift"], np.integer):`

pypeit/metadata.py
debora-pe: Not a fan of `-9999` in the pypeit file either. It seems to me that `0` would work in the same way, unless (I guess) somebody want to force a 0 shift...

debora-pe (changes requested):
A few comments from me.

pypeit/core/flexure.py
rcooke-ast: Thanks for spotting that! I've updated accordingly, and will redirect this PR to the `order_sync` branch 👍

pypeit/core/tracewave.py
rcooke-ast: You are right - I've made the changes and propagated this through the code.

pypeit/flatfield.py
rcooke-ast: Well spotted, again! They should indeed be the same. Updated.

pypeit/images/pypeitimage.py
rcooke-ast: I'm sure there's a fancier way of doing this, but I've updated this with the cleanest approach that I know  :-)

pypeit/par/pypeitpar.py
rcooke-ast: Both good suggestions - implemented, and propagated throughout.

pypeit/pypeit.py
rcooke-ast: I'm not quite sure about this. My understanding of the previous code is:

`(manual_flexure or manual_flexure == 0)`
if `manual_flexure` is anything other than `None`

`and not (np.issubdtype(self.fitstbl[frames[0]]["shift"], np.integer))`
and it's set to something other than an integer.

The first line seems like this is always true, since the previous default in `metadata.py` would set `shift=0` as the default (unset) value. I think the `MASKED_VALUE` clears this up. Then, the second line has me more confused why this is there... For example, the simple examples:
```
In [2]: not np.issubdtype(np.zeros(2, dtype=float).dtype, np.integer)
Out[2]: True

In [3]: not np.issubdtype(np.zeros(2, dtype=int).dtype, np.integer)
Out[3]: False
```
So, I think this `if` statement is only being triggered when the `shift` value is a float. I don't really understand why this is, but I assumed folks would input a integer pixel value (hence my change). Perhaps this should just be:

`if manual_flexure != self.fitstbl.MASKED_VALUE:`

pypeit/metadata.py
rcooke-ast: That's correct - if someone sets `manual_shift=0`, then the manual shift wouldn't be applied, because the `manual_shift` is set to the (hard-coded) masked value.

I am also not a huge fan, but I don't have a better solution. Although, I thought the `-9999` wouldn't appear in the `.pypeit` file, it's only something that's hidden in the code. Am I mistaken about this?

rcooke-ast (commented):
Thanks @debora-pe - I've responded to all comments, updated and rebuilt the docs. There's two points that still need to be considered:

* What to do with the metadata `MASKED_VALUE`. Are there better options?
* I'm not quite sure we're handling the `manual_flexure` correct, so would be great to get some clarity on this.

I'll re-run the dev suite to make sure I haven't broken anything obvious!

pypeit/metadata.py
debora-pe: The -9999 would appear in the `.pypeit` file if you run `pypeit_setup -f` (the `shift` column will be added). Currently that column would have all the values to 0 and then the user would change those values as they prefer. So technically, that would work the same with -9999, but I would prefer to not see -9999 in the `.pypeit` file, if possible. I wonder if we could just set the default to `None`? There are other metadata that are sometime `None`, so I don't think it should be a problem. In addition, the condition in `pypeit.py` would catch it and not use it anyway.

What do you guys think? @rcooke-ast @kbwestfall

pypeit/pypeit.py
debora-pe: Ok. I'm confused too. 
I think you are right, just checking that `self.fitstbl[frames[0]]['shift']` is not equal to `self.fitstbl.MASKED_VALUE` (or possibly `None`) should suffice.

BTW, it just occurred to me that, currently, the manual flexure can be only applied to the science/standard frame, since this piece of code is here and not in `rawimage.py`. This is a bit misleading because a `shift` column is added to the `.pypeit` file and the users could think that they can set the shift also for other frames. Sorry, this has nothing to do with your PR, but I think it's something that we need to fix at certain point.

pypeit/metadata.py
rcooke-ast: OK - I think I have a better solution -- let me know what you think!  Previously, we were getting rid of masked columns. So, if we reinstate masked columns in the `fitstbl`, then we can simply use masked values. This means that the `shift` column of the pypeit file will just be empty for masked values, and spatial flexure would only be applied to frames with non-empty (i.e. non-masked) values. I don't think this has any undesired effects... 😬

pypeit/images/rawimage.py
```diff
@@ -669,8 +676,12 @@ def process(self, par, bpm=None, scattlight=None, flatimages=None, bias=None, sl
         # bias and dark subtraction) and before field flattening.  Also the
         # function checks that the slits exist if running the spatial flexure
         # correction, so no need to do it again here.
-        self.spat_flexure_shift = self.spatial_flexure_shift(slits, debug=debug) \
-                                    if self.par['spat_flexure_correct'] else None
+        self.spat_flexure_shift = None
+        if self.par['spat_flexure_method'] != "skip" or not np.ma.is_masked(manual_spat_flexure):
+            self.spat_flexure_shift = self.spatial_flexure_shift(slits, method=self.par['spat_flexure_method'],
```
debora-pe: Hey @rcooke-ast . Just a little confusion here.
- `maxlag` is not a parameter of `self.spatial_flexure_shift()` anymore. `self.par['spat_flexure_maxlag']` is passed directly to `flexure.spat_flexure_shift()`
- you can do the same with `method`. Since it is defined in `self.par`, you can just pass it directly to `flexure.spat_flexure_shift()`

rcooke-ast: Ah, yes, good call!  I'll update it 👍

PR #1858, opened 2024-10-05
Add ginga Spec1dView plugin to view 1dspec files 

ejeschke:
Creates a new Ginga plugin that is distributed with PypeIt to view 1dspec files.

NOTE: this PR does *not* deprecate `linetools`, which is still used to look up line lists.  It would be fairly straightforward to make a new line list that is native to `PypeIt` and I might submit that as a future PR.

To test this you will need at least version 5.2.0 of Ginga (will be installed by default)

Then, install this PR branch. 

You can view your Science/1dspec files in one of two ways:

1) start ginga from the command line (e.g. ginga --loglevel=20 --stderr) and then drag your 1dspec file to the viewer panel.  This will open the table viewer.  From the "Plugins" or Operations menu, then choose "PypeIt"=>"Spec1dView".

2) run "pypeit_show_1dspec --ginga  <1dspec_file>


ejeschke:
A few tips for testing (documentation TBD!):

- holding the `Ctrl` key while zooming will constrain the zoom to the X axis
- holding the `Shift` key while zooming will constrain the zoom to the Y axis
- the magnifying glass icons are used to zoom or zoom_fit; the one with "X" only fits the X axis, the one with "Y" only fits the Y axis and the one with the square inside fits the whole plot
- once zoomed, scrollbars can be used to pan
- mouse wheel click can be used to set the pan position
- use the "Low" and "High" boxes for X and Y axes in the "Synopsis" panel to precisely set the view for either axis

ejeschke:
It should look something like this if it is working correctly:
Screenshot from 2024-10-04 15-37-38

pypeit/display/__init__.py
```diff
@@ -30,3 +30,9 @@ def setup_SlitWavelength():
                  module='ginga_plugins', klass='SlitWavelength',
                  ptype='global', workspace='right', start=False,
                  category='PypeIt', menu='SlitWavelength', tab='SlitWavelength')
+
+def setup_Spec1dView():
+    return Bunch(path=os.path.join(os.path.split(__file__)[0], 'ginga_spec1dview.py'),
+                 module='ginga_spec1dview', klass='Spec1dView',
+                 ptype='local', workspace='right', start=False,
```
kbwestfall, line 37: Can you clarify the difference between `ptype='local'` and `ptype='global'`?  The `setup_SlitWavelength` uses `global`; should we change that to `local`?

pypeit/display/display.py
```diff
@@ -666,3 +666,24 @@ def show_scattered_light(image_list, slits=None, wcs_match=True):
         if clear:
             clear = False
 
+
+def show_1dspec(filename, ext=0, masked=True, fluxed=False, extraction='OPT'):
```
kbwestfall, line 671: It doesn't look like any of the keyword arguments are used.  Is this a mistake or something left to a future PR?

pypeit/display/ginga_spec1dview.py
```diff
@@ -0,0 +1,380 @@
+# This is open-source software licensed under a BSD license.
+# Please see the file LICENSE.txt for details.
```
kbwestfall: Our license file is `LICENSE.rst`, but we (as a project) need to figure out our header and add it to all our source files (which we don't have at the moment).

pypeit/display/ginga_spec1dview.py
```diff
@@ -0,0 +1,380 @@
...
+"""
+Spec1dView is a plugin for the Ginga viewer that provides functionality for
+visualizing and analyzing 1D spectra from FITS files. The plugin allows users
+to plot spectra, identify spectral lines from various line lists, and
+customize the display according to different parameters.
+
+**Plugin Type: Local**
```
kbwestfall, line 7: Does it make sense to make these rst-style section headers, instead of just bold text?

pypeit/display/ginga_spec1dview.py
```diff
@@ -0,0 +1,380 @@
...
+from linetools.lists.linelist import LineList
+
+__all__ = ['Spec1dView']
+
+
+class Spec1dView(GingaPlugin.LocalPlugin):
+
+    def __init__(self, fv, fitsimage):
+        # superclass defines some variables for us, like logger
+        super(Spec1dView, self).__init__(fv, fitsimage)
```
kbwestfall: Nit-picking:  You don't need the arguments in super.  I.e.: `super().__init__(fv, fitsimage)` will work.

pypeit/display/ginga_spec1dview.py
```diff
@@ -0,0 +1,380 @@
...
+from pypeit import utils
+# TODO: need to make PypeIt LineList class to deprecate linetools
+from linetools.lists.linelist import LineList
+
+__all__ = ['Spec1dView']
+
+
+class Spec1dView(GingaPlugin.LocalPlugin):
+
+    def __init__(self, fv, fitsimage):
```
kbwestfall, line 62: Can you provide short docstrings for each method?

pypeit/scripts/show_1dspec.py
```diff
@@ -29,6 +29,8 @@ def get_parser(cls, width=None):
 #        parser.add_argument('--jdaviz', default=False, action='store_true',
 #                            help='Open the spectrum in jdaviz (requires specutils and jdaviz '
 #                                 'to be installed)')
+        parser.add_argument('--ginga', default=False, action='store_true',
+                            help='Open the spectrum in ginga')
```
kbwestfall, line 33: Nice!

kbwestfall (approved):
Looks good!  I haven't tested the code.  We should talk about ways that we can test it using the dev-suite, assuming it's possible.

I have a few requests (primarily docstrings), but I'm approving now.

pypeit/display/__init__.py
ejeschke: @kbwestfall, a local plugin can have an instance open for each channel in Ginga (a channel is an organizing device for a bunch of files that get cycled through a viewer in a window). A global plugin has only one instance open for all channels. So basically, a singleton instance for many viewers or one instance per viewer. Which you choose depends on how you are organizing your code and what it needs to do.  In this case, I'd prefer to track things by viewer, since we might have multiple plot viewers open. It's easier to do that by having it be a local plugin.

pypeit/display/ginga_spec1dview.py
ejeschke: Shall I just remove this comment then? Or change it to `LICENSE.rst`?

pypeit/display/display.py
ejeschke: Good catch--overlooked TBD, I will push an update to handle these.

pypeit/display/ginga_spec1dview.py
ejeschke: I believe they do become headers when processed as RST.  We use these docstrings in Ginga and they become headers in the Ginga manual. But maybe you have a different way of processing them?

ejeschke:
@debora-pe , I changed the selector to use extension names, as you suggested. Thanks for the tip.

pypeit/display/ginga_spec1dview.py
ejeschke: Added. LMK if need to expand upon them.

pypeit/display/ginga_spec1dview.py
ejeschke: Fixed

pypeit/display/ginga_spec1dview.py
ejeschke: Removed

pypeit/display/display.py
ejeschke: It now handles the command line parameters.

ejeschke:
@kbwestfall , I have addressed the issues you listed. It looks like some of the CI tests are failing due to unrelated issues?

To invoke this viewer with `pypeit_show_1dspec --ginga` does require the latest ginga. I am going to release that in the next couple of days as ginga v5.2.0.

debora-pe (approved):
Thanks @ejeschke!

ejeschke:
Ginga v5.2.0 released, so I updated the requirement in `setup.cfg`. This should complete the necessary changes.

pypeit/display/ginga_spec1dview.py
```diff
@@ -0,0 +1,482 @@
...
+        canvas.add(p1, tag='spectrum', redraw=False)
+
+        # optionally, plot error vs. wavelength
+        if self.settings.get('plot_error', False):
+            points = np.array((self.data.wave, self.data.sig)).T
+            p2 = self.dc.Path(points, linewidth=2, linestyle='solid',
+                              alpha=1.0, color='red')
+            canvas.add(p2, tag='error', redraw=False)
+
+        self.plot.set_titles(title="Spec1dView",
```
MNBrod: Is there value in setting this title to something that matches the filename/extension being plotted, rather than a static Spec1dViewer?

MNBrod:
No major comments - this look great!

This should be getting merged into `develop` though, right?

pypeit/display/ginga_spec1dview.py
kbwestfall: Sorry, I didn't realize these would be formatted as headings!  In other docs, we always use some sort of underlining characters; e.g., 
```
Plugin Type: Local
------------------
```

PR #1860, opened 2024-10-14
Spatial flexure reorganisation

rcooke-ast:
This delta PR moves the manual spatial flexure correction to the `RawImage`. The benefit of this approach is that manual spatial flexure can now be assigned for any frame (including calibrations). Previously, it was only science/standard/background frames that could have a manual flexure. Tests pass:

image

### DIFF
--- a/pypeit/calibrations.py
+++ b/pypeit/calibrations.py
@@ -771,8 +771,8 @@ class Calibrations:
 
         # get pixel flat frames
         pixel_frame = {'type': 'pixelflat', 'class': flatfield.FlatImages}
-        raw_pixel_files, pixel_cal_file, pixel_calib_key, pixel_setup, pixel_calib_id, detname \
-            = [], None, None, illum_setup, None, detname
+        raw_pixel_files, raw_pixel_index, pixel_cal_file, pixel_calib_key, pixel_setup, pixel_calib_id, detname \
+            = [], [], None, None, illum_setup, None, detname
         # read in the raw pixelflat frames only if the user has not provided a pixelflat_file
         if self.par['flatfield']['pixelflat_file'] is None:
             raw_pixel_files, raw_pixel_index, pixel_cal_file, pixel_calib_key, pixel_setup, pixel_calib_id, detname \
--- a/pypeit/calibrations.py
+++ b/pypeit/calibrations.py
@@ -392,7 +392,7 @@ class Calibrations:
         self.msarc = buildimage.buildimage_fromlist(self.spectrograph, self.det,
                                                     self.par['arcframe'], raw_files,
                                                     bias=self.msbias, bpm=self.msbpm,
-                                                    dark=self.msdark, calib_dir=self.calib_dir,
+                                                    dark=self.msdark, slits=self.slits, calib_dir=self.calib_dir,
                                                     setup=setup, calib_id=calib_id,
                                                     manual_spat_flexure=manual_spat_flexure)
         # Save the result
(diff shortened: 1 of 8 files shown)


pypeit/calibrations.py
```diff
@@ -381,13 +384,17 @@ def get_arc(self):
...
 
+        # If a manual spatial flexure is requested for any of the frames, collect it now
+        manual_spat_flexure = self.fitstbl.get_shifts(raw_index)
+
         # Otherwise, create the processed file.
         msgs.info(f'Preparing a {frame["class"].calib_type} calibration frame.')
         self.msarc = buildimage.buildimage_fromlist(self.spectrograph, self.det,
                                                     self.par['arcframe'], raw_files,
                                                     bias=self.msbias, bpm=self.msbpm,
                                                     dark=self.msdark, calib_dir=self.calib_dir,
```
debora-pe: Found a bug. Obviously not introduced here.

```suggestion
                                                    dark=self.msdark, slits=self.slits, calib_dir=self.calib_dir,
```

pypeit/calibrations.py
```diff
@@ -756,11 +775,12 @@ def get_flats(self):
             = [], None, None, illum_setup, None, detname
```
debora-pe: `raw_pixel_index` should be added here too, as an empty list (just for consistency)

pypeit/images/buildimage.py
```diff
@@ -160,7 +160,7 @@ def construct_file_name(cls, calib_key, calib_dir=None, basename=None):
 
 def buildimage_fromlist(spectrograph, det, frame_par, file_list, bias=None, bpm=None, dark=None,
                         scattlight=None, flatimages=None, maxiters=5, ignore_saturation=True,
-                        slits=None, mosaic=None, calib_dir=None, setup=None, calib_id=None):
+                        slits=None, mosaic=None, manual_spat_flexure=None, calib_dir=None, setup=None, calib_id=None):
```
debora-pe, line 163: add `manual_spat_flexure` description in the docstring

pypeit/images/rawimage.py
```diff
@@ -401,7 +401,7 @@ def build_rn2img(self, units='e-', digitization=False):
         return np.array(rn2)
 
     def process(self, par, bpm=None, scattlight=None, flatimages=None, bias=None, slits=None, dark=None,
-                mosaic=False, debug=False):
+                mosaic=False, manual_spat_flexure=None, debug=False):
```
debora-pe, line 404: add `manual_spat_flexure` description in the docstring

debora-pe (approved):
@rcooke-ast Thank you for doing this.
I tested the code a bit and it looks good. I only found a bug, which was not introduced here, but your code exposed it! And I have a few more comments/requests.
Please take a look at them, but I am approving anyway.

pypeit/calibrations.py
rcooke-ast: Agreed - good catch

pypeit/images/buildimage.py
rcooke-ast: Thanks, added 👍

pypeit/images/rawimage.py
rcooke-ast: Also done  :-)

rcooke-ast (commented):
Thanks for the review @debora-pe, and for catching that bug!  All comments addressed, and merging.

PR #1863, opened 2024-10-23
HOTFIX to slitless pixflat

debora-pe:
Set `slit_illum_finecorr = False` when computing the slitless pixelflat


bpholden (approved):

rcooke-ast (approved):
Good call 👍

PR #1876, opened 2024-11-28
Fixes the WCS used for weighted combination

rcooke-ast:
The number of RA points used in the white light image generated when calculating the weights was incorrect. This is now fixed.

Running tests...


kbwestfall (approved):
Looks fine to me.  Can you add a new test?

rcooke-ast:
Tests pass... There's a corresponding [dev suite PR](https://github.com/pypeit/PypeIt-development-suite/pull/340) as well.

image

rcooke-ast:
> Can you add a new test?

Thanks for the suggestion @kbwestfall. I'm quite keen to add tests for KCWI, to avoid breaking things I've put in the code long ago! But, I don't think this is really something where a test would help. The issue was that arcsec^2 was a negative number (because the WCS is flipped E-W), and the number of bins was incorrectly set.

debora-pe (approved):
Thanks @rcooke-ast

PR #1883, opened 2024-12-18
HOTFIX :: Datacube variance should be squared, not cubed

rcooke-ast:
As pointed out by @jhennawi, the cubed term is likely wrong, and has been changed to a squared term. This checks out based on some simple checks.


kbwestfall (approved):
Without digging too deeply, this looks right to me.

profxj (approved):
happy to approve

jhennawi (approved):
Thanks @rcooke-ast

PR #1889, opened 2025-01-16
Remove Spectrograph `idname()` method

tbowers7:
The base class method raises a `NotImplementedError`, but only Keck/Deimos actually implements this method -- no other spectrograph does.

All frame types are defined in `pypeit/core/framematch.py`, and individual spectrograph classes now match raw frames with types from this list.

Since the `idname()` method is unused and unneeded, remove it from the base class and Keck/Deimos.


kbwestfall (approved):
Thanks for cleaning this up!

badpandabear (approved):

Issue #1890, opened 2025-01-21
Manual Extraction on Coadd2d Spectra: Unexpected AttributeError in numpy.fmod with dtype=object Arrays


jasonpeng17:
When I attempted to manual extraction on my coadd2d spectra from Pypeit 1.17.1, I got the following error message:

```
Traceback (most recent call last):
 File "/Users/zixuanpeng/opt/anaconda3/envs/pypeit/bin/pypeit_coadd_2dspec", line 8, in <module>
  sys.exit(CoAdd2DSpec.entry_point())
       ^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/Users/zixuanpeng/opt/anaconda3/envs/pypeit/lib/python3.11/site-packages/pypeit/scripts/scriptbase.py", line 114, in entry_point
  cls.main(cls.parse_args())
 File "/Users/zixuanpeng/opt/anaconda3/envs/pypeit/lib/python3.11/site-packages/pypeit/scripts/coadd_2dspec.py", line 196, in main
  = coadd.reduce(pseudo_dict, show=args.show, show_peaks=args.peaks, basename=basename)
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/Users/zixuanpeng/opt/anaconda3/envs/pypeit/lib/python3.11/site-packages/pypeit/coadd2d.py", line 834, in reduce
  global_sky_pseudo, sobjs_obj = objFind.run(show_peaks=show or show_peaks, show_skysub_fit=show_skysub_fit)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/Users/zixuanpeng/opt/anaconda3/envs/pypeit/lib/python3.11/site-packages/pypeit/find_objects.py", line 351, in run
  sobjs_obj, self.nobj = self.find_objects(self.sciImg.image, self.sciImg.ivar,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/Users/zixuanpeng/opt/anaconda3/envs/pypeit/lib/python3.11/site-packages/pypeit/find_objects.py", line 442, in find_objects
[...]

PR #1894, opened 2025-01-31
Fix for manual extraction in coadd 2D

kbwestfall:
Addresses #1890 .

There were two underlying issues.

1. The use of `scipy.interpolate.RectBivariateSpline.__call__` function expects array-like inputs.  If you give it single values, it will return things like `array(0.43390805)`, which was presumably the reason why the code was explicitly performing the conversion to a single float.  I've changed the code so that it now passes array-like objects (lists) to the call function and selects the single element returned.  See the change in `findobj_skymask.py`.
2. Attributes of `SpecObjs` are pooled together from its component `SpecObj` using the `specobjs.lst_to_array` function.  This is the root cause of why the `SPAT_FRACPOS` attribute was an array with `dtype=object`.  The `lst_to_array` function wasn't smart enough to know that a list with mixed numpy and native float types can be converted to a numpy array without having to set `dtype=object`.  I've updated the function so that it can do this conversion by leaning on the fact that numpy *is* smart enough.  The `try` block is a much cleaner way of doing the conversion by leaning on numpy to make the correct casting, but the try block is still necessary to accommodate lists that are a mix of arrays and `None` values.

Unit tests pass.  I'll set off a dev-suite run today.


profxj (approved):
nice fix

debora-pe (approved):
Thanks @kbwestfall . These are very good fixes.

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  259 passed, 6579 warnings in 589.06s (0:09:49) ---
--- PYTEST UNIT TESTS PASSED  152 passed, 136953 warnings in 1075.07s (0:17:55) ---
--- PYTEST VET TESTS PASSED  66 passed, 108215 warnings in 6274.60s (1:44:34) ---
Testing Started at 2025-02-05T22:05:06.951486
Testing Completed at 2025-02-06T20:43:34.134317
Total Time: 22:38:27.182831
```

Will merge once CI finishes.

PR #1898, opened 2025-02-06
HOTFIX to sensfunc fluxed std QA

debora-pe:
The QA plot for the fluxed standard star, which is generated as a check when the sensitivity function is computed, was showing some inconsistency in the comparison with the true (archived) flux. The reason for this was that the flux was not telluric corrected. Now the telluric correction is applied. See attached example before (left) and after (right) these changes.
Screen Shot 2025-02-06 at 11 19 34 AM


kbwestfall (approved):
Looks good!  Also add a comment in the release doc 😄

jhennawi (approved):
Thanks @debora-pe

PR #1906, opened 2025-03-05
Update packaging infrastructure

kbwestfall:
Updates the packaging approach to remove the `setup.cfg` file and use only the `pyproject.toml` file.  We still require the `setup.py` file to build the C components of the code.

This also updates the development doc to reflect these (and a few other) changes.

Note this is currently a PR into the branch that removes the `misc` directories that have now been moved to their own repository.


badpandabear (approved):
Looks good. I'm wondering if we should update our development status to "Production/Stable" in pyproject.toml but that's probably a discussion for another time.

tepickering (approved):
looks good to me! i'm going to borrow some of those `tomllib` tricks.

pypeit/scripts/chk_plugins.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+
+"""
+Check that the installation of python is C enabled.
```
debora-pe, line 4: You may want to change this description

debora-pe (approved):
Thanks @kbwestfall

PR #1907, opened 2025-03-05
Adds SECURITY.md

kbwestfall:
As titled.  These have a standard format, and I modeled ours after:

https://github.com/matplotlib/matplotlib/blob/main/SECURITY.md
https://github.com/astropy/astropy/blob/main/SECURITY.md

Many repos (notably numpy and scipy) do not have these...

I think the pypeit@ucolick.org e-mail not longer exists, but I'm planning to set it up again.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,25 @@
+
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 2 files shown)


profxj (approved):
Thanks!

badpandabear (approved):

PR #1915, opened 2025-03-31
p200_ngps

FrankVerdi:
Added support for the Next Generation Palomar Spectrograph (NGPS)


pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -226,15 +226,13 @@ def config_specific_par(self, scifile, inp_par=None):
         # Start with instrument wide
         par = super().config_specific_par(scifile, inp_par=inp_par)
 
-        if (self.get_meta_value(scifile, 'dispname')=='HK') and (self.get_meta_value(scifile, 'dichroic')=='zJ'):
+        if (self.get_meta_value(scifile, 'SPEC_ID')=='R'):
```
kbwestfall: Were these edits meant for NGPS?

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+from pypeit import msgs
+from pypeit import io
+from pypeit import telescopes
+from pypeit.core import framematch
+from pypeit.spectrographs import spectrograph
+from pypeit.core import parse
+from pypeit.images import detector_container
+
+
+def flip_fits_slice(s: str) -> str:
```
kbwestfall: Can you put this into `pypeit/core/parse.py` and update both this file and `p200_dbsp.py` to use one function?

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+            return retval
+        
+        if meta_key == 'mjd':
+            return Time(headarr[0]['UTSHUT']).mjd
+        elif meta_key == 'dispangle':
+            try:
+                return 0
+            except Exception as e:
+                msgs.warn("Could not read dispangle from header:" + msgs.newline() + str(headarr[0]['ANGLE']))
+                raise e
```
kbwestfall: You don't need the try block.  `return 0` should always work.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+        # If there is no target keyword, return image type
+        elif meta_key == 'target':
+            if headarr[0]['TARGET'] is not None:
+                return headarr[0]['TARGET']
+            else:
+                return headarr[0]['IMGTYPE']
+        elif meta_key == 'dichroic': 
+            return None
+        else:
+            msgs.error("Not ready for this compound meta: ", meta_key)
```
kbwestfall: Change to:
```
msgs.error(f"Not ready for this compound meta: {meta_key}")
```
There is a second argument, but it should be a string with the type of class to use for the error.  If you copied this from somewhere else in the code, we should fix it there, too!

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+            numamplifiers   = 1, 
+            gain            = np.atleast_1d(2.8),
+            ronoise         = np.atleast_1d(8.5),
+            datasec         = np.atleast_1d(flip_fits_slice(hdu[1].header['DATASEC'])),
+            oscansec        = np.atleast_1d(flip_fits_slice(hdu[1].header['BIASSEC']))
+        )
+
+        # Instantiate
+        detector1 = detector_container.DetectorContainer(**detector_dict1)
+        return detector1
```
kbwestfall: Skip the middle man: `return detector_container.DetectorContainer(**detector_dict1)`

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+            all of PypeIt methods.
+        """
+        par = super().default_pypeit_par()
+
+        par['calibrations']['slitedges']['sync_predict'] = 'nearest'
+        par['calibrations']['slitedges']['edge_thresh'] = 10. # Lower edge tracing thresdhold to catch leftmost slit
+
+        par['calibrations']['slitedges']['minimum_slit_length'] = 100 # Set minimum slit length 
+
+        par['calibrations']['slitedges']['add_slits'] = ['1:2090:24:153']
```
kbwestfall: You _always_ need to add a slit?

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+        par['scienceframe']['process']['sigclip'] = 4.0
+        par['scienceframe']['process']['objlim'] = 5.0
+
+        # Make a bad pixel mask
+        par['calibrations']['bpm_usebias'] = True
+
+        # Set pixel flat combination method
+        par['calibrations']['pixelflatframe']['process']['combine'] = 'median'
+
+        par['calibrations']['wavelengths']['lamps'] = ['ThAr'] # FeAr and ThAR lamps for NGPS (ThAr)
```
kbwestfall, line 300: I think this means that you will be the `ThAr_lines.dat` file by default.  Do you need the `ThAr_NGPS_lines.dat` file?

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+        par['calibrations']['wavelengths']['method'] = 'full_template' # Use wavelength template
+        par['calibrations']['wavelengths']['reid_arxiv'] = 'wvarxiv_p200_ngps_20250131T1227.fits' # R Channel
+
+        par['calibrations']['wavelengths']['rms_thresh_frac_fwhm'] = 1.0 
+
+        # Do not flux calibrate
+        #par['fluxcalib'] = None
+        par['sensfunc']['algorithm'] = 'UVIS' # CHANGE #
```
kbwestfall: Does this need to change (i.e., based on the comment)?

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+
+        # Set the default exposure time ranges for the frame typing
+        par['calibrations']['biasframe']['exprng'] = [None, 0.001]
+        par['calibrations']['darkframe']['exprng'] = [999999, None]     # No dark frames
+        par['calibrations']['pinholeframe']['exprng'] = [999999, None]  # No pinhole frames
```
kbwestfall: I don't think these two lines are necessary, given that `check_frame_type` skips these two frame types anyway.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+                use :func:`default_pypeit_par`.
+
+        Returns:
+            :class:`~pypeit.par.parset.ParSet`: The PypeIt parameter set
+            adjusted for configuration specific parameter values.
+        """
+        # Start with instrument wide
+        par = super().config_specific_par(scifile, inp_par=inp_par)
+
+        return par
```
kbwestfall: If you're always using the base class function, you can remove it from this class.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+            0.
+        """
+
+        # Call the base-class method to generate the empty bpm
+        bpm_img = super().bpm(filename, det, shape=shape, msbias=msbias)
+
+        #msgs.info("Using hard-coded BPM for NGPS") # Can create actual BPM later if necessary
+        #bpm_img[:, 0] = 1
+
+        return bpm_img
```
kbwestfall: Same comment here about always using the base class function.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+            numamplifiers   = 1, # Updated
+            gain            = np.atleast_1d(2.8),
+            ronoise         = np.atleast_1d(8.5),
+            datasec         = np.atleast_1d(flip_fits_slice(hdu[2].header['DATASEC'])),
+            oscansec        = np.atleast_1d(flip_fits_slice(hdu[2].header['BIASSEC']))
+        )
+
+        # Instantiate
+        detector2 = detector_container.DetectorContainer(**detector_dict2)
+        return detector2
```
kbwestfall: Skip the middle man.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+            all of PypeIt methods.
+        """
+        par = super().default_pypeit_par()
+
+        par['calibrations']['slitedges']['sync_predict'] = 'nearest'
+        par['calibrations']['slitedges']['edge_thresh'] = 10. # Lower edge tracing thresdhold to catch leftmost slit
+       
+        par['calibrations']['slitedges']['minimum_slit_length'] = 100 # Set minimum slit length 
+
+        par['calibrations']['slitedges']['add_slits'] = ['1:2035:123:251']
```
kbwestfall: Same question here about always adding a slit.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+        par['calibrations']['wavelengths']['rms_thresh_frac_fwhm'] = 1.0 
+
+        # Do not flux calibrate  
+        #par['fluxcalib'] = None
+        par['sensfunc']['algorithm'] = 'UVIS' # CHANGE #
+
+        # Set the default exposure time ranges for the frame typing
+        par['calibrations']['biasframe']['exprng'] = [None, 0.001]
+        par['calibrations']['darkframe']['exprng'] = [999999, None]     # No dark frames
+        par['calibrations']['pinholeframe']['exprng'] = [999999, None]  # No pinhole frames
```
kbwestfall: same comments here about not needing the two lines above.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+                use :func:`default_pypeit_par`.
+
+        Returns:
+            :class:`~pypeit.par.parset.ParSet`: The PypeIt parameter set
+            adjusted for configuration specific parameter values.
+        """
+        # Start with instrument wide
+        par = super().config_specific_par(scifile, inp_par=inp_par)
+
+        return par
```
kbwestfall: Same comment here about removing this function.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,619 @@
...
+            0.
+        """
+
+        # Call the base-class method to generate the empty bpm
+        bpm_img = super().bpm(filename, det, shape=shape, msbias=msbias)
+
+        #msgs.info("Using hard-coded BPM for NGPS") 
+        #bpm_img[:, 0] = 1
+
+        return bpm_img
```
kbwestfall: And here.

kbwestfall (changes requested):
Thanks for submitting the PR!  I have a number of comments for you.

Two other issues:

1. Do you need the `ThAr_NGPS_lines.dat` file?  It looks like it has the same length as the `ThAr_lines.dat` file, and it doesn't look like you use the new file by default.
2. I assume that deleting the Keck/HIRES pixel flat was a mistake?  Can you restore that file so that it is not deleted in this PR?

We also need you to add test data to the dev-suite, and we're happy to help you through that process.

pypeit/spectrographs/p200_ngps.py
kbwestfall: Thanks!

kbwestfall:
Thanks for addressing the comments!

I'm still confused about why you need to always add a slit.  Is this slit always missed?  Do you know why the automated tracing algorithm is not finding it?

Thanks for the fix to the ThAr_lines.data file!

kbwestfall:
Sorry, two more things:

 - Please merge the `develop` branch of the main repo into this branch of your fork.  I recently added some guidance on this here: https://pypeit.readthedocs.io/en/develop/dev/development.html#installing-and-working-with-your-fork
 - Then redirect this PR to the `pypeit:develop` branch.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,485 @@
...
+        :func:`pypeit.spectrographs.spectrograph.Spectrograph.get_rawimage` that
+        sets ``sec_includes_binning`` to True.  See the base-class function for
+        the detailed descriptions of the input parameters and returned objects.
+        """
+
+        return super().get_rawimage(raw_file, det=1, sec_includes_binning=True)
+    
+
+
+    def compound_meta(self, headarr: List[fits.Header], meta_key: str):
```
debora-pe, line 185: It seems to me that `compound_meta()` is the same for both `p200_ngps_R` and `p200_ngps_I`. If yes, it would be better to put it in the superclass `P200NGPSSpectrograph` (unless you foresee it would not be true when the other 2 channels will be added). This helps to have less code to maintain.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,485 @@
...
+        if meta_key == 'mjd':
+            return Time(headarr[0]['UTSHUT']).mjd
+        elif meta_key == 'dispangle':
+            return 0
+            
+        elif meta_key == 'binning':
+            # Always the same binning for DET01 and DET02    
+            binspat = headarr[1]['BINSPAT'] 
+            binspec = headarr[1]['BINSPEC']
+            return f"{binspec},{binspat}"
```
debora-pe: The general PypeIt-way of returning the binning here is: `return parse.binning2string(binspec, binspat)`

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,485 @@
...
+        
+        # If there is no target keyword, return image type
+        elif meta_key == 'target':
+            if headarr[0]['TARGET'] is not None:
+                return headarr[0]['TARGET']
```
debora-pe, line 220: Do you mean here that the key `TARGET` may not exist at all or that it may be None? Because if it does not exist at all in the header, I think `headarr[0]['TARGET']` will raise an error. In that case you should use:
`if headarr[0].get('TARGET') is not None:`

debora-pe (changes requested):
Thanks @FrankVerdi for this. I only have a few requested changes, in addition to the point raised by Kyle about adding a slit.
I also have noticed that the wavelength templates that you generated are not included in this PR. Don't remember how everything works with the cache system. @kbwestfall do you mind telling Frank where the templates should be added.

Last thing, yes, we'll need to have a test data (one per each channel) in the dev-suite.
Thanks!

kbwestfall:
They finished a while ago, but tests pass (apart from the known failures):

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  259 passed, 6581 warnings in 501.78s (0:08:21) ---
--- PYTEST UNIT TESTS PASSED  152 passed, 136953 warnings in 979.82s (0:16:19) ---
--- PYTEST VET TESTS PASSED  66 passed, 108215 warnings in 6122.58s (1:42:02) ---
--- PYPEIT DEVELOPMENT SUITE FAILED 2/263 TESTS  ---
Failed tests:
    lbt_mods/MODS2R_Longslit pypeit
    lbt_mods/MODS1R_Longslit pypeit
Skipped tests:
Testing Started at 2025-05-22T17:40:56.671082
Testing Completed at 2025-05-23T10:40:21.294629
Total Time: 16:59:24.623547
```

kbwestfall (approved):
Approving, but please make sure you've addressed Debora's comments.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -0,0 +1,483 @@
...
+        retval = super().compound_meta(headarr, meta_key)
+        
+        # If superclass could not handle the meta key
+        if retval is not None:
+            return retval
+        
+        if meta_key == 'mjd':
+            return Time(headarr[0]['UTSHUT']).mjd
+        elif meta_key == 'dispangle':
+            try:
```
debora-pe: What do you mean to do here with this try/except? You don't really need to try to return 0.

debora-pe (approved):
Thanks @FrankVerdi for addressing my comments. I see that the compound_meta key `binning` is actually different between the R and I channels. Just for future knowledge, you could still use only one `compound_meta()` function in the superclass P200NGPSSpectrograph and update only the `binning` key in the subclasses.

I'm approving.

PR #1918, opened 2025-04-16
Update development docs to reflect migration to forks

kbwestfall:
As titled.  I also migrated all the developer-specific installation instructions from the main installation page to the developer page.

This is a doc-only PR.


badpandabear (approved):
These are great! I learned somethings from your updates.

I did notice a lot of warnings when I tried to build these docs but I assume you're already aware of those.

kbwestfall:
> I did notice a lot of warnings when I tried to build these docs but I assume you're already aware of those.

I checked and it seems to compile fine for me on python 3.13 with Sphinx 7.4.7.

badpandabear:
> I checked and it seems to compile fine for me on python 3.13 with Sphinx 7.4.7.

That was my mistake. I didn't re-install when I cloned your repo. I've confirmed it works on 3.13 and 3.12. Sorry.

kbwestfall:
No worries.  Thanks for checking!

debora-pe (approved):
@kbwestfall Thank you so much!!! I don't know how long it would have taken me to figure out all the fork stuff. This looks great!

PR #1922, opened 2025-04-25
Bump numpy and astropy dependencies and clean up tox environments

kbwestfall:
As titled.

We should also add a warning about the error in sigma clipping for astropy <= v6.1.6.  Any suggestions where?


tbowers7 (approved):
Onward into the NumPy 2+ only world.

tbowers7:
> We should also add a warning about the error in sigma clipping for astropy <= v6.1.6. Any suggestions where?

Per @debora-pe , the issue was only with v6.1.6, and was fixed with v6.1.7.  The present update requires `astropy>=7.0`, so this becomes a moot point?

https://docs.astropy.org/en/v6.1.7/changelog.html

debora-pe (approved):
Thanks @kbwestfall!
It's true that if the requirement is now that `astropy>=7.0`,the problem with <= v6.1.6 should not exist. I just wonder what happen if somebody uses an older version of PypeIt, it is worth to add a note in the docs for that case?

kbwestfall:
I added a short sentence to the known issues doc.  I agree with @debora-pe that it's worth having it there, at least for a while.

PR #1925, opened 2025-05-12
Telluric-related docs cleanup + minor TelluricPar tweaks

freddavies:
Cleaning up some old documentation that still referred to the old atmospheric grids, including various mentions of having to download the heavy files ahead of time, etc. which are no longer relevant (I wonder if we should just add the standard PCA files to github?). The old grid-based telluric method does still work with a small parameter tweak, and I'd rather not deprecate it entirely, but this way I think there will be less confusion. Would be nice to get some feedback to tighten it up a bit further.

I also went ahead and threw in a couple of tiny tweaks to the telluric functionality. First, I reduced the range of `pix_stretch_bounds` from `(0.9,1.1)` to `(0.98,1.02)` as it is unlikely for the wavelength solution to be so bad, and allowing the wider range has led to occasional problems (we discussed this in the developer Slack some time ago). Second, I have added `resln_frac_bounds` to the `qso/star/poly_telluric` calls in the tellfit script since I have run into a few cases where it was important to fiddle with. 

This second change may have more of an effect than it seems: the default bounds in the `qso/star/poly_telluric` function calls were actually somewhat wider (`(0.3,1.5)`) than the global default for the parameter (`(0.6,1.4)`), and the latter would only be used by default in the sensfunc script. For now I have not touched the global default but we should consider adjusting some instrument defaults if they tend to be oversampled (e.g. DEIMOS?), or maybe make `(0.3,1.5)` the global default. The afterburn tests which call both sensfunc and tellfit (gemini_gnirs_echelle 32_SB_SXD + gemini_gmos GS_HAM_R400_700) pass successfully and look great; happy to run more if requested.


doc/installing.rst
```diff
@@ -384,50 +384,36 @@ use the scripts below.
 
 .. _install_atmosphere:
 
-Atmospheric Model Grids
+Atmospheric Models
```
kbwestfall, line 387: Can you check that the length of the dashes below this line doesn't matter?  I think Sphinx complains if the length is too short, but it might not care if it's too long.  It should typically match the length of the title; i.e.
```
Atmospheric Models
------------------
```

pypeit/scripts/tellfit.py
```diff
@@ -166,6 +166,7 @@ def main(args):
                                            tell_norm_thresh=par['telluric']['tell_norm_thresh'],
                                            only_orders=par['telluric']['only_orders'],
                                            bal_wv_min_max=par['telluric']['bal_wv_min_max'],
+                                           resln_frac_bounds=par['telluric']['resln_frac_bounds'],
```
kbwestfall: Is this defined by `TelluricPar` in `par/pypeitpar.py`?

kbwestfall (changes requested):
Thanks, @freddavies !!  If you're willing, it would be great if you could add a little more to `telluric.rst`.  Specifically, I have 2 requests:

 - add a short description of how the telluric correction works
 - move/add anything relevant from the PR you referenced into the doc.  This will help us keep the doc up-to-date moving forward (i.e., we can update the doc, but not the PR)

Also, you updated the `tellfit` parameter help descriptions for the script, but actually the descriptions in the `TelluricPar` parameter set are arguably more important.  We need these to match (I think we should actually remove the script parameters in a future PR).

Lastly, my opinion is that we should minimize the times we explicitly give defaults in the docs.  The defaults are parsed by Sphinx directly from the function calling sequence into the API docs.  Anytime we define the defaults by hand means we have to keep the defaults consistent in multiple places.  (Stepping off my soap box now...)

doc/installing.rst
freddavies: There are no Sphinx warnings, but I can change this anyway. I've only seen warnings when the length of the dashes was _less_ than the length of the text.

pypeit/scripts/tellfit.py
freddavies: Yep, it's already there.

freddavies:
I hope I have addressed your comments @kbwestfall !

While I agree with your general sentiment regarding parameter defaults, I would prefer to keep the default number of PCA components mentioned in the documentation, as it is a key part of the model. I am happy to take on the responsibility of updating that number in case it changes in the future ;)

kbwestfall (approved):

debora-pe (commented):
Thank you @freddavies for all of this. Really appreciated.
I'm approving, but I have 2 small change requests that it would be great if you can address:
1. Can you also update this [README](https://github.com/pypeit/PypeIt/blob/f7b67ad8f51cbbb2f61806a99ac7a851267a2e63/pypeit/data/telluric/atm_grids/README)? 
2.  It has happened many times that users decide to use the telluric grid instead of the pca but then they don't really know that they also have to change the parameter `teltype` (which is by default set to `pca`). The true problem is that they then get an obscure error, like that the `NCOMP` keyword does not exist, and they usually get confused and unsure of what to do. So, I wonder if we can add something in the docs or adding a more useful error for when users try to load a telluric grid without changing the `teltype` parameter. 

Thanks!

debora-pe (approved):

PR #1928, opened 2025-05-21
fixed a bug where the outmask is not passed correctly when use_2dmode…

jhennawi:
This fixes a bug in local_skysub_extract when use_2dmodel_mask is False.


kbwestfall (approved):
Thanks!  I'll set off a dev-suite run now.

profxj (approved):
yeeps, good catch!

Issue #1809, opened 2024-05-22
Masking sensfunc outside user-specified wavelength range


rcooke-ast:
It would be useful to introduce a user-specified parameter that limits the wavelength range to compute the sensitivity function. The spectra on some detectors (some gratings of GTC/OSIRIS, for example) don't cover the full detector, and this introduces spurious sensfunc solutions near the edges of the spectrum.

jhennawi:
Hi @rcooke-ast take a look at tweak_standard in the keck_mosfire.py spectrograph file. It does what you want.

rcooke-ast:
Addressed for GTC/OSIRIS+ in PR #1932

PR #1932, opened 2025-05-27
GTC/OSIRIS+ sensfunc updates

rcooke-ast:
For the low resolution gratings there are two factors that make it difficult to calculate the sensitivity function:

* second order light
* partial detector coverage

This PR addresses these issues automatically with the `tweak_standard` function. Addresses issue #1809 

Running all tests...


rcooke-ast:
I call this a pass... The MODS2R and MODS1R fails are because of missing files that are unrelated to this PR.
image

profxj (approved):
great stuff!

Raindogjones (commented):
Thanks, @rcooke-ast!
A few comments:
-I think this should be applied to osiris, osiris_plus and maat. Right now, I think it is just osiris_plus, correct?
-Maybe my ignorance but are the wave_red and wave_blue parameters exposed as user-level parameters anywhere? I can see several cases where one would want to override the defaults but not have to edit gtc_osiris.py
-Finally, the debug plot from the test I did was pretty crazy...
Figure_1

pypeit/core/flux_calib.py
```diff
@@ -1080,6 +1080,7 @@ def get_mask(wave_star, flux_star, ivar_star, mask_star,
     msgs.info(" Masking Below the atmospheric cutoff")
     atms_cutoff = wave_star <= 3000.0
     mask_bad[atms_cutoff] = False
+    gpm = np.where(mask_bad)
```
kbwestfall: (A bit nit-picky)  If it makes sense, I'd like to keep the convention that `bpm` and `gpm` are boolean arrays.  Might be easiest to change this to `gpm = mask_bad.copy()`?  Otherwise, we could replace `mask_bad` with `gpm` throughout the function?

pypeit/core/wavecal/wvutils.py
```diff
@@ -92,6 +92,8 @@ def get_delta_wave(wave, wave_gpm, frac_spec_med_filter=0.03):
     gauss_kernel = convolution.Gaussian1DKernel(sig_res)
     wave_diff_smooth = convolution.convolve(wave_diff_filt, gauss_kernel, boundary='extend')
     delta_wave[wave_gpm] = wave_diff_smooth
+    # Set any regions with delta_wave = 0 to the median value of the data
+    delta_wave[np.logical_not(wave_gpm)] = np.median(delta_wave[wave_gpm])
```
kbwestfall: I'd like @jhennawi 's or @feigewang 's thoughts on this.  I don't know the code well enough to know how/if the `delta_wave = 0.` points are used (if at all)?  I see all the tests passed, but are we sure this doesn't have any unintended follow-on effects?

pypeit/sensfunc.py
```diff
@@ -713,15 +730,16 @@ def write_QA(self):
...
                       color=(rr, gg, bb), drawstyle='steps-mid', linewidth=1.0,
                       label=thru_title[iorddet], zorder=idet, alpha=0.7)
 
 
-        wave_gpm_global = self.sens['SENS_FLUXED_STD_WAVE'] > 1.0
-        wave_min = 0.98*(self.sens['SENS_FLUXED_STD_WAVE'][wave_gpm_global]).min()
-        wave_max = 1.02*(self.sens['SENS_FLUXED_STD_WAVE'][wave_gpm_global]).max()
+        # wave_gpm_global = self.sens['SENS_FLUXED_STD_WAVE'] > 1.0
+        wave_min = 0.98*_wave_min #(self.sens['SENS_FLUXED_STD_WAVE'][wave_gpm_global]).min()
+        wave_max = 1.02*_wave_max #(self.sens['SENS_FLUXED_STD_WAVE'][wave_gpm_global]).max()
```
kbwestfall: Remove these commented lines?

kbwestfall (changes requested):
Thanks!  A couple of minor things from me, plus one that I'd like input from @jhennawi and/or @feigewang to look at.  Thanks for posting the test results!

pypeit/core/flux_calib.py
rcooke-ast: Good catch -- I agree. I've updated this to be a `gpm` proper, and I've also renamed `mask_bad` to `bpm`

pypeit/sensfunc.py
rcooke-ast: I've made a slight change, just to catch `None` cases. So, both versions are still there.

pypeit/core/wavecal/wvutils.py
rcooke-ast: Everywhere `delta_wave` is used, the input `wave_gpm` is used to mask out bad values. However, after looking at this a little closer, it seems that this implementation could introduce significant edge effects, because wave_diff could have a few masked values. I think setting the bad pixels to a typical value (as done in the latest version of the code), should reduce the impact of these edge effects.

I have checked a few examples manually, and this works, as before. I will run the tests again before merging, though.

rcooke-ast (commented):
Thanks @kbwestfall for the comments! I've made changes to address all of your comments 👍 

Thanks also to @Raindogjones, for the review and providing additional testing data. A few comments on your queries:

> -I think this should be applied to osiris, osiris_plus and maat. Right now, I think it is just osiris_plus, correct?

Correct. It was previously only applied to `osiris_plus` and `maat` (the latter by inheritance). I've included this code for (the old) `osiris`, too.

> -Maybe my ignorance but are the wave_red and wave_blue parameters exposed as user-level parameters anywhere? I can see several cases where one would want to override the defaults but not have to edit gtc_osiris.py

This is a great suggestion, and something that's been implemented by @debora-pe in PR #1941. Once that PR clears, it will be merged into this PR, and both functionality will be supported, with the user requested values taking precedence.

> -Finally, the debug plot from the test I did was pretty crazy...

Yep, and thanks for sharing data. This is because the data was OSIRIS, not OSIRIS+, and I hadn't implemented the trimmed wavelength solutions for OSIRIS when you tested out the code. After doing this, the `pypeit_sensfunc` script was much better behaved.

pypeit/core/flux_calib.py
```diff
@@ -1070,16 +1070,17 @@ def get_mask(wave_star, flux_star, ivar_star, mask_star,
 
     # masking bad entries
     msgs.info(" Masking bad pixels")
-    mask_bad = mask_star.copy()
-    mask_bad[ivar_star <= 0.] = False
-    mask_bad[flux_star <= 0.] = False
+    bpm = mask_star.copy()
```
kbwestfall: Isn't `mask_bad` (and `mask_star`) as previously used a good-pixel mask?  I.e., the point of the lines below is to exclude the values where `ivar_star` and `flux_star` are <=0, right?

rcooke-ast:
Tests still passing, with some strange vet tests failures, that I _think_ are unrelated, but I'll check.
image

pypeit/core/flux_calib.py
rcooke-ast: You're quite right -- well-spotted! I've made changes to the code so that this mask is now clearly defined.

rcooke-ast:
Tests pass... one of the vet tests fails, but since @profxj is deprecating specutils, I don't see this as an issue.
image

pypeit/spectrographs/gtc_osiris.py
```diff
@@ -439,6 +450,101 @@ def bpm(self, filename, det, shape=None, msbias=None):
 
         return bpm_img
 
+    def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table, log10_blaze_function=None, debug=False):
```
debora-pe: Hi @rcooke-ast , I was look over this PR again and I noticed that you should add the parameter `trim_std_pixs` in this `tweak_standard()` function. See an example [here](https://github.com/pypeit/PypeIt/blob/b624425cfa28972b72eaa05dad65b1e71272df9a/pypeit/spectrographs/keck_mosfire.py#L757-L808).
Could you please do that before merging?

pypeit/spectrographs/gtc_osiris.py
```diff
@@ -1030,3 +1146,97 @@ def bpm(self, filename, det, shape=None, msbias=None):
             bpm_img[bc[bb][2]:bc[bb][3] + 1, bc[bb][0]:bc[bb][1] + 1] = 1
 
         return bpm_img
+
+    def tweak_standard(self, wave_in, counts_in, counts_ivar_in, gpm_in, meta_table, log10_blaze_function=None, debug=False):
```
debora-pe: Same here.

debora-pe (commented):
@rcooke-ast I noticed something missing in this PR. Could you please take a look?

pypeit/spectrographs/gtc_osiris.py
rcooke-ast: Great catch - thanks @debora-pe! I've just added  :-)

pypeit/spectrographs/gtc_osiris.py
rcooke-ast: Also added here too 👍

pypeit/spectrographs/gtc_osiris.py
debora-pe: Thank you for the quick response and action @rcooke-ast . I don't if you were about to do it anyway, but I want to mention that also the following lines of code need to be added for the parameter to be actually used.
```
        if trim_std_pixs is not None:
            return super().tweak_standard(wave_in, counts_in, counts_ivar_in, gpm_in, meta_table,
                                          trim_std_pixs=trim_std_pixs, log10_blaze_function=log10_blaze_function)
```

rcooke-ast: Quite right, I hadn't spotted that  👍  Now implemented, thanks!

pypeit/core/wavecal/wvutils.py
kbwestfall: I'm happy with this change as long as the tests pass!  I think we can merge without input from @jhennawi and @feigewang .

kbwestfall (approved):

PR #1942, opened 2025-06-10
Hotfix for deprecation of fromstring in numpy 2.3.0

kbwestfall:
As titled.

### DIFF
--- a/pypeit/display/display.py
+++ b/pypeit/display/display.py
@@ -141,6 +141,9 @@ def show_image(inp, chname='Image', waveimg=None, mask=None, exten=0, cuts=None,
     if clear:
         clear_all()
 
+    sh = viewer.shell()
+    sh.add_channel_on_demand(chname)
+
     ch = viewer.channel(chname)
     # Header
     header = {}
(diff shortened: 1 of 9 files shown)


profxj (approved):
do it!

ejeschke (approved):
This looks good and necessary.  However, I do think you should add the two-line fix that I suggested in the slack, because I think that error about trying to clear the canvas of a channel that doesn't exist could continue to be an issue.  That fix ensures that the channel is created.

ejeschke:
I can make a separate PR for that if you like...

tbowers7:
Write-access approval.

@ejeschke , probably yes create a PR onto @kbwestfall 's branch with the changes you suggested on the Slack.

tbowers7 (approved):

kbwestfall:
Fair point, @ejeschke .  I've added the lines.

pypeit/display/display.py
```diff
@@ -141,6 +141,9 @@ def show_image(inp, chname='Image', waveimg=None, mask=None, exten=0, cuts=None,
     if clear:
         clear_all()
 
+    sh = viewer.shell()
+    sh.add_channel_on_demand(chname)
```
ejeschke: This needs to be `get_channel_on_demand`.

ejeschke (changes requested):

pypeit/display/display.py
kbwestfall: Thanks!

PR #1943, opened 2025-06-11
Reduce nsnippet for R2500U grating

Raindogjones:
The R2500U grating has very few lines at the blue end. Using the default nsnippet=2 leads to those lines not being identified and a wavelength solution that is out by as much as 10A.


rcooke-ast (approved):
Happy to approve this, and it makes sense, but it would be good to confirm that you have tested this with R2500U data, and it works better (more often) than `nsnippet=2`? Thanks very much!

Raindogjones:
See attached examples. `nsnippet=2` fails 100% of the time in my tests, unless you set `match_tol` to a very large number (and even then it usually identifies the weaker of the blue doublet with the wavelength of the brighter line).
Arc_1dfit_A_0_DET02_S0467_standardparams
Arc_1dfit_A_0_DET02_S0467_matchtol_10

Raindogjones:
Meanwhile, `nsnippet=1` seems to always work correctly:
Arc_1dfit_A_0_DET02_S0467

profxj (approved):
very good

PR #1944, opened 2025-07-01
Update collate1d docs 

badpandabear:
Fix the collate 1d documentation problems described in Issue https://github.com/pypeit/PypeIt/issues/1811


kbwestfall (approved):
Looks great.  Thanks!

debora-pe (commented):
Thank you for this @badpandabear! 
Just one clarification: the user mentioned in the issue that it would be good to have the name of parameter `exclude_slit_trace_bm` to be consistent between the  .collate1d file and the command line flag. Did you decide that you prefer to not do that?

badpandabear:
@debora-pe Originally I had decided not to do that change, but looking at it again I changed my mind since all the other parameters are the same.

debora-pe (approved):
Thanks @badpandabear

PR #1945, opened 2025-07-03
Enable OpenMP support for Apple's XCode compiler suite

tepickering:
The B-spline utilities in `pypeit` are coded in C and use OpenMP pragmas to run some potentially large loops in parallel. With modern multi-core processors, the speedup with OpenMP is very significant. The catch has been that OpenMP is not enabled by default within XCode and the required `libomp` library is not provided. Our suggested workaround has been for Apple users to install `gcc` via a 3rd party package manager like `homebrew`. 

One of those missing pieces is now provided by `conda-forge` in the form of the `llvm-openmp` package. This PR makes the modifications necessary to take advantage of that and unlock XCode's latent OpenMP support. The `environment.yml` file is updated to include `llvm-openmp` and to install some core dependencies via `conda-forge` so that they're easier to keep up-to-date. Work is ongoing to resuscitate the `pypeit` feedstock for `conda-forge`. With the fixes in this PR in place in a `conda-forge` build, then all Apple users would need to do for full multi-core support is `conda install -c conda-forge pypeit`.


pypeit/bspline/setup_package.py
```diff
@@ -7,21 +7,41 @@
...
+if "CONDA_PREFIX" in os.environ:
+    # If we are in a conda environment, we need to add the include and lib directories
+    # so that the libraries from the llvm-openmp package can be found.
+    extra_compile_args.append("-I" + os.environ["CONDA_PREFIX"] + "/include")
+    extra_link_args.append("-L" + os.environ["CONDA_PREFIX"] + "/lib")
+
+if "darwin" in sys.platform:
+    # On macOS, we need manually add the -Xclang and -fopenmp flags to unlock OpenMP.
+    extra_compile_args.append("-Xclang")
+    extra_compile_args.append("-fopenmp")
```
tbowers7: Consider combining these into one line:
```
extra_compile_args.extend(["-Xclang", "-fopenmp"])
```

tbowers7 (approved):
These changes worked on my Intel Mac (macOS 14) to add OpenMP capability without using homebrew.  Thanks for thinking this through!

kbwestfall (commented):
Looks good to me.

I found that homebrew can also be used to install `libomp`: https://formulae.brew.sh/formula/libomp

Is that a relevant installation option?

tepickering:
> Looks good to me.
> 
> I found that homebrew can also be used to install `libomp`: https://formulae.brew.sh/formula/libomp
> 
> Is that a relevant installation option?

i have looked into that. unfortunately, it looks like `libomp` is keg-only which means it gets stuck in the "Cellar" and not linked into `$HOMEBREW_PREFIX/lib`. so the flags you'd need to set would be dependent on the specific version of `libomp`. there might be a trick to query if it's installed and build the flags from that information. not sure if that's worth the hassle, though.

kbwestfall (approved):

PR #1954, opened 2025-07-28
Show arxiv hot fix

profxj:
As titled

Simple fix for this simple script

A test will appear in the DevSuite in a moment.


profxj:
https://github.com/pypeit/PypeIt-development-suite/pull/365

kbwestfall (approved):
I changed the code a bit and added a unit test (it doesn't need the dev-suite).

badpandabear (approved):

PR #1963, opened 2025-08-11
Fix port 9000 blockage issue with Ginga

ejeschke:
Change the default port of 9000 for Ginga remote control interaction to a new Ginga default port number.  Also makes it possible to pass a custom port number to the `connect_to_ginga` function and it will configure Ginga to use that port dynamically.

- bumps Ginga requirement to v5.4.0

### DIFF
--- a/pypeit/display/display.py
+++ b/pypeit/display/display.py
@@ -26,7 +26,8 @@ from pypeit import msgs
 from pypeit import io
 from pypeit import utils
 
-def connect_to_ginga(host='localhost', port=9000, raise_err=False, allow_new=False):
+def connect_to_ginga(host='localhost', port=grc.default_rc_port,
+                     raise_err=False, allow_new=False):
     """
     Connect to a RC Ginga.
 
@@ -34,7 +35,7 @@ def connect_to_ginga(host='localhost', port=9000, raise_err=False, allow_new=Fal
         host (:obj:`str`, optional):
             Host name.
         port (:obj:`int`, optional):
-            Probably should remain at 9000
+            Probably should remain at Ginga default
         raise_err (:obj:`bool`, optional):
             Raise an error if no connection is made, otherwise just
             raise a warning and continue
@@ -53,7 +54,8 @@ def connect_to_ginga(host='localhost', port=9000, raise_err=False, allow_new=Fal
         tmp = sh.get_current_workspace()
     except:
         if allow_new:
-            subprocess.Popen(['ginga', '--modules=RC,SlitWavelength'])
+            subprocess.Popen(['ginga', f'--rcport={port}',
+                              '--modules=RC,SlitWavelength'])
 
             # NOTE: time.sleep(3) is now insufficient. The loop below
             # continues to try to connect with the ginga viewer that
@@ -72,7 +74,7 @@ def connect_to_ginga(host='localhost', port=9000, raise_err=False, allow_new=Fal
                     break
             if i == maxiter-1:
                 msgs.error('Timeout waiting for ginga to start.  If window does not appear, type '
-                           '`ginga --modules=RC,SlitWavelength` on the command line.  In either '
+                           f'`ginga --rcport={port} --modules=RC,SlitWavelength` on the command line.  In either '
                            'case, wait for the ginga viewer to open and try the pypeit command '
                            'again.')
             return viewer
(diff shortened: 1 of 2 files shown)


ejeschke:
This doesn't add a port number option to the 'pypeit_show*' scripts, but that could be added later, if desired.

Please test!

profxj (approved):
Looks good to me, thanks

pypeit/display/display.py
```diff
@@ -72,7 +74,7 @@ def connect_to_ginga(host='localhost', port=9000, raise_err=False, allow_new=Fal
                     break
             if i == maxiter-1:
                 msgs.error('Timeout waiting for ginga to start.  If window does not appear, type '
-                           '`ginga --modules=RC,SlitWavelength` on the command line.  In either '
+                           f'`ginga --rcport={port} --modules=RC,SlitWavelength` on the command line.  In either '
```
kbwestfall, line 77: We may need to be more detailed.  How  about this:

```
msgs.error(
    'Timeout waiting for ginga to start.  If the window does not appear automatically, type '
    '`ginga --modules=RC,SlitWavelength` on the command line.  If running the script does '
    'not show the images, there may be a problem with the port being used (because a '
    'different program (e.g., VSCode) is already using it.  You can set the port directly using '
    'the `--rcport` option: e.g., `ginga --rcport=9100 --modules=RC, SlitWavelength`.  In '
    'any case, execute the ginga command then wait for the ginga viewer to open before '
    'running the pypeit command again.'
)
```

pypeit/display/display.py
```diff
@@ -81,7 +83,7 @@ def connect_to_ginga(host='localhost', port=9000, raise_err=False, allow_new=Fal
             raise ValueError
         else:
             msgs.warn('Problem connecting to Ginga.  Launch an RC Ginga viewer and '
-                      'then continue: \n    ginga --modules=RC,SlitWavelength')
+                      f'then continue: \n    ginga --rcport={port} --modules=RC,SlitWavelength')
```
kbwestfall, line 86: Same comment here.  I'm worried people are just going to copy/paste what's here instead of replacing `{port}` with a port number...

kbwestfall (approved):
Thanks!  I had one minor comment about the error/warning messages.

badpandabear (approved):

pypeit/display/display.py
ejeschke: Hmm, but advising the user to use a non-standard port (9100) invoking ginga separately, without a way to specify the port in `pypeit_show_1dspec` is going to mean that the ports don't match on both sides.

The error message that I put in there shows the port that Pypeit is using, so I think that is the right message.

pypeit/display/display.py
ejeschke: What we have now is the ability for Pypeit to control the port--so I think the error messages should show that port, as is being done in the PR.

pypeit/display/display.py
kbwestfall: Sorry.  You're right.  I didn't realize it was pulling the port value from what was being passed to the function.  Do they need to include the port in the calling sequence?

ejeschke: Calling sequence for `connect_to_ginga`? No, that will default to the new port Ginga switched to in version 5.4.  However, now that Pypeit has the ability to choose it's own port, it might be a good idea to define that somewhere and provide the ability to set it or override it on the command line.  But that doesn't need to be part of this PR.

kbwestfall: No, I mean the calling sequence on the command line.  Ideally, I don't want the users to have to care about the port number.  Along the same lines, is there a way to determine the port used by an open ginga window?  What happens if the port number passed to `connect_to_ginga` is different from the one used by an open window?  Does it open a second ginga window?  Now that we can specify the port, I'm worried about failure modes and possible mismatches between port numbers.

ejeschke: In this PR, Pypeit defaults to the `new` Ginga default RC port number, so users don't need to care about it in the usual case.

There is only one port used for any one instantiation of ginga. If you start a second instance of ginga running (not just a new window in the same instance) you would need to specify a different port number on the ginga command line. Because Pypeit can now specify the port it wants to use, that is now possible.  But error messages should show which port is being used.

> Along the same lines, is there a way to determine the port used by an open ginga window?

Not without trying to figure out some other form of RPC or writing a file with the number in it or something like that. But now Pypeit can specify and control the port.

> Now that we can specify the port, I'm worried about failure modes and possible mismatches between port numbers.

I think this will have exactly the opposite effect--now that you can control and coordinate the port numbers there will be less potential problems

ejeschke:
@kbwestfall , this has been rebased against `develop`.

PR #1964, opened 2025-08-12
Enhance ginga spec1d

### LINKED ISSUE #1955: Ginga displaying values with excessive number of decimal places
Ginga displays values with an excessive number of decimal places - 10 or 12 - and the most significant part of the value gets shifted to the left and out of the box. The low/high x and y values in the attached screenshot of a Ginga window used to display a 1D spectrum give an example.  Having such a long number to delete/replace when one wants to zoom about a particular line or spectral region can be awkward. Could the values be displayed to fewer decimal places? For context, in this case, x is wavelength and y is counts. 

Image

ejeschke:
Enhances the Ginga 1Dspec viewer plugin

- plots labels on lines so that they are always visible when the line itself is visible
- plot updates when zooming or panning changes the range of the plot
- requires Ginga v5.4.0

### DIFF
--- a/pypeit/scripts/show_1dspec.py
+++ b/pypeit/scripts/show_1dspec.py
@@ -4,7 +4,7 @@ Wrapper to the linetools XSpecGUI
 .. include common links, assuming primary doc root is up one directory
 .. include:: ../include/links.rst
 """
-import os.path
+from pathlib import Path
 from IPython import embed
 
 from pypeit.scripts import scriptbase
@@ -123,7 +123,7 @@ class Show1DSpec(scriptbase.ScriptBase):
             from pypeit.display.display import show_1dspec
 
             # in case Ginga is invoked in another directory
-            file_path = os.path.abspath(args.file)
+            file_path = str(Path(args.file).absolute())
             show_1dspec(file_path, ext=exten,
                         masked=args.masked, extraction=args.extract,
                         fluxed=args.flux)
--- a/pypeit/scripts/show_1dspec.py
+++ b/pypeit/scripts/show_1dspec.py
@@ -4,6 +4,7 @@ Wrapper to the linetools XSpecGUI
 .. include common links, assuming primary doc root is up one directory
 .. include:: ../include/links.rst
 """
+import os.path
 from IPython import embed
 
 from pypeit.scripts import scriptbase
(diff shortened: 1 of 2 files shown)


ejeschke:
NOTE: It would be best to merge #1963 first, but this is not necessary. However, if you are testing this without having merged #1963, then it is best to invoke ginga separately like so (first), before running `pypeit_show_1dspec`:
```bash
$ ginga --modules=RC,SlitWavelength --rcport=9000
```
because the default remote control port changed in Ginga v5.4.0

profxj (approved):
Thanks for these!

pypeit/scripts/show_1dspec.py
```diff
@@ -121,7 +122,9 @@ def main(args):
             # show the 1d spectrum in Ginga
             from pypeit.display.display import show_1dspec
 
-            show_1dspec(args.file, ext=exten,
+            # in case Ginga is invoked in another directory
+            file_path = os.path.abspath(args.file)
```
kbwestfall: Is this necessary?  We prefer using `pathlib` in new code.  Can we change this to:
```
from pathlib import Path
...
    file_path = Path(args.file).absolute()
```

kbwestfall (changes requested):
Thanks!  I have one request concerning the use of `os.path` vs. `pathlib.Path`.

pypeit/scripts/show_1dspec.py
ejeschke: Sure !

ejeschke: Turns out we need to turn the pathlib object into a string to get it to pass cleanly over the remote procedure call interface.

ejeschke:
If there are other `pypeit_show_<X>` commands that ask Ginga to load a file from disk rather than passing a buffer it would be good to make this same sort of change of passing an absolute filename. It prevents the error that if you invoked Ginga separately from a different directory the load will fail on a relative path.

I can poke around in the `pypeit/scripts` directory and see what other ones I can find...

ejeschke:
Looks like most of the other scripts pass a buffer

kbwestfall (approved):
Thanks!

debora-pe (approved):
Thanks!

ejeschke:
@kbwestfall , this has been rebased against `develop`.

kbwestfall:
Sorry to ask you to do this.  But I just merged #1964 , so you'll need to rebase again.

ejeschke:
Rebased.  You may want to squash and merge to subsume the 2 fix commits.

PR #1972, opened 2025-08-27
fix for binning 2x2

bpholden:
This adds a fix for reducing data that is binned 2x2, by changing a default parameter.


kbwestfall (approved):
Looks good to me!

tbowers7 (approved):
Looks good.

PR #1976, opened 2025-09-03
Avoid extension checks for spec2d (PypeIt) files

rcooke-ast:
As titled. Addresses an issue with SOAR/Goodman highlighted on the Users Slack.

### DIFF
--- a/pypeit/display/display.py
+++ b/pypeit/display/display.py
@@ -26,7 +26,7 @@ from pypeit import msgs
 from pypeit import io
 from pypeit import utils
 
-def connect_to_ginga(host='localhost', port=grc.default_rc_port,
+def connect_to_ginga(host='localhost', port=9000,
                      raise_err=False, allow_new=False):
     """
     Connect to a RC Ginga.
--- a/pypeit/display/display.py
+++ b/pypeit/display/display.py
@@ -26,7 +26,7 @@ from pypeit import msgs
 from pypeit import io
 from pypeit import utils
 
-def connect_to_ginga(host='localhost', port=9000,
+def connect_to_ginga(host='localhost', port=grc.default_rc_port,
                      raise_err=False, allow_new=False):
     """
     Connect to a RC Ginga.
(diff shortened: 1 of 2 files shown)


pypeit/display/display.py
```diff
@@ -26,7 +26,7 @@
 from pypeit import io
 from pypeit import utils
 
-def connect_to_ginga(host='localhost', port=grc.default_rc_port,
+def connect_to_ginga(host='localhost', port=9000,
```
profxj: for @ejeschke to comment on

pypeit/spectrographs/spectrograph.py
```diff
@@ -320,6 +320,16 @@ def _check_extensions(self, filename):
         """
         if self.allowed_extensions is not None:
             _filename = Path(filename).absolute()
+            # Don't check PypeIt spec2d files
+            if _filename.name.startswith("spec2d_"):
```
profxj, line 324: not the most elegant work around, but fine by me.

profxj (changes requested):
Will approve once Eric comments on the `port` change
which I know has been a "whack a mole" issue.

pypeit/spectrographs/spectrograph.py
kbwestfall: Agreed.  I took a look, though, and a cleaner solution would require much more effort.  I suggest we use this for now and refactor soon.

kbwestfall (approved):
Thanks for the quick fix!

pypeit/display/display.py
rcooke-ast: I've reverted this to the previous version  😄

pypeit/spectrographs/spectrograph.py
rcooke-ast: I agree... it also raises other improvements that we could make (e.g. at the moment, the `spec2d_` prefix is hard-coded in `pypeit.py` and really should be set in `Spec2DObj` for visibility to the entire code, as needed).

Thanks for the quick comments!

profxj (approved):
thx!

pypeit/display/display.py
ejeschke: If you revert this to 9000, it is going to fail for people with Ginga v5.4.0 installed and will re-introduce the problem with vscode

PR #1985, opened 2025-09-14
Updates to soar_goodman to correctly identify standard stars

### LINKED ISSUE #1984: SOAR/Goodman configurations currently ignore standards
The SOAR/Goodman configurations currently ignore standard frames (e.g., frames in a science setup with short exposure times) by default:

```
        if ftype in ['standard']:
            # Don't type pinhole or dark frames
            return np.zeros(len(fitstbl), dtype=bool) & self.lamps(fitstbl, 'off')
```

I would suggest adapting this to match the behavior in Keck/LRIS where pypeit checks for exposures in the correct standard frame range and within some radius of a standard star in the pypeit library, along with the other checks that science frames undergo:

```
        if ftype in ['standard']:
            std = np.zeros(len(fitstbl), dtype=bool)
            # Don't type pinhole or dark frames
            if 'ra' in fitstbl.keys() and 'dec' in fitstbl.keys():
                std = np.array([flux_calib.find_standard_file(ra, dec, toler=10.*units.arcmin, check=True)
                                for ra, dec in zip(fitstbl['ra'], fitstbl['dec'])])
            return good_exp & std & (fitstbl['idname'] == 'SPECTRUM') & (self.lamps(fitstbl, 'off'))
```

This will also require adding these imports to `soar_goodman.py`:

```
from astropy import units
from pypeit.core import flux_calib
```

charliekilpatrick:
Editing soar_goodman so that standard star files are handled properly.  Changes:

- Updates `check_frame_type` in `soar_goodman.py` so that standard star files are identified from the database in `flux_calib` similar to other spectrographs
- Update to `test_metadata` unit test to include explicit testing for dummy standard, science, arc, and flat frames.

Closes: https://github.com/pypeit/PypeIt/issues/1984


kbwestfall (approved):
Looks good to me.  Thanks for the contribution!  We should add a test (to the dev-suite).  If you're up for it, please do so.  If not, I'll at least make sure the dev-suite passes for the SOAR/Goodman data.

bpholden (approved):
Thanks for doing this

rcooke-ast (approved):
I'm also happy to approve - thanks @charliekilpatrick!

kbwestfall (changes requested):
Just marking this as requesting changes until we've run the tests.

charliekilpatrick:
> Looks good to me. Thanks for the contribution! We should add a test (to the dev-suite). If you're up for it, please do so. If not, I'll at least make sure the dev-suite passes for the SOAR/Goodman data.

Thanks @kbwestfall.  Looking at the dev-suite, I can add a test similar to `test_metadata.test_nirspec_lamps`.  I'll commit the changes later today.

kbwestfall:
> Thanks @kbwestfall. Looking at the dev-suite, I can add a test similar to `test_metadata.test_nirspec_lamps`. I'll commit the changes later today.

Awesome!  Thanks, @charliekilpatrick !

charliekilpatrick:
@kbwestfall - `test_metadata` is updated and is passing for me locally.  I've made the commit, so ready for review whenever is convenient.

pypeit/tests/test_metadata.py
```diff
@@ -86,6 +86,36 @@ def test_nirspec_lamps():
...
+    fitstbl.add_row((5.0208,8.5102,'supernova','1.0_LONG_SLIT','2,2',400.0,60931.1597,1.46,'400_SYZY','400_M1',5.7999,'SPECTRUM','FALSE','FALSE','FALSE','FALSE','FALSE','FALSE','FALSE','FALSE','/home','supernova.fits.fz','',0))
+    # Arc lamp
+    fitstbl.add_row((5.0212,8.5116,'supernova_comp','1.0_LONG_SLIT','2,2',0.5,60931.1544,1.48,'400_SYZY','400_M1',5.7995,'ARC','TRUE','TRUE','FALSE','FALSE','FALSE','FALSE','FALSE','FALSE','/home','supernova.fits.fz','',0))
+    # Flat
+    fitstbl.add_row((148.3446,-17.6491,'dflat','1.0_LONG_SLIT','2,2',7.0,60930.7888,1.66,'400_SYZY','400_M1',5.7997,'LAMPFLAT','FALSE','FALSE','FALSE','FALSE','FALSE','FALSE','FALSE','TRUE','/home','dflat.fits.fz','',0))
+
+    # Check standard
+    tst = spectrograph.check_frame_type('standard',fitstbl)
+    assert np.array_equal(tst, np.array([True, False, False, False]))
+    # Check science - note that the standard will also pass this test
```
kbwestfall: We've been talking about this over Slack this morning.  We're not sure,  but it's possible that typing something as both `science,standard` is a problem.  I think the nominal control flow in the code assumes that observations are typed as one or the other, not both.  Have you tested this with an example reduction?  If `pypeit_setup` appropriately types things correctly (without double typing), is it because of the exposure-time constraint?  What if people have particularly long standard exposures or short science exposures?  I expect we need to change the `keck_lris.py` code, as well, but what do you think of the following way of ensuring that observations are only typed as `standard` or `science`, not both?

```
if ftype in ['science', 'standard']:
    std = np.zeros(len(fitstbl), dtype=bool)
    # Identify standard stars from flux_calib
    if 'ra' in fitstbl.keys() and 'dec' in fitstbl.keys():
        std = np.array([flux_calib.find_standard_file(ra, dec, toler=10.*units.arcmin, check=True)
                        for ra, dec in zip(fitstbl['ra'], fitstbl['dec'])])
    base = good_exp & (fitstbl['idname'] == 'SPECTRUM') & self.lamps(fitstbl, 'off')
    if ftype == 'science':
        return base & np.logical_not(std)
    return base & std
```

kbwestfall (changes requested):
Your comment in the test sparked a question that might mean we need to revisit the `keck_lris.py` code, as well.  Let me know what you think.

pypeit/tests/test_metadata.py
charliekilpatrick: Hi @kbwestfall - yes, I wasn't sure if this was intentional or not.  In retrospect, I see that this is not ideal because you would get different behavior if you tested for science or standard first and pypeit explicitly allows for different behavior on science versus standard frames.

I tested this locally and get the expected behavior in the unit test (and removed that comment about the standard also passing the science test).  Do you want me to update keck_lris (I haven't done an exhaustive search for all spectrographs have this, but it might be a lot of them) as well or should that be a separate PR?

kbwestfall: No, I wouldn't worry about updating the `keck_lris.py` file in this PR.  I'd like to redirect this to the `develop` branch, so that we can more fully address the issue before putting it in the `release` branch.

Just to state it here, I think we need to check how many instruments do this, and I expect we should move the check that creates `std` into `metadata.py` and add the results as a new column in `fitstbl` so that we don't have to recreate the vector every time we check the type of a single frame.

kbwestfall (approved):
Thanks!!

kbwestfall:
Hi @charliekilpatrick .  Looks like the tests are failing.  Can you check that they pass locally for you?

charliekilpatrick:
Hi @kbwestfall - apologies, I must have deleted the last return when I committed to my fork.  I've fixed this.

kbwestfall:
Checks pass.  Thanks @charliekilpatrick !  Merging...

PR #1854, opened 2024-09-06
Package dependency updates

kbwestfall:
Deprecates support for python <= 3.10.  I also updated other package dependencies, trying to at least synchronize them to when 3.11 was released.

This also includes some doc fixes.

We'll need to run tests.

PR #1994, opened 2025-09-21
Remove ``__UTC__`` crutch for Python 3.10 support

tbowers7:
When ``datetime.datetime.utcnow()`` was deprecated in Python 3.12, we needed some crutches (#1785) to be able to successfully transition to the new ``datetime.UTC`` constant (introduced in Python 3.11).  Since PypeIt no longer supports Python 3.10 (#1854), we can safely remove these crutches to provide for cleaner code relying solely on the ``datetime.UTC`` constant.

DevSuite tests pass (Nautilus report ``tbowers-py310-sep19.report``):

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  261 passed, 443 warnings in 508.66s (0:08:28) ---
--- PYTEST UNIT TESTS PASSED  157 passed, 93167 warnings in 959.23s (0:15:59) ---
--- PYTEST VET TESTS PASSED  69 passed, 135631 warnings in 4766.40s (1:19:26) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 280/280 TESTS  ---
Testing Started at 2025-09-19T17:10:25.540237
Testing Completed at 2025-09-20T11:39:28.966919
Total Time: 18:29:03.426682
```

### DIFF
--- a/pypeit/calibrations.py
+++ b/pypeit/calibrations.py
@@ -221,7 +221,7 @@ class Calibrations:
         # Loop on the files
         for ii, ifile in enumerate(file_list):
             # Save the lamp status
-            headarr = copy.deepcopy(self.spectrograph.get_headarr(ifile))
+            headarr = [h.copy() for h in self.spectrograph.get_headarr(ifile)]
             lampstat[ii] = self.spectrograph.get_lamps_status(headarr)
 
         # Check that the lamps being combined are all the same
--- a/pypeit/calibrations.py
+++ b/pypeit/calibrations.py
@@ -4,25 +4,15 @@ Class for guiding calibration object generation in PypeIt.
 .. include common links, assuming primary doc root is up one directory
 .. include:: ../include/links.rst
 """
-import os
-from pathlib import Path
-from datetime import datetime
-from copy import deepcopy
 from abc import ABCMeta
 from collections import Counter
-import yaml
-
-# TODO: datetime.UTC is not defined in python 3.10.  Remove this when we decide
-# to no longer support it.
-try:
-    __UTC__ = datetime.UTC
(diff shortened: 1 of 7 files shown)


pypeit/calibrations.py
```diff
@@ -233,7 +221,7 @@ def check_calibrations(self, file_list, check_lamps=True):
         # Loop on the files
         for ii, ifile in enumerate(file_list):
             # Save the lamp status
-            headarr = deepcopy(self.spectrograph.get_headarr(ifile))
+            headarr = copy.deepcopy(self.spectrograph.get_headarr(ifile))
```
kbwestfall: For these header copies, I think we should be using the class method explicitly: https://docs.astropy.org/en/latest/io/fits/api/headers.html#astropy.io.fits.Header.copy (even though the above line of code does this implicitly).

kbwestfall (approved):
I had one minor comment about using the copy class method for headers, but otherwise this looks good to go.  Thanks!

pypeit/calibrations.py
tepickering: in the interest of being explicit and clear, i am inclined to agree, but don't feel super strongly about it.

tepickering (approved):
thanks for this long-overdue cleanup! looks good and thanks also for some of the import cleanups.

bpholden (approved):
Looks good according to the other Tim

pypeit/calibrations.py
tbowers7: Now reads:
```
headarr = [h.copy() for h in self.spectrograph.get_headarr(ifile)]
```

PR #2000, opened 2025-09-26
Adds automatically generated tables that list the available telluric models

kbwestfall:
As titled.

This is a doc only PR, also associated with a larger refactor effort.


debora-pe (approved):
Thanks Kyle! This looks great!

doc/scripts/build_telluric_table.py
bpholden, line 1: Please check the result of subprocess.run()

bpholden, line 1: If you are going to capture the output of the exception as a variable (e) then you should do something with it.
Examples include:
print(f'{type(e) e')

bpholden (changes requested):
I have quibbles regarding the exception handling.

doc/scripts/build_telluric_table.py
kbwestfall: I did a couple of things.  I now check the return code, and I also check that each returned line parses into the expected number of entries.  Is this what you meant?  Or did you have something else in mind?

kbwestfall: You're right that I don't need to capture the exception.  But I don't need to print the exception because I'm re-raising it at the end of the block.  I.e., the code will fault if the `subprocess.run` call faults, it will just add a brief message before faulting.

bpholden (approved):

PR #1980, opened 2025-09-05
LDT/DeVeny Documentation Additions

tbowers7:
Spectrograph-specific documentation for LDT/DeVeny.  This documentation was written in support of POSE grant goals for instrument support.

Includes sphinx package ``sphinx-design`` for creating "subfigures".

Documentation only; no DevSuite run.

Will add release notes once ``v1.18.0`` posts.

PR #2006, opened 2025-09-26
Last bit of the new LDT/DeVeny docs

tbowers7:
Last-minute add of coadd2D comparison.  Also, fixing last PR requests from #1980 .


pypeit/scripts/setup_coadd2d.py
```diff
@@ -147,7 +147,7 @@ def main(args):
                                 else [f.name.split('-')[1].split('_')[0] for f in spec2d_files])
         if args.obj is not None:
             # Limit to the selected objects
-            _objects = [o for o in objects if o == args.obj]
+            _objects = [o for o in objects if o in args.obj]
```
kbwestfall, line 150: Can you say more about this change?  The original idea was that the object name should be visible via the spec2d file names, and the user should be able to set it exactly.  I appreciate there could be an issue where, e.g., human error led to a slightly changed the object name written to the header.  Is the idea to allow for slight deviations of the string as long as at least some fraction of it is common for all files?  And, if so, wouldn't you want to do `if args.obj in o` instead of the other way around?

kbwestfall (commented):
I like the JWST demo!

I wanted more information about the change to `setup_coadd.py`, but the rest looks good.

pypeit/scripts/setup_coadd2d.py
tbowers7: ``args.obj`` is a list -- I found the bug because I got a strange error:
```
pypeit.pypmsgs.PypeItError: Unable to find relevant objects.  Unique objects are ['obj1', 'obj2', 'obj3', 'JWST', 'obj4', 'obj5']; you requested ['JWST'].
```

I know ``args.obj`` is a list because the parser line in the script is:
```
        parser.add_argument('--obj', type=str, nargs='+',
```

tbowers7: Changing ``==`` to ``in`` works for both single object arguments to ``--obj`` and multiple objects, since ``args.obj`` is a list.  This still requires an exact match between what's in the ``target`` metadata tag (and hence the filename) and what the user inputs on the command line for this script.  I think the option to add multiple objects (``nargs='+'``) was possibly added after the ``--obj`` option was tested -- and not retested after.

tbowers7: Also, there's no test for this script.

tbowers7: We can put this in a separate PR, if you want.  It was just a very small bug I found while putting together this example for the docs.

kbwestfall: Ah, good point.  I had forgotten that `args.obj` could be a list.  There's a vet test in the dev-suite that tests this cript, but none of the test calls use the `--obj` option!  Given that you got an error when you tried to run the code, I think we should leave it in this PR.  Can you run the tests on this branch, just to check that everything passes?  If it passes, we'll merge and then add an issue that we need to add a test of the `--obj` option!  If it doesn't pass, let's revert that single change, then merge, and then add the issue :)

tbowers7: DevSuite launched.

tbowers7:
Ghosts in the machines.  Full Nautilus DevSuite run had a weird (bspline?) failure in LBT/LUCI, but then a re-run of just that instrument both on Nautilus and locally passed cleanly.  🤷
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  250 passed, 471 warnings in 470.61s (0:07:50) ---
--- PYTEST UNIT TESTS PASSED  157 passed, 95147 warnings in 1019.01s (0:16:59) ---
--- PYTEST VET TESTS PASSED  68 passed, 1612 warnings in 6483.11s (1:48:03) ---
--- PYPEIT DEVELOPMENT SUITE FAILED 1/280 TESTS  ---
Failed tests:
    lbt_luci/LUCI-I pypeit
Skipped tests:
Testing Started at 2025-09-27T15:12:33.328169
Testing Completed at 2025-09-28T11:09:08.815992
Total Time: 19:56:35.487823
```
Then just LBT/LUCI:
```
Reduced data for the following setups:
    lbt_luci/LUCI-I
    lbt_luci/LUCI-II
...
Test Summary
--- PYPEIT DEVELOPMENT SUITE PASSED 2/2 TESTS  ---
Testing Started at 2025-09-29T16:59:20.504892
Testing Completed at 2025-09-29T17:16:41.931613
Total Time: 0:17:21.426721
```

kbwestfall (approved):
Thanks!

profxj (approved):
thanks!

PR #2007, opened 2025-10-02
Refactor pypeit

profxj:
As titled, this PR is a major refactor of the primary module of PypeIt.

The main goals are to:

- Expose the primary methods currently in `pypeit.py` which perform image processing, object finding, and extraction
- Enable users to perform "step by step" reductions on individual detectors/mosaics.
- Clean up some of the chaff that has grown in `pypeit.py` over the years
- Enable spectrographs to effectively skip the processing step, e.g. if they have their own tried+true processing codes
- Put in the hooks for the Dashboard that will allow step by step running and monitoring of PypeIt

The new script is `pypeit_reduce_by_step` which runs on a single science/standard frame (or a set of frames, if `comb_id` is set in the PypeIt file) and a single detector/mosaic of that frame. 

Docs are provided with warnings about running PypeIt in anything but the end-to-end mode

I have been testing in the [dev_pypeit_refactor](https://github.com/pypeit/PypeIt-development-suite/tree/dev_pypeit_refactor) branch in the DevSuite.  Thus far, these are ad hoc tests and I'll be
begging @badpandabear 's help to meld them into standard DevSuite runs.  Those tests do check
that we are achieving nearly identical results with the new vs. old PypeIt as well as with the new
`pypeit_reduce_by_step` script.

I should add that the `state.py` code should be ignored as well as any hooks to it,
unless you see a way things could break.  For now, there are hooks but none of that
code should be active.

PR #1971, opened 2025-08-27
Designate configuration-specific parameters based on PypeIt metadata

tbowers7:
#### Initial Description

As described in #1958 , configuration-specific parameters had been polled from FITS headers directly.  This overrides edits to the ``.pypeit`` files.  As discussed in the developer telecon on 25 August 2025, the primacy of the Pypeit Reduction File as the source for TRUTH in a reduction should be maintained.  This PR aims to restore this primacy.

Of the calls to ``.config_specific_par()``, some really do need to send the raw data file / header rather than the metadata table (*e.g.* from `sensfunc`, where no metadata table exists).  As best I can figure, only the calls from ``inputfiles.py`` and ``chk_for_calibs.py`` are under conditions where the metadata table (from, *e.g.*, the ``.pypeit`` file) is required to possibly correct for bad FITS headers.

The solution was to check for the *type* of the input to the method, where the callers that need to rely upon editable metadata can send a row of the metadata `Table` and others can continue to send a filename or `HDUList` or whatever.  The method extracts the needed value as appropriate given the *type* of the input and sets and returns parameters as before.

The bulk of the work was to propagate these ``config_specific_par()`` changes to **ALL** relevant spectrographs!  In the process, I also did some import cleanup / standardization in the affected spectrograph classes.  In instruments with multiple possible configurations, ``if..elif..elif..else`` syntaxes were moved to a ``match..case..case`` syntax both for modernization and for clarity for the reader.

Finally, a unit test was added to ensure the functionality of both *type*s of input and that when changing metadata keys used to sort configurations, the ``config_specific_par`` actually changes.

#### Additions to the PR made along the way:

It is important to be sure metadata keys needed for ``config_specific_par()`` are actually in the PypeIt Reduction File or are configuration keys.  Various keys added to various spectrographs.

Added companion DevSuite PR (https://github.com/pypeit/PypeIt-development-suite/pull/369) that ~~updates the canned ``.pypeit`` files in the DevSuite to include the new columns written to satisfy ``config_specific_par()`` when passed a metadata table row~~ fixes some Gemini/GMOS MJD issues.

A couple issues were stinkers. The first was Python's dynamic typing and Bok/B&C not following the usual scheme of "300GPM" or "300L" for grating names... just "300" -- ``int(300)`` is not the same as ``str(300)``. The second was older Gemini/GMOS .pypeit files using observing epoch in the MJD metadata column... which makes for Bad Things when trying to use that value as an actual MJD...

[...]

PR #2012, opened 2025-10-10
Logging refactor

kbwestfall:
This refactors the logging infrastructure for PypeIt so that it now uses the standard python `logging` library (I don't know why I torture myself...).  The new logger draws heavily from the ones implemented by `astropy` and `sdsstools`.

The python logging package gives the following levels, DEBUG, INFO, WARNING, ERROR, and CRITICAL.  In their parlance, we basically consider all ERRORs as CRITICAL, so the `log.critical` function isn't really useful to us.  For the scripts, I've mapped the `verbosity` to these levels as DEBUG = 2, INFO = 1, and WARNING = 0.  I.e., if `verbosity == 0`, we only log warnings and errors, whereas `verbosity == 1` adds info message, and `verbosity == 2` adds debug messages.  All the scripts now also allow us to set the level specifically for the file log (using the same integer mapping as the console verbosity), which can be different from the console log.

The key changes are:

- `pypmsgs.py` is effectively replaced by `logger.py`.  There is some functionality loss (I think this is primarily limited to the fact that we will no longer catch Ctrl-c events and the QA html will not be automatically constructed when failures happen), so let me know if you think we're losing something critical
- instead of `from pypeit import msgs`, basically every file will need to `from pypeit import log, PypeItError`.
- anywhere you would have put `msgs.error(...)`, you should instead `raise PypeItError(...)`.  *All* exceptions (and warnings) are now caught and processed by the `PypeItLogger` (not just `PypeItError` instances); so if you think its more appropriate to raise a different error type, go for it.  The logging class has `error` and `critical` functions, but they're not that useful for pypeit.
- the `warn` function in the `logging` package has been deprecated, so all warnings should use `log.warning` instead of `msgs.warn`.
- we should now consider making more use of the `debug` message level.  Anything that used to use `msgs.work`, I've now converted to `log.debug`, but there are some conceptual differences there.
- the `msgs.newline()` function no longer exists.  If you want a new line in the log message, include the `\n` character directly.  I was thinking of adding some line-wrapping tools for messages in the future, but formatted output that suits everyone's terminal width choices is tricky.
- all the scripts now add verbosity and log file options by default (see above)
- the default logging level is set to `DEBUG`, which displays all messages and adds the function file, name, line number information.
- the `INFO` level does not include the calling function information
- if the previous scripts had a default log file, I kept it, but now effectively any script can produce a log file with a user-specified name.
- The format of the console stream logging and the file logging are different
    - the calling function information is always included in the file log and
    - the file log includes date/time information.
- The logging level can be different between the console stream and the file stream.  I.e., one could set the console to only display warnings and errors, but include all the messages in the file log (or vice versa).
- most of the changes are just find/replace, so the key files to review are `pypeit/logger.py`, `pypeit/__init__.py`, `pypeit/tests/test_log.py`.

I'm submitting this as a draft PR, but I think it's mostly done.  But, because this PR basically touches everything, we should discuss when it gets merged.  I expect it should at least come after #2007 and #1971 , and I will deal with the likely conflicts.  So if you have another big PR in the works and don't want to deal with the conflicts, let's try to get those in before this one.


ejeschke:
Good move to integrate with standard library logger.  One cool feature is the use of the `exc_info` flag when you call `logger.error()` (or possibly other levels as well).  If you call the logger in a `except` clause, and pass the keyword argument `exc_info=True`, then in addition to logging the error message it will also log a stack trace. Very handy.

pypeit/logger.py
```diff
@@ -0,0 +1,378 @@
...
+from pathlib import Path
+import re
+import sys
+import traceback
+import warnings
+
+from IPython import embed
+
+# TODO: Can we put this *inside* the logger?
+def short_warning(message, category, filename, lineno, file=None, line=None):
```
rcooke-ast: Should we add these parameters to the docstring?

rcooke-ast (approved):
I'm OK with this. It's great that you've brought this part of the code into the present times!

I only reviewed `__init__.py`, `test_logger.py` and `logger.py`.

I couldn't checkout a test drive of the code, just to see what it looks like as terminal printout. Could you please post a representative screenshot? Thanks!

kbwestfall:
Hi @rcooke-ast .  No problem.  Here are some screen grabs of the different levels when sent to the screen, and what the level 2 output looks like in the file.

Level 2 file output
level2_file

Level 0 screen output
level0_screen

Level 1 screen output
level1_screen

Level 2 screen output
level2_screen

pypeit/logger.py
kbwestfall: Good call.  I don't do a formal definition of the parameters, but I point to the main doc for `warnings.formatwarning`.  And I actually realized that the calling sequence doesn't identically match the default function.  Fixed!

doc/conf.py
```diff
@@ -351,5 +354,10 @@ def get_package_version():
 
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/3', None),
+    'numpy': ('https://numpy.org/doc/stable/', None),
+    'scipy': ('https://docs.scipy.org/doc/scipy/', None),
+    'astropy': ('https://docs.astropy.org/en/stable/', None),
```
tbowers7, line 362: Consider adding:
```
    'matplotlib': ('https://matplotlib.org/stable/', None),
```

doc/pypeit_par.rst
```diff
@@ -661,22 +661,22 @@ ReduxPar Keywords
...
+``qadir``               str             ..       ``QA``                                        Directory relative to calling directory to write quality assessment files.                                                                                                                                                                                                                                                                                                                                
+``quicklook``           bool            ..       False                                         Run a quick look reduction? This is usually good if you want to quickly reduce the data (usually at the telescope in real time) to get an initial estimate of the data quality.                                                                                                                                                                                                                           
+``redux_path``          str             ..       ``/Users/westfall/Work/packages/pypeit/doc``  Path to folder for performing reductions.  Default is the current working directory.
```
tbowers7: Consider changing these instances to something static when building the docs so that these sections don't change based on **_who_** compiles the docs.

pypeit/core/wavecal/autoid.py
```diff
@@ -29,7 +29,8 @@
 from pypeit.core import pca
 from pypeit import utils
 
-from pypeit import msgs
+from pypeit import log
+from pypeit import PypeItError
```
tbowers7, line 33: In the PR description, you indicate `from pypeit import log, PypeItError` as one line.  I know consistency on this point is pedantic, but thought I'd ask.

pypeit/core/wavecal/waveio.py
```diff
@@ -9,7 +9,8 @@
 import linetools.utils
 import numpy as np
 
-from pypeit import msgs
+from pypeit import log
+from pypeit import PypeItError
```
tbowers7, line 13: I'm sure we'll see this again, so I'll stop commenting past here.

pypeit/core/flat.py
```diff
@@ -15,7 +15,7 @@
 
 from IPython import embed
 
-from pypeit import msgs
+from pypeit import log
```
tbowers7, line 18: Add `PypeItError`

pypeit/core/qa.py
```diff
@@ -10,9 +10,9 @@
 
 from IPython import embed
 
-# CANNOT INCLUDE msgs IN THIS MODULE AS
```
tbowers7, line 14: Could consider placing the html-generation calls in a separate module, so that this module can `import log`.

pypeit/core/wave.py
```diff
@@ -18,7 +18,8 @@
 
 
 
-from pypeit import msgs
+from pypeit import log
+from pypeit import PypeItError
```
tbowers7: Technically, this module doesn't even need `import log`.

pypeit/images/detector_container.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 
 from pypeit import datamodel
-from pypeit import msgs
+from pypeit import log
```
tbowers7: Not used.

pypeit/par/util.py
```diff
@@ -10,8 +10,8 @@
 
 from configobj import ConfigObj
 
-from pypeit import msgs, __version__
-from pypeit.pypmsgs import PypeItError
+from pypeit import log, __version__
```
tbowers7: Unused.

pypeit/spectrographs/aat_uhrf.py
```diff
@@ -227,7 +228,7 @@ def check_frame_type(self, ftype, fitstbl, exprng=None):
         if ftype in ['arc', 'tilt']:
             return good_exp
 
-        msgs.warn('Cannot determine if frames are of type {0}.'.format(ftype))
+        log.debug('Cannot determine if frames are of type {0}.'.format(ftype))
```
tbowers7, line 233: Nice.  As warnings, these weren't terribly helpful for the user.

pypeit/tests/test_pydl.py
```diff
@@ -1,12 +0,0 @@
-"""
```
tbowers7, line 1: Ought we test these?  Are they nice and stable and don't need testing?

pypeit/coadd2d.py
```diff
@@ -1606,16 +1607,16 @@ def snr_report(self, slitid, spat_pixpos, snr_bar):
...
+        msg_string += '-------------------------------------'
+        msg_string += '       exp#   spat_pixpos     S/N'
+        msg_string += '-------------------------------------'
         for iexp, (spat,snr) in enumerate(zip(spat_pixpos, snr_bar)):
-            msg_string += msgs.newline() + '       {:2d}      {:7.1f}      {:5.2f}'.format(iexp, spat, snr)
-        msg_string += msgs.newline() + '-------------------------------------'
-        msgs.info(msg_string)
+            msg_string += '       {:2d}      {:7.1f}      {:5.2f}'.format(iexp, spat, snr)
+        msg_string += '-------------------------------------'
+        log.info(msg_string)
```
tbowers7: Add `\n` to replace the `msgs.newline()`

pypeit/exceptions.py
```diff
@@ -0,0 +1,22 @@
+"""
```
tbowers7, line 1: Nice!

pyproject.toml
```diff
@@ -68,11 +68,12 @@ test = [
     "coverage",
 ]
 docs = [
-    "sphinx>=1.6,<8",
+    "sphinx>6",
     "docutils<0.22",
     "sphinx-automodapi",
     "sphinx-design",
-    "sphinx_rtd_theme==2.0.0",
+    "sphinx_rtd_theme==3.0.0",
```
tbowers7, line 75: Any reason to peg at `3.0.0` and not allow it to float `>=3`?

pypeit/__init__.py
```diff
@@ -5,88 +5,37 @@
...
+## Send all signals to messages to be dealt with (i.e. someone hits ctrl+c)
+#def signal_handler(signalnum, handler):
+#    """
+#    Handle signals sent by the keyboard during code execution
+#    """
+#    if signalnum == 2:
+#        log.info('Ctrl+C was pressed. Ending processes...')
+#        sys.exit()
+#
+#signal.signal(signal.SIGINT, signal_handler)
```
tbowers7, line 40: Can uncomment these, in conjunction with adding an `atexit.register()` in, say, the QA module or PypeIt module that registers the `close_qa()` function to be handled upon exit.

I think it would be very helpful to catch the Ctrl+C and gracefully exit rather than puking with the `KeyboardInterrupt` exception + traceback.

pypeit/logger.py
```diff
@@ -0,0 +1,381 @@
...
+        log_file_level
+            The logging level specific to the log file.  If None, adopt the
+            console logging level.
+        """
+        # NOTE: I originally included these as options in the class.  I've
+        # removed them for now (i.e., we'll always catch warnings and
+        # exceptions), but I've left the if statements in place below in case we
+        # want to make these things options in the future.
+        capture_exceptions = True
+        capture_warnings = True
```
tbowers7, line 185: Good call.

pypeit/logger.py
```diff
@@ -0,0 +1,381 @@
...
+        if len(value.args) > 0:
+            message = f"{etype.__name__}: {str(value)}"
+        else:
+            message = str(etype.__name__)
+
+        # Log the error
+        self.error(message, exc_info=exc_info)
+
+        # Call the original exception hook
+        self._excepthook_orig(etype, value, trace)
```
tbowers7, line 270: Does this function mean that we log errors but still allow them to crash the code?

Is there a more graceful way to print useful traceback without a full-on crash?

pypeit/scripts/scriptbase.py
```diff
@@ -113,15 +117,14 @@ def entry_point(cls):
...
+    # TODO: This could also be combined with the property decorator; see
+    # https://docs.python.org/3.9/library/functions.html#classmethod
+    # Order matters.
     #
     # @classmethod
     # @property
     #
-    # Because we're not requiring python 3.9 yet, we have to leave this as a
-    # classmethod only:
+    # Leave as is for now.
```
tbowers7: This was deprecated again in Python 3.11: https://docs.python.org/3.13/library/functions.html#classmethod

pypeit/scripts/scriptbase.py
```diff
@@ -177,30 +182,100 @@ def get_parser(cls, description=None, width=None,
...
+            _log_file = cls.default_log_file()
+        elif args.log_file in ['None', None]:
+            _log_file = None
+        else:
+            _log_file = args.log_file
+        log.init(level=level,
+                 log_file=_log_file,
+                 log_file_level=log_file_level)
+
+    @staticmethod
```
tbowers7: Nice!

pypeit/scripts/scriptbase.py
```diff
@@ -177,30 +182,100 @@ def get_parser(cls, description=None, width=None,
...
+                raise PypeItError(f'Verbosity level must be 0, 1, or 2, not {v}.')
+
+    @classmethod
+    def default_log_file(cls):
+        """
+        Set the default name for the log file.
+        """
+        # Create a UT timestamp (to the minute) for the log filename
+        timestamp = datetime.datetime.now(datetime.UTC).strftime("%Y%m%d-%H%M")
+        return f'{cls.name()}_{timestamp}.log'
```
tbowers7, line 255: Nice.

tbowers7 (approved):
This is a great advancement for PypeIt.  There were several small things I noticed, but nothing major.  Merging other PRs into this one will be challenging and will require patience.  Also, there was a blanket addition of `log` and `PypeItError` to files that previously had `msgs`, but both were not needed all of the time.  May be worth a fine-tooth comb to remove spurious imports, by maybe not?

The only request I have is to re-include the Ctrl+C catcher and provide a graceful shutdown.  With the use of `atexit.register()`, it should be possible to close out QA's and logging gracefully before `sys.exit()`-ing.

doc/conf.py
kbwestfall: Good call.  Added.

doc/pypeit_par.rst
kbwestfall: Good call.  These should now appear as `$PWD`.  See the changes to `parset.py`

pypeit/core/wavecal/autoid.py
kbwestfall: Yeah, I've since decided I prefer one line for each imported module, but I could be convinced otherwise.  Leaving for now.

pypeit/core/wavecal/waveio.py
kbwestfall: Same response.

pypeit/core/flat.py
kbwestfall: Good catch!  Fixed.

pypeit/core/qa.py
kbwestfall: Actually, this points out something we need to revisit.  This PR removes the automatic construction of the html file.  The `pypeit_qa_html` script still exists and can build it *after* `run_pypeit` is finished, but it's not longer par of the logging system.  Let's discuss this on a development call and/or over Slack.

pypeit/core/wave.py
kbwestfall: Good call!  Removed.

pypeit/images/detector_container.py
kbwestfall: Good catch!  Removed.

pypeit/par/util.py
kbwestfall: Good call!  I also removed the numpy import (also not used).

pypeit/scripts/scriptbase.py
kbwestfall: Thanks for pointing that out!  I have removed the comment.

pypeit/tests/test_pydl.py
kbwestfall: No, we definitely need tests, but this file was useless.  The fact that it existed gave me a false sense that we were actually testing the pydl module...

pypeit/__init__.py
kbwestfall: I've never used the `atexit` package, but it seems pretty useful.  I also don't have a good sense of why we should prefer to exit gracefully (I'm not sure what that would look like) when someone intentionally kills the program using Ctrl+C.

Let's discuss at the next telecon and/or over Slack.

pypeit/coadd2d.py
kbwestfall: Good catch!  Done.

pypeit/logger.py
kbwestfall: Right, the purpose of this function is basically to add the log entry before executing the same exception code that would have been executed without the bespoke logger.

Similar to a comment above, I'm not sure what you mean by exiting more gracefully, but let's discuss.

pyproject.toml
kbwestfall: Yeah, the RTD theme can be picky about the docutils and sphinx versions.  So rather than being caught out by those, I have typically pinned this package to a specific version and keep track of new releases.  It's more maintenance overhead, but the docs can fail pretty silently (i.e., someone has to notice that the webpages aren't right) and this is my attempt to limit those silent failures.

There's a way we can add a doc compilation check to the GitHub CI (e.g., astropy does it), but I haven't figured out how to implement that yet.

kbwestfall (commented):
Thanks for the detailed review, @tbowers7 !  I'd like to talk through how we exit because I don't understand the details well enough.  Let's talk at the next development telecon about that and/or over Slack.

pypeit/cache.py
```diff
@@ -260,26 +260,26 @@ def fetch_remote_file(
...
             )
 
         elif filetype == "arc_lines/lists":
             err_msg = (
-                f"Cannot find local arc line list {filename}{msgs.newline()}"
-                f"Use the script `pypeit_install_linelist` to install{msgs.newline()}"
-                f"your custom line list into the cache.  See instructions at{msgs.newline()}"
+                f"Cannot find local arc line list {filename}\n"
+                f"Use the script `pypeit_install_linelist` to install"
+                f"your custom line list into the cache.  See instructions at"
```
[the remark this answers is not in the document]
kbwestfall: Carriage returns aren't needed here.

pypeit/cache.py
```diff
@@ -260,26 +260,26 @@ def fetch_remote_file(
...
             )
 
         elif filetype == "extinction":
             err_msg = (
-                f"Cannot find local extinction file {filename}{msgs.newline()}"
-                f"Use the script `pypeit_install_extinctfile` to install{msgs.newline()}"
-                f"your custom extinction file into the cache.  See instructions at{msgs.newline()}"
+                f"Cannot find local extinction file {filename}\n"
+                f"Use the script `pypeit_install_extinctfile` to install"
+                f"your custom extinction file into the cache.  See instructions at"
```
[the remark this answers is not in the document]
kbwestfall: Carriage returns aren't needed here.

pypeit/cache.py
```diff
@@ -288,19 +288,19 @@ def fetch_remote_file(
...
-                f"Error downloading {filename}: {error}{msgs.newline()}"
-                f"URL attempted: {remote_url}{msgs.newline()}"
-                f"If the error relates to the server not being found,{msgs.newline()}"
-                f"check your internet connection.  If the remote server{msgs.newline()}"
-                f"name has changed, please contact the PypeIt development{msgs.newline()}"
+                f"Error downloading {filename}: {error}\n"
+                f"URL attempted: {remote_url}\n"
+                f"If the error relates to the server not being found,"
+                f"check your internet connection.  If the remote server"
+                f"name has changed, please contact the PypeIt development"
```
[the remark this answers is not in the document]
kbwestfall: Carriage returns aren't needed here.

kbwestfall:
Dev-suite tests started.

kbwestfall:
Dev-suite restarted.

kbwestfall:
Dusty has the magic touch.  The tests finally went all the way through.

```
Test Summary
--- PYTEST UNIT TESTS PASSED  156 passed, 1573 warnings in 548.25s (0:09:08) ---
--- PYTEST VET TESTS FAILED  1 failed, 72 passed, 1606 warnings in 3059.76s (0:50:59) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 291/291 TESTS  ---
Total disk usage: 362.02 GiB
Testing Started at 2026-01-29T19:55:40.479705
Testing Completed at 2026-01-30T12:46:36.065432
Total Time: 16:50:55.585727
```

The vet-test failure was because I missed a use of the now defunct `verbosity` kwarg in `pypeit.pypeit.PypeIt`.  I fixed that and it now runs successfully:

```
% pytest test_skysub.py::test_
skysub
platform darwin -- Python 3.13.11, pytest-8.4.2, pluggy-1.6.0
PyQt6 6.9.1 -- Qt runtime 6.9.2 -- Qt compiled 6.9.0
rootdir: /Users/westfall/Work/packages/PypeIt-development-suite
configfile: pyproject.toml
plugins: astropy-0.11.0, mock-3.15.1, hypothesis-6.142.1, remotedata-0.4.1, filter-subpackage-0.2.0, doctestplus-1.4.0, astropy-header-0.2.2, qt-4.5.0, arraydiff-0.6.1, cov-7.0.0
collected 1 item

test_skysub.py .                                                         [100%]

=============================== warnings summary ===============================
vet_tests/test_skysub.py: 72 warnings
  RankWarning: The fit may be poorly conditioned

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================= 1 passed, 72 warnings in 46.52s ========================
```

Merging.

PR #2029, opened 2025-10-28
Slitless Flats HOTFIX

debora-pe:
HOTFIX for when the slitless flat is created for a dataset that has mask design information.
This is very simple change. I don't think we need to run the Dev Suite.


kbwestfall (approved):
Looks good!

bpholden (approved):
I wish all fixes were this easy.

debora-pe:
Merging this.

PR #2032, opened 2025-10-31
Replace all instances of ``np.invert`` with ``np.logical_not``

tbowers7:
As titled.  Goes along with the ``np.in1d`` --> ``np.isin`` changes.


profxj (approved):
thanks for this!

PR #1929, opened 2025-05-21
Kcwi dec 2024

jhennawi:
This PR involves major upgrades to KCWI and specifically KCRM functionality.

PR #2034, opened 2025-11-12
Merges release and develop into #1929

kbwestfall:
As titled.

The primary files with conflicts were `pypeit/core/datacube.py` and `pypeit/coadd3d.py`.


jhennawi (approved):
Looks great. Thanks for doing this.

PR #2039, opened 2025-12-06
Edits to Refactor PypeIt

debora-pe:
I think I have resolved most of the problems.
We need to consult @rcooke-ast for some IFU specific code that is hard to fit in the current flow.


pypeit/pypeit_steps.py
```diff
@@ -678,29 +710,35 @@ def extract_det(spectrograph, fitstbl, par,
...
+            sobjs_obj
         slitgpm = (slits.mask == 0)
-        slitshift = objFind.slitshift
-        # If waveImg has not yet been created, make it now
-        if waveImg is None:
-            waveImg = caliBrate.wv_calib.build_waveimg(tilts, slits, spat_flexure=objFind.spat_flexure_shift)
+        # Check these below
+        tilts = caliBrate.wavetilts.fit2tiltimg(slits.slitmask, flexure=sciImg.spat_flexure - caliBrate.wavetilts.spat_flexure)
+        waveImg = caliBrate.wv_calib.build_waveimg(tilts, slits, spat_flexure=sciImg.spat_flexure)
+        slitshift = objFind.slitshift #TODO: this is not correct. Find a way to propagate this information from objFind
```
rcooke-ast: So, the `slitshift` and the `scaleImg` are both determined from the sky model generated in the background fitting. The `slitshift` is effectively the same thing as finding/applying the spectral flexure for each slit. The `scaleImg` is unique to IFU (as far as I'm aware), and calculates the relative throughout of the sky for each slit. It really shouldn't be needed (i.e. you should be able to get this information from the flats), but for some reason the throughput of KCWI cannot be easily measured, even from external (dome) flats. It may be better to eventually move this to a calibration step, but for now, the best place is the object finding where the sky needs to be iteratively subtracted and scaled.

debora-pe: @rcooke-ast Thank you so much for your input!
The current problem with this new refactoring is that the `objFind` class is not propagated anymore to the extraction class, so all the useful info generated in `objFind` are saved in the`SpecObj` class, `SlitTraceSet` class and the final sky image. See [these lines](https://github.com/pypeit/PypeIt/blob/a6fbcadd1415dce29e8403200b9cd3d97f072339/pypeit/exposure.py#L433-L457).
Do you have any suggestion on how we can propagate  `slitshift` and `scaleImg` without using the `objFind` class?

pypeit/wavecalib.py
```diff
@@ -76,7 +76,13 @@ class WaveCalib(calibframe.CalibFrame):
                                    descr='Echelle order ID numbers.  Defined only for echelle.'),
                  'strpar': dict(otype=str, descr='Parameters as a string'),
                  'lamps': dict(otype=str,
-                               descr='List of arc lamps used for the wavelength calibration')}
+                               descr='List of arc lamps used for the wavelength calibration'),
+                 'flex_shift': dict(otype=np.ndarray, atype=np.floating, descr='Global spectral shift (pixels) of '
```
rcooke-ast: I might be missing something, but this correction depends on the science frame, and I think the slitshift is being written to the calibration information (which should be independent of the science frame)?

debora-pe: Good point @rcooke-ast!

pypeit/exposure.py
```diff
@@ -37,7 +37,7 @@ def adjust_for_slitmask(sciImg_dict:dict, spectrograph, fitstbl, par,
             Spectrograph object
         fitstbl (:class:`~pypeit.metadata.PypeItMetaData`):
             The class holding the metadata for all the frames in this PypeIt run.
-        par (:class:`~pypeit.par.pypeitpar.CalibrationsPar`):
+        par (:class:`~pypeit.par.pypeitpar.PypeItPar`):
```
profxj, line 40: good catch

profxj (approved):
thanks for this!!

debora-pe:
@rcooke-ast thank you so much for your input here. It's very helpful. I moved `flex_shift` from wavecalib to sciimg. Do you mind taking a look at those changes? Also, is there any test in the Dev Suite that tests the `slitshift` and `scaleImg`? If not, can you suggest a test that I can add? Thanks!

pypeit/scripts/reduce_by_step.py
```diff
@@ -156,75 +158,99 @@ def main(args):
...
-                    sciImg_dict, pypeIt.spectrograph, pypeIt.fitstbl, pypeIt.par, 
-                    frames[0], pypeIt.fitstbl['binning'][frames[0]],
-                    specobjs_objfind, calib_slits)
-                # Update this_calib_silts
-                idx = mosaics.index(det)
-                this_calib_silts = calib_slits[idx]
-
-            # Proceed with extraction
-
+            # Load a lot of stuff
```
rcooke-ast: Perhaps change this to `Load frames related to the science exposure`

pypeit/exposure.py
rcooke-ast: Perhaps add the additional two returned items?

pypeit/exposure.py
```diff
@@ -199,21 +209,59 @@ def findobj_on_exposure(sciImg_dict:dict, spectrograph, fitstbl, par,
...
+                     det, this_objfind, initial_sky, all_specobjs_objfind,
+                     bkg_redux_sciimg=bkg_redux_sciimg, bkg_redux=bkg_redux, show=show)
+
+        # store the final skies
+        final_sky_dict[det] = final_global_sky
+        bkg_redux_final_sky_dict[det] = bkg_redux_global_sky
+        # Update the sciImg with the scaleImg information
+        sciImg_dict[det].rel_scaleImg = this_objfind.scaleimg
+        # and the global spectral flexure shift
+        sciImg_dict[det].flex_shift = this_objfind.slitshift
```
rcooke-ast, line 254: I can't quite follow the logic to see if this `slitshift` should be assigned or added to the value already in the `sciImg`. This might need to be checked, or you can add a TODO item for me to check 👍

rcooke-ast (approved):
This all looks good to me, thanks @debora-pe for making the many changes! I only have a few minor things to consider. I don't believe there is currently a test that will ensure these features still hold for the IFU data reductions, but adding more tests is on my TODO list.

pypeit/scripts/reduce_by_step.py
debora-pe: Changed

pypeit/exposure.py
debora-pe: Thanks! Done.

pypeit/exposure.py
debora-pe: Added the TODO! Thanks!

kbwestfall:
@debora-pe Just a note to request that you merge in `develop` and resolve any conflicts before you merge this PR, which should bring `refactor_pypeit` up-to-date at the same time.

debora-pe:
merging

PR #2040, opened 2025-12-09
APF frame typing

bpholden:
This additional automatically modifies the pypeit file based on the setup of the instrument. This allows the user to have
a FITS image with one classification for the "long" slit setup and a different classification for the "short" slit setup.
Because all frame typing is local (the instrument setups are built up from the types) this is done by calling a spectrograph
specific method that modifies the fitstbl object just before writing out that object to the  pypeit file.


pypeit/spectrographs/apf_levy.py
```diff
@@ -464,3 +464,16 @@ def get_rawimage(self, raw_file, det, sec_includes_binning=True):
...
+        """
+        Correct the table frametype values for the given setup, if necessary.
+        """
+
+        is_wideflat = table['frametype'] == 'pixelflat,trace'
+        is_narrowflat = table['frametype'] == 'trace'
+        if np.any(is_narrowflat) and setup['decker'] == '3.0':
+            table['frametype'][is_wideflat] = 'pixelflat'
+
+        return
```
kbwestfall: You don't need the return statement.

pypeit/spectrographs/spectrograph.py
```diff
@@ -962,6 +962,13 @@ def config_independent_frames(self):
         """
         return {'bias': 'binning', 'dark': 'binning'}
 
+    def final_config_frametypes(self, setup, table):
+        """
+        Correct the table frametype values for the given setup, if necessary.
+        """
+
+        return
```
kbwestfall: Given that the function never returns anything (for any spectrograph class), I think I would use `pass` here instead.  But I'm interested in other opinions.

pypeit/metadata.py
```diff
@@ -1771,6 +1771,9 @@ def write_pypeit(self, output_path=None, cfg_lines=None,
 
             # Get the data lines
             subtbl = self.table[output_cols][in_cfg]
+
+            self.spectrograph.final_config_frametypes(setup_dict[f'Setup {setup}'], subtbl)
```
kbwestfall, line 1775: I'll preface this by saying that I'm not sure what I'm about to say matters in a practical sense:

I'm unsure about the fact that this effectively means the `fitstbl` for APF is always wrong until it's written to and read from a pypeit file.

In trying to find if there was a different/better place to use this new `spectrograph` function, I came across the `vet_assigned_ftypes` function.  Would this function work for what you're trying to accomplish?  The whole `PypeItMetaData` object is passed to the function so it will have access to the configurations, and you can loop over them in the APF-specific function.

kbwestfall (changes requested):
I have a couple minor questions about some `return` statements, but my main question is whether or not we could use the `vet_assigned_ftypes` function instead of adding a new function.  Sorry if we already talked this through and I forgot!

pypeit/metadata.py
bpholden: Yes, a fitstbl is not correct until it is written, which is why we have so many existing functions that correct that object. The whole point of `vet_assigned_types` is to address the potential confusion between a standard frame and a science frame. 

The larger issue with `vet_assigned_types` is that it is called as part of `get_frame_types` which is in the `finalize_usr_build` method. That is called BEFORE `set_configurations` which is where the setups are determined. 

The `final_config_frametypes` needs to be called AFTER the `set_configurations` is called. 

The fundamental issue is that Wideflat images are used as only pixel flats for the short slit configurations but are used as trace and pixel flats for the long slit cofigurations. Until you know what configuration you are in, you do not know what frame types you need.

kbwestfall: Got it.  Thanks.  Given that, I'm fine with letting this go, but I think we should put this on the growing "we should refactor this" list.  Your PR is a case when we need the configurations before we can properly set the frame types, but there are other cases when we need the frame types to properly count configurations and assign frames to them (e.g., bias frames are mostly configuration independent).

kbwestfall (approved):
Thanks for the quick response.  We should probably run the tests, but I'm approving now.

debora-pe (approved):
I have a vague feeling that the code in `final_config_frametypes` could be included in `check_frame_type`, but I may be wrong. I don't have any opposition here. I'm approving.

bpholden:
The suggestion of check_frame_type is similar to Kyle's, as it is run in
get_frame_types.
What is needed to make these changes are the setups, which are only
computed after the frames are all typed.
There is this egg/chicken problem.

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  281 passed, 465 warnings in 334.83s (0:05:34) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 93617 warnings in 815.75s (0:13:35) ---
--- PYTEST VET TESTS PASSED  70 passed, 115248 warnings in 4708.71s (1:18:28) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 284/284 TESTS  ---
Testing Started at 2025-12-09T19:50:28.564930
Testing Completed at 2025-12-10T07:58:08.223391
Total Time: 12:07:39.658461
```

PR #2050, opened 2026-01-06
Allow for different sizes on det 1 and 2

bpholden:
The `get_rawimage()` method used for the `keck_lris_red` spectrometer (not the original red nor the current Mark IV) 
handles windowed images incorrectly. 
For example, when reading in the following windowed 1x1 image from the dev suite 
(data taken with LRIS-R Science mosaic Mark 2: CCD  1: 19-3  CCD 2: 19-2)
`long_400_8500_longread/LR.20181003.56288.fits.gz`
there are two image extensions 
1 (4116, 635)
2 (4116, 739)
Running `get_rawimage()` on that frame and retrieving the dat for detector 1 and 2 yields out data with following shapes:
1 (4116, 687)
2 (4116, 687)

This PR fixes that while maintaining the rest of the functionality (such as correctly mapping the individual amplifiers and
making a full image with the `det=None` option.)

### DIFF
--- a/pypeit/spectrographs/keck_lris.py
+++ b/pypeit/spectrographs/keck_lris.py
@@ -580,19 +580,19 @@ class KeckLRISSpectrograph(spectrograph.Spectrograph):
             if 'VidInp' in ihdu.name:
                 extensions.append(kk)
         n_ext = len(extensions)
-        xcol = []
+        xcol = np.zeros(n_ext, dtype=int)
         xmax = 0
         ymax = 0
         xmin = 10000
         ymin = 10000
-        xmins = []
-        xmaxs = []
-        ymins = []
-        ymaxs = []
+        xmins = np.zeros(n_ext, dtype=int)
+        xmaxs = np.zeros(n_ext, dtype=int)
+        ymins = np.zeros(n_ext, dtype=int)
+        ymaxs = np.zeros(n_ext, dtype=int)
 
-        for i in extensions:
+        for i, ext in enumerate(extensions):
 
-            theader = hdu[i].header
+            theader = hdu[ext].header
             detsec = theader['DETSEC']
             if detsec != '0':
                 # parse the DETSEC keyword to determine the size of the array.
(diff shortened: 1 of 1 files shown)


kbwestfall (approved):
Looks good to me.  Let's run the dev-suite tests for `keck_lris`.

pypeit/spectrographs/keck_lris.py
```diff
@@ -585,8 +585,13 @@ def get_rawimage(self, raw_file, det):
         ymax = 0
         xmin = 10000
         ymin = 10000
+        xmins = []
```
rcooke-ast: This is all good, but I wonder if it's a bit cleaner to initialise these as arrays, and then you could remove lines 616-619. I don't think this is absolutely necessary, I'm just not a fan of lists! I've made a few suggested changes, in case you agree with this, otherwise, feel free to leave it as is.

```
xmins, xmaxs = np.zeros(len(extensions)), np.zeros(len(extensions))
ymins, ymaxs = np.zeros(len(extensions)), np.zeros(len(extensions))
```

pypeit/spectrographs/keck_lris.py
```diff
@@ -585,8 +585,13 @@ def get_rawimage(self, raw_file, det):
...
+        xmaxs = []
+        ymins = []
+        ymaxs = []
 
         for i in extensions:
```
rcooke-ast: `for i, ext in enumerate(extensions):`

pypeit/spectrographs/keck_lris.py
```diff
@@ -585,8 +585,13 @@ def get_rawimage(self, raw_file, det):
...
+
             theader = hdu[i].header
```
rcooke-ast: `theader = hdu[ext].header`

pypeit/spectrographs/keck_lris.py
```diff
@@ -596,18 +601,39 @@ def get_rawimage(self, raw_file, det):
                 # find the range of detector space occupied by the data
                 # [xmin:xmax,ymin:ymax]
                 xt = max(x2, x1)
-                xmax = max(xt, xmax)
+                xmaxs.append(max(xt, xmax))
```
rcooke-ast: `xmaxs[i] = max(xt, xmax)`

(and similarly for the other instances)

rcooke-ast (approved):
Approving with a minor comment that you can ignore. It's a personal preference to intitialise arrays from the beginning if the size is known _a priori_. So, I don't mind you ignoring all these suggestions, and I'm approving!

pypeit/spectrographs/keck_lris.py
bpholden: OK, I have done this, and changed the `xcol` list (which was original) into an array as well. Note this requires `dtype=int`

pypeit/spectrographs/keck_lris.py
bpholden: I did this because one has to but this does seem odd to have two integers that are just off by one.

pypeit/spectrographs/keck_lris.py
bpholden: I wrote this as += which is an old habit.

kbwestfall: Sorry to be annoying, but why are you using `+=`?  I realize the integers start out at 0, so I guess it shouldn't matter, but can you just use `=`?

bpholden:
This is a leftover from learning about views vs copies.
A += forces a change but I could just make it = if you really want to be
picky and promise me it will work.

kbwestfall:
Personally, I'd prefer it if you use `=`, and I can promise it will work.

bpholden:
Updated.

PR #2056, opened 2026-01-22
updated np.is1d to np.isin

Cnmoya:
Was trying to reduce Magellan FIRE spectra an ran into an error with the np.is1d command when running run_pypeit. The error specifically mentioned that np.isin was the correct function to use. Turns out np.is1d was deprecated and np.isin is the way to go now. This was in a fresh installation of pypeit using an anaconda environment with python 3.11.14


tbowers7:
Hi @Cnmoya .  Thank you.  We are aware of this issue and it will be fixed with the imminent release of PypeIt 2.0.

kbwestfall:
Thanks for the effort, @Cnmoya !  Like @tbowers7 said, we have fixed this in an upcoming release branch.  Feel free to jump on the `develop` branch until PypeIt 2.0 is released.

PR #2058, opened 2026-01-29
Deprecate linetools

profxj:
Strip `linetools` out of `pypeit`, for once and for all. :)

I am having one test fail on my workstation, which I think
is unrelated.  Let's see how CI goes..

This PR also contains new Claude Code related bits and pieces.
Nothing in the main code base, but likely useful to those who use it.

PR #2059, opened 2026-01-30
Support for python 3.14

kbwestfall:
Bumps the python support limit to include 3.14, and adds python3.14 tests to the CI matrix.

Note this includes edits from #2058.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,8 +3,8 @@
 
 [tox]
 envlist =
-    {3.11,3.12,3.13}-test{,-alldeps}{,-cov}
-    {3.11,3.12,3.13}-test-{numpy,astropy,ginga}dev
+    {3.11,3.12,3.13,3.14}-test{,-alldeps}{,-cov}
+    {3.11,3.12,3.13,3.14}-test-{numpy,astropy,ginga}dev
     codestyle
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -3,8 +3,8 @@
 
 [tox]
 envlist =
-    {3.11,3.12,3.13}-test{,-alldeps}{,-cov}
-    {3.11,3.12,3.13}-test-{numpy,astropy,ginga}dev
+    {3.11,3.12,3.13,3.14}-test{,-alldeps}{,-cov}
+    {3.11,3.12,3.13,3.14}-test-{numpy,astropy,ginga}dev
     codestyle
 requires =
     setuptools >= 65.0
```
kbwestfall, line 10: Should we bump this?  Current version of setuptools is 80.10.x.

tbowers7:
Testing like this asks the question, "What is in our code that is no longer supported by recent Python versions?"  This is important, and is related to the upper bound of supported versions.

We should also be thinking toward the question, "What is not in our code that is now supported by recent Python versions?"  There may be efficiencies or clearer ways of doing things that are available now.  This second question depends on the lower bound of supported versions.

Issue #2062, opened 2026-02-05
Ginga 1d viewer: Problem showing `--object SPATnnnn-SLITnnnn-DETnn`


tbowers7:
I am having two (likely distinct) problems with the Ginga 1d viewer, in particular as it exists when launched from the `deprecate_linetools` branch:

1. (More problematic) When I request a particular object using the `--object SPATnnnn-SLITnnnn-DETnn` argument, the Ginga viewer never refreshes from the tabular view to the actual drawn spectrum (as it does without this argument).

2. (Annoying, but tolerable) When a new Ginga viewer is opened using the `pypeit_show_1dspec` command, the default "Image" pane is loaded, and the Spec1dView pane is loaded as a second panel.  The "Image" pane must be removed in order to show the spectrum in the entirety of the viewer.

Example:
<img width="1440" height="814" alt="Image" src="https://github.com/user-attachments/assets/a273d3c4-74fe-47c2-87a2-b3c3b8c6cc36" />

The preferred outcome would be a full-viewer spectrum being displayed.

---

Note: When the `--exten n` option is used instead of `--object SPATnnnn-SLITnnnn-DETnn`, the table is converted into a plot, but the "Image" pane is still extant.

Example:

<img width="1440" height="814" alt="Image" src="https://github.com/user-attachments/assets/e6f97b97-a88f-4573-a795-10895684d98f" />

ejeschke:
Regarding #2, if ginga is invoked with the `--channels` (or `-c`) option with a list of the names of channels, it will only create those.  Probably just need to modify the script that is invoking ginga to add that.

tbowers7:
> Regarding [#2](https://github.com/pypeit/PypeIt/pull/2), if ginga is invoked with the `--channels` (or `-c`) option with a list of the names of channels, it will only create those. Probably just need to modify the script that is invoking ginga to add that.

This is where `ginga` is called.  Any suggestions @ejeschke ?

https://github.com/pypeit/PypeIt/blob/824328774afdeb2f225507d10c8d2a08add22a79/pypeit/display/display.py#L676-L699

ejeschke:
Actually, I have a better fix.  In `pypeit.display.display`, change the `show_1dspec` function at the bottom as follows:
```python
def show_1dspec(filename, ext=0, masked=True, fluxed=False, extraction='OPT',
                clear=True):
    """
    Interface to ginga to show a 1dspec and manipulate with Spec1dView plugin.

    Parameters
    ----------
[...]

PR #2063, opened 2026-02-05
Fix an issue where --obj with pypeit_show_1dspec wasn't working with …

ejeschke:
Fixes an issue when using `pypeit_show_1dspec` with `--ginga` and the `--obj` option doesn't show the correct HDU extension.

Partial fix for #2062


tbowers7 (approved):
Thanks for this!  It works for me to display a spectrum using the `--obj` option, and I'm glad the fix was so easy!

kbwestfall (approved):
Thanks for the quick fix!

ejeschke:
Rebased

PR #1940, opened 2025-06-06
Eris dev

wmorgali:
Integrating work done to create pipeline for VLT/Eris. For followup, Shane Bechtel and Joe Hennawi will be point of contact.

PR #2067, opened 2026-02-06
Adds support for VLT/ERIS

kbwestfall:
As titled.

Re-issue of PR #1940.

Comments in that PR should be address in this branch.


debora-pe:
I did some tests on this PR to understand the status. It is definitely non ready to be merged nor to be consider as supported by PypaIt.
`pypeit_setup` works (I think) OK, but the reduction is not good and it crashes. We either keep this PR open indefinitely, or we merge it as a "NOT SUPPORTED" spectrograph.

kbwestfall:
This PR has been reverted to a draft PR because it is not in active development.  If you have VLT/ERIS data you would like to reduce, we'd be interested to work with you to update pypeit to do so, building from this in initial effort!

Issue #2064, opened 2026-02-06
Ginga 1d viewer: Rescale y axis when switching `Fluxed`, `Extraction`, and `Masked`


tbowers7:
When a new `Extension` is selected, the viewer rescales the y axis to match the new data.  This functionality needs to be added for the `Fluxed`, `Extraction`, and `Masked` drop-downs, as well.

Example of switching from `Fluxed = false` to `Fluxed = true` for a given spectrum:

<img width="1465" height="814" alt="Image" src="https://github.com/user-attachments/assets/a893d6b9-cf0f-4aff-8399-2c5b6efa676e" />

<img width="1465" height="814" alt="Image" src="https://github.com/user-attachments/assets/0095560b-5df1-4fc0-a6a0-0a8d88c08863" />

ejeschke:
Merged

PR #2069, opened 2026-02-07
Fit the Y-axis after the user changes extraction, mask or fluxed

ejeschke:
Zoom fits the Y-axis after the user performs a change to extraction, mask or fluxed.

- possible fix for #2064

### DIFF
--- a/pypeit/display/spec1dview.py
+++ b/pypeit/display/spec1dview.py
@@ -297,6 +297,11 @@ class Spec1dView(GingaPlugin.LocalPlugin):
 
         self.fv.gui_do(self.plot_lines)
 
+    def fit_y(self):
+        """Zoom to fit Y axis"""
+        viewer = self.channel.get_viewer('Ginga Plot')
+        viewer.zoom_fit(axis='y')
+
     def set_extraction_cb(self, w, idx):
         """Callback for changing the `extraction` option in the plugin.
 
@@ -314,6 +319,7 @@ class Spec1dView(GingaPlugin.LocalPlugin):
         self.fluxed = self.fluxed_options[idx]
         self.logger.debug(f"Selected fluxed option: {self.fluxed}")
         self.recalc()
+        self.fit_y()
 
     def set_masked_cb(self, w, idx):
         """Callback for changing the `masked` option in the plugin.


ejeschke:
@tbowers7, does this accomplish what you are suggesting in #2064 ?

I'm wondering if everyone would agree that the plot should automatically be fit to the Y axis.  It might be better to make this a preference with a default (maybe using an options pane or menu with a checkable option).

tbowers7:
> @tbowers7, does this accomplish what you are suggesting in #2064 ?
> 
> I'm wondering if everyone would agree that the plot should automatically be fit to the Y axis. It might be better to make this a preference with a default (maybe using an options pane or menu with a checkable option).

Yes, this is exactly what I was going for, thank you!

In response to your question:
* For the fluxed = true/false, always rescale because these are different units entirely and should not be compared on the same scale (as you already autorescale when a different object is selected).
* For the masked and extraction toggles, I can see arguments for both rescaling and not, since the toggled options share a common physical unit.

pypeit/display/spec1dview.py
```diff
@@ -314,6 +320,7 @@ def set_fluxed_cb(self, w, idx):
         self.fluxed = self.fluxed_options[idx]
         self.logger.debug(f"Selected fluxed option: {self.fluxed}")
         self.recalc()
```
tbowers7, line 321: For changing between fluxed and not, why not use `self.autozoom_plot()`?  Would this reset any scaling of the X-axis, and you are thinking users may want to examine changes in subregions?

tbowers7 (approved):
Thanks, @ejeschke !  Just one small comment, but otherwise this meets what I was looking for.

kbwestfall (approved):
Looks good to me!

We talked about this during the telecon on 9 Feb, and I thought it may be useful to enable users to effectively blink between boxcar and optimal extractions, and spectra with and without masking.  In those cases, we would want the y axis to stay the same.  But it definitely makes sense to automatically rescale when switching between spectra that are and are not flux calibrated.

I'm approving, but interested in anyone's thoughts on the above.  We could add an issue for it that is addressed in a later PR.

pypeit/display/spec1dview.py
ejeschke: > For changing between fluxed and not, why not use `self.autozoom_plot()`? Would this reset any scaling of the X-axis, and you are thinking users may want to examine changes in subregions?

Not being a regular user of spectrographic reductions, I can't comment on what would be appropriate. I'm guessing that if a user is zoomed in to a subregion (some particular feature they are interested in, say an emission line or absorption) and you suddenly rescale the X axis because you've adjusted one of the reduction options that that could become very irritating.  

`autozoom_plot()` would rescale the X axis.

ejeschke:
> We talked about this during the telecon on 9 Feb, and I thought it may be useful to enable users to effectively blink between boxcar and optimal extractions, and spectra with and without masking. In those cases, we would want the y axis to stay the same. 

Hmm, it's better to think this through now and come up with a sensible plan, because it could be difficult/problematic to change the behavior of the viewer later.

pypeit/display/spec1dview.py
```diff
@@ -323,6 +330,7 @@ def set_masked_cb(self, w, idx):
         self.masked = self.masked_options[idx]
         self.logger.debug(f"Selected masked option: {self.masked}")
         self.recalc()
+        self.fit_y()
```
tbowers7: Remove this -- keep same scaling for masked / unmasked.

pypeit/display/spec1dview.py
```diff
@@ -305,6 +310,7 @@ def set_extraction_cb(self, w, idx):
         self.extraction = self.extraction_types[idx]
         self.logger.debug(f"Selected extraction type: {self.extraction}")
         self.recalc()
+        self.fit_y()
```
tbowers7: Remove this -- keep same scaling for optimal / boxcar.

tbowers7 (changes requested):
Here are the two changes discussed at the PypeIt telecon on 23 Feb.

ejeschke:
Modified and rebased against develop

tbowers7 (approved):
Looks great!  Thanks!

tbowers7:
@ejeschke , could you rebase one more time so we can merge?  Thanks!

PR #2071, opened 2026-02-10
Adds a class that creates a ParSet from a function signature

kbwestfall:
As titled.

This isn't used in the code yet, but I will use it in the upcoming refactor of the flux-calibration/telluric code.  This allows us to easily create consistent parameter sets that expose the full list of keyword arguments in functions like `djs_reject`, which get used throughout the code.


kbwestfall:
This PR also now updates `develop` for the recent tag, and bumps the change log doc.

kbwestfall:
Closing for now...

PR #2073, opened 2026-02-17
Updates `develop` for changes to `release` and bumps dev doc

kbwestfall:
As titled.

I decided to separate this from #2071 .

### DIFF
--- a/doc/help/run_pypeit.rst
+++ b/doc/help/run_pypeit.rst
@@ -6,7 +6,7 @@
                       pypeit_file
     
     PypeIt: The Python Spectroscopic Data Reduction Pipeline
-    Version 1.18.2.dev643+g76c0191d0
+    Version 1.18.2.dev637+g6503accfd
     
     Available spectrographs include:
         aat_uhrf, apf_levy, bok_bc, gemini_flamingos1, gemini_flamingos2,
--- a/tox.ini
+++ b/tox.ini
@@ -52,7 +52,8 @@ deps =
 # The following indicates which extras_require from setup.cfg will be installed
 extras =
     test
-    alldeps: specutils, scikit-image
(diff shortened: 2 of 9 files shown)


bpholden (approved):
looks great

bpholden (approved):

doc/help/run_pypeit.rst
```diff
@@ -6,7 +6,7 @@
                       pypeit_file
     
     PypeIt: The Python Spectroscopic Data Reduction Pipeline
-    Version 1.18.2.dev643+g76c0191d0
+    Version 1.18.2.dev637+g6503accfd
```
tbowers7, line 9: Do you want to do the same trick as you do for the other "include" files of rounding the version?  Therefore, this would say "2.0.0"?

tox.ini
```diff
@@ -52,7 +52,8 @@ deps =
 # The following indicates which extras_require from setup.cfg will be installed
 extras =
     test
-    alldeps: shapely,specutils,scikit-image
```
tbowers7, line 55: What was shapely used for (and now isn't)?

tbowers7 (approved):

doc/help/run_pypeit.rst
kbwestfall: Good call, but I'll punt to a different PR.

tox.ini
kbwestfall: I think it was used for some geometry calculations for the IFU data, but the optional dependency was removed at some point.

I haven't dug into it, but it feels like something recently changed with either tox itself or the tox.ini syntax because this was passing not that long ago.  The real change has to do with fixing the syntax; removing shapely was just because I realized we no longer import it anywhere.

PR #2078, opened 2026-02-21
Platescale issue in coadd2d

debora-pe:
As pointed out by a user, when performing coadd2d using the `spat_samp_fact` parameter (i.e., applying a re-binning), the spatial sampling factor was not correctly propagated to objfind and extraction, specifically in the conversion from arcsec to pixels for boxcar and fwhm values.

This PR fixes the problem by changing the value of `platescale` in the 2D coadded data to account for `spat_samp_fact`. This has also the advantage to save the correct value of the pixel sampling in the spec2d and spec1d files for coadded data.

I have added some tests here: https://github.com/pypeit/PypeIt-development-suite/pull/390

I am running the Dev Suite.


kbwestfall (commented):
Thanks for the fix!  Conceptually, I'm wary of directly altering the `platescale` attribute of the detector objects.  Is there a work-around where we can incorporate the sampling factor without having to directly edit the platescale?

debora-pe:
> Thanks for the fix! Conceptually, I'm wary of directly altering the `platescale` attribute of the detector objects. Is there a work-around where we can incorporate the sampling factor without having to directly edit the platescale?

Hi @kbwestfall. Thank you for your comment. 
Yes, I also struggled with that. I first thought to change the `binning` attribute of the detector object. To me it seemed more reasonable, but binning cannot be a float number, so I had to give up that idea. 
 
I then thought to add a new attribute specifically for the resampling in coadd2d, but that requires to make more changes in the whole code to include that attribute when converting from arcsec to pixels, and requires to remember the existence of this new attribute whenever, in future development, we use platescale and binning. To me this seems to make it a little more difficult to properly maintenance the code.

So, then I thought that changing only `platescale` was the more straightforward route. Please, let me know if you think of better alternative to this. Thanks!

kbwestfall:
> So, then I thought that changing only `platescale` was the more straightforward route. Please, let me know if you think of better alternative to this. Thanks!

Okay, let's go ahead with this solution then.  Thanks for thinking this through!

kbwestfall (approved):
Thanks!

debora-pe:
Tests pass - the one vet test failure is the usual ghost i my machine (`FAILED ../test_scripts.py::test_identify - AssertionError: Bad auto ID`).

```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  292 passed, 437 warnings in 295.74s (0:04:55) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 1569 warnings in 1039.47s (0:17:19) ---
--- PYTEST VET TESTS FAILED  1 failed, 73 passed, 1686 warnings in 6475.60s (1:47:55) ---
Total disk usage: 226.596 GiB
Testing Started at 2026-02-21T05:33:10.069878
Testing Completed at 2026-02-21T23:48:55.511563
Total Time: 18:15:45.441685
```

bpholden (approved):
Looks sensible to me.

debora-pe:
Thanks all. I'm merging.

PR #2079, opened 2026-02-22
APF tweaks

bpholden:
There are two small pieces.

1. Small changes to default parameter values.
2. Change a raised exception to a log warning that occurs when running `pypeit_setup` on some APF data.


kbwestfall (approved):
We had multiple discussions about this in person.   Approving.

badpandabear (approved):

debora-pe (approved):
Sorry, I am late on this. Thanks!

Issue #2081, opened 2026-02-24
Circular imports


kbwestfall:
I used AI to help identify circular imports in pypeit.  At first it said there aren't any, but I knew of at least one (pypeit.core.fitting <-> pypeit.core.pydl).  Using that to correct its first pass, it came up with the following circular imports:

```
pypeit.core.fitting -> pypeit.core.pydl -> pypeit.core.fitting

pypeit.specobj -> pypeit.core.flexure -> pypeit.specobjs -> pypeit.specobj

pypeit.core.wavecal.kdtree_generator -> pypeit.core.wavecal.waveio -> pypeit.core.wavecal.kdtree_generator

pypeit.core.wavecal.autoid -> pypeit.core.wavecal.wv_fitting -> pypeit.core.wavecal.autoid

pypeit.specobj -> pypeit.core.flexure -> pypeit.specobj
```

I have a branch in progress that fixes the first one, but we need to fix the other four.  It's also likely worth improving the query/prompt to make sure there aren't any more!

kbwestfall:
```
pypeit.core.fitting -> pypeit.core.pydl -> pypeit.core.fitting
```
Should be solved by #2082

kbwestfall:
```
pypeit.core.wavecal.autoid -> pypeit.core.wavecal.wv_fitting -> pypeit.core.wavecal.autoid
```
Should be solved by #2083

kbwestfall:
```
pypeit.specobj -> pypeit.core.flexure -> pypeit.specobjs -> pypeit.specobj
pypeit.specobj -> pypeit.core.flexure -> pypeit.specobj
```

Should be solved by #2084

kbwestfall:
```
pypeit.core.wavecal.kdtree_generator -> pypeit.core.wavecal.waveio -> pypeit.core.wavecal.kdtree_generator
```

Should be solved by #2085

kbwestfall:
All of these are now fixed, given the PRs listed above.

PR #2082, opened 2026-02-25
Updates `core.pydl`, adds tests, and deals with circular import

kbwestfall:
This PR does two main things:
 - It (modestly) "pythonizes" some of the pydl code, namely `djs_reject` and `djs_laxisnum`/`djs_laxisgen`
 - It removes any use of `pydl.TraceSet` (which was pretty limited) and replaces it with `fitting.PypeItFitCollection`.  The latter is primarily done to avoid a circular import (see #2081).

In the process, I added a large number of tests for `djs_reject`.  These were written with the help of AI *before* any changes were made to the function.  The tests passed both before and after my changes.

Importantly, replacing `pydl.TraceSet` with `fitting.PypeItFitCollection` uncovered a significant bug with how we were using the former:  When calling `fitting.robust_fit` at line 442 in the previous version of `core.pydl`, we were passing it the *normalized* x-coordinate vector *and* the minimum and maximum x values before the normalization.  This means the x coordinates in the fits are significantly messed up.  I can reproduce the fits between `pydl.TraceSet` and `fitting.PypeItFitCollection` (see one of the added tests) when I introduce the same bug.  However, doing the "correct" thing of passing the un-normalized x-coordinates or the normalized minimum and maximum x values leads to significantly different fits.  This PR maintains the existing (bugged) functionality because if we do the "correct" thing, everyone will need to adjust the order of the polynomial used to fit the edge traces.  This is something we should fix, but it represents a significant testing effort.

I've commented some of the previous code just so you can see it, but I'll remove it before this gets merged.


kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  306 passed, 451 warnings in 459.29s (0:07:39) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 1569 warnings in 1029.28s (0:17:09) ---
--- PYTEST VET TESTS PASSED  73 passed, 1668 warnings in 7524.73s (2:05:24) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 291/291 TESTS  ---
Total disk usage: 324.872 GiB
Testing Started at 2026-02-25T01:39:02.752640
Testing Completed at 2026-02-26T03:27:37.639426
Total Time: 1 day, 1:48:34.886786
```

pypeit/core/tracewave.py
```diff
@@ -352,21 +352,25 @@ def trace_tilts_work(arcimg, lines_spec, lines_spat, thismask, slit_cen, inmask=
...
+#                tilts_sub_mask_box = np.ones_like(tilts_sub_mask_box)
+#            # Re-measure using Gaussian weighting and refit
+#            tilts_sub_fit_out, tilts_sub_out, tilts_sub_err_out, tilts_sub_bpm_out, _ \
+#                = trace.fit_trace(sub_img, tilts_sub_fit_out, spat_order,
+#                                  bpm=np.logical_not(sub_inmask.astype(bool)),
+#                                  trace_bpm=np.logical_not(tilts_sub_mask_box),
+#                                  weighting='gaussian', fwhm=fwhm, maxdev=maxdev, niter=3,
+#                                  idx=str(iline), debug=show_tracefits, xmin=0.0,
+#                                  xmax=float(nsub - 1))
+#            tilts_sub_mask_box = moment1d(sub_thismask, tilts_sub_fit_out, fwhm)[0] > 0.99 * fwhm
```
tbowers7, line 373: So, is `gauss` never actually used in the codebase at the moment?  What is/was the purpose of such a refinement, and do we project ever needing it?  Does it make sense to remove the "gauss" input to this function (_i.e._, is it user-settable at all?), or just leave the `NotImplementedError` for the hapless developer who stumbles upon it at some point in the future?

pypeit/tests/test_fitting.py
```diff
@@ -51,6 +54,31 @@ def test_legendre():
     np.testing.assert_allclose(pypeitFit.fitc, np.array([  6.37115652e-01,   6.83317251e-17,
                                                    -6.84581686e-01, -7.59352737e-17]), atol=1e-9)
 
-def test_robust_fit():
-    # NEED A TEST!!
-    pass
```
tbowers7, line 56: Do we not still need a test for `robust_fit`?

pypeit/core/fitting.py
```diff
@@ -21,6 +21,170 @@
 
 from IPython import embed
 
+# TODO:
+#   - Allow xpos to be 1D and ypos to be 2D?
+#   - Make this a data container?
+#   - Parallelize the fits?
```
tbowers7: Are any of these TODO's things you want to include in this PR, or wait for later?

pypeit/core/fitting.py
```diff
@@ -21,6 +21,170 @@
...
+    Allowed functional forms for fitting.
+    """
+
+    def __init__(
+        self, xpos, ypos, ivar=None, gpm=None, func='legendre', order=3, xmin=None, xmax=None,
+        maxiter=10, maxdev=None, lower=None, upper=None
+    ):
+
+        self.xpos = xpos
+        self.nfit = xpos.shape[0]
```
tbowers7, line 95: Out of curiosity, would there be any case when the input is a "transposed" array (_e.g._, with shape `(1,N)` or `(,N)`)?  This particular description of number of points would be invalid.

pypeit/core/fitting.py
```diff
@@ -21,6 +21,170 @@
...
+
+        self.ypos = ypos
+        self.ivar = ivar
```
tbowers7, line 107: Add checks to ensure same dimensionality as `xpos`?

pypeit/core/pydl.py
```diff
@@ -638,120 +640,128 @@ def djs_reject(data, model, outmask=None, inmask=None,
...
+    # errors in the rejection, i.e. compute the rejection threhsold using the
+    # mad.
 
     if upper is None and lower is None and maxdev is None:
-        log.warning('upper, lower, and maxdev are all set to None. No rejection performed since no rejection criteria were specified.')
+        log.warning(
+            'upper, lower, and maxdev are all set to None.  No rejection performed since no '
+            'rejection criteria were specified.'
+        )
+        # TODO: Should the function return here?
```
tbowers7, line 389: I would say "yes", since the whole point of this function is to perform rejection based on these input criteria.  Makes for cleaner execution and reading.

pypeit/core/pydl.py
```diff
@@ -638,120 +640,128 @@ def djs_reject(data, model, outmask=None, inmask=None,
...
     # Create outmask setting = True for good data.
-    #
-    # ToDo JFH: I think it would actually make more sense for outmask be a required input parameter (named lastmask or something like that).
     if outmask is None:
         outmask = np.ones(data.shape, dtype='bool')
-        log.warning('outmask was not specified as an input parameter. Cannot asess convergence of rejection -- qdone is automatically True')
+        log.warning(
+            'outmask was not specified as an input parameter.  Cannot assess convergence of '
+            'rejection -- qdone is automatically True'
+        )
```
tbowers7, line 406: So, does this mean that rejection is still performed, but the "all done" flag is set this pass through because no convergence assessment is possible?

pypeit/core/pydl.py
```diff
@@ -638,120 +640,128 @@ def djs_reject(data, model, outmask=None, inmask=None,
...
     else:
-        if data.shape != outmask.shape:
-            raise ValueError('Dimensions of data and outmask do not agree.')
-    #
-    # Check other inputs.
-    #
-    if model is None:
-        if inmask is not None:
-            outmask = inmask
-        return (outmask, False)
```
tbowers7, line 710: Have we lost this check?  If so, did we need it?

pypeit/core/pydl.py
```diff
@@ -761,60 +771,39 @@ def djs_reject(data, model, outmask=None, inmask=None,
...
-            if upper is not None:
-                upper_chi = np.percentile(chi[igood], upper)
-            else:
-                upper_chi = np.inf
-    #
+            igood &= inmask
+        if np.sum(igood)> 1:
+            lower_chi = -np.inf if lower is None else np.percentile(chi[igood],lower)
+            upper_chi = np.inf if upper is None else np.percentile(chi[igood], upper)
+
```
tbowers7, line 467: Thank you for all of this... it's much easier to read!

pypeit/core/pydl.py
```diff
@@ -945,88 +941,93 @@ def djs_laxisnum(dims, iaxis=0):
...
-        else:
-            raise ValueError("Bad value for iaxis: {0:d}".format(iaxis))
-    else:
-        raise ValueError("{0:d} dimensions not supported.".format(ndimen))
-    return result
-
+    _dims = tuple(np.atleast_1d(dims).tolist())
+    if len(_dims) == 1:
+        return np.zeros(dims, dtype=int)
+    return arange_ndim(dims, axis=iaxis)
```
tbowers7, line 661: !!!!

tbowers7 (approved):
Looks great!

This PR is a good reminder related to "supported versions of Python", where it is important to go back through the codebase and find places where more modern syntax and dependency usage can both clean things up and make them more future-proof.

pypeit/core/fitting.py
kbwestfall: For later, but I'm actually going to remove them.  At the moment, I have no intention to follow through on any of these...

pypeit/core/fitting.py
kbwestfall: Good insight, but I don't think we should check for them.  A shape of `(1,N)` is fine; there will just be one polynomial fit, and I'm not sure you can actually get a shape of `(,N)`.  E.g.:
```
>>> import numpy as np
>>> x = np.arange(10)
>>> x.T.shape
(10,)
```

Leaving this as is, unless you know of a specific case we should check for.

pypeit/core/fitting.py
kbwestfall: Good call!  Added.

pypeit/core/pydl.py
kbwestfall: I agree, but I'm going to punt for now.  The stuff that happens after this is complex enough that there is non-zero effort in figuring out what the return statements should be (ensuring the behavior is unchanged).

pypeit/core/pydl.py
kbwestfall: Yeah, that's my read of it.

pypeit/core/pydl.py
kbwestfall: I basically removed this as an option because I now force `model` to be a numpy array.  I.e., the check that the data and model have the same shape is done without checking that the model is not None.

pypeit/core/tracewave.py
kbwestfall: All good questions, but one's that I didn't want to deal with at the moment :smile: .  We don't use this option in the code (or at least there was nothing in the dev-suite that tripped it), and how the results of this fit are used vs. their intent doesn't make sense.  Specifically, the if-block changes `tilts_sub_mask_box`, but then it continues to use the polynomial results from the fit done before the "Gaussian refinement" (i.e., `tset_out` is from the first fit, not the Gaussian refit).

Given the confusion about this, I decided I would effectively remove the option and see if it caused any havoc.  And it hasn't ... yet.  I think we should leave this in for a while, but I could be convinced to remove the relevant code and comments.

pypeit/tests/test_fitting.py
kbwestfall: Yes, but we need an actual test instead of one that always passes.

kbwestfall (commented):
Thanks for the review, @tbowers7 !

debora-pe (approved):
Wow! This is a lot! Thank you for doing that @kbwestfall!

For as much as I can understand of all of this, it looks good!

PR #2085, opened 2026-02-27
Avoid a circular import by moving `load_tree` into `kdtree_generator`

kbwestfall:
As titled.


rcooke-ast (approved):
This looks good to me 👍 

I burned many hours attempting to solve the ThAr automatic wavelength calibration problem, and I think this is not the solution. Perhaps there's a better solution that we could reconsider at some point, but I think for now, perhaps we could raise an error if someone attempts to load a kdtree, to say that this is currently not supported?

kbwestfall:
> This looks good to me 👍
> 
> I burned many hours attempting to solve the ThAr automatic wavelength calibration problem, and I think this is not the solution. Perhaps there's a better solution that we could reconsider at some point, but I think for now, perhaps we could raise an error if someone attempts to load a kdtree, to say that this is currently not supported?

Thanks for the quick review, @rcooke-ast !  I added an exception if the user tries to use 'holy-grail' with ThAr arcs.  Is this what you were thinking?

rcooke-ast:
Precisely, thanks for implementing that!

profxj (approved):
Indeed, this is nearly deprecated code, but no harm in keeping
it around for now.

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  330 passed, 450 warnings in 321.21s (0:05:21) ---
--- PYTEST UNIT TESTS PASSED  157 passed, 1449 warnings in 867.75s (0:14:27) ---
--- PYTEST VET TESTS PASSED  74 passed, 1671 warnings in 5037.93s (1:23:57) ---
Total disk usage: 353.696 GiB
Testing Started at 2026-03-23T19:15:11.626513
Testing Completed at 2026-03-24T09:42:06.047664
Total Time: 14:26:54.421151
```

Merging.

PR #2088, opened 2026-03-06
Setup gui updates

badpandabear:
Update the setup gui to:

No longer use specific log messages to track progress.
Re-enable the "cancel" button.
Use a dragable button bar
Increase the font size

This requires an associated dev-suite PR: https://github.com/pypeit/PypeIt-development-suite/pull/394


badpandabear:
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  330 passed, 448 warnings in 793.52s (0:13:13) ---
--- PYTEST UNIT TESTS PASSED  159 passed, 1450 warnings in 1808.29s (0:30:08) ---
--- PYTEST VET TESTS FAILED  1 failed, 73 passed, 1672 warnings in 10847.33s (3:00:47) ---

```
The one vet test failure was an issue fixed by PR 2086

debora-pe (commented):
Hey @badpandabear. I played a bit with this - I cannot really understand the code itself. But it looks good. All the new functionalities work for me. 
I only had a question. 
I did not understand what you mean with "Re-enable the "cancel" button.", so I don't think I tested that. Could you explain a little bit more what you mean? Thanks!

badpandabear:
>  I did not understand what you mean with "Re-enable the "cancel" button.", so I don't think I tested that. Could you explain a little bit more what you mean? Thanks!

The cancel button is part of the in progress dialog that only comes up when running a long setup process. To test this I would typically give it the entire RAW_DATA/keck_deimos directory.  To  to cancel a background PypeIt thread in code that doesn't know anything about the GUI, I use the "requestInterruption" method on the Qt thread running the setup. Then a wrapper I put around the logging methods that runs inside that thread checks for that interruption and throws an exception to cancel the operation.

This had stopped working when Kyle made his logging refactor, so it was disabled. I discovered that it was Python's logging infrastructure catching the cancel exception, so I decided to derive my "OpCanceledError" exception from BaseException instead of Exception.  Which means normal exception handlers that just catch Exception won't catch it. And then re-enable the feature.

The other progress dialog change I made was based on earlier feedback from Kyle. He didn't like that I was looking for a specific log message for progress as that introduced a hidden dependency that would break if the log message  changed, and I agreed.  So now it just shows a progress bar going back and forth to show that something is going on, but not a specific percentage complete.  I'm hoping the state feature X and Austin are working on for the Dashboard will allow for a cleaner solution but I don't want to wait.

I probably should have put this level of detail into the original PR text, sorry about that.

debora-pe (approved):
Ok @badpandabear! No worries, thank you for the explanation. I was indeed only using small datasets and I did not notice the progress dialog. I, now, tested with a larger dataset and I can see the cancel button and the progress bar.
All looks good to me.
Thanks!

kbwestfall:
Hi @badpandabear .  I'll try to get my review in for this today.  When working on something else, I came across the GUI specific code in `pypeit/scripts/obslog.py`, which is out of date.  Can you update it and include that as part of this PR?

pypeit/setup_gui/model.py
```diff
@@ -77,6 +78,9 @@ def write(self, message):
             message (str): The log message being written to the log.
         """
 
+        # Get rid of Terminal escape codes in message
+        message = clear_text_color(message)
```
kbwestfall: I'm not sure it matters, but, by assigning the altered version to it self, does this alter the input `message` in the calling function?  Can we instead combine the lines?

```
self._buffer.append(clear_text_color(message))
```

pypeit/setup_gui/view.py
```diff
@@ -1441,18 +1441,19 @@ def _hideTabCloseButton(self, tab_pos, always=False):
...
-class SetupGUIMainWindow(QWidget):
+class SetupGUIMainWindow(QMainWindow):
     """Main window widget for the PypeIt Setup GUI
 
     Args:
         model (:class:`pypeit.setup_gui.model.SetupGUIStateModel`): The model for the PypeitSetupGUI.
         controller (:class:`pypeit.setup_gui.controller.SetupGUIController`): The controller for the PypeitSetupGUI.
     """
 
+    helpURL = QUrl("https://pypeit.readthedocs.io/en/latest/tutorials/setup_gui.html")
```
kbwestfall: We might want this to point to the "stable" branch instead:

https://pypeit.readthedocs.io/en/stable/tutorials/setup_gui.html

?

kbwestfall (changes requested):
Looks good to me!  Although, like @debora-pe, I'm out of my depth with the details of the code.  I had two minor requests.

pypeit/setup_gui/model.py
badpandabear: It doesn't affect the calling function, as strings are immutable, it just creates a copy. But I can easily combine the lines, so I did.

kbwestfall (approved):
Looks great.  Thanks!

PR #2089, opened 2026-03-11
Move key-binding descriptions from class docstring to `pypeit_show_1dspec` description

kbwestfall:
As titled.  This is primarily a documentation PR that is directly to `release`, and it includes some changelog comments related to #2079.

I'll tag and release version 2.0.1, assuming this passes tests.

### DIFF
--- a/doc/out_spec1D.rst
+++ b/doc/out_spec1D.rst
@@ -117,8 +117,16 @@ Here is a typical call:
 Interacting with the display
 ----------------------------
 
-The special ginga bindings (mode) implemented for interacting with PypeIt spec1d
-files is as follows:
+The spec1d mode is a mode enabled by ginga to interact with the spectrum plot.
+**To get a listing of the key-bindings in the ginga window, type 'h'.**  If that
+doesn't work, or if the key bindings themselves don't seem to work, you may have
+exited the mode.  To restart the mode, hit the space-bar and then the 1 key.  If
+you're still having trouble, please submit an issue via GitHub or post a
+question to the Users Slack.
+
+The key-bindings below are copied from the relevant docstring; if anything is
+different from the dialog you get when pressing 'h' in ginga, you should defer
+to what is shown there.  The key-bindigs are as follows:
 
 Enter the mode
 ++++++++++++++
--- a/doc/out_spec1D.rst
+++ b/doc/out_spec1D.rst
@@ -112,54 +112,69 @@ Here is a typical call:
     file is typically written to the ``VERSPYP`` header keyword) or re-reduce
     the data with the new PypeIt version.
 
-Options
--------
+.. _pypeit_show_1dspec_interact:
 
-Here are the typical options you will use:
+Interacting with the display
+----------------------------
 
---list
-++++++
+The special ginga bindings (mode) implemented for interacting with PypeIt spec1d
+files is as follows:
 
-This prints a list to the screen of all the objects extracted.  An example:
+Enter the mode
+++++++++++++++
+* Space, then "1"
 
-.. code-block:: console
+Exit the mode
++++++++++++++
+* Esc
 
-    EXT0000001 = SPAT0351-SLIT0000-DET01
-    EXT0000002 = SPAT0392-SLIT0001-DET01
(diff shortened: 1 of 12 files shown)


kbwestfall:
@ejeschke Somewhat related to this, the `connect_to_ginga` function in `pypeit/display/display.py` will call
```
subprocess.Popen(['ginga', f'--rcport={port}',  '--modules=RC,SlitWavelength'])
```
If it can't connect to an already open ginga window.  Should this instead be:
```
subprocess.Popen(['ginga', f'--rcport={port}',  '--modules=RC,SlitWavelength,Spec1dView'])
```
?

badpandabear (approved):

doc/out_spec1D.rst
```diff
@@ -112,54 +112,69 @@ Here is a typical call:
...
-This indicates the extension of the object with this :ref:`spec1d-naming`.
-
---exten
-+++++++
+The special ginga bindings (mode) implemented for interacting with PypeIt spec1d
+files is as follows:
 
-This is a short-cut of sorts to pull the object you want without
-typing in its name.
+Enter the mode
```
debora-pe, line 131: Thank you for this @kbwestfall!
One thing that I keep forgetting when I use `pypeit_show_1dspec` is how to show the help screen on the side with all the relevant bindings.
Do you mind adding here that, after entering the mode, by typing `h` in the plot window ginga will show the help on the side?

pypeit/scripts/show_1dspec.py
```diff
@@ -107,7 +122,7 @@ def main(cls, args):
         if args.obj is not None:
```
debora-pe, line 122: Should this be:
```
        if args.obj is not None:
            exten = np.where(sobjs.NAME == args.obj)[0][0]
        else:
            exten = args.exten-1 # 1-index in FITS file
        if exten < 0:
            raise PypeItError(f"Bad input extension/object name: {args.obj}")

```

debora-pe (approved):
Thanks @kbwestfall!

pypeit/display/pypeit_modes.py
```diff
@@ -14,64 +14,7 @@ class Spec1DMode(Plot2DMode):
...
-    * [ : mark lower X boundary of region
-    * ] : mark upper X boundary of region
-    * backslash : clear region
-    * singlequote : (zoom) set X range to region
-    * f : fit region and show result
-    * c : clear fit result
+    See :ref:`pypeit_show_1dspec_interact` for interactivity implemented here.
     """
 
     # Needs to be set by reference viewer (via set_shell_ref) before any
```
ejeschke, line 79: @kbwestfall , I don't think you should remove this help docstring. It is used by Ginga to show help for that pypeit special mode when a person types 'h'.  I suppose we could add a special command to fetch the help from somewhere else, but I would be a little concerned that we bypassed the internal help system.

ejeschke (changes requested):

ejeschke:
> If it can't connect to an already open ginga window. Should this instead be:
> 
> subprocess.Popen(['ginga', f'--rcport={port}',  '--modules=RC,SlitWavelength,Spec1dView'])
> 
> ?

No, doesn't need to because `Spec1dView` registers itself via `PypeIt` as a plugin, and the plugin is available and invoked by `pypeit_show_1dspec`. I think `SlitWavelength` could probably be treated the same way, if we modify the relevant pypeit "show_" scripts.  The RC plugin needs to be passed with --modules because it is not set to auto-start by default. That also very likely could be automated in the "show" scripts; I could take a look at that.

doc/out_spec1D.rst
kbwestfall: Done!

pypeit/display/pypeit_modes.py
kbwestfall: Sorry, I didn't know about the 'h' command!  I've restored this docstring to the old version, and added a comment about the 'h' command to the main doc page.

pypeit/scripts/show_1dspec.py
kbwestfall: I think so, yes!  Good catch.  Updated.

PR #2091, opened 2026-03-21
Add TEST logging level

tbowers7:
For code development, add TEST logging level, between INFO and WARNING.

I find it helpful to have an extra logging level above INFO that I can use in code development for test messages.

Invoke with `log.test("<string>")`.


pypeit/logger.py
```diff
@@ -19,6 +19,16 @@
...
 
+# Add a logging TEST level between INFO and WARNING
+TEST_LEVEL_NUM = 25
+logging.addLevelName(TEST_LEVEL_NUM, "TEST")
+def logtest(self, message, *args, **kwargs):
+    """Define TEST log level for calling"""
+    # Skip this wrapper frame so filename/lineno point at the caller of log.test()
+    kwargs.setdefault("stacklevel", 2)
+    self.log(TEST_LEVEL_NUM, message, *args, **kwargs)
+logging.Logger.test = logtest
```
kbwestfall: Instead of altering the top-level logger, can we move this into `PypeItLogger`?  I'm wary of directly editing the default `Logger`.

pypeit/logger.py
```diff
@@ -102,6 +112,7 @@ def format(self, record):
         level_colors = {
             'debug': [116, 173, 209],
             'info': [49, 54, 149],
+            'test': [223, 255, 0],
```
kbwestfall, line 105: Do you have a sample of what this color looks like?

kbwestfall (changes requested):
Thanks, @tbowers7 !  I'm fine with this if you find it helpful.  Some high-level questions/requests:

- If possible, can we put the new function inside the definition of `PypeItLogger`?  I'm wary of altering the default logger.  I would have to Google and play around to know how to do this (or what the best practice is).
- How do you go about ensuring this level is printed when you're running pypeit?  The `PypeItLogger.convert_verbosity_to_logging_level` function converts from the verbosity level entered by a user to a logging level.  Which verbosity level do you set, or do you have some other way of ensuring that the test messages are printed?
- I presume we should avoid `log.test()` statements in production-level code?
- (Unrelated to this PR), can you merge in the `release` branch so that we can both add your changes and sync `develop` and `release` using your PR?

pypeit/logger.py
tbowers7: [Charteuse](https://convertingcolors.com/rgb-color-223_255_0.html).  Bright green/yellow -- easy to see when running testing of new code.

pypeit/logger.py
tbowers7: Changed.

tbowers7:
> * If possible, can we put the new function inside the definition of `PypeItLogger`?  I'm wary of altering the default logger.  I would have to Google and play around to know how to do this (or what the best practice is).

Done.

> * How do you go about ensuring this level is printed when you're running pypeit?  The `PypeItLogger.convert_verbosity_to_logging_level` function converts from the verbosity level entered by a user to a logging level.  Which verbosity level do you set, or do you have some other way of ensuring that the test messages are printed?

By setting a log level (integer) between INFO and WARNING -- the Python logger includes levels at the set integer value and above.  Setting verbosity to INFO will include TEST, setting verbosity to WARNING will not.

> * I presume we should avoid `log.test()` statements in production-level code?

That is the idea.  I often find it very helpful, when developing, to have test messages that are easy to find in the output -- but not including all DEBUG-level messages.  I would advocate for not including `log.test()` in production code.  The nice thing about how this is now added, we can add other levels to the logger for eye-catching statements we need users to pay attention to, since much of the logging is not necessarily of much use to the user.  There may be INFO statements in the code we want to the user to pay attention to, and could alter them to a new color, or even make something above warning and set the default output to WARNING.  There are lots of ways this could go.

> * (Unrelated to this PR), can you merge in the `release` branch so that we can both add your changes and sync `develop` and `release` using your PR?

Done.

pypeit/logger.py
```diff
@@ -359,6 +382,8 @@ def close_file(self):
         if self.fh in self.warnings_logger.handlers:
             self.warnings_logger.removeHandler(self.fh)
 
+# Register a logging TEST level between INFO and WARNING
+logging.addLevelName(PypeItLogger.TEST, "TEST")
```
kbwestfall, line 386: Sorry to keep nit-picking, but can we move this inside `get_logger`, so that the relevant lines would be something like:

```
logging.setLoggerClass(PypeItLogger)
logging.addLevelName(PypeItLogger.TEST, "TEST")

try:
   ...
```

If something goes wrong in the try block, I assume this would mean that any calls to `log.test` would raise an exception.  But that might be okay if we're assuming there will be no calls to `log.test` in production level code.

kbwestfall (commented):
> There may be INFO statements in the code we want to the user to pay attention to, and could alter them to a new color, or even make something above warning and set the default output to WARNING. There are lots of ways this could go.

I agree, at some level, but two quick thoughts on this: (1) I think we should limit our bespoke logging levels.  (2) If not instead of, at least in addition to, adding new logging levels, we should aim to make the existing INFO messages more meaningful.

kbwestfall:
Sorry, I also meant to ask: Can you also add a short test of the `log.test` call in `tests/test_log.py`?

tbowers7:
> I agree, at some level, but two quick thoughts on this: (1) I think we should limit our bespoke logging levels. (2) If not instead of, at least in addition to, adding new logging levels, we should aim to make the existing INFO messages more meaningful.

This is an important point.

Since `log.test()` may be something only I will use, maybe I close this PR and we think about how to use the existing logging levels more effectively.

PR #2100, opened 2026-04-02
Show RMS in SlitWavelength Plugin

MNBrod:
Currently, the slitwavelength Ginga plugin shows wavelength information at a given pixel based off of a wavelength image by hijacking the RA readout to show wavelength instead. This PR additionally hijacks DEC to also allow the display of the RMS of the wavelength fit for each slit, if provided. By default, there won't be any changes seen, to take advantage of this feature a display script needs to construct an rms_img in addition to the wave_img.

Note that there is a bug in Ginga in which only one character can be displayed in the RA/DEC label. This isn't an issue for the wavelength label (since α become λ), but this implementation tries to replace δ with "RMS", so it currently only displays "R". This can only be fixed on the Ginga side, I will submit an issue there shortly.


kbwestfall (changes requested):
Looks good to me.

Two requests:

- Add a comment on this in the relevant release doc
- Add some text to the show_2dspec documentation that describes how this RMS value is determined.  Is it just one value per slit?  Is the value in pixels?  etc.

debora-pe (approved):
Looks good.

PR #2105, opened 2026-04-13
Added VLT/UVES (echelle) data reduction support to PypeIt

rcooke-ast:
With this PR, we include support of the VLT/UVES echelle spectrograph. Much of the preparatory work for this reduction was provided by Aldric Wong (@aldric98) (many thanks!!).

I also identified a bug in the way PypeIt stores the slitmask. Non slits are assigned a value of `-1`, but for echelle slits, there are some slits that can be even more negative values (or, indeed, in exceptionally rare cases, an echelle slit could be assigned a value of `-1`). I will raise an issue about this. For now, I have implemented a quick fix to ensure that the ONLY failed slits are those with `slitmask==-1`, rather than only considering positive slits.

There is a companion dev-suite PR, where the tests pass (I'm now running a full dev-suite run to check nothing else has broken in the meantime):
image


rcooke-ast:
All tests pass. Note that the failed test is because of extra files in LDT/DEVENY, which I highly suspect is unrelated to this PR.

image

pypeit/spectrographs/vlt_uves.py
```diff
@@ -0,0 +1,1189 @@
...
+        if ftype == 'pinhole':
+            # Don't type pinhole
+            return np.zeros(len(fitstbl), dtype=bool)
+        if ftype in ['arc', 'tilt']:
+            return good_exp & (fitstbl['idname'] == 'LAMP,WAVE')
+
+        log.warning('Cannot determine if frames are of type {0}.'.format(ftype))
+        return np.zeros(len(fitstbl), dtype=bool)
+
+    # IS THIS NEEDED??
```
kbwestfall: It looks like this can be removed.  I don't think it's doing anything, so this subclass can just rely on the base-class function.

pypeit/spectrographs/vlt_uves.py
```diff
@@ -0,0 +1,1189 @@
...
+    #     # 346
+    #     # spec_max = np.asarray([3000]*32 + [925])#, 2460
+    #     # spec_min = np.asarray([635] + [0]*32)
+    #     # 390
+    #     spec_max = np.asarray([3000]*38 + [920, 2740])
+    #     spec_min = np.asarray([1650, 330] + [0]*38)
+    #     # 437
+    #     # spec_max = np.asarray([3000]*30 + [2260])
+    #     # spec_min = np.asarray([1060] + [0]*30)
+    #     return np.vstack((spec_min, spec_max))
```
kbwestfall: remove the commented code?

pypeit/spectrographs/vlt_uves.py
```diff
@@ -0,0 +1,1189 @@
...
+
+        # Adjustments to parameters for Keck HIRES
+        turn_off_on = dict(use_biasimage=False, use_overscan=True, overscan_method='median')
+        par.reset_all_processimages_par(**turn_off_on)
+        # Right now we are using the overscan and not biases becuase the
+        # standards are read with a different read mode and we don't yet have
+        # the option to use different sets of biases for different standards,
+        # or use the overscan for standards but not for science frames
+
+        # Setup dependent -- This is only temporary until we have the reidentification files for all the red settings
```
kbwestfall: remove?

pypeit/spectrographs/vlt_uves.py
```diff
@@ -0,0 +1,1189 @@
...
+        # Telluric parameters
+        # HIRES is usually oversampled, so the helio shift can be large
+        par['telluric']['pix_shift_bounds'] = (-40.0,40.0)
+        # Similarly, the resolution guess is higher than it should be
+        par['telluric']['resln_frac_bounds'] = (0.25,1.25)
+
+        # Coadding
+        par['coadd1d']['wave_method'] = 'log10'
+
+        return par
```
kbwestfall, line 528: How much of this is the same as used in the blue arm?  Can much of this be moved to the `VLTUVESSpectrograph` base class, such that only a few changes are made in the relevant subclass functions for each arm?

pypeit/spectrographs/vlt_uves.py
```diff
@@ -0,0 +1,1189 @@
...
+        Return the list of allowed detector mosaics.
+
+        Only red arm on VLT/UVES requires mosaicing.
+
+        Returns:
+            :obj:`list`: List of tuples, where each tuple provides the 1-indexed
+            detector numbers that can be combined into a mosaic and processed by
+            PypeIt.
+        """
+        return [(1,), (1,2)]
```
kbwestfall: Do we need to allow the single detector tuple, `(1,)`?  I thought people could just set `detnum=1`.

pypeit/spectrographs/vlt_uves.py
```diff
@@ -0,0 +1,1189 @@
...
+#         # 760u
+#         # spec_max = np.asarray([4096]*16)
+#         # spec_min = np.asarray([2700] + [0]*15)
+#         # 860l
+#         # spec_max = np.asarray([4096]*19 + [3800])
+#         # spec_min = np.asarray([800] + [0]*19)
+#         # 860u
+#         spec_max = np.asarray([4096]*12 + [1440])
+#         spec_min = np.asarray([0]*13)
+#         return np.vstack((spec_min, spec_max))
```
kbwestfall: Remove all this commented code?

kbwestfall (changes requested):
Looks good!

I'm mostly requesting some code clean-up, and I'm wondering if we can consolidate some of the parameter definitions in the subclass for each arm into the main `VLTUVESSpectrograph` base class.

Can you also add a comment to the release file and update the documentation?  I can help with one or both of those if needed.

Thanks!

pypeit/spectrographs/vlt_uves.py
rcooke-ast: Agreed - I have removed this 👍

pypeit/spectrographs/vlt_uves.py
rcooke-ast: I'd like to keep this stored somewhere, in case I ever need to regenerate the wavelength solution. However, I don't think it needs to be stored in the main code base. So, I've moved this to the dev-suite, since it's likely it will never be needed again!

pypeit/spectrographs/vlt_uves.py
rcooke-ast: As above, I have moved this to the dev-suite

pypeit/spectrographs/vlt_uves.py
rcooke-ast: Good point... I kept this flexible while developing, but now is a good time to consolidate the common values 👍

pypeit/spectrographs/vlt_uves.py
rcooke-ast: As above, I have moved this to the dev-suite  😊

pypeit/spectrographs/vlt_uves.py
rcooke-ast: You're right, we don't need that, I've removed this part, and updated the default mosaic below.

rcooke-ast (commented):
Thanks @kbwestfall! All comments addressed, and I've added the documentation + release file info, too. I'll rerun the dev-suite to make sure nothing has broken in the process 👍

doc/tutorials/tutorials.rst
```diff
@@ -45,6 +45,7 @@ bottom of the :doc:`../spectrographs/spectrographs` page.
    LDT DeVeny <deveny_howto>
    Shane Kast <kast_howto>
    Subaru FOCAS <focas_howto>
+   VLT UVES (follow a similar procedure to Keck/HIRES) <hires_howto>
```
kbwestfall: I'm not sure about this.  I.e., having a "VLT/UVES" tutorial included this way feels confusing.  Can we instead add a section to the docs that list the suggested tutorials to follow for instruments that don't have explicit tutorials?  Is that overly pedantic?  Let me know what you think.

kbwestfall (approved):
Thanks, @rcooke-ast !  One lingering comment, but I'm approving now.

doc/tutorials/tutorials.rst
debora-pe: Agree with Kyle. I also think that what is already written in `vlt_uves.rst‎` under `Additional Reading` should be enough.

pypeit/spectrographs/vlt_uves.py
```diff
@@ -0,0 +1,761 @@
...
+
+
+class UVESMosaicLookUp:
+    """
+    Provides the geometry required to mosaic VLT/UVES data.
+    Similar to :class:`~pypeit.spectrographs.gemini_gmos.GeminiGMOSMosaicLookUp`
+
+    """
+    # Original
+    geometry = {
```
debora-pe, line 33: Since UVES blue and red arms are 2 different PypeIt spectrographs, and the blue one is only one detector, I think you don't need to have here both MSC01 and MSC02. You can have only MSC02 (and maybe call it MSC01) and use it only for the red spectrograph (as it is already done  below).

debora-pe (approved):
I just have small comment on the `UVESMosaicLookUp`, but I'm approving.
Looks good!

doc/tutorials/tutorials.rst
rcooke-ast: Yes, good point 👍 I've removed this (but kept in the note in the `vlt_uves.rst` file). Perhaps we should think more about how to include tutorials for instruments that don't currently have a tutorial.

pypeit/spectrographs/vlt_uves.py
rcooke-ast: Good point -- I've updated this, and checked that the reduction still works.

rcooke-ast (commented):
Addressed comments - thanks!  I have checked that I can still reduce UVES data on my machine, and I am now running the dev suite again just to make sure nothing else is broken.

rcooke-ast:
Tests pass...

image

PR #2112, opened 2026-05-06
Add wavelength solution lbt_luci2_G210K_22000.fits

opkuhn:
This PR (to the develop branch) is to add the LUCI2 G210 K wavelength solution under pypeit/data/arc_lines/reid_arxiv. This wavelength solution is referenced by the pypeit file to reduce the sample dataset, [luci2_G210K](https://drive.google.com/drive/folders/1wYH57fV9rHvBQfq8SY6bNyD2pMklEDoX?usp=drive_link). See [Pypeit-development-suite PR #403](https://github.com/pypeit/PypeIt-development-suite/pull/403).

pypeit_identify: 4th order polynomial fit, rms = 0.075 pixels, although post-reduction pypeit_chk_wavecalib gives a higher but acceptable rms (~0.2 pixels).

lbt_luci2_G210K_22000


pypeit/core/standard.py
```diff
@@ -24,9 +24,8 @@
 
 
 def mAB_to_cgs(wave, mAB):
-    r"""
-    Convert AB magnitudes to :math:`F_\lambda` in the cgs units :math:`{\rm
-    erg/cm}^2{\rm/s}/\AA`.
+    """
```
kbwestfall, line 27: I think you need the `r` character (creating a Python "raw string") so that sphinx knows how to compile the math.

pypeit/core/standard.py
```diff
@@ -883,8 +902,9 @@ def get_model_standard(spectral_type, V_mag):
             return KuruczModelStandard(V_mag, spectral_type)
 
 
-def get_standard_spectrum(spectral_type=None, V_mag=None, ra=None, dec=None, tol=20., unit=None,
-                          archives='default'):
+def get_standard_spectrum(archives, spectral_type=None, V_mag=None, ra=None, dec=None, tol=20., unit=None):
+#def get_standard_spectrum(spectral_type=None, V_mag=None, ra=None, dec=None, tol=20., unit=None,
+#                          archives=None):
```
kbwestfall, line 907: Is there a reason you needed to change the calling sequence?  I.e., why did you need to make `archives` a positional argument?

pypeit/core/standard.py
```diff
@@ -925,5 +945,7 @@ def get_standard_spectrum(spectral_type=None, V_mag=None, ra=None, dec=None, tol
         raise PypeItError('Insufficient data provided to determine the appropriate standard spectrum.  '
                    'Provide either the coordinates of the standard or a stellar type and '
                    'magnitude.')
+# add star_arxiv here
+#   if archives != 'default':
```
kbwestfall, line 949: remove?

kbwestfall (changes requested):
Thanks!  Just a few minor comments.  The main change request is to add the `r` character back for the `mAB_to_cgs` docstring.

kbwestfall:
And we should run the dev-suite tests.  I can set that going if you like.

pypeit/core/standard.py
kbwestfall: PS - This change is what's causing the check failures.

opkuhn:
Closing this PR in favor of [2114](https://github.com/pypeit/PypeIt/pull/2114) which _only_ adds the lbt_luci2_G210K_22000.fits wavelength solution.

PR #2117, opened 2026-05-12
Core refactor - move spline

bpholden:
This moves bspline into core and update the dependencies. Test pass.


.gitignore
```diff
@@ -87,3 +87,4 @@ pypeit/data/QL_CALIB
 doc/sphinx_warnings.out
 pypeit/tests/files/setup_files/*
 .DS_Store
+*.png
```
badpandabear: I'm worried this will ignore .png files needed by the GUI and the docs. Were there png files in specific directories that you wanted to ignore? If so you could put a .gitignore into that directory to ignore them.

badpandabear (changes requested):

.gitignore
bpholden: Excellent point.
I restored the .gitignore and moved the pydeps PNG files out of the source tree.

badpandabear (approved):

PR #2119, opened 2026-05-13
Improvements to `update_docs`

kbwestfall:
This improves the usability of the `update_docs` script.

First, I cleaned up the `doc/Makefile`, removing all the unused macros.  Then I added new macros that (1) isolate the script that needs access to the dev-suite and (2) allow building the docs without that script.

Second, I improved the `update_docs` script so that it can be run with or without access to the dev-suite and will work appropriately.

I also updated the development doc, added a release note, and removed the `sphinx.readme` file.

This is a doc-only PR, so we don't need to run the test/dev-suite.


doc/Makefile
```diff
@@ -2,56 +2,18 @@
...
-# Internal variables.
-PAPEROPT_a4     = -D latex_paper_size=a4
-PAPEROPT_letter = -D latex_paper_size=letter
-ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
-# the i18n builder cannot share the environment and doctrees with the others
-I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
-
-.PHONY: help clean html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest coverage gettext
-
-help:
```
badpandabear, line 28: I kinda like having "make help" be available to print the acceptable options. Would you consider leaving it and only having those targets we use (like html, htmlonly, htmlnoex)?

badpandabear (approved):
Looks good. I have a minor request about leaving some "make help" documentation.

Another thing I noticed, unrelated to this change, is that when I cloned your repo and did a build in a fresh environment, it failed because `PypeIt/doc/scripts/build_telluric_table.py` uses awscli.  Since that requires the user have aws credentials in their .aws directory, do we really want to require that?  Do the telluric tables change often enough to need dynamic building?

doc/Makefile
kbwestfall: Good idea.  I've rewritten the help so that it's actually useful 😄

kbwestfall (commented):
Both are good points, @badpandabear .

I added back the `help` command and made it actually reflect what's in the Makefile. 

On the use of `aws`, I think you're right that we likely don't need to update these files every time.  But I thought I had written the code so that if the `aws` subprocess command fails that it should just not update the telluric table and keep going.  Do you agree?  I could move that script into another macro that we run more sporadically.

bpholden (approved):
I will approve.

This gets rid of most of the documentation options. That is probably fine, I doubt anyone builds LaTeX files, but just in case, are there options to add them back?

kbwestfall:
> This gets rid of most of the documentation options. That is probably fine, I doubt anyone builds LaTeX files, but just in case, are there options to add them back?

Yeah, we can bring back the option, if necessary (we can just add the macro back).  But, like you say, I don't think anyone has actually used the `make latexpdf` option; I'm not even sure it worked!

PR #2120, opened 2026-05-13
Remove instances of `log.error`

kbwestfall:
Unlike the old `msgs.error`, `log.error` does not raise an exception.  That means we basically should never be using it.  Instead we should be raising exceptions, typically `raise PypeItError`, which will be caught and handled by the new logging code.

This PR replaces the few instances I found in the code.


badpandabear (approved):

rcooke-ast (approved):
Sorry about that -- my bad!

PR #2123, opened 2026-05-19
Updates to `kcwi_dec_2024` to resolve comments.

rcooke-ast:
As titled. See #1929 for more details.


pypeit/coadd3d.py
```diff
@@ -1591,7 +1531,7 @@ def run_align(self, fwhm=1.5, show_qa=False):
         # Initialize the RA and Dec offset arrays
         ra_offsets, dec_offsets = [0.0]*self.numfiles, [0.0]*self.numfiles
         # Register spatial offsets between all frames
-        if self.user_alignment:
+        if self.alignment_method == "user":
             # The user has specified offsets - update these values accounting for the difference in header RA/DEC
             ra_offsets, dec_offsets = datacube.align_user_offsets(self.ifu_ra, self.ifu_dec,
                                                                   self.ra_offsets, self.dec_offsets)
```
kbwestfall: Inside this `if` block, you could:
```
return datacube.align_user_offsets(
    self.ifu_ra, self.ifu_dec, self.ra_offsets, self.dec_offsets
)
```
so that the rest of the function doesn't need to be in the `else` block.

rcooke-ast: Yep, good call, I've updated this 👍

jhennawi (approved):

Issue #2113, opened 2026-05-06
Datacube transposes


kbwestfall:
Is there a way to refactor the code so that we don't need to constantly deal with transposing the arrays?

rcooke-ast:
Addressed by PR #2132 (also, see Issue #1699)

Issue #1699, opened 2023-10-07
Avoid transposing when saving DataCubes


rcooke-ast:
As discussed in PR #1669, it would be beneficial in the long term to avoid transposing the datacubes when saving them. The current best option is to generate the cubes in transposed forms (which involves preparing the bin edges and voxel coordinates in transposed form).

rcooke-ast:
Addressed by PR #2132.

PR #2132, opened 2026-05-21
Remove transpose operation in datacube generation

rcooke-ast:
This PR removes a long-standing TODO to remove the transposing of datacubes when being saved. It has a lot of knock on effects, and I don't expect to have caught all instances where this will need to be corrected, but I have checked that the main output cubes (and ivar, gpm, etc.) are identical between the old version and this new version. The speedup for the cubes I have tested is about 2-3x faster.

Addresses Issue #1699 (as discussed in PR #1669). Also discussed in Issue #2113 (and PR #1929)

### DIFF
--- a/pypeit/core/datacube.py
+++ b/pypeit/core/datacube.py
@@ -212,7 +212,7 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
     initial_guess = (1, _init_obj_position[0], _init_obj_position[1], fwhm*fwhm2sigma, fwhm*fwhm2sigma, 0, 0)
     bounds = ([0,      _init_obj_position[0]-fwhm/3.0, _init_obj_position[1]-fwhm/3.0, fwhm/6.0, fwhm/6.0, -np.pi, -np.inf],
               [np.inf, _init_obj_position[0]+fwhm/3.0, _init_obj_position[1]+fwhm/3.0, fwhm    , fwhm    , np.pi , np.inf])
-    print(_init_obj_position[0], _init_obj_position[1])
+
     # Perform the fit
     # TODO :: May want to generate the image on a finer pixel scale first
     # TODO JFH: The 2D Gaussian fitting should be using the noise and the gpm. This should be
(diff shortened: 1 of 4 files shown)


pypeit/core/datacube.py
```diff
@@ -212,6 +212,7 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
     initial_guess = (1, _init_obj_position[0], _init_obj_position[1], fwhm*fwhm2sigma, fwhm*fwhm2sigma, 0, 0)
     bounds = ([0,      _init_obj_position[0]-fwhm/3.0, _init_obj_position[1]-fwhm/3.0, fwhm/6.0, fwhm/6.0, -np.pi, -np.inf],
               [np.inf, _init_obj_position[0]+fwhm/3.0, _init_obj_position[1]+fwhm/3.0, fwhm    , fwhm    , np.pi , np.inf])
+    print(_init_obj_position[0], _init_obj_position[1])
```
kbwestfall: Remove the print?

pypeit/core/datacube.py
```diff
@@ -551,12 +552,12 @@ def extract_point_source(
...
-        skyspec = skycube.sum(axis=(0,1))
-        nrmsky = skymask.sum(axis=(0,1))
+        skyspec = skycube.sum(axis=(1,2))
+        nrmsky = skymask.sum(axis=(1,2))
         skyspec *= utils.inverse(nrmsky)
-        _flxcube -= skyspec.reshape((1, 1, numwave))
+        _flxcube -= skyspec.reshape((numwave, 1, 1))
         # Now subtract the residual sky from the white light image
-        sky_val = np.sum(wl_img[:, :, np.newaxis] * smask) / np.sum(smask)
+        sky_val = np.sum(wl_img[np.newaxis, :, :] * smask) / np.sum(smask)
```
kbwestfall: Happy for you to leave it as `np.newaxis`, but just thought I'd point out (in case you didn't know) that this is identical to `wl_img[None,:,:]`.

pypeit/core/datacube.py
```diff
@@ -644,7 +645,7 @@ def extract_point_source(
         optkern = apo_smooth_wl_img/np.sum(apo_smooth_wl_img)
 
 
-    optkern_masked = optkern * mask[:,:,0]
+    optkern_masked = optkern * mask[0,:,:]
```
kbwestfall, line 648: FWIW, you could just do `mask[0]` here, instead of `mask[0,:,:]`.

kbwestfall (approved):
I have a few very minor comments.  I can't say I checked everything in detail, but everything I saw made sense in terms of reordering the operations.

Thanks so much for taking the time to do this!

pypeit/core/datacube.py
rcooke-ast: Thanks, indeed!

pypeit/core/datacube.py
rcooke-ast: If it's OK, I prefer this because it shows that it's a 3D array.

PR #2133, opened 2026-05-22
Wavelength template line detection

kbwestfall:
I wanted to reduce the data in the `RAW_DATA/keck_deimos/830G_L_8410` directory as a way of producing a test for flux calibration that would use the `multi_spec_det` option.  Wavelength calibration on detector 3 was failing because `pypeit.core.wavecal.wvutils.arc_lines_from_spec` was not detecting lines in the template spectrum `keck_deimos_830G.fits`.  This is because the FWHM is too small.  I.e., the FWHM of the observed data is smaller than in the template, which is causing the lines in the template to be rejected.

This is a short term fix that adds a loop that iteratively increments the FWHM until the number of good template lines reaches some minimum fraction of all lines detected.  This works for this data set, but we'll need to run the dev-suite to check how this affects other instruments.

In the longer term, I think we should
   - only detect the lines in the template once
   - or, better yet, require that the pixel and wavelength coordinates of the lines used in calibrating the template be included in the template file.  I.e., we shouldn't need to re-detect the lines.

The primary changes are in `pypeit.core.wavecal.wvutils.arc_lines_from_spec` and `pypeit.core.wavecal.autoid.reidentify`.  Particularly for the latter, I took the opportunity to clean-up the code and docs a bit.


pypeit/core/wavecal/autoid.py
```diff
@@ -377,193 +377,238 @@ def match_qa(arc_spec, tcent, line_list, IDs, scores, outfile = None, title=None
...
-    if cont_sub:
-        # use the continuum subtracted arc spectrum for the rest of the code
-        use_spec = spec_cont_sub
+    if len(detections) == 0:
+        raise PypeItError('No lines were detected in the observed arc spectrum!')
+
+    # Use the continuum subtracted arc spectrum for the rest of the code, if requested
+    use_spec = spec_cont_sub if cont_sub else spec
+
+    # TODO: Everytime we re-run reidentify, we're re-detecting the lines in the
```
profxj, line 577: yes, good point!

pypeit/core/wavecal/autoid.py
```diff
@@ -657,63 +737,94 @@ def reidentify(spec, spec_arxiv_in, wave_soln_arxiv_in, line_list,
...
-                        line_iarxiv = np.append(line_iarxiv,iarxiv)
+                if pdiff[bstpx] > match_toler:
+                    # Match is not within tolerance, keep going
+                    continue
+
+                # Using the arxiv arc wavelength solution, search for the
+                # nearest line in the line list
+                bstwv = np.abs(wvdata - wvval_arxiv[bstpx])
+
+                if bstwv[np.argmin(bstwv)] > match_toler*disp_arxiv[iarxiv]:
```
profxj, line 762: Oh what tangled webs we weave;  has been awhile since I looked at this gnarly code..

profxj (approved):
Excellent addition.  I might give Debora a head's up about it too
  If you haven't already

pypeit/core/wavecal/autoid.py
tbowers7: Second!

pypeit/core/wavecal/autoid.py
```diff
@@ -377,193 +377,238 @@ def match_qa(arc_spec, tcent, line_list, IDs, scores, outfile = None, title=None
...
     spec_arxiv_cont_sub = np.zeros_like(spec_arxiv)
     det_arxiv1 = {}
     for iarxiv in range(narxiv):
-        tcent_arxiv, ecent_arxiv, cut_tcent_arxiv, icut_arxiv, spec_cont_sub_now = wvutils.arc_lines_from_spec(
-            spec_arxiv[:, iarxiv], sigdetect=sigdetect, nonlinear_counts=nonlinear_counts, fwhm=fwhm,
-            debug=debug_peaks)
-        spec_arxiv_cont_sub[:, iarxiv] = spec_cont_sub_now
+        # NOTE: I've hard-coded the values for good_frac, fwhm_incr, and
+        # max_good_iter.  We may want to revisit this during testing and/or as
+        # we continue to develop.
```
tbowers7, line 586: Is there any reason at all to expose these as user-tunable parameters, or are they strictly algorithm-specific parameters that need to be tested and set for good?

pypeit/core/wavecal/autoid.py
```diff
@@ -590,45 +630,82 @@ def reidentify(spec, spec_arxiv_in, wave_soln_arxiv_in, line_list,
...
+            success, shift_vec[iarxiv], stretch_vec[iarxiv], stretch2_vec[iarxiv],
+            ccorr_vec[iarxiv], _, _
+        ) = wvutils.xcorr_shift_stretch(
+            use_spec, use_spec_arxiv[:, iarxiv], sigdetect=sigdetect, lag_range=cc_shift_range,
+            cc_thresh=cc_thresh, fwhm=fwhm, seed=rng, debug=debug_xcorr, percent_ceil=percent_ceil,
+            max_lag_frac=max_lag_frac, stretch_func=stretch_func
+        )
+        log.info(
+            f'shift = {shift_vec[iarxiv]:5.3f}, stretch = {stretch_vec[iarxiv]:5.3f}, cc = '
+            f'{ccorr_vec[iarxiv]:5.3f}'
```
tbowers7: Formatting annoyance: place the "cc =" on the next line -- doesn't affect anything other than me reading the code.

pypeit/core/wavecal/autoid.py
```diff
@@ -1146,6 +1262,7 @@ def full_template(spec, lamps, par, ok_mask, det, binspectral, nsnippet=2, slit_
         # Grab the observed arc spectrum
         obs_spec_i = spec[:,slit]
         # get FWHM for this slit
+        # TODO: Doesn't this mean measured_fwhms *cannot* be None?
```
tbowers7: That's how I read it.  Maybe add a defensive check for `None`?

pypeit/core/wavecal/autoid.py
```diff
@@ -1267,36 +1384,41 @@ def full_template(spec, lamps, par, ok_mask, det, binspectral, nsnippet=2, slit_
...
             mwvsnippet = mwv[j0:j1]
-            # TODO: JFH This continue statement deals with the case when the msnippet derives from *entirely* zero-padded
-            #  pixels, and allows the code to continue with crashing. This code is constantly causing reidentify to crash
-            #  by passing in these junk snippets that are almost entirely zero-padded for large shifts. We should
-            #  be checking for this intelligently rather than constantly calling reidentify with basically junk arxiv
-            #  spectral snippets.
+
+            # TODO: JFH This continue statement deals with the case when the
+            # msnippet derives from *entirely* zero-padded pixels, and allows
+            # the code to continue with crashing. This code is constantly
```
tbowers7: *without crashing*?

pypeit/core/wavecal/wvutils.py
```diff
@@ -317,38 +317,111 @@ def get_wave_grid(waves=None, gpms=None, wave_method='linear', iref=0, wave_grid
...
-        debug=debug)
+        spec, sigdetect=sigdetect, fwhm=fwhm, fit_frac_fwhm=fit_frac_fwhm,
+        cont_frac_fwhm=cont_frac_fwhm, max_frac_fwhm=max_frac_fwhm, cont_samp=cont_samp,
+        niter_cont=niter_cont, nonlinear_counts=nonlinear_counts, debug=debug
+    )
+
+    if all(iter_args) and len(w)/len(tampl) < good_frac:
+        # Iteratively increase the FWHM in an attempt to improve the line detection
+        for i in range(max_good_iter):
+            _fwhm = fwhm*fwhm_incr**(i+1)
```
tbowers7, line 418: So, if `fwhm_incr` were less than 1, we would be _decreasing_ the FWHM with each iteration.  Is there a case where this would be useful at all?

pypeit/core/wavecal/wvutils.py
```diff
@@ -784,8 +859,9 @@ def xcorr_shift_stretch(inspec1, inspec2, cc_thresh=-1.0, percent_ceil=50.0, use
         else:
             raise PypeItError('Unrecognized stretch_func')
         result = scipy.optimize.differential_evolution(
-                zerolag_shift_stretch, args=(y1,y2), x0=x0_guess, tol=toler, 
-                bounds=bounds, disp=False, polish=True, seed=seed)
+            zerolag_shift_stretch, args=(y1,y2), x0=x0_guess, tol=toler, bounds=bounds, disp=False,
+            polish=True, rng=seed
```
tbowers7: `seed = rng` instead?

pypeit/core/wavecal/autoid.py
```diff
@@ -942,6 +1056,7 @@ def map_fwhm(image, gpm, slits_left, slits_right, slitmask, npixel=None, nsample
                 continue
             # Detect lines and store the spectral FWHM
             _, _, cent, wdth, _, best, _, nsig = arc.detect_lines(arc_spec.squeeze(), sigdetect=sigdetect, fwhm=fwhm, bpm=arc_spec_bpm.squeeze())
+            # NOTE: "best" is a list of line indices, NOT a boolean array!
```
tbowers7, line 1059: This kind of `NOTE` occurs often in this PR...

This indicates that a boolean array is expected for this kind of variable... is there a good reason it is not and/or would it be prohibitively expensive (code changes / in-process computation / etc.) to make it thus?

tbowers7 (approved):
Thanks for figuring out this issue.  There is only one place that needs an actual change -- the rest are questions and comments.  Approving now.

pypeit/core/wavecal/autoid.py
kbwestfall: My hope is that these are actually (somewhat) temporary parameters, so I think the answer to your first question is no.  Particularly because it's would be difficult to add a way for a user to assess what the parameters should be without digging into the code.

pypeit/core/wavecal/autoid.py
kbwestfall: Done!

pypeit/core/wavecal/autoid.py
kbwestfall: Personally, I often expect that these kinds of arrays are boolean.  So I added these (admittedly obnoxious) notes mostly as a reminder to me.  At the same time, I didn't think it was worth actually trying to change the object from an int array to a bool array.  Punting...

pypeit/core/wavecal/autoid.py
kbwestfall: Instead of checking for None, I made it a positional argument.

pypeit/core/wavecal/autoid.py
kbwestfall: Good catch!  Fixed.

pypeit/core/wavecal/wvutils.py
kbwestfall: In the limited testing I've done, it seems like you tend to miss lines if the FWHM is too small, but not if it is too big.  That means there is not much to be gained by iteratively making the FWHM smaller.  But, I didn't think it was worth imposing that `fwhm_incr` be greater than 1, particularly given that it's effectively a developer-only parameter in the current workflow.  I've added a comment along these lines.

pypeit/core/wavecal/wvutils.py
kbwestfall: The `seed` keyword is being deprecated by scipy, in favor of `rng`.

kbwestfall (commented):
Thanks for the reviews!  All comments addressed.  Next is to run the tests!

PR #2134, opened 2026-05-22
Add Palomar NGPS (Next Generation Spectrograph) support

cfremling:
## Summary

Adds the four per-channel spectrograph classes for Palomar / NGPS
(Next Generation Spectrograph) -- the four-channel optical
spectrograph at the Hale 200-inch that replaces DBSP -- plus the
single-slice central-slit ``full_template`` reidentify archives,
user docs, and a CHANGES entry.

NGPS covers 3050-10400 Å at R ~ 4000 via a 3-slice image-slicer
and a u/g/r/i dichroic train.  All four channels share one raw
multi-extension FITS file per exposure (one image extension per
channel keyed by ``EXTNAME``).

### Spectrographs added
- ``p200_ngps_u`` (3050-4430 Å)
- ``p200_ngps_g`` (4250-5960 Å)
- ``p200_ngps_r`` (5620-7950 Å)
- ``p200_ngps_i`` (7530-10400 Å)

PR #2135, opened 2026-05-22
Add Palomar NGPS (Next Generation Spectrograph) support

cfremling:
*Re-opens [#2134](https://github.com/pypeit/PypeIt/pull/2134), which I accidentally closed by renaming the head branch on my fork.  Same content; just a different head ref.*

## Summary

Adds the four per-channel spectrograph classes for Palomar / NGPS
(Next Generation Spectrograph) -- the four-channel optical
spectrograph at the Hale 200-inch that replaces DBSP -- plus the
single-slice central-slit ``full_template`` reidentify archives, user
docs, and a CHANGES entry.

NGPS covers 3050-10400 Å at R ~ 4000 via a 3-slice image slicer and a
u/g/r/i dichroic train.  All four channels share one raw
multi-extension FITS file per exposure (one image extension per
channel keyed by ``EXTNAME``).

### Spectrographs added
- ``p200_ngps_u`` (3050-4430 Å)
- ``p200_ngps_g`` (4250-5960 Å)
- ``p200_ngps_r`` (5620-7950 Å)
- ``p200_ngps_i`` (7530-10400 Å)

### Companion PR

[pypeit/PypeIt-development-suite#409](https://github.com/pypeit/PypeIt-development-suite/pull/409) registers a ``1.5_2x3`` reduce test setup for each of the four NGPS channels.  Raw data uploaded; ``pypeit_test reduce`` passes 8/8 locally against this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cfremling:
Dev-suite raw data uploaded; companion PR pypeit/PypeIt-development-suite#409 now runs end-to-end on this branch:

\`\`\`
$ ./pypeit_test reduce -i p200_ngps_u p200_ngps_g p200_ngps_r p200_ngps_i -t 4
--- PYPEIT DEVELOPMENT SUITE PASSED 8/8 TESTS  ---
\`\`\`

Added a follow-up commit (\`f837fb74\`) that fixed three issues surfaced by running pypeit_setup auto-discovery on a mixed-extension raw dir (long ``getcalib_ug`` G+U-only exposures alongside short G+I+R+U exposures):

1. Override ``find_raw_files`` to filter raw files by ``EXTNAME`` so each per-channel spectrograph only sees the files that carry its own image extension.
2. ``compound_meta('binning')`` simplified to just read BINSPEC/BINSPAT from the channel's extension (no sentinels, no primary-header fallback).
3. ``check_frame_type`` on u/g now requires arc/flat frames' extension set to be exactly ``{G, U}`` -- the short 4-channel cal frames have ~zero flux at u/g wavelengths.

Also removed the ``paranal_sky_air.fits`` flexure reference (custom non-stock sky template); spectral flexure stays at PypeIt's default ``spec_method='skip'`` on r/i for now.

If preferred I can squash the two commits before merge.

doc/spectrographs/p200_ngps.rst
```diff
@@ -0,0 +1,78 @@
+.. highlight:: rest
+.. _p200_ngps_doc:
+
+***************
+Palomar / NGPS
```
profxj, line 5: Thanks for these!!

If Claude is feeling frisky, you might generate a HOWTO doc too, although that needs graphics..

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
+            extname = (hdr.get('EXTNAME') or '').strip()
+            if extname == self._extname:
+                return i
+        return None
+
+    def _canonicalize(self, raw):
+        """
+        Apply the per-channel geometric transform that puts the raw
+        image into the canonical "dispersion along axis 1" orientation.
+        Default: identity (R, I).  G overrides with ``np.fliplr``; U
```
profxj: These are usually handled in the detector dict; but this is fine

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
+        self.meta['idname'] = dict(ext=0, card='IMGTYPE')
+        self.meta['instrument'] = dict(ext=0, card='INSTRUME')
+
+        self.meta['lampstat01'] = dict(ext=0, card='LAMPBLUC')
+        self.meta['lampstat02'] = dict(ext=0, card='LAMPFEAR')
+        self.meta['lampstat03'] = dict(ext=0, card='LAMPREDC')
+        self.meta['lampstat04'] = dict(ext=0, card='LAMPTHAR')
 
     def configuration_keys(self):
-        """
```
profxj, line 62: did you mean to wipe out this doc string altogether?

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
-        Returns:
-            :obj:`list`: List of keywords of data pulled from file headers
-            and used to constuct the :class:`~pypeit.metadata.PypeItMetaData`
-            object
-        """
         return ['binning']
-    
 
     def raw_header_cards(self):
-        """
```
profxj, line 79: same question

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
-        if ftype == 'bias':
-            return good_exp & (fitstbl['idname'] == 'BIAS')
-        
-        if ftype in ['pixelflat', 'trace', 'illumflat']:
-            return ((good_exp & (fitstbl['idname'] == 'DOMEFLAT')) | (good_exp & (fitstbl['idname'] == 'CONT')))
-        
-        if ftype in ['pinhole', 'dark']:
-            # Don't type pinhole or dark frames
+
+    _standard_targets = (
```
profxj: why customize this for NGPS?  There is a larger list of ok standards with PypeIt

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
-        Define the list of keys to be output into a standard PypeIt file.
-
-        Returns:
-            :obj:`list`: The list of keywords in the relevant
-            :class:`~pypeit.metadata.PypeItMetaData` instance to print to the
-            :ref:`pypeit_file`.
-        """
         return super().pypeit_file_keys()
-    
-    def check_frame_type(self, ftype, fitstbl, exprng=None):
```
profxj, line 109: Is this method removed altogether?  Seems problematic..

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
+        per-channel spectrograph.  Filtering here means pypeit never
+        attempts to read this channel's BINSPEC/BINSPAT from a file
+        where that extension doesn't exist."""
+        files = super().find_raw_files(root, extension=extension)
+        if not cls._extname:
+            return files
+        return [p for p in files if cls._extname.upper() in cls._file_extnames(p)]
 
-    def get_rawimage(self, raw_file, det):
+    def check_frame_type(self, ftype, fitstbl, exprng=None):
```
profxj, line 352: ok, it is back

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
+            oscansec_str = self._oscansec_string(ext_hdr)
+            oscansec = (np.atleast_1d(parse.flip_fits_slice(oscansec_str))
+                        if oscansec_str else None)
+
+        det_dict = dict(
+            binning=binning,
+            det=1,
+            dataext=dataext,
+            specaxis=1,
+            specflip=False,
```
profxj: here is where you could account for flip, rotation.  if desired

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
+        # unconditionally, so this switch alone is not sufficient if
+        # any slit has no detectable lines; we rely on
+        # ``exclude_regions`` above to keep the slits clean.
+        par['calibrations']['wavelengths']['fwhm_fromlines'] = False
+
         par['scienceframe']['process']['combine'] = 'median'
         par['calibrations']['standardframe']['process']['combine'] = 'median'
 
-        par['scienceframe']['process']['use_overscan'] = True
+        # Apply per-channel overscan policy to every frame-type processing
```
profxj, line 667: These inline comments are super helpful;  thanks for providing them!

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
+        # bspline-with-sigma-rejection is non-deterministic and the
+        # ~10 % swing in PIXELFLAT_NORM was driving the sky-sub
+        # gradients we were chasing.
+        slit_illum_finecorr=False,
+        spec_samp_coarse=1.0,
+    )
+    def _canonicalize(self, raw):
+        return np.fliplr(raw)
+
+    def _canonicalize_section(self, sec_str, ext_hdr):
```
profxj: This is novel;  Might be useful for other spectrographs where folks have struggled with `datasec`

profxj (changes requested):
Mainly minor comments, but I've put a Request changes on just in case I caught anything important (unlikely).  Will approve without further comments

doc/spectrographs/p200_ngps.rst
```diff
@@ -0,0 +1,78 @@
...
+***************
+
+.. _ngps_overview:
+
+Overview
+========
+
+The Palomar Next Generation Spectrograph (NGPS) is the four-channel
```
rcooke-ast: According to the website, I think this should be `Next Generation Palomar Spectrograph (NGPS)`

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
+    All subclasses present their data to PypeIt with dispersion along
+    array axis 1 (i.e. ``specaxis=1``) regardless of native CCD
+    orientation, so that downstream reduction (slit edge tracing,
+    wavelength calibration, sky subtraction) sees a uniform layout.
     """
-    ndet = 1 
+    ndet = 1
     telescope = telescopes.P200TelescopePar()
 
+    # -- Per-channel overrides -----------------------------------------
```
rcooke-ast, line 68: These per-channel overrides are not the standard approach for other spectrographs. I'm not entirely opposed, but I wonder if there's a way that we can stay consistent with other spectrographs? My feeling is that we have a parent class that defines the relevant variables, and then these can be changed in the child class.

EDIT: OK, I now see the beauty in this approach. I can see the benefits, and it really puts a lot more in the parent class, which I like a lot. I wonder how easy it would be to apply this same logic to other spectrographs. In particular, I really like how easy it is to see what's different with each child spectrograph.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,469 +42,953 @@
...
-        self.meta['lampstat01'] = dict(ext=0, card='LAMPBLUC') # Blue Xe
-        self.meta['lampstat02'] = dict(ext=0, card='LAMPFEAR') # FeAr
-        self.meta['lampstat03'] = dict(ext=0, card='LAMPREDC') # Red Continuum
-        self.meta['lampstat04'] = dict(ext=0, card='LAMPTHAR') # ThAr
+        # ``SHUTTIME`` is the actual shutter-open time in seconds; the
+        # primary-header ``EXPTIME`` is in *milliseconds* for some
+        # frames (it's the requested integration time written by the
+        # detector controller).  Use SHUTTIME when present, otherwise
+        # fall back to EXPTIME with an automatic ms-to-s conversion
+        # if the value looks like ms (>1e4).  Implemented in
```
rcooke-ast: This is a slight concern to me. Is there a way that units are stored in the header, or some standard that represents ms or s? If it is truly arbitrary (e.g. up to the observer whether the time is stored as ms or s, then perhaps we can request that the units are stored as well)?

pypeit/spectrographs/p200_ngps.py
rcooke-ast: This seems to be grouped in the "Per-channel overrides" mentioned earlier. Personally, I think this might be better to remove it.

doc/spectrographs/p200_ngps.rst
kbwestfall: Agreed.  These are very nice docs!

I wondering if you have plans to work on recommendations for coadding spectra across channels and between slices such that users can get a single extracted spectrum that covers the full wavelength range of all channels.  Of course, no need to include that in this PR, but it would be great to have this included in a future PR.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -19,44 +42,236 @@
...
+    Each per-channel subclass sets ``_extname`` to the EXTNAME of its
+    image extension in the raw FITS file ('U', 'G', 'R', or 'I') and
+    optionally overrides ``_canonicalize`` to apply a per-channel
+    geometric transform (no-op for R/I, ``fliplr`` for G, ``rot90(-1)``
+    for U).
+
+    All subclasses present their data to PypeIt with dispersion along
+    array axis 1 (i.e. ``specaxis=1``) regardless of native CCD
+    orientation, so that downstream reduction (slit edge tracing,
+    wavelength calibration, sky subtraction) sees a uniform layout.
```
kbwestfall: Similar to @profxj 's comment, I'm not sure what to recommend here.  I like your implementation, but it's different from what all the other spectrographs do.  It also means the NGPS entries in [this table](https://pypeit.readthedocs.io/en/stable/detectors.html#instrument-specific-data) won't be correct.

Am I right that you would get the same result if you used the following?

- `specaxis = 0, specflip = False`, for the U band
- `specaxis = 1, specflip = True`, for the G band

I don't know if this orients the slices in the same order though.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -101,387 +310,691 @@ def pypeit_file_keys(self):
...
 
+    @classmethod
+    def _file_extnames(cls, path) -> frozenset[str]:
+        """Cached EXTNAME set of the image HDUs in this raw FITS."""
+        key = str(path)
+        cached = cls._extname_cache.get(key)
+        if cached is None:
+            from astropy.io import fits as _fits
+            try:
+                with _fits.open(path, memmap=True) as h:
```
kbwestfall: In the short term, please move the import to the top of the file and don't use `as`.  In the longer term, we should improve `pypeit.io.fits_open` so that it allows for this kind of context management.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -101,387 +310,691 @@ def pypeit_file_keys(self):
...
+        # 4-channel cal frames have ~zero flux at u/g wavelengths.
+        if (self._require_ug_only_cals
+                and ftype in ('arc', 'tilt', 'pixelflat',
+                              'trace', 'illumflat')):
+            import os as _os
+            for i in range(len(fitstbl)):
+                if not mask[i]:
+                    continue
+                fpath = _os.path.join(str(fitstbl['directory'][i]),
+                                       str(fitstbl['filename'][i]))
```
kbwestfall: Similar to above, please put the import at the top of the file and don't use `as`.  We prefer to use `pathlib.Path` instead of `os.path` in new code, if possible.

kbwestfall (changes requested):
Thanks!  I only have a few minor comments and requests.

I particularly like that the subclasses of the main NGPS spectrograph class are so clean!  Makes me think we should adopt this approach for other multi-channel spectrographs...

pypeit/spectrographs/p200_ngps.py
```diff
@@ -101,387 +310,691 @@ def pypeit_file_keys(self):
...
+            # can't be a plausible visible-band exposure in seconds
+            # (ZTF/SN spectra are 60-3600 s; values >= 10000 are ms).
+            if ph is not None:
+                shut = ph.get('SHUTTIME')
+                if shut is not None:
+                    try:
+                        return float(shut)
+                    except (TypeError, ValueError):
+                        pass
+                ex = ph.get('EXPTIME')
```
rcooke-ast, line 440: I couldn't find out online what the actual cut off is (or if there are units available in the header), but there must be a standard approach that we can employ. I feel uncomfortable with this approach, as it could throw off the flux calibration without the user knowing.

pypeit/spectrographs/p200_ngps.py
rcooke-ast: I agree -- I think it's better to use the functionality already available, if possible.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -101,387 +310,691 @@ def pypeit_file_keys(self):
...
+        # ``_use_overscan=False`` must take effect for biasframe /
+        # arcframe / pixelflatframe processing too, not just science.
+        # Spatial flexure correction is enabled for every science /
+        # standard frame: PypeIt cross-correlates the dome-flat slit
+        # pattern with the slit pattern in each science exposure to
+        # measure the spatial shift in pixels and propagates it
+        # through extraction.  This is essential at airmass >1.2 where
+        # atmospheric refraction shifts the slit relative to the
+        # dome-flat reference position.
+        for ftype in ('scienceframe', 'biasframe', 'arcframe',
```
rcooke-ast, line 679: I think this is covered with the `baseprocess`?

`par['baseprocess']['use_overscan'] = cls._use_overscan`

rcooke-ast (commented):
Thanks very much for this great new implementation! Overall, this is a very interesting approach to the spectrograph class design. I'm not opposed, but it deserves some more discussion/thought if we should implement something like this more widely.

Also, I think we'll need some data in the dev-suite to test each of the detectors with this new class.

pypeit/spectrographs/p200_ngps.py
kbwestfall: Chiming in here just to say this suggestion won't work.  The `baseprocess` parameter group is only something that can be used in the `pypeit` file.  It's a programmatic instruction that tells pypeit to propagate the provided parameters to all frame types, so that users don't have to explicitly include the same parameter changes for each type.  The `PypeItPar` class doesn't actually have a `baseprocess` element.

doc/spectrographs/p200_ngps.rst
cfremling: The current slicer unit unfortunately has enough shortcomings (gaps between slices and throughput loss) that we do not believe it worthwhile to attempt signal stacking (for the purpose of increasing total SNR) at the moment. It is of course still possible for the user to try to stack the resulting 1d spectra to determine if any SNR gain is possible or not in their particular science observation. 

On cross channel stitching, we are keeping the post processing contained here https://github.com/cfremling/ngps_pipeline. to get reasonably clean and calibrated stitched 1d spec there are a number of steps needed and putting all this in the main class does not seem appropriate.

pypeit/spectrographs/p200_ngps.py
cfremling: will be fixed in upcoming commit

pypeit/spectrographs/p200_ngps.py
cfremling: will be fixed in upcoming commit

pypeit/spectrographs/p200_ngps.py
cfremling: tested and we need 

specaxis = 0, specflip = True, for the U band
specaxis = 1, specflip = True, for the G band

for bit identical outcome. this is changed in the upcoming commit

pypeit/spectrographs/p200_ngps.py
cfremling: This comment isnt accurate -- there is no uncertainty element here. SHUTTIME is seconds, EXPTIME is ms, always. updated comment in next commit

pypeit/spectrographs/p200_ngps.py
cfremling: addressed in latest commit

pypeit/spectrographs/p200_ngps.py
cfremling: Restored in 4950259 — these (along with `raw_header_cards` and `pypeit_file_keys`) were collateral damage from the comment-stripping pass, not intentional. Back to the upstream-style docstrings now.

pypeit/spectrographs/p200_ngps.py
cfremling: Same as above — restored in 4950259.

pypeit/spectrographs/p200_ngps.py
cfremling: Done in 4950259: dropped the hand-rolled `_standard_targets` / `_standard_target_mask` pair. `check_frame_type` now returns the same `SCI` mask for 'science' and 'standard', leaving the discrimination to PypeIt's stock coordinate matching against the archived flux-standard registry.

pypeit/spectrographs/p200_ngps.py
cfremling: Not removed — `check_frame_type` just moved below the new EXTNAME helpers in the diff (you spotted it again at the line below). No functionality dropped.

pypeit/spectrographs/p200_ngps.py
cfremling: Right — just relocated in the diff, not removed. Thanks for double-checking.

pypeit/spectrographs/p200_ngps.py
cfremling: Agreed these usually live in the detector dict; they feed straight into `get_detector_par`. I kept them as class attrs so each per-channel subclass reads as a flat list of just-the-differences, but happy to inline them into the dict if you'd prefer.

pypeit/spectrographs/p200_ngps.py
cfremling: Thanks!

pypeit/spectrographs/p200_ngps.py
cfremling: Thanks — glad it might be reusable. Happy to factor the malformed-DATASEC fallback out into something more general if that would help other instruments.

pypeit/spectrographs/p200_ngps.py
cfremling: Done in 2d9a6c8 — switched to PypeIt's stock `specaxis`/`specflip` (g: `specaxis=1, specflip=True`; u: `specaxis=0, specflip=True`; r/i unchanged). `get_rawimage` now returns the native-frame image and `orient_image` does the geometry. I verified the oriented image **and** the DATASEC/OSCANSEC pixel maps are byte-for-byte identical to the previous `_canonicalize` output for all four channels (end-to-end spec2d differences are within the pipeline's pre-existing run-to-run non-determinism), and the detectors table now reports the true per-channel orientation.

doc/spectrographs/p200_ngps.rst
cfremling: Fixed in 2d9a6c8 — now "Next Generation Palomar Spectrograph (NGPS)".

pypeit/spectrographs/p200_ngps.py
cfremling: Agreed with @kbwestfall — `baseprocess` is a pypeit-file-only directive and `PypeItPar` has no `baseprocess` element, so the per-frametype loop in `default_pypeit_par` (setting `use_overscan` on each frame type) is the right mechanism. Left as-is.

profxj (approved):
Thanks for all the edits, looks great!

kbwestfall:
Just a heads up that I started the dev-suite tests for this branch.  They should finish tomorrow, and then I think this is ready to merge!

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  353 passed, 448 warnings in 136.65s (0:02:16) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 1449 warnings in 430.42s (0:07:10) ---
--- PYTEST VET TESTS PASSED  75 passed, 1671 warnings in 2219.55s (0:36:59) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 296/296 TESTS  ---
Total disk usage: 236.697 GiB
Testing Started at 2026-06-12T17:29:14.647214
Testing Completed at 2026-06-12T23:28:31.902459
Total Time: 5:59:17.255245
```
Merging

kbwestfall (approved):

PR #2136, opened 2026-05-22
Clean out development-only scripts in `pypeit/core/wavecal`

kbwestfall:
As titled.

There is a new `core_refactor` branch in the dev-suite, where the files deleted from here have been moved to `pypeitdev/wavecal`.


bpholden (approved):
Moving sale.

PR #2140, opened 2026-06-01
Add wavelength template for VLT FORS2 GRIS_1200R

Mao-PH:
(1) spectral binning = 2
(2) slit spat_id = 481


kbwestfall:
Thanks, @Mao-PH !

I'm basically ready to approve this, but I'd like to better integrate the new template into the code.  To that end:

- Do you have a 1200R dataset that you're willing to share so that we can add it to our dev-suite?  It just needs to be a minimal dataset, like those that already exist for the other gratings; see [here](https://drive.google.com/drive/folders/1JL7VWjvNDhjaVI5y75_H30cM1thq-Jz2?usp=sharing).  If so, we can upload it to the Google drive and integrate it into our testing suite.
- It would be nice to change the code so that it uses this template by default.  To do so, you would add lines [here](https://github.com/pypeit/PypeIt/blob/d5f8d74da16e131d8c1513bae4c193df280002d4/pypeit/spectrographs/vlt_fors.py#L359), similar to the parameter changes used for other gratings.
- I think we'd actually prefer this be merged into the `develop` branch.  If you're willing to recreate your branch starting from `develop` and reissue the PR, that would be great.  Otherwise, we can do this for you.

PR #2141, opened 2026-06-01
Dashboard prototype

NominalAnxiety:
This is the dashboard prototype.


profxj:
As this is simply a PR to the `status` branch, I am merging it in
and will pick up the pieces there

Thanks!

Issue #2116, opened 2026-05-11
Revisit cube construction for KCWI/KCRM to make sure both arms are consistent


kbwestfall:
As titled.  See https://github.com/pypeit/PypeIt/pull/1929#discussion_r3203387394

rcooke-ast:
Addressed by PR #2147

PR #2147, opened 2026-06-05
WCS and alignment operations fixed for datacubes

rcooke-ast:
This is an extension to PR #1929 that fixes the alignment and WCS values discussed there. I will run the dev-suite to make sure I haven't broken everything, but all of the modes that I have tested work. This PR also includes a test (it's not ideal, because it doesn't touch the `coadd3d` file, but at least it's a fast way to confirm the core alignments codes). This addresses (I think!) Issue #2116.

I have also made some coding improvements that increased the speed (by almost 2x) of creating datacubes.

### DIFF
--- a/pypeit/core/datacube.py
+++ b/pypeit/core/datacube.py
@@ -2365,11 +2365,9 @@ def subpixellate(
                 # Interpolate the RA/Dec over the subpixel spatial positions
                 tmp_ra = this_ra[this_sl]
                 tmp_dec = this_dec[this_sl]
-                ra_spl = interp1d(spatpos[ssrt], tmp_ra[ssrt], kind='linear', bounds_error=False, fill_value='extrapolate')
-                dec_spl = interp1d(spatpos[ssrt], tmp_dec[ssrt], kind='linear', bounds_error=False, fill_value='extrapolate')
                 # Evaluate the RA/Dec at the subpixel spatial positions
-                this_ra_int = ra_spl(spatpos_subpix)
-                this_dec_int = dec_spl(spatpos_subpix)
+                this_ra_int = utils.linear_interpolate_extrapolate(spatpos_subpix, spatpos[ssrt], tmp_ra[ssrt])
+                this_dec_int = utils.linear_interpolate_extrapolate(spatpos_subpix, spatpos[ssrt], tmp_dec[ssrt])
                 # Now apply the DAR correction and any user-supplied offsets
                 this_ra_int += ra_corr + _ra_offset[fr]
                 this_dec_int += dec_corr + _dec_offset[fr]
--- a/pypeit/core/datacube.py
+++ b/pypeit/core/datacube.py
@@ -692,7 +692,7 @@ def extract_point_source(
     if fluxed:
         sobj.OPT_FLAM = sobj.OPT_COUNTS
         sobj.OPT_FLAM_SIG = sobj.OPT_COUNTS_SIG
-        sobj.OPT_FLAM_IVAR = sobj.OPT_COUNTS_IVARf
+        sobj.OPT_FLAM_IVAR = sobj.OPT_COUNTS_IVAR
 
     # Make a specobjs object
     sobjs = specobjs.SpecObjs()
--- a/pypeit/core/datacube.py
+++ b/pypeit/core/datacube.py
@@ -97,7 +97,7 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
         The inverse variance of the image. Optional. If not passed, the standard deviation computed
         from the image will be used to compute the inverse variance. Default is None.
     gpm : `numpy.ndarray`_, optional
-        A good pixel mask. Pixels that are True are good. Default is None,
+        A good pixel mask. Pixels that are True are good. Default is None.
     init_obj_position : tuple, optional
         The initial guess for the object position in the image with format
         (x, y). If set, the 2D Gaussian fit will be performed with the position constrainted
@@ -150,7 +150,7 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
     # Normalise if requested
     wlscl = np.max(image) if norm else 1.0
     if ivar is None: 
-        mean, median, std = sigma_clipped_stats(image[np.logical_not(totmask)], sigma=3.0)
+        mean, median, std = sigma_clipped_stats(image[_gpm], sigma=3.0)
         if std > 0:
             _ivar = np.full_like(image, 1.0/std**2)
         else:
--- a/pypeit/core/datacube.py
+++ b/pypeit/core/datacube.py
@@ -149,15 +149,6 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
     sigma = fwhm*fwhm2sigma
     # Normalise if requested
     wlscl = np.max(image) if norm else 1.0
-    if ivar is None: 
-        mean, median, std = sigma_clipped_stats(image[_gpm], sigma=3.0)
-        if std > 0:
-            _ivar = np.full_like(image, 1.0/std**2)
-        else:
-            log.warning('Could not measure standard deviation from image.  Assuming 1.')
-            _ivar = np.ones_like(image)
-    else: 
-        _ivar = ivar
 
     ## Find the objects
     ximg = np.tile(np.arange(image.shape[1]), (image.shape[0], 1))
@@ -165,6 +156,16 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
     edgemask = (ximg < mask_edge) | (ximg >= image.shape[1] - mask_edge) | \
                 (yimg < mask_edge) | (yimg >= image.shape[0] - mask_edge)
     totmask = edgemask | np.logical_not(_gpm)
+
+    if ivar is None:
+        mean, median, std = sigma_clipped_stats(image[np.logical_not(totmask)], sigma=3.0)
+        if std > 0:
+            _ivar = np.full_like(image, 1.0/std**2)
+        else:
+            log.warning('Could not measure standard deviation from image.  Assuming 1.')
+            _ivar = np.ones_like(image)
+    else:
+        _ivar = ivar
 
     if init_obj_position is None: 
         if DAOStarFinder is None:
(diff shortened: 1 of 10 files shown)


pypeit/core/datacube.py
```diff
@@ -5,6 +5,7 @@
 """
 
 import os
+# import line_profiler
```
kbwestfall, line 8: Remove?

pypeit/core/datacube.py
```diff
@@ -198,14 +200,15 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
                     sources[col].info.format = '%.2f'  # for consistent table output
             sources.pprint(max_width=76)
         if sources is None:
+            embed()
```
kbwestfall: Remove

pypeit/core/datacube.py
```diff
@@ -198,14 +200,15 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
...
             display.show_image((objfind_image*np.logical_not(totmask)*np.sqrt(ivar_objfind)),
                             chname='S/N objfind_image', cuts=(-2.0, 5.0))
             raise PypeItError(
                 "No sources found in the image. Try lowering the significance threshold, "
                 f"nsigma = {nsigma:.1f} or adjust the DAOStarFinder parameters."
             )
 
-        _init_obj_position = sources['ycentroid'][0], sources['xcentroid'][0]
+        _init_obj_position = sources['y_centroid'][0], sources['x_centroid'][0]
```
kbwestfall, line 210: Was this a bug? Or has the column name changed in different versions of photutils?  I.e., is this something that we need to think about in terms of the version requirement for photutils?

pypeit/core/datacube.py
```diff
@@ -1393,14 +1396,14 @@ def wcs_bounds(raImg, decImg, waveImg, slitid_img_gpm, ra_offsets=None, dec_offs
         # Get the RA, Dec, and wavelength of the pixels on the slit
         if ra_min is None or ra_max is None:
             this_ra = _raImg[fr][_slitid_img_gpm[fr] > 0]
-            tmp_min, tmp_max = np.min(this_ra)+_ra_offsets[fr], np.max(this_ra)+_ra_offsets[fr]
+            tmp_min, tmp_max = np.min(this_ra)-_ra_offsets[fr], np.max(this_ra)-_ra_offsets[fr]
```
kbwestfall: Can we separate these into two lines?

pypeit/core/datacube.py
```diff
@@ -1393,14 +1396,14 @@ def wcs_bounds(raImg, decImg, waveImg, slitid_img_gpm, ra_offsets=None, dec_offs
...
             if fr == 0 or tmp_min < _ra_min:
                 _ra_min = tmp_min
             if fr == 0 or tmp_max > _ra_max:
                 _ra_max = tmp_max
         if dec_min is None or dec_max is None:
             this_dec = _decImg[fr][_slitid_img_gpm[fr] > 0]
-            tmp_min, tmp_max = np.min(this_dec)+_dec_offsets[fr], np.max(this_dec)+_dec_offsets[fr]
+            tmp_min, tmp_max = np.min(this_dec)-_dec_offsets[fr], np.max(this_dec)-_dec_offsets[fr]
```
kbwestfall: Same here.

pypeit/core/datacube.py
```diff
@@ -2164,7 +2167,7 @@ def generate_cube_subpixel(
     return flxcube, sigcube, bpmcube, normcube, wave
 
 
-
+# @line_profiler.profile
```
kbwestfall, line 2173: Remove?  I'm okay with you leaving these `line_profiler` lines, but I would add a comment just reminding other developers how they were used.

pypeit/core/datacube.py
```diff
@@ -2310,6 +2315,13 @@ def subpixellate(
         this_sci = _sciImg[fr][this_onslit_gpm]
         this_var = utils.inverse(_ivarImg[fr][this_onslit_gpm])
         this_wav = _waveImg[fr][this_onslit_gpm]
+        slshape = (this_slits.nspec, this_slits.nspat, slice_subpixel)
+        raimg_slc, decimg_slc = np.zeros(slshape, dtype=float), np.zeros(slshape, dtype=float)
```
kbwestfall: split this into two lines?  Sorry if I'm being obnoxious, but I find it more clear to have these kinds of statements on two separate lines.

pypeit/core/datacube.py
```diff
@@ -2380,12 +2383,13 @@ def subpixellate(
             else:
                 if verbose: 
                     log.info("Preparing subpixel weights")
-                vox_index = np.floor(outshape * (vox_coord - binrng[:,0].reshape((1, 1, 3))) /
-                                                (binrng[:,1] - binrng[:,0]).reshape((1, 1, 3))).astype(int)
+                vox_index = np.floor(vox_coord * voxscale - voxoffset).astype(int)
+                # vox_index = np.floor(outshape * (vox_coord - binrng[:,0].reshape((1, 1, 3))) /
+                #                                 (binrng[:,1] - binrng[:,0]).reshape((1, 1, 3))).astype(int)
```
kbwestfall: remove?

pypeit/tests/test_datacube.py
```diff
@@ -0,0 +1,193 @@
...
+        dec_shifts = (dec_pix_star[ref_idx] - dec_pix_star) * _dspat
+        ra_offsets = [ra_offsets[ff] - ra_shifts[ff] for ff in range(numfiles)]
+        dec_offsets = [dec_offsets[ff] - dec_shifts[ff] for ff in range(numfiles)]
+
+    # Check that the offsets agree to within 1/3 of a pixel
+    atol = _dspat.to(u.arcsec).value/3
+    assert np.isclose(ra_offsets[0].to(u.arcsec).value, 0.0, atol=atol)
+    assert np.isclose(dec_offsets[0].to(u.arcsec).value, 0.0, atol=atol)
+    assert np.isclose(ra_offsets[1].to(u.arcsec).value, offs_ra.to(u.arcsec).value, atol=atol)
+    assert np.isclose(dec_offsets[1].to(u.arcsec).value, offs_dec.to(u.arcsec).value, atol=atol)
```
kbwestfall, line 248: Thanks so much for the new tests!!  Can we break this single test function into multiple separate functions, or does the entire test need to run for it to make sense?

pypeit/utils.py
```diff
@@ -1481,6 +1522,30 @@ def linear_interpolate(x1, y1, x2, y2, x):
     return y1 if np.isclose(x1,x2) else y1 + (x-x1)*(y2-y1)/(x2-x1)
 
 
+def linear_interpolate_extrapolate(x, xp, fp):
```
kbwestfall, line 1525: Do we need another function for this?  Is it faster than `scipy.interpolate.interp1d`?

kbwestfall (commented):
Thanks, @rcooke-ast !  Two main comments from me:

 - Can we split up the new datacube test?
 - Do we need the new `linear_interpolate_extrapolate` function?

The rest are some code clean-up comments.

pypeit/core/datacube.py
rcooke-ast: If it's OK, I'm keen to keep this in, as I hope to continue making improvements to the speed of this code.

pypeit/core/datacube.py
rcooke-ast: Indeed - thanks for spotting that!

pypeit/core/datacube.py
rcooke-ast: The previous code raised a deprecation warning, so I implemented this to "future proof" the code. So, it wasn't a bug, and `photutils` is not a dependency of PypeIt, so I've changed the warning raised to explicitly mention the required version of `photutils`.

pypeit/core/datacube.py
rcooke-ast: Yep, done!

pypeit/core/datacube.py
rcooke-ast: Done as well!

pypeit/core/datacube.py
rcooke-ast: I'll leave it in and commented out, if that's OK? I've added some developer notes so that folks don't use it in an unintended way.

pypeit/core/datacube.py
rcooke-ast: Absolutely, no problems at all. Done.

pypeit/core/datacube.py
rcooke-ast: Thanks for spotting this - removed 👍

pypeit/tests/test_datacube.py
rcooke-ast: Sure, I can break it up into three separate tests. That makes more sense. They are pseudo-independent, and it would be good to know if one crashes when another doesn't.

pypeit/utils.py
rcooke-ast: This one is purpose-built, so it's faster than `scipy.interpolate.interp1d` based on my tests. I'm happy to revert if you prefer, but this is one of the bottlenecks and so speeding things up by 10% makes a bit of a difference.

rcooke-ast (commented):
Thanks @kbwestfall -- comments addressed!

pypeit/core/datacube.py
kbwestfall: Just to be clear, it's an optional dependency.  I.e., in `pyproject.toml`, we have
```
photutils = [
    "photutils>=2.0",
]
```
and `photutils` is also in the `dev` list of dependencies.  Thanks for future proofing the code!!

rcooke-ast: I've updated the toml file -- please check to make sure I've done this correctly!!

pypeit/utils.py
kbwestfall: Yep, makes sense.  Side rant:  I find it frustrating that the scipy team has designated `scipy.interpolate.interp1d` as "legacy" code, suggesting people use `numpy.interp` instead.  `numpy.interp` effectively can't extrapolate, meaning people will constantly be writing their own extrapolation code.

PR #2149, opened 2026-06-09
Kcwi dec 2024 weights

rcooke-ast:
An extension to the sequence of PRs related to #1929. In this PR, I have:

* Added weights as an optional parameter to `weighted_combine()` (and propagated changes).
* Moved the combination of datacubes in `coadd3d.run()` to a new routine. The code flow is largely unchanged, but I made some modifications (moved some bits outside of a for loop that didn't need to be updated every loop).
* Added a `save_separate` parameter to output individual datacubes (on the final combined WCS) if requested. Previously this was output by default.

Dev-suite is running on the cloud, but the relevant parts pass on my machine. There is a related dev-suite PR.


rcooke-ast:
Tests pass...
image

pypeit/images/combineimage.py
```diff
@@ -215,12 +215,13 @@ def run(self, ignore_saturation=False, maxiters=5):
...
             # calculate the mean of the images
-            [mean_img], _, mean_gpm, _ = combine.weighted_combine(np.ones(self.nimgs, dtype=float)/self.nimgs,
-                                                                  [img_stack],
+            [mean_img], _, mean_gpm, _ = combine.weighted_combine([img_stack],
                                                                   [rn2img_stack],
                                                                   # var_list is added because it is
                                                                   # required by the function but not used
-                                                                  gpm_stack, sigma_clip=self.par['clip'],
+                                                                  gpm_stack,
+                                                                  weights=None,
```
kbwestfall, line 223: Adding weights to the function call isn't necessary, right?  I assume this is mostly included just to make clear to a developer that the stack is unweighted?

pypeit/images/combineimage.py
```diff
@@ -240,12 +241,10 @@ def run(self, ignore_saturation=False, maxiters=5):
...
         # Coadd them
         if self.par['combine'] == 'mean':
-            weights = np.ones(self.nimgs, dtype=float)/self.nimgs
             img_list_out, var_list_out, gpm, nframes \
-                    = combine.weighted_combine(weights,
-                                               [img_stack, scl_stack],  # images to stack
+                    = combine.weighted_combine([img_stack, scl_stack],  # images to stack
                                                [rn2img_stack, basev_stack], # variances to stack
-                                               gpm_stack, sigma_clip=self.par['clip'],
+                                               gpm_stack, weights=None, sigma_clip=self.par['clip'],
```
kbwestfall, line 247: Same here?

pypeit/coadd3d.py
```diff
@@ -1739,6 +1740,149 @@ def compute_weights(self, show_qa=False):
...
+        # 1. Performing an intiial sigma clipping of the cubes.
+        # 2. Computing an initial preliminary stacked cube.
+        # 3. Generate a whitelight image from the preliminary stacked cube.
+        # 4. Perform object finding on this stacked cube.
+        # 5. Compute the weights at the location of the object by extracting spectra from the individual cubes,
+        #    probably there should be an option to use the optimal extraction method  (extract_point_source)
+        #    or one can use the single pixel computation (extended sources) in compute_weights.
+        # 6. Re-combined the cubes using these weights, again performing the final round of sigma clipping
+        # 7. Write out the individual cubes with their sigma clipped pixels masked (?)
+        # 8. Write out the final combined cube.
```
kbwestfall: I assume this is still a "TODO"?  I'd prefer we move the comments out of the docstring, or actually use docstring syntax.

kbwestfall (approved):
Very minor comments.  The only actual change requested is related to a docstring.  I'll make this change myself.

pypeit/images/combineimage.py
rcooke-ast: That's correct. I think it's a bit strange that we have a function called `weighed_combine()` where the weights are not currently being used, so just being explicit (and to make sure this remains supported the way it was originally intended in case the defaults change somewhere down the line.

pypeit/images/combineimage.py
rcooke-ast: As above  :-)  Happy to remove these, if preferred.

PR #2150, opened 2026-06-09
Fix datacube extraction x-y coordinates

jhennawi:
This PR fixes some coordinate transpose errors that existed in the extraction code after the WCS and transpose fixes were implemented.


pypeit/tests/test_datacube.py
```diff
@@ -7,12 +7,79 @@
...
 from astropy.wcs import WCS
 import astropy.units as u
 from astropy.coordinates import SkyCoord
 
 from pypeit.core import datacube
 from pypeit.core.flexure import calculate_image_phase
+from pypeit.coadd3d import DataCube
+
+
+def make_test_datacube(whitelight_range=None):
```
rcooke-ast, line 20: Great addition!

pypeit/tests/test_datacube.py
```diff
@@ -7,12 +7,79 @@
...
+    wcs.wcs.cunit = ['deg', 'deg', 'Angstrom']
+    wcs.wcs.ctype = ['RA---TAN', 'DEC--TAN', 'WAVE']
+    return wcs
+
+
+def test_datacube_whitelight_range_datamodel(tmp_path):
+    wl_range = np.array([5004.0, 5012.0])
+    cube = make_test_datacube(whitelight_range=wl_range)
+
+    assert np.allclose(cube.resolve_whitelight_range([None, None]), wl_range)
```
rcooke-ast, line 53: For all of these asserts, should we add the optional string as well so that an appropriate error is reported in the event that an assertion fails?

pypeit/tests/test_datacube.py
```diff
@@ -232,9 +299,9 @@ def test_align_fit():
                     # ivar=utils.inverse(np.square(sig_imgs[ff])),
                     # gpm=np.logical_not(bpm_imgs[ff]),
                     fwhm=fwhm.value/_dspat.value, norm=False
-                )
+            )
             gaussian_position = popt[1], popt[2]
-            dec_pix_star[ff], ra_pix_star[ff] = gaussian_position
+            ra_pix_star[ff], dec_pix_star[ff] = gaussian_position
```
rcooke-ast, line 304: With the new transposed cubes, the original version is correct. `dec` corresponds to the first element, and `ra` corresponds to the second element. I suspect this test will fail.

pypeit/coadd3d.py
```diff
@@ -45,10 +45,12 @@ class DataCube(datamodel.DataContainer):
             Wavelength array of the spectral blaze function
         blaze_spec (`numpy.ndarray`_):
             The spectral blaze function
-        sensfunc (`numpy.ndarray`_, None):
-            Sensitivity function (nwave,). Only saved if the data are fluxed.
         PYP_SPEC (str):
             Name of the PypeIt Spectrograph
+        whitelight_range (`numpy.ndarray`_, None):
+            Default wavelength range to use for white-light images and extraction.
+        sensfunc (`numpy.ndarray`_, None):
```
rcooke-ast, line 52: Thanks for including this 👍

pypeit/coadd3d.py
```diff
@@ -203,6 +219,45 @@ def from_file(cls, ifile, verbose=True, chk_version=True, **kwargs):
...
         return self
 
+    def resolve_whitelight_range(self, whitelight_range):
+        """
+        Resolve the white-light wavelength range for extraction.
+
+        Any user-provided wavelength bound overrides the cube default.  Missing
+        bounds are filled from the white-light range stored in the datacube; if
+        that metadata is unavailable, the full cube wavelength range is used.
+        """
```
rcooke-ast, line 229: Need to add `Parameters` and `Returns` to the docstring. Also, I suspect `whitelight_range` should be an optional input, with the default value being `None`?

pypeit/coadd3d.py
```diff
@@ -1604,7 +1688,7 @@ def run_align(self, fwhm=1.5, show_qa=False):
                             np.logical_not(bpm_imgs[:, :, ff]), model, gaussian_position,
                             init_obj_position, channel_prefix = f'Img_{ff}'
                         )
-                    dec_pix_star[ff], ra_pix_star[ff] = gaussian_position
+                    ra_pix_star[ff], dec_pix_star[ff] = gaussian_position
```
rcooke-ast, line 1691: Same comment as above. I think you need to revert this to the original. The transposed operation means dec is the first index, ra is the second.

pypeit/core/datacube.py
```diff
@@ -193,21 +202,21 @@ def fitGaussian2D(image, ivar=None, gpm=None, init_obj_position=None,
...
             )
+        if verbose:
+            log.info('DAOStarFinder brightest source properties')
+            for col in sources.colnames:
+                if col not in ('id', 'npix'):
+                    sources[col].info.format = '%.2f'  # for consistent table output
+            sources.pprint(max_width=76)
 
-        _init_obj_position = sources['y_centroid'][0], sources['x_centroid'][0]
+        _init_obj_position = sources['x_centroid'][0], sources['y_centroid'][0]
```
rcooke-ast, line 219: I think this needs to be reverted. The recent change to the transposed cubes means that dec is always first index, ra is second.

rcooke-ast (changes requested):
Thanks @jhennawi - lots of good additions here (and I hope the run went well)!

We can chat about this some more, but I believe there is an issue with the main changes that you've made. You can double check, but I think the `test_datacube` test will now fail. The new transposed datacubes have `dec` as the first index, and `ra` as the second index. We can't change that. So, the `test_datacube` ordering was correct (but, by all means check this). This means that I think we will need to revert many (but not all) of the changes you've made here. For the record, I never checked the ginga display, so I think there might be some bugs with the displayed/reported values in the `extract spectrum` code.

pypeit/core/datacube.py
jhennawi: There is no WCS implied here. I verified that the extraction in the previous version was broken because X and Y were flipped, and these modifications fix it. The issue is simply image conventions between DAOstarfind and what was previously transposed.

pypeit/tests/test_datacube.py
jhennawi: Thanks Codex (not me)!

pypeit/tests/test_datacube.py
jhennawi: I agree that should be done. This test was written by AI and I didn't inspect it that carefully.

pypeit/tests/test_datacube.py
jhennawi: Okay, I think this issue may be specific to this test, and should probably be reverted.

pypeit/coadd3d.py
jhennawi: Yeah this was important so that the conventions for creating these images would symmetric between cube construction and extraction.

pypeit/coadd3d.py
jhennawi: Okay I think the issue is what DAOStarfind returns, and its convention. I think this may be correct. Perhaps it is more clear to transpose the image input there?

kbwestfall:
Superseded by #2184 .

PR #2151, opened 2026-06-11
Error in rescaled coordinates for quasi-2D bspline construction

kbwestfall:
When using the `x2` functionality in `bspline`, the default xmin and xmax are set to 0 and 1.  You can set them to something different by directly accessing the `bspline` attributes after an instance is created, as is done in `bspline_profile`.  However, if you don't, the rescaling of the `x2` data will not correctly map to be within the range [-1,1] as needed for the orthogonal basis functions, except when `x2` is in the range [0,1].

This PR sets the default `xmin` and `xmax` to None so that they are dynamically defined by the `action` method.  That method also checks that the range of `x2` is actually within the limits of `xmin` and `xmax`; it raises a warning if not.

This should be mostly invisible to the code, but I need to run the dev-suite to check.

### DIFF
--- a/pypeit/bspline/bspline.py
+++ b/pypeit/bspline/bspline.py
@@ -428,6 +428,10 @@ class bspline(datamodel.DataContainer):
             self.xmax = x2.max()
         if self.xmin == self.xmax:
             self.xmax = self.xmin + 1
+            warnings.warn(
+                'Minimum and maximum value used to rescale the range for x2 are identical.  '
+                'Adjusting xmax but be wary of the output!'
+            )
         if self.xmin > x2.min() or self.xmax < x2.max():
             warnings.warn('Rescaled range for x2 will not remap linearly to the range [-1,1]!')
 
--- a/pypeit/bspline/bspline.py
+++ b/pypeit/bspline/bspline.py
@@ -137,8 +137,8 @@ class bspline(datamodel.DataContainer):
         else:
             self.coeff = np.zeros((nc,), dtype=float)
             self.icoeff = np.zeros((nc,), dtype=float)
-        self.xmin = 0.0
-        self.xmax = 1.0
+        self.xmin = None
+        self.xmax = None
         self.funcname = funcname
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


pypeit/bspline/bspline.py
```diff
@@ -423,6 +422,15 @@ def action(self, x, x2=None):
         if x2.size != nx:
             raise ValueError('Dimensions of x and x2 do not match.')
 
+        if self.xmin is None:
+            self.xmin = x2.min()
+        if self.xmax is None:
+            self.xmax = x2.max()
+        if self.xmin == self.xmax:
+            self.xmax = self.xmin + 1
```
tbowers7, line 430: Will this produce the proper output range for arbitrary `self.xmin = self.xmax`?

tbowers7 (approved):
I think this looks okay.  I left one question related to `self.xmin = self.xmax`.

Approving.

pypeit/bspline/bspline.py
kbwestfall: Good call.  I added a warning.  I'm wary of raising an error...

profxj (approved):
Whoa, nice catch.  

I'd suggest we add a test, but I think
we are going to catch any issues with
the code itself

kbwestfall:
Tests pass.  The NGPS failure is because the tests were started before the new NGPS branch was merged.
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  353 passed, 448 warnings in 348.54s (0:05:48) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 1449 warnings in 836.71s (0:13:56) ---
--- PYTEST VET TESTS PASSED  75 passed, 1663 warnings in 5408.51s (1:30:08) ---
--- PYPEIT DEVELOPMENT SUITE FAILED 1/300 TESTS  ---
Failed tests:
    p200_ngps_i/1.5_2x3 pypeit
Skipped tests:
Total disk usage: 311.181 GiB
Testing Started at 2026-06-12T17:38:34.528548
Testing Completed at 2026-06-13T08:20:38.345698
Total Time: 14:42:03.817150
```

PR #2155, opened 2026-06-26
Limit the number of bspline breakpoints when fitting slit illumination profiles

kbwestfall:
As titled.

When testing the BSpline refactor, the MMT/MMIRS dataset was failing because the flat-fielding was bad.  I traced this back to the fit to the illumination profile.  The MMT/MMIRS slit fills the detector meaning that the slit coordinates are identical for all spectral rows.  This was causing the bspline fitter to mask nearly all of the breakpoints.  This PR boils down to one change:

The bspline spacing for fitting the illumination profile used to be set to:
```python
bkspace=np.fmax(1.0 / median_slit_width / 10.0, 1.2 * np.median(np.diff(spat_coo_data)))
```

This PR changes it to:
```python
bkspace=np.fmax(1.0 / median_slit_width, 1.2 * np.median(np.diff(spat_coo, axis=1))
```

This successfully avoids the MMT/MMIRS issue and has no material effect on the Shane/Kast reductions.  Now to test all the other spectrographs...

Here's are the fits before and after this change for Shane/Kast:
shane_before
shane_after

And for MMT/MMIRS:
mmt_mmirs_before
mmt_mmirs_after


kbwestfall:
Dev-suite running.

Note that most of the changes are associated with a doc update.  I also tried to clean up the `pypeit/core/flat.py` file a bit...

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  358 passed, 9403 warnings in 333.80s (0:05:33) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 2823 warnings in 877.19s (0:14:37) ---
--- PYTEST VET TESTS PASSED  77 passed, 179506 warnings in 4958.67s (1:22:38) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 298/298 TESTS  ---
Total disk usage: 370.725 GiB
Testing Started at 2026-06-26T21:29:59.906569
Testing Completed at 2026-06-27T11:50:17.359868
Total Time: 14:20:17.453299
```

pypeit/flatfield.py
```diff
@@ -1420,23 +1423,34 @@ def spatial_fit(self, norm_spec, spat_coo, median_slit_width, spat_gpm, gpm, deb
...
+#        bkspace = np.fmax(1.0 / median_slit_width / 10.0, 1.2 * np.median(np.diff(spat_coo_data))))
+        # UPDATE: For some datasets where a long-slit fills the detector, the
+        # spat_coo array is effectively identical for each row because the slit
+        # edges are exactly aligned with the pixel coordinates.  Choosing a
+        # sampling following the original approach above was causing the bspline
+        # fitting to mask nearly all of the breakpoints.  This is fixed by
+        # increasing the bspline spacing.
+        bkspace = np.fmax(1.0 / median_slit_width, 1.2 * np.median(np.diff(spat_coo, axis=1)))
+        spat_bspl = bspline.bspline(spat_coo_data, nord=4, bkspace=bkspace)
+
```
tbowers7, line 1442: I admit this feels rather counterintuitive to me given the examples provided in the PR description.  By _increasing_ the spacing, this allows MMIRS to fit?  What, specifically, was happening to the all of the (more closely spaced?) breakpoints in the old scheme?  For the Kast example, I can see the effect of growing the breakpoint spacing.

If the DevSuite run proves that this changes works fine for all the other spectrographs, however, I'll be satisfied.

tbowers7 (approved):

pypeit/flatfield.py
kbwestfall: First, just to be clear: For the Kast example, the slits have some curvature meaning that the spatial coordinates are different at each spectral coordinate.  For the MMIRS data, every spectral row has exactly the same set of spatial coordinates.

Second, the old approach was effectively placing a breakpoint at every 1/10th of a pixel.  We can get away with this for cases like Kast, but for MMIRS this was effectively putting in a lot of break points that were unconstrained by a separation in spatial coordinates.  I think of it like trying to fit a line to two points that have the same x coordinate; the slope is infinite so the solution is undefined.  The B-spline fitting algorithm has a catch for this where it finds breakpoints that yield undefined solutions and masks them.  This masking was basically causing all of the breakpoints to be removed for the MMIRS reduction.  The solution I chose was to reduce the number of breakpoints.  Another way around this issue would be add small random perturbations to all the spatial coordinates before performing the fit.  I preferred reducing the number of breakpoints, but we could explore the other option if we're worried about the sparser sampling of the illumination profile.

The dev-suite tests all pass, so I don't think this change has a strongly detrimental effect on the reduction, but it does mean that we're not as finely sampling the edges of the illumination profile.

tepickering: to paraphrase gordon gekko, "Curvature is Good" when it comes to bspline fitting. which is indeed counterintuitive at first, but it's how you get the sub-pixel sampling out of it. it might be worth spot-checking the sampling for some other uncurved spectrographs for sanity's sake. binospec comes to mind, but i think it has just enough wiggle to get by. MMIRS also has a pretty whack illumination profile...

tepickering (approved):
as i mentioned in my other comment, it's probably worth doing a systematic, quantitative look at the performance of the B-spline definitions and fitting across a range of configurations and more optimal ways to configure B-spline breakpoints. for now, though, this definitely sorts out the one pathological case and appears to work as well elsewhere. we may have been more breakpoints than we needed to previously...

bpholden (approved):
You answered Tim's question and mostly this is fine.
I do think it would be best if we could come up with a way to estimate the correct spacing based on the properties of the data.

PR #2152, opened 2026-06-12
B-spline refactor

kbwestfall:
Prompted by recent struggles trying to to understand/debug the code, Claude and I had some fun doing something I had wanted to do for a long time: refactor `bspline`.  This represents a few days of work that would have been weeks (if not months) of work otherwise.

I'm issuing this as a draft PR to start.  I have not integrated any of the new code into the main pypeit workflow, but I will before taking it out of draft.  Note this currently includes the changes made in #2151 (and what prompted that PR to begin with).

The best way to understand the changes is to look at the `bspline.rst` description of how the new classes work.  The file also contains migration instructions.  Note that I haven't worked the `DataContainer` wrapper classes into the documentation yet.

There will be an associated PR to the dev-suite with some reference documentation.  Of interest here is the [performance comparison](https://github.com/pypeit/PypeIt-development-suite/blob/bspline_refactor/bspline_refactor/performance_results.md).  In addition to making the code more pythonic, this also has pretty substantial performance gains.  I'll get a better understanding of how much this translates to real applications of pypeit through next week.

PR #2157, opened 2026-06-29
B-spline Refactor with integration

### LINKED ISSUE #2033: Add `try`/`except` around calls to `pypeit.core.fittting.iterfit()`
In the fitting function `iterfit()`, a `ValueError` is raised if there are no valid data points for the fitting:

https://github.com/pypeit/PypeIt/blob/81057b02e550a923361b0d0caf97aa0ef3b269e8/pypeit/core/fitting.py#L952-L955

This should be caught by the calling functions and should yield a context-specific error message that will help the user diagnose **_why_** there were no valid points in that particular fit.  As the code stands, the above lines result in a semi-useless traceback.  For consideration, this function also raises a `ValueError` if various inputs do not match length as they should -- which is a coding error rather than a problem with the input data.  I suggest the above lines raise a different type of error that can be caught by the calling functions (perhaps a core-specific custom `Exception`).

-----

As of the creation of this issue, `pypeit.core.fittting.iterfit()` is called 5 times from 2 different modules:

https://github.com/pypeit/PypeIt/blob/81057b02e550a923361b0d0caf97aa0ef3b269e8/pypeit/core/extract.py#L900-L905

https://github.com/pypeit/PypeIt/blob/81057b02e550a923361b0d0caf97aa0ef3b269e8/pypeit/core/extract.py#L1055-L1056

https://github.com/pypeit/PypeIt/blob/81057b02e550a923361b0d0caf97aa0ef3b269e8/pypeit/core/flux_calib.py#L1023-L1029

kbwestfall:
Introduces new classes that replace the old `bspline` class and its primary use in `bspline_profile`.  Here's a quick summary of the components of the PR:

- `pypeit.core.bspline.bspline.bspline` is deprecated and replaced by `pypeit.core.bspine.BSpline` and `pypeit.core.bspine.BSpline2D`.  These new classes are more performant and more pythonic, removing all the matrix gymnastics used to maintain the column-major (Fortran) memory ordering.
- The two new B-spline classes are basically fully independent of pypeit.  There are two subclasses -- `pypeit.containers.bspline.BSplineContainer` and `pypeit.containers.bspline.BSpline2DContainer` -- that handle I/O.
- The functions `pypeit.fitting.iterfit` (which was rarely used) and `pypeit.fitting.bspline_profile` are replaced by `pypeit.fitting.iterative_bspline_fit`.  The function signature handles all the previous use cases more cleanly, including allowing `gpm`, `ivar`, and `profile_basis` to be optional inputs, such that `iterative_bspline_fit` 

Notes:

- This PR supersedes #2152 .  This branch integrates the refactor into the code and deprecates all the old code.
- This branch already includes the changes in #2155 .
- The dev-suite passes except for some vet-test failures I'm looking into.  I will leave this PR in draft mode until those are fixed.
- I intend to update the doc-strings so that they are not constantly referencing the old code; i.e., these need to be independently understandable.  I'll also improve the doc page that helps transitioning from the old code.


kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  528 passed, 9413 warnings in 208.26s (0:03:28) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 2802 warnings in 783.55s (0:13:03) ---
--- PYTEST VET TESTS PASSED  77 passed, 180205 warnings in 4447.24s (1:14:07) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 304/304 TESTS  ---
Total disk usage: 361.699 GiB
Testing Started at 2026-07-06T16:36:55.620290
Testing Completed at 2026-07-07T02:34:36.935353
Total Time: 9:57:41.315063
```

pypeit/core/skysub.py
```diff
@@ -341,57 +334,45 @@ def skyoptimal(piximg, data, ivar, oprof, sigrej=3.0, npoly=1, spatial_img=None,
...
+
+    # Get the sky-only model
+    sky_bmodel, _ = sset.value(
+        piximg, basis=profile_basis[:, nobj:], coeff=sset.coeff[:, nobj:]
+    )
+
+    # Get the source-only model
+    obj_bmodel, _ = sset.value(
+        piximg, basis=profile_basis[:, :nobj], coeff=sset.coeff[:, :nobj]
+    )
```
tbowers7: Okay, so the sky-only portion is stacked after all of the objects in the `profile_basis` and `sset.coeff`.  It took some mental gymnastics to see that.

pypeit/core/skysub.py
```diff
@@ -440,8 +421,7 @@ def optimal_bkpts(bkpts_optimal, bsp_min, piximg, sampmask, samp_frac=0.80,
     pix = pix[isrt]
     piximg_min = pix.min()
     piximg_max = pix.max()
-    bset0 = bspline.bspline(pix, nord=4, bkspace=bsp_min)
-    fullbkpt_grid = bset0.breakpoints
+    fullbkpt_grid = Knots(spacing=bsp_min, x=pix, nord=4).breakpoints
```
tbowers7, line 422: I like the `Knots` object here!

pypeit/core/fitting.py
```diff
@@ -1536,3 +1132,197 @@ def bspline_qa(xdata, ydata, sset, gpm, yfit, xlabel=None, ylabel=None, title=No
...
+        bspl_cls = bspline.BSpline2D
+
+    outmask = np.ones(y.shape, dtype=bool)
+    maskwork = outmask.copy()
+    if gpm is not None:
+        maskwork &= gpm
+    if ivar is not None:
+        maskwork &= ivar > 0
+    if not maskwork.any():
+        return None, outmask, np.zeros(y.shape), 0., 4
```
tbowers7, line 1265: It looks like this change from raising an exception to returning a sensible set of "zeros" fixes #2033 .

pypeit/core/spatialprofile.py
```diff
@@ -391,12 +391,18 @@ def fit_profile(image, ivar, waveimg, thismask, spat_img, trace_in, wave,
...
-    c_answer, cmask   = fitting.iterfit(wave[indsp], flux_sm[indsp], invvar = fluxivar_sm[indsp]*bmask2,
-                                     kwargs_bspline={'everyn': 30}, kwargs_reject={'groupbadpix':True,'maxrej':1})
+    b_answer, bmask, *_ = iterative_bspline_fit(
+        wave[indsp], flux_sm[indsp], ivar=fluxivar_sm[indsp],
+        kwargs_knots={'stride': 1.5}, kwargs_reject={'groupbadpix': True, 'maxrej': 1}
+    )
+    b_answer, bmask2, *_ = iterative_bspline_fit(
+        wave[indsp], flux_sm[indsp], ivar=fluxivar_sm[indsp]*bmask,
+        kwargs_knots={'stride': 1.5}, kwargs_reject={'groupbadpix': True, 'maxrej': 1}
+    )
```
tbowers7, line 401: Out of curiosity, why have both calls return `b_answer` when the second one overwrites the first?

It would make for easier reading to have the first call return `_, bmask, *_`.

And more of a semantic issue than code, why do you call the first return value here `answer` whereas elsewhere it's called `set`?  I'm sure it's historical, but it may be worth enforcing a common output language for the new `iterative_bspline_fit()`.

pypeit/core/spatialprofile.py
```diff
@@ -623,9 +631,10 @@ def fit_profile(image, ivar, waveimg, thismask, spat_img, trace_in, wave,
...
-        mode_shift_out = fitting.bspline_profile(xtemp.flat[inside], norm_obj.flat[inside],
-                                               norm_ivar.flat[inside], profile_basis,
-                                               maxiter=1, kwargs_bspline={'nbkpts':nbkpts})
+        mode_shift_out = iterative_bspline_fit(
+            xtemp.flat[inside], norm_obj.flat[inside], ivar=norm_ivar.flat[inside],
+            basis=profile_basis, maxiter=1, kwargs_knots={'count': nbkpts}
+        )
         # Check to see if the mode fit failed, if so punt and return a Gaussian
         if not np.any(mode_shift_out[1]):
             log.info('B-spline fit to trace correction failed for fit to ninside = {:}'.format(ninside) + ' pixels')
```
tbowers7, line 640: Why keep everything in the `mode_shift_out` tuple here, rather than breaking things out explicitely?
```
mode_shift_set, mode_shift_mask, *_ = ...

if not np.any(mode_shift_mask):
   ...
```

pypeit/core/spatialprofile.py
```diff
@@ -636,20 +645,22 @@ def fit_profile(image, ivar, waveimg, thismask, spat_img, trace_in, wave,
...
 
         profile_basis = np.column_stack((mode_zero,mode_stretch))
-        mode_stretch_out = fitting.bspline_profile(xtemp.flat[inside], norm_obj.flat[inside],
-                                                 norm_ivar.flat[inside], profile_basis, maxiter=1,
-                                                 fullbkpt=mode_shift_set.breakpoints)
+        mode_stretch_out = iterative_bspline_fit(
+            xtemp.flat[inside], norm_obj.flat[inside], ivar=norm_ivar.flat[inside],
+            basis=profile_basis, maxiter=1,
+            kwargs_knots={'full': mode_shift_set.breakpoints}
+        )
```
tbowers7, line 663: Same comment here regarding the tuple `mode_stretch_out` and simply breaking it apart in the return unpacking.

pypeit/core/spatialprofile.py
```diff
@@ -684,17 +695,24 @@ def fit_profile(image, ivar, waveimg, thismask, spat_img, trace_in, wave,
...
                                                                   show_profile, ind=good, l_limit=l_limit,r_limit=r_limit, xlim=7.0)
                 return (profile_model, trace_in, fwhmfit, med_sn2)
 
-            bset = bset_out[0] # This updated bset used for the next set of trace corrections
+            # bset is BSpline2D (npoly=1, funcname=None); extract 1D view for
+            # value() calls in the next iteration at arbitrary sigma_x points
+            _bset2d = bset_out[0]
+            bset = BSpline(knots=Knots(full=_bset2d.breakpoints), nord=_bset2d.nord)
+            bset.bkpt_gpm = _bset2d.bkpt_gpm.copy()
+            bset.coeff = _bset2d.coeff[:, 0]
```
tbowers7, line 715: Same comment about tuple unpacking.

pypeit/core/spatialprofile.py
```diff
@@ -710,10 +728,17 @@ def fit_profile(image, ivar, waveimg, thismask, spat_img, trace_in, wave,
...
+        ivar=norm_ivar.flat[ss[inside]], basis=pb.flat[ss[inside]], nord=4,
+        kwargs_knots={'interior': bkpt}, upper=10, lower=10
+    )
+    # Convert BSpline2D (npoly=1, funcname=None) to 1D BSpline so value() works
+    # at arbitrary evaluation points throughout the rest of this function
+    _bset2d = bset_out[0]
+    bset = BSpline(knots=Knots(full=_bset2d.breakpoints), nord=_bset2d.nord)
+    bset.bkpt_gpm = _bset2d.bkpt_gpm.copy()
+    bset.coeff = _bset2d.coeff[:, 0]
     outmask = bset_out[1]
```
tbowers7, line 742:
```
_bset2d, outmask, *_ = ...
```

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+class Knots:
+    """
+    Construct and store breakpoints for B-spline fitting.
+
+    The breakpoint knot vector is built lazily when :meth:`build` is called and
+    then cached in :attr:`breakpoints`.
+
+    When building the knots for any given fit, the parameters of this class are
+    with respect to to the independent coordinate (``x``) and the order
+    (``nord``).  Note that ``x`` and ``nord`` to not need to be provided at
```
tbowers7: "**do** not need to be provided"

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+            _interior = np.array([sx, ex]) if np.sum(w) < 2 else _interior[w]
+
+        if _interior.size < 2:
+            _interior = np.array([sx, ex])
+        if _interior[0] > sx:
+            _interior[0] = sx
+        if _interior[-1] < ex:
+            _interior[-1] = ex
+
+        self._breakpoints = Knots._pad(_interior, nord, self.spread).astype(float)
```
tbowers7: Style question: as a `@staticmethod`, `_pad()` could be called with either `self._pad()` or `Knots._pad()`.  Any particular reason for the preference of the latter over the former?  For instance, when I was reading this, it took me a minute to realize I was already in `Knots`, so `self` would have been more intuitive for me -- plus I don't like calling "private" methods from outside a class, so `Knots._pad()` raised my hackles.

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+            )
+
+        # Objects used during fitting, which hold the results and cache useful
+        # data that can be used across multiple fits
+        self.bkpt_gpm = None
+        self.coeff = None
+        self.icoeff = None
+        self.x = None
+        self.yfit = None
+        self._cached_design = None
```
tbowers7, line 413: This is a nice addition to cache these values across cyclic calls to the fitting.

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+        """
+        n = breakpoints.size - nord
+        return np.clip(np.searchsorted(breakpoints, x, side='right') - 1, nord - 1, n - 1)
+        # indx = np.zeros(x.size, dtype=int)
+        # ileft = nord - 1
+        # for i in range(x.size):
+        #     while x[i] > breakpoints[ileft + 1] and ileft < n - 1:
+        #         ileft += 1
+        #     indx[i] = ileft
+        # return indx
```
tbowers7: Vestigial code?

pypeit/core/spatialprofile.py
```diff
@@ -636,20 +645,22 @@ def fit_profile(image, ivar, waveimg, thismask, spat_img, trace_in, wave,
...
         mode_shift_set = mode_shift_out[0]
-        temp_set = bspline.bspline(None, fullbkpt=mode_shift_set.breakpoints,
-                                   nord=mode_shift_set.nord)
-        temp_set.coeff = mode_shift_set.coeff[0, :]
+        temp_set = BSpline(knots=Knots(full=mode_shift_set.breakpoints), nord=mode_shift_set.nord)
+        temp_set.bkpt_gpm = mode_shift_set.bkpt_gpm
+        temp_set.coeff = mode_shift_set.coeff[:, 0]
         h0, _ = temp_set.value(xx)
-        temp_set.coeff = mode_shift_set.coeff[1, :]
+        temp_set.coeff = mode_shift_set.coeff[:, 1]
```
tbowers7, line 652: Note to self: The axes swap for `coeff` is a deliberate design choice in the new `Bspline` class.

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+
+        nx = x.size
+        n = nbkpt - self.nord
+        lower = np.zeros(n - self.nord + 1, dtype=int)
+        upper = np.zeros(n - self.nord + 1, dtype=int) - 1
+
+        # NOTE: _find_spans is used by both BSpline and BSpline2D, and there's
+        # no need for the latter to override this base class method.  However,
+        # just to somewhat future-proof the code, I specify type(self) here,
+        # instead of using `BSpline._find_spans`.
```
tbowers7, line 601: This follows on the comment about how to address `@staticmethod`s.  Could this not just be `self._find_spans()`?

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+        Parameters
+        ----------
+        bad_cols : :class:`numpy.ndarray`
+            Column indices flagged by :meth:`_solve_banded`.
+
+        Returns
+        -------
+        int
+            Value is -1 if masking succeeded (caller should retry :meth:`fit`),
+            or -2 if too few breakpoints remain (caller should abort).
```
tbowers7, line 811: I know these are all internal computation algorithms, so as long as these are all internally documented and conformed to it's fine.  I can't help thinking, though, that it would be easier for long-term maintenance to have these as `enum` values or something like C's `#define` variables.

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+        reset_knots : bool, optional
+            Regardless of any existing breakpoints, reset the breakpoints using
+            :attr:`knots` and ``x``.
+
+        Returns
+        -------
+        err : int
+            Set to 0 on success; -1 if breakpoints were masked (Cholesky was
+            degenerate — caller should retry); and -2 on failure (too few active
+            breakpoints).
```
tbowers7, line 930: For the Public API, I feel even stronger about having named `int` or `enum` values for return codes.

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+            Fitted B-spline evaluated at ``x``.
+        """
+        # Breakpoints must be established before reset_coeff (which reads
+        # breakpoints.size to allocate the coefficient arrays).
+        if reset_knots or self.breakpoints is None:
+            self.reset_knots(x, required=True)
+
+        # Build or retrieve the cached design matrix.  Must follow reset_knots
+        # so the cache-empty state is visible when determining whether to rebuild.
+        if self._cached_design is None or x is not self.x:
```
tbowers7, line 941: Note to self: `x is not self.x` checks object identity in memory not array values.

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+        self.yfit = self._evaluate_model(A, lower, upper)
+        if bad_cols[0] != -1:
+            return self._mask_breakpoints(bad_cols), self.yfit
+
+        goodbk_idx = goodbk.nonzero()[0]
+        self._update_coefficients(sol, chol, goodbk_idx)
+        self.yfit = self._evaluate_model(A, lower, upper)
+        return 0, self.yfit
+
+    def _fit_gpm(self, x):
```
tbowers7: Maybe move this internal function out of the "Public API" section of the class?

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1867 @@
...
+
+        _P = self.P if P is None else P
+        _npoly = self.npoly if npoly is None else npoly
+
+        nx = x.size
+        n = nbkpt - self.nord
+        lower = np.zeros(n - self.nord + 1, dtype=int)
+        upper = np.zeros(n - self.nord + 1, dtype=int) - 1
+
+        indx = type(self)._find_spans(x, self.breakpoints[self.bkpt_gpm], self.nord)
```
tbowers7, line 1366: Similar `@staticmethod` question/comment.

pypeit/containers/bspline.py
```diff
@@ -0,0 +1,355 @@
...
+
+.. note::
+
+    The :attr:`~pypeit.core.bspline.BSpline.breakpoints` property defined on
+    :class:`~pypeit.core.bspline.BSpline` is a read-only data descriptor that
+    delegates to ``self.knots.breakpoints``.  Using ``breakpoints`` as a datamodel
+    key would create an irresolvable naming conflict (the property intercepts all
+    attribute-level reads, making the stored value invisible).  The breakpoint
+    array is therefore stored under the distinct key ``bkpt_full``, and the
+    property continues to serve all algorithmic code via ``self.knots``.
```
tbowers7, line 20: Can you address the pros and cons of this approach as compared to removing the delegated property, allowing the datamodel to use the `breakpoints` key, and accessing the breakpoints via `<object>.knots.breakpoints`?

pypeit/containers/bspline.py
```diff
@@ -0,0 +1,355 @@
...
+    def _validate(self):
+        if self.bkpt_full is not None and self.knots is None:
+            self.knots = Knots(full=self.bkpt_full, nord=self.nord)
+            if self.bkpt_gpm is None:
+                self.bkpt_gpm = np.ones(self.knots.breakpoints.size, dtype=bool)
+
+    def _bundle(self):
+        if self.knots is not None and self.knots.breakpoints is not None:
+            self.bkpt_full = self.knots.breakpoints
+        return super()._bundle(ext='BSPLINE')
```
tbowers7, line 191: Indicate that these are overrides of the `DataContainer` methods.

pypeit/core/fitting.py
```diff
@@ -1536,3 +1132,197 @@ def bspline_qa(xdata, ydata, sset, gpm, yfit, xlabel=None, ylabel=None, title=No
...
+    outmask : :class:`numpy.ndarray` of bool
+        Final good-pixel mask after all rejection iterations.
+    yfit : :class:`numpy.ndarray`
+        Best-fit model values at ``x``.
+    reduced_chi : float
+        Reduced chi-squared of the final fit.
+    exit_status : int
+        0 — converged normally; 1 — reached ``maxiter``; 2 — too few
+        good points; 3 — singular / degenerate fit; 4 — fewer good
+        points than ``nord`` on entry.
```
tbowers7: Continued comment about `enum` for `exit_status`.  See similar comments elsewhere for more details.

pypeit/core/fitting.py
```diff
@@ -1536,3 +1132,197 @@ def bspline_qa(xdata, ydata, sset, gpm, yfit, xlabel=None, ylabel=None, title=No
...
+    err = -1
+    qdone = False
+    iiter = 0
+    exit_status = 0
+    reduced_chi = 0.
+    relative_factor = 1.0
+    nrel = 0 if relative is None else len(relative)
+    _gpm = None if gpm is None else gpm.copy()
+
+    termwidth = 80 - 13
```
tbowers7, line 1286: Ought `termwidth` be hardwired, or should there be some way to adjust based on... anything?

pypeit/core/fitting.py
```diff
@@ -1536,3 +1132,197 @@ def bspline_qa(xdata, ydata, sset, gpm, yfit, xlabel=None, ylabel=None, title=No
...
+        # number of iterations it is allowed to perform while masking bad
+        # breakpoints.
+        iiter += 1
+
+        if err == -2:
+            log.warning('B-spline fit failed: All break points lost!')
+            return bspl, np.zeros(x.shape, dtype=bool), np.zeros(x.shape), reduced_chi, 3
+        if err != 0:
+            log.info(f' {iiter:4d}{nullval}'.center(termwidth))
+            continue
```
tbowers7, line 1326: Having `err` as something semantic would help with readability.

doc/bspline.rst
```diff
@@ -0,0 +1,376 @@
...
+Because the column grouping preserves the banded structure of
+:math:`\mathbf{A}` in the knot index, the normal equations remain banded with
+half-bandwidth :math:`np` and are again solved via
+:func:`scipy.linalg.cholesky_banded` and
+:func:`scipy.linalg.cho_solve_banded`.
+
+Iterative Sigma-Clipping
+------------------------
+
+:func:`~pypeit.bspline.refactor.bspline_profile_refactor` wraps either class
```
tbowers7: I presume a forthcoming doc update will align these references with the locations of classes in this PR.

pypeit/tests/test_bspline.py
```diff
@@ -1,205 +1,1915 @@
...
+# 
+#     new = BSpline(knots=old.breakpoints.copy(), nord=4)
+#     new.bkpt_gpm[:] = old.mask
+#     new._cached_design = None
+#     new.fit(x, y, ivar=invvar)
+# 
+#     x_eval = np.sort(rng.uniform(0, 10, 50))
+#     old_val, _ = old.value(x_eval)
+#     new_val, _ = new.value(x_eval)
+#     np.testing.assert_allclose(new_val, old_val, rtol=1e-10)
```
tbowers7: It is helpful to see this in the PR, but maybe remove before merging?

tbowers7 (approved):
This PR provides a solid refactor of the basis spline code underlying much of what PypeIt does.

I did not do a careful check of the mathematical refactoring -- fairly or unfairly trusting the combination of Claude and the set of cross-check tests in `test_bspline.py`.

This biggest issue I noticed is the need for a doc sweep in `bspline.rst` to align function name and location references with the as-built versions in this PR.

Most of my comments are fairly minor, so I am approving now.  The one issue I would like a response on is having a semantic set of return codes from the various fitting routines for readability and maintainability.

doc/bspline.rst
kbwestfall: Yep, sorry.  I've now updated these old references.

pypeit/containers/bspline.py
kbwestfall: Personally, I think we want the delegated property because it makes the code more clear.  But, another alternative would be to rename the delegated property instead of renaming the datamodel key.  Regardless of this choice, the new code requires a change to the bspline datamodel, but I'm open to changing the name of the delegated property so that we maintain `breakpoints` as the datamodel key.  What do you think?

pypeit/containers/bspline.py
kbwestfall: Done!  Full docstrings added.

pypeit/core/bspline.py
kbwestfall: Good catch!  Fixed.

pypeit/core/bspline.py
kbwestfall: Fair question.  Personally, I've typically preferred this form because I immediately read it as a static or class method that does not need or alter `self`.  But I see where you're coming from.  In a bit of Googling about the difference, I decided I should at least change this to `type(self)` (another pattern that note later).  Not that we would ever subclass from `Knots`, but this guards against that to ensure it's always calling the subclass function.  I'm not necessarily against just using `self` (I'm almost positive I do in other places in the code), but I'd rather use `type(self)`, unless you have a strong objection.

pypeit/core/bspline.py
kbwestfall: A Claude suggestion!

pypeit/core/bspline.py
kbwestfall: Yep, I didn't want to remove Claude's code until I was sure my change worked.  Removed.

pypeit/core/bspline.py
kbwestfall: See previous response.

pypeit/core/bspline.py
kbwestfall: I'm not opposed to this change.  You'll see in one of the new edits to `bspline.rst` that a check for `err==-1` is added, meaning that `fit` needs to be called again because a breakpoint was removed.  A more symmantic approach would make that easier to read.  But, at least for now, I'd like keep the historical precedent and punt on this, but I'm happy to revisit.

pypeit/core/bspline.py
kbwestfall: See previous response.

pypeit/core/bspline.py
kbwestfall: Good call.  I moved things around in both `BSpline` and `BSpline2D`.

pypeit/core/bspline.py
kbwestfall: See previous response.

pypeit/core/fitting.py
kbwestfall: See previous response.

pypeit/core/fitting.py
kbwestfall: Nice!  Let's remember to close that when this is merged!

pypeit/core/fitting.py
kbwestfall: We could definitely get fancier with this by using something that automatically determines the terminal width.  But I didn't think it was worth it.

pypeit/core/fitting.py
kbwestfall: See previous response.

pypeit/core/skysub.py
kbwestfall: Good attention to detail!  I added a comment that explains this.

pypeit/core/spatialprofile.py
kbwestfall: Not necessarily on the names (i.e., answer vs. set), but I come back to this in a follow-up PR.  Bottom line is that I keep the returned model (instead of reconstructing it).  At the moment, I don't remember if I keep `b_answer` and `c_answer` or not.

pypeit/core/spatialprofile.py
kbwestfall: I change this in an upcoming PR.  See previous response.

pypeit/core/spatialprofile.py
kbwestfall: Yeah, this has to do with the change from FORTRAN-native memory order to C-native memory order.

pypeit/core/spatialprofile.py
kbwestfall: Same response :)

pypeit/core/spatialprofile.py
kbwestfall: Same response :)

pypeit/core/spatialprofile.py
kbwestfall: Same response.

pypeit/tests/test_bspline.py
kbwestfall: Good catch.  Yes, I'll remove them before merging (or at least I'll try to remember to so).

kbwestfall:
Thanks for the detailed review, @tbowers7 .  I've address most of them, but left some open questions to you about the use of self for staticmethods and about whether or not to change the `breakpoints` element in the bspline datacontainer.

pypeit/core/bspline.py
tbowers7: Yes, I can see the usefulness of indicating a static call at the call site.  The `type(self)` addresses both the "I'm calling a function in this class" needs of `self` and the "I'm calling a static or class method" visual cue.  I hereby withdraw my other objections to `type(self)`.

pypeit/core/spatialprofile.py
tbowers7: I put this note in to remind myself not to complain about it.

pypeit/core/bspline.py
tbowers7: Punting is fine, but I wanted to at least bring it up while this change is fresh.

pypeit/core/fitting.py
tepickering: it might auto-close from that comment. or add a new stand-alone comment to specifically do so.

pypeit/containers/bspline.py
tbowers7: I guess I'm of two minds about this one.  The biggest question is whether the Container or underlying class will be used more.  Ideally, both would have access to the same attributes, since that's how they're built.  I'm thinking mostly of the "principle of least surprise".  Is there any way to have our cake and eat it, too?  (Meaning, allowing the datamodel use `breakpoints`, but also have it as an accessible property in `BSpline` and children?)

pypeit/core/spatialprofile.py
tepickering: i, for one, am thankful for finally switching this to python/C-native memory order...

tbowers7:
> Thanks for the detailed review, @tbowers7 . I've address most of them, but left some open questions to you about the use of self for staticmethods and about whether or not to change the `breakpoints` element in the bspline datacontainer.

Responded inline about both.  Basically, I'm happy with the `type(self)` approach for the former, and I want to have my cake and eat it, too, for the latter.

pypeit/core/bspline.py
```diff
@@ -0,0 +1,1832 @@
...
+----------
+Algorithms adapted from IDL/PYDL bspline utilities:
+  https://doi.org/10.5281/zenodo.1095150
+"""
+
+import warnings
+
+from IPython import embed
+import numpy as np
+from scipy.linalg import cho_solve_banded, get_lapack_funcs
```
tepickering, line 43: using these routines for linear algebra opens up optimization by swapping `openblas` for hardware-accelerated libraries like `Accelerate` and intel's `MKL`.

pypeit/core/pydl.py
```diff
@@ -175,117 +175,117 @@ def djs_maskinterp(yval, mask, xval=None, axis=None, const=False):
...
-            # uu,ww,vv = np.linalg.svd(alpha, full_matrices=False)
-            res[nonfix] = np.linalg.solve(alpha, beta)
-        else:
-            # res[nonfix] = (ysub * (invvar > 0) * finalarr).sum()/alpha
-            res[nonfix] = (ysub * invvar * finalarr).sum()/alpha
-        if len(fixed) > 0:
-            res[fixed] = inputans[fixed]
-        yfit = np.dot(legarr.T, res[0:ncfit])
-    return res, yfit
+# def func_fit(x, y, ncoeff, invvar=None, function_name='legendre', ia=None,
```
tepickering: should probably nuke this now if it's deprecated. the code lives on in the git history if it needs to be referred to.

pypeit/core/pydl.py
```diff
@@ -1275,110 +1275,110 @@ def spheregroup(ra, dec, linklength, chunksize=None):
...
-                match1[nmatch] = omatch1[s[i]]
-                match2[nmatch] = omatch2[s[i]]
-                distance12[nmatch] = odistance12[s[i]]
-                nmatch += 1
-    else:
-        match1 = omatch1[s]
-        match2 = omatch2[s]
-        distance12 = odistance12[s]
-    return (match1, match2, distance12)
+# def spherematch(ra1, dec1, ra2, dec2, matchlength, chunksize=None,
```
tepickering: remove vs comment out...

pypeit/tests/test_bspline.py
```diff
@@ -1,205 +1,2004 @@
...
+    cp = spl.copy()
+    orig_val = cp.coeff[0, 0]
+    spl.coeff[0, 0] += 999.0
+    assert cp.coeff[0, 0] == orig_val, 'mutating source coeff must not affect the 2D copy'
+
+
+# ============================================================================
+# Cross-check tests: new vs. original implementation
+# ============================================================================
+
```
tepickering: if the commented out tests are no longer runnable, i would remove them. however, it also doesn't look like they live in git history because this is a new file.

tepickering (approved):
this is really, really great! the new documentation is a big help as well. makes the bspline machinery much less of a black box than it was before. i'm curious to see if or how much real-world performance is improved, especially with hardware-accelerated BLAS/LAPACK.

bpholden (approved):

pypeit/containers/bspline.py
kbwestfall: Unfortunately, the components of the `DataContainer` datamodel must be attributes of the class that exist in its `__dict__`, and python class `property`s aren't contained in that dictionary.  I.e., we can't have our cake and eat it, too.

The only place the `BSplineContainer` objects are currently used is in `flatfield.py`, and this is primarily as I/O infrastructure.  The sites that access/use the `BSpline` parent of the `Container` are only using it to sample the the model (as far as I can tell).  So I expect the underlying class will be (and is currently) used more in the code.  I doubt users were actually reading in the breakpoints from the flat-field output file, meaning I expect that users will not really notice the difference.

I think we should leave this change as is.  Thoughts?

pypeit/core/bspline.py
kbwestfall: Added issue https://github.com/pypeit/PypeIt/issues/2170 .

kbwestfall:
I re-ran tests just to be sure and I'm getting a vet test failure:
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  574 passed, 9432 warnings in 197.35s (0:03:17) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 2802 warnings in 772.93s (0:12:52) ---
--- PYTEST VET TESTS FAILED  1 failed, 80 passed, 289270 warnings in 4315.70s (1:11:55) ---
Total disk usage: 354.215 GiB
Testing Started at 2026-07-21T17:38:33.624195
Testing Completed at 2026-07-22T03:32:16.291958
Total Time: 9:53:42.667763
```

The failing vet test is
```
FAILED ../PypeIt-development-suite/vet_tests/test_binospec_ifu.py::test_binospec_ifu_sky_fiber_skymo
del - AssertionError: SKYMODEL has no positive sky counts
```
Please don't merge this until I figure it out.

kbwestfall:
Bug found.  Dev-suite restarted.

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  574 passed, 9432 warnings in 198.30s (0:03:18) ---
--- PYTEST UNIT TESTS PASSED  158 passed, 2802 warnings in 761.22s (0:12:41) ---
--- PYTEST VET TESTS PASSED  81 passed, 289195 warnings in 4309.56s (1:11:49) ---
Total disk usage: 349.599 GiB
Testing Started at 2026-07-23T14:51:40.948545
Testing Completed at 2026-07-24T00:42:06.181558
Total Time: 9:50:25.233013
```

PR #2160, opened 2026-07-01
Modify python version

profxj:
As titled

Just don't want folks to think they should be using `3.11`

### DIFF
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -37,7 +37,7 @@ You can set up a new python environment using either `conda`_:
 
 .. code-block:: console
 
-    conda create -n pypeit python=3.11
+    conda create -n pypeit python=3.13
     conda activate pypeit
 
 or `venv`_:


rcooke-ast (approved):
Agree with this 👍

doc/installing.rst
```diff
@@ -37,7 +37,7 @@ You can set up a new python environment using either `conda`_:
 
 .. code-block:: console
 
-    conda create -n pypeit python=3.11
+    conda create -n pypeit python=3.13
```
tbowers7, line 40: Should think about asking Claude to write a script to always insert the upper-end of supported Python into this doc?  Or, we just need to keep on top of it.

tbowers7 (approved):

kbwestfall:
Maybe we should just remove the python version.  I think you can just run `conda create -n pypeit`.  It's also worth noting that the python dependencies are listed in a table lower down in the page.

bpholden:
Do we still support conda?
I am not be facetious here, I am asking if anyone has tested the conda
instructions in the past few years?

I would actually move that we delete entirely and just use pip.

profxj:
this is a hotfix; do a real fix later :)

PR #2166, opened 2026-07-20
Implement support for up-the-ramp fitting, dithering, MOS, and slit-mask definitions for MMIRS

tepickering:
This PR overhauls MMIRS support in `mmt_mmirs`, adding up-the-ramp fitting, automatic A–B (nod) background subtraction, and ingestion of MMIRS slit-mask design (`.msk`) files so multi-object (MOS) reductions carry real target identities through to per-target coadds. It also completes automatic wavelength calibration for every observatory-recommended grism/filter combination. The ramp-fitting work builds on https://github.com/zhechenghu/mmt-mmirs-up-the-ramp-pypeit and the references therein.

## Ramp fitting

- Up-the-ramp fitting with likelihood-based jump (cosmic-ray) detection uses the algorithm of [Brandt (2024)](https://arxiv.org/abs/2404.01326), replacing the correlated double sampling (first minus last read) used previously. Read noise is significantly reduced and cosmic rays are much more effectively rejected. The upstream MIT-licensed module is vendored under `pypeit/ext/fitramp/`, kept **byte-identical** to [t-brandt/fitramp](https://github.com/t-brandt/fitramp) so upstream fixes can be diffed in cleanly; the directory carries its `LICENSE` and a `README` documenting the provenance and the do-not-refactor policy. Frames with at least 5 reads are ramp-fit; frames with fewer fall back to correlated double sampling (fewer than 5 reads do not sample the ramp well enough to fit reliably). Each read is reference-pixel corrected before fitting.

### Effective read-noise calibration

The per-read noise the fit needs is not the instantaneous detector read noise but an **effective** noise: read noise plus the shot noise from dark current (and any flux) that accumulates over the ramp. It therefore *grows with exposure time* — measured on MMIRS darks, roughly 5.7 e⁻ for a 3-read (3 s) ramp, 8.3 e⁻ for 8 reads (10 s), and 9.4 e⁻ for 69 reads (300 s) — so it must be measured at the science exposure time.

- PypeIt calibrates it from the dark frames listed in the pypeit file. Only darks whose exposure time **matches the science frame** and that have at least 10 reads are used; each is calibrated independently by rescaling the ramp-fit chi-squared, and the results are combined as an **inverse-variance weighted mean** (each dark weighted by the bootstrap uncertainty on its own calibrated noise). With no matching dark, a frame with ≥10 reads self-calibrates from itself; a frame with fewer reads is too poorly constrained and instead uses a starting-guess effective noise of 9 e⁻.
- The calibrated value is floored at the instantaneous read noise from the header `RDNOISE` card (a derived noise below the physical read noise is unphysical); `RDNOISE` is also used as the correlated-double-sampling read noise for short-ramp frames. The effective read noise of the fitted image, `sigma * sqrt(12 (N-1) / (N (N+1)))` for `N` reads, is propagated to the detector parameters.

## Preprocessed ramp images

The results of the ramp-fitting are stored in a `RampFit` directory within the reduction directory. When loading a raw data cube, `get_rawimage` will look there to see if it has been ramp-fitted. If not, it will process the cube and save the result in `RampFit` (re-fitting automatically if the raw cube's modification time changes). The `pypeit_fit_ramp` script can be used to preprocess data cubes prior to reduction, which has the advantage of providing 2D raw images that are easier to inspect and preview:

```bash
pypeit_fit_ramp mmt_mmirs raw/*.fits
```

The script is deliberately generic rather than instrument-specific: it takes the spectrograph as its first argument and guards on an allow-list of ramp-capable spectrographs (currently just `mmt_mmirs`), so enabling a future instrument is a one-line change instead of a new script.

## Parallel fitting and performance tuning

The per-pixel ramp fit is independent across the detector, so it is performed in parallel: rows are split into small blocks fit concurrently in a thread pool (NumPy releases the GIL during the element-wise arithmetic that dominates the fit). Output is numerically identical to the serial row-by-row fit. On a 69-read 2048×2048 cube this is ~3× faster (≈99 s → ≈32 s) with the default of `min(6, os.cpu_count())` threads. The vendored `fitramp.py` is untouched and no process pool is used — the fit is memory-bandwidth bound, so threads already reach that ceiling.

Two new `[rdx]` parameters let users tune this for their hardware:

```ini
[rdx]
    spectrograph = mmt_mmirs
    ramp_fit_cores = 12        # worker threads (None -> min(6, os.cpu_count()); 1 disables)
    ramp_fit_chunk_rows = 16   # detector rows per fit call (expert knob)
```

Because the fit is memory-bandwidth bound, the speedup plateaus at ~6 cores on a typical machine and raising `ramp_fit_cores` pays off mainly on workstations with more memory bandwidth. Peak memory scales roughly as `ramp_fit_cores × ramp_fit_chunk_rows`.

## A–B nod background subtraction

MMIRS science frames are commonly taken as an A–B nod sequence (dithering the target along the slit between exposures), for both long-slit and MOS masks. The MMIRS header has no dither keyword, so PypeIt derives the along-slit dither offset of each frame from the telescope pointing relative to the catalog target and records it in the pypeit file as `dithoff`, along with `dithpos` (A/B/A'/B' labels), `dithpat` (e.g. `ABA'B'`), and `frameno`.

During `pypeit_setup`, PypeIt automatically fills the `comb_id` and `bkg_id` columns so that each science frame is background-subtracted using the temporally-adjacent frame at the opposite nod position. Each frame keeps a unique `comb_id` (frames are not 2D-coadded across the small sub-dither), so PypeIt produces one `spec2d`/`spec1d` per nod pair. A sequence whose dither offsets span less than ~1 arcsec is treated as a stare and left unpaired (`bkg_id = -1`). The columns can be edited by hand afterward like any other A–B differencing setup.

## Slit-mask design (MOS)

For multi-object masks, PypeIt can now ingest the MMIRS `.msk` mask-design file (the xfitmask output delivered with the raw data) so that every slit, `SpecObj`, and coadd carries the real target identity. During `pypeit_setup`, if a file named `<MOSID>.msk` (matching the `APERTURE` header keyword and the `decker` metadata column) is found in the raw-data directory, PypeIt automatically:

- enables slitmask-design tracing (`[calibrations][slitedges] use_maskdesign = True`, with `maskdesign_filename` pointing at the `.msk`);
- stamps `MASKDEF_ID`, `RA`, `DEC`, and `MASKDEF_OBJNAME` (the catalog target name) on each extracted object;
- flags the alignment-box (`BOX`) slits as alignment rather than science and uses them to register the mask-to-detector offset; and
- force-extracts designed targets that were not auto-detected, at their predicted positions.

The mask `y` coordinate (mm) maps linearly, and anti-aligned, to the detector spatial pixel (`spat = C − scale·y`, with `scale = (1/arc2mm)/platescale/bin_spat` px/mm); the mask `x` coordinate is the dispersion axis and does not enter the spatial edge prediction. The absolute registration offset is fit by PypeIt's existing `EdgeTraceSet.maskdesign_matching` from the detected slits, so only the relative scale must be correct (validated at ~2.5 px RMS on `nep.as1`). When no `.msk` file is present, PypeIt falls back to generic edge tracing with `SPATxxxx-SLITyyyy` object names.

New parser module `pypeit/spectrographs/mmirs_maskfile.py` reads the `.msk`; the MMIRS spectrograph class gains `get_slitmask`, `get_maskdef_slitedges`, and `.msk` auto-discovery in `config_specific_par`.

## Coadding MOS spectra by target

A new `[collate1d] outfile_from` parameter controls how `pypeit_collate_1d` names its coadded output files: `coord` (the default, unchanged for all other spectrographs) names each file after the source sky coordinate or spatial pixel position, while `maskdef_objname` names it after the slit-mask-design object name (`MASKDEF_OBJNAME`), falling back to coordinate naming for sources with no mask-design name (e.g. serendips). **MMIRS defaults to `maskdef_objname`**, so grouping the per-nod-pair spectra across exposures with `pypeit_collate_1d` (ra/dec matching) yields one coadd per designed target, named after the real catalog target (e.g. `qso_172239.955+655201.69_MMIRS_<date>.fits`). The `SpecObj` name itself is left as the standard `SPATxxxx-SLITyyyy-DET`, matching the multislit convention used by Keck/DEIMOS and MOSFIRE; identity lives in `MASKDEF_OBJNAME`/`RA`/`DEC`, which are shown by `pypeit_show_1dspec --list` and in the `.txt` object summary next to each spec1d.

## Validation against the IDL pipeline

The full `nep.as1` MOS reduction (24-frame ABA'B' nod, zJ/J, mask `nep.as1.msk`) was cross-checked against the official IDL MMIRS pipeline reduction of the same data. 22 of the 23 PypeIt coadds match an IDL object by sky coordinate (the two pipelines each extracted a small number of objects the other did not).

- **Flux and wavelength.** On the bright continuum sources the smoothed spectra track feature-for-feature (Pearson *r* ≈ 0.78–0.87), including emission bumps and absorption troughs, with **sub-pixel wavelength agreement** (cross-correlation lag 0–2 Å, ≤ one 2.4 Å IDL bin). The flux differs only by a single constant multiplier (this reduction is in counts, `flux_value = False`; IDL uses its own units) — there is no wavelength-dependent throughput discrepancy.
- **Uncertainties.** The two pipelines deliver the same *real* noise — empirical S/N (continuum ÷ measured pixel-to-pixel scatter, in a common wavelength bin) agrees to ~5% (median PypeIt/IDL ≈ 1.05). But PypeIt's reported error array is realistic (formal S/N ≈ 1.4× the true scatter) whereas IDL's underestimates the noise by ~4.5×, so downstream error propagation (line fluxes, redshift confidence, χ²) is trustworthy here in a way it is not with the IDL errors.
- **Sky subtraction.** Measured directly in the 2D frames, the residual in blank-sky pixels is flat and sub-count at all wavelengths — **≈0.2% of the subtracted sky even at the red end** (13500–15000 Å), i.e. no red-end sky-subtraction failure. A faint apparent red-end excess relative to IDL is a wavelength-independent ~few-count pedestal (well below the per-pixel noise) that only becomes fractionally visible where a faint object's spectrum falls toward zero; IDL removes it with an additional local-sky subtraction that over-subtracts to ≤0.
- **Robustness.** Where a bright full-slit-width artifact (uncanceled sky line / persistence) starves the sky fit of clean pixels, the affected slit is flagged `BADSKYSUB` and that exposure is dropped from extraction; verified across all frames that such frames never reach the coadd (e.g. one bright target's coadd used exactly the 20 clean of 24 exposures).

## Wavelength coverage for all supported grism/filter combinations

Added `full_template` wavelength-solution templates and `config_specific_par` handling for the two remaining observatory-recommended grism/filter combinations, **H3000/H** and **HK/HK3**. PypeIt now auto-selects an archival solution for all supported combinations on the [MMTO recommendations page](https://www.mmto.org/mmirs-recommendations/) — `J`/`zJ`, `K3000`/`Kspec`, `H3000`/`H`, `HK`/`HK3` — plus the `HK`/`zJ` second-order setup.

Both new templates (`mmt_mmirs_H3000_H.fits`, `mmt_mmirs_HK_HK3.fits`) were built from real OH-airglow lines in science exposures and validated three ways:

- **Self-reidentification** of the source data (cc = 1.0).
- **Reidentification of an independent dataset**: H3000/H against a different science target and night (cc ≈ 0.997); HK/HK3 against standard star data with lower sky line S/N (cc ≈ 0.94–0.96), each reproducing the templates' central wavelengths, dispersions, and ranges.
- **Full end-to-end reductions** producing `spec2d`/`spec1d` with correct H-band (14434–18592 Å) and H+K-band (11418–25122 Å) wavelengths.

New dev-suite setups `H3000_H` and `HK_HK3` exercise both modes. These are also the first **post-2019** MMIRS data in the development suite, so they exercise deriving the dither offsets from the newer header format.

## Other fixes and notes

- `pypeit_setup` no longer crashes on MMIRS data missing expected FITS extensions; when `strict=False` skips an unreadable file, its row is also dropped from the user metadata so the metadata table stays aligned.
- The MMIRS docs note a display artifact in coadded 1D spectra: bins where no input exposure contributed a good pixel are correctly masked (`flux`/`ivar`/`mask` all zero) but their stored wavelength is left at 0, so a viewer that ignores the good-pixel mask (e.g. `lt_xspec`) draws a spurious spike to `wave = 0`. Loading the file through `specutils` (which honors the mask) avoids it; no coadd math is changed.

### Correctness fixes from code review

A branch code review surfaced several input-dependent bugs, now fixed and covered by unit tests:

- **`.msk` `offset`/`theta` retention and geometry.** The mask parser previously dropped the per-slit `offset` (field 10) and `theta` (field 11) columns, so the design geometry always treated a target as centered in an axis-aligned slit. The parser now retains both; `get_slitmask` rotates each slit's corners by `theta` and makes a target's `OBJ_TOPDIST`/`OBJ_BOTDIST` asymmetric per `offset`. All masks tested so far (167 slits across five masks) have zero offset/tilt, so this is a defensive correction with no effect on current reductions — `offset == theta == 0` reproduces the prior geometry exactly (locked in by a regression test).
- **A–B nod RA wrap.** The along-slit dither offset now wraps the RA difference into (−180°, 180°], so a nod sequence that straddles 0h RA yields a small offset instead of one near ±360° (which would mis-pair the `comb_id`/`bkg_id` columns).
- **`CAT-RA` unit handling in the dither offset.** `CAT-RA` is written either as sexagesimal degrees (the MOS mask tool) or as decimal hours (long-slit pointings, matching the hours-valued `RA` card); parsing the hours form as degrees mis-scaled RA by 15× and produced nonsensical long-slit `dithoff` values (e.g. −335831″). The derivation now selects whichever interpretation lands the catalog position within a degree of the actual pointing. Surfaced while regenerating the dev-suite MMIRS long-slit setups.
- **Sexagesimal RA/Dec in the dither offset.** MMIRS changed its coordinate header format after 2019: newer long-slit *and* MOS data write `RA`/`CAT-RA` as sexagesimal strings rather than decimal. The former `float()` parse raised on these and silently returned a zero offset, disabling nod pairing for all recent data. Coordinates are now parsed with `astropy.coordinates.Angle` and the signed along-slit offset is computed with `SkyCoord.spherical_offsets_to`; off-sky calibration frames carrying a `DEC = -100` sentinel are caught and return zero. The entire pre-existing dev suite is decimal-format, so this was validated against recent frostig (long-slit) and igor (MOS) data.
- **`MASKDEF_OBJNAME` filename safety.** `pypeit_collate_1d` now sanitizes the mask-design object name used for `outfile_from = maskdef_objname` coadd names to a single safe path component, so a catalog target name containing a path separator or other invalid character can no longer fail the source or write outside the output directory.
- **Disambiguated names in the collate report.** When two distinct sources build the same base coadd name, the second is written under a disambiguated name; `collate_report.dat` now records that actual name (rather than a recomputed base name), so the report no longer points two sources at the same file.

Documentation (MMIRS spectrograph page, parameter tables, and release notes) has been updated throughout, and unit tests cover the ramp fitting, nod-pairing metadata (including the sexagesimal-header format), `.msk` parsing/geometry, wavelength-arxiv selection, and coadd file naming.


doc/releases/2.1.0dev.rst
```diff
@@ -127,6 +136,49 @@ Instrument-specific Updates
...
+  temporally-adjacent frame at the opposite nod position.  Pairing is done
+  separately for each target within an instrument configuration, and a sequence
+  spanning less than ~1 arcsec is left unpaired as a stare.  See the A-B nod
+  section of the :ref:`mmt_mmirs` documentation.
+- MMIRS multi-object (MOS) masks: ``pypeit_setup`` now ingests the MMIRS
+  ``.msk`` mask-design file (the ``xfitmask`` output) when a matching
+  ``<MOSID>.msk`` is found in the raw-data directory, enabling slitmask-design
+  tracing, alignment-box registration, and stamping of ``MASKDEF_ID``, ``RA``,
+  ``DEC``, and ``MASKDEF_OBJNAME`` (the catalog target name) on each extracted
+  object.  See the mask-definition section of the :ref:`mmt_mmirs` documentation.
```
kbwestfall, line 189: Can we move much of the information content in these bullet points to the `mmt_mmirs.rst` doc?  I.e., can we make these change-log comments more brief?

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
 up-the-ramp fitting
 +++++++++++++++++++
 
-The up-the-ramp fitting has not been implemented yet for MMIRS.
+MMIRS raw frames store the non-destructive reads of the HAWAII-2 detector as
+separate FITS extensions.  For frames with at least 5 reads, PypeIt performs
+up-the-ramp fitting with likelihood-based jump (cosmic-ray) detection using
+the algorithm of `Brandt (2024) <https://arxiv.org/abs/2404.01326>`__,
```
kbwestfall, line 22: Also add a reference to the `fitramp` repo here?

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+was inspired by the prototype at
+`mmt-mmirs-up-the-ramp-pypeit
+<https://github.com/zhechenghu/mmt-mmirs-up-the-ramp-pypeit>`__.
+
+The per-read noise needed by the fit is not the instantaneous detector read
+noise but an **effective** noise: the read noise plus the shot noise from dark
+current (and any flux) that accumulates over the ramp.  It therefore *grows
+with exposure time* -- measured on MMIRS darks, roughly 5.7 e- for a 3-read
+(3 s) ramp, 8.3 e- for 8 reads (10 s), and 9.4 e- for 69 reads (300 s) -- so it
+must be measured at the science exposure time.
```
kbwestfall, line 36: I'm not sure how much of the answer to this question needs to be added to this doc, but it's unclear to me how all of this propagates through the image processing used by the nominal pypeit workflow; see `doc/calibrations/image_proc.rst`.  In particular, how does this fit in with the error modeling done by `pypeit.core.procimg.variance_model`?

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+Preprocessed ramp images
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Fitting a many-read ramp takes minutes, and PypeIt loads each science frame
+several times during a reduction.  To avoid re-fitting, the fitted 2D
+count-rate image (units of e-/s) is written to a ``RampFit`` directory
+inside the reduction directory (alongside ``Calibrations``, ``Science``,
+and ``QA``) the first time each cube is loaded, and reused on subsequent
+loads.  A preprocessed image is re-fit automatically if the raw cube's
+modification time changes; if the ``RampFit`` directory is not writable,
```
kbwestfall, line 103: When would the directory not be writable?  We don't have this catch for the other directories produced by pypeit.  Why do we need it for this one?

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+the fit proceeds in memory with a warning and nothing is cached.
+
+The fitted images can also be created (and inspected) ahead of a reduction
+with:
+
+.. code-block:: bash
+
+    pypeit_fit_ramp mmt_mmirs raw/*.fits
+
+which writes into the ``RampFit`` directory under ``--odir`` (default: the
```
kbwestfall, line 113: Is `RampFit` hard-coded?  Technically, the other directories aren't, and part of the point of https://github.com/pypeit/PypeIt/pull/2192 is to ensure that the output directories are consistently treated throughout the codebase (not true at the moment).

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+current directory, so run it from the reduction directory) and accepts
+``--sig`` to force the single-read noise, ``--dark`` to calibrate it from a
+dark cube, and ``--force`` to re-fit existing outputs.  Preprocessed files
+carry the fit parameters in header cards (``RAMPSIG``, ``RAMPRON``,
+``NGROUPS``) and preserve all raw metadata, so :ref:`pypeit_setup` can be
+run directly on a ``RampFit`` directory if preferred.  ``run_pypeit`` never
+requires the manual step.
```
kbwestfall, line 120: I'm unsure about being able to point pypeit at either the truly raw data or the `RampFit` output.  I'd prefer consistency in the pypeit file with some auto-detection used to decide whether or not fitramp needs to be run, and then looks for output files that already exist.

I find the last sentence too terse.  Maybe: "``run_pypeit`` never requires preprocessing of the ramp files using ``pypeit_fit_ramp``."?

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+++++++++++++++++++++++++++++++
+
+MMIRS science frames are commonly taken as an A-B nod sequence (dithering the
+target along the slit between exposures), for both longslit and multi-object
+(MOS) masks. PypeIt derives the along-slit dither offset of each frame from the
+telescope pointing relative to the catalog target (the MMIRS header has no
+dither keyword) and records it in the ``dithoff`` column, along with ``dithpos``
+(A/B/A'/B' labels), ``dithpat`` (e.g. ``ABA'B'``), and ``frameno``.
+
+During :ref:`pypeit_setup`, PypeIt automatically fills the ``comb_id`` and
```
kbwestfall, line 187: Have you tested any of this when using the setup GUI?  I don't think the behavior should be different, but it's worth checking.

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+and left unpaired (``bkg_id = -1``).
+
+Combine the resulting 1D spectra with :ref:`pypeit_coadd_1dspec`.
+
+Mask definition (MOS)
++++++++++++++++++++++
+
+For multi-object (MOS) masks, PypeIt can ingest the MMIRS ``.msk`` mask-design
+file (the xfitmask output delivered with the raw data) so that every slit,
+``SpecObj``, and coadd carries the real target identity. During
```
kbwestfall, line 207: I'm not sure if the docs are that consistent with the terminology here, but I expect `spec1d` is better than `SpecObj`.

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+:ref:`pypeit_setup`, if a file named ``<MOSID>.msk`` (matching the ``APERTURE``
+header keyword and the ``decker`` metadata column) is found in the raw-data
+directory, PypeIt automatically:
+
+- enables slitmask-design tracing
+  (``[calibrations][slitedges] use_maskdesign = True``,
+  ``maskdesign_filename`` pointing at the ``.msk``);
```
kbwestfall, line 214: Show this in its true, pypeit-file form:
```
[calibrations]
    [[slitedges]]
        use_maskdesign = True
        maskdesign_filename = blah.msk
```
Maybe also emphasize that mask changes through the night should automatically be separated into different instrument setups/configurations, meaning you only have to define one mask-design file per pypeit file.

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+- stamps ``MASKDEF_ID``, ``RA``, ``DEC``, and ``MASKDEF_OBJNAME`` (the catalog
+  target name) on each extracted object
+  (``[reduce][slitmask] assign_obj = True``);
```
kbwestfall, line 217: Similar statement here about the format.  Maybe point to the parameters doc.

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+    left at 0 rather than the grid wavelength. A viewer that plots the raw
+    ``wave`` array *without* honoring the mask (such as ``lt_xspec``) therefore
+    draws a spurious spike back toward ``wave = 0``. The data are fine — this is
+    purely a display artifact. Load the file through `specutils`_ instead, which
+    applies the mask and returns a strictly monotonic wavelength axis::
+
+        from specutils import Spectrum
+        import pypeit.specutils   # registers the PypeIt loaders
+        spec = Spectrum.read('coadd/qso_172239.955+655201.69_MMIRS_<date>.fits')
+        # spec.spectral_axis (Angstrom) and spec.flux have the masked bins dropped
```
kbwestfall, line 260: We should make this consistent with `doc/include/specutils_usage.rst`.  I don't mind if we edit this code or that, but we should keep it consistent.

doc/spectrographs/mmt_mmirs.rst
```diff
@@ -12,24 +16,250 @@ items for the MMT/MMIRS spectrograph.
...
+
+.. note::
+
+    A coadded 1D spectrum (``OneSpec``) can contain masked bins where no input
+    exposure contributed a good pixel (e.g. a persistent bad pixel or an OH
+    residual that lands at the same wavelength in every flexure-controlled nod
+    exposure). These bins are correctly flagged in the good-pixel mask (``flux``,
+    ``ivar``, and ``mask`` are all zero there), but their stored wavelength is
+    left at 0 rather than the grid wavelength. A viewer that plots the raw
+    ``wave`` array *without* honoring the mask (such as ``lt_xspec``) therefore
```
kbwestfall, line 252: Should we be referencing `lt_xspec`?  Does `pypeit_show_1dspec` handle this properly?

pypeit/scripts/fit_ramp.py
```diff
@@ -0,0 +1,133 @@
...
+        parser = super().get_parser(
+            description='Preprocess up-the-ramp cubes into 2D count-rate '
+                        'images (e-/s), written to the RampFit directory '
+                        'inside the reduction directory.  Currently only '
+                        f'supports: {", ".join(RAMP_SPECTROGRAPHS)}.',
+            width=width,
+            default_log_file=True)
+        parser.add_argument('spectrograph', type=str,
+                            help='A valid spectrograph identifier: {0}'.format(
+                                 ', '.join(available_spectrographs)))
```
kbwestfall, line 46: If this is only viable for `mmt_mmirs`, why are we showing all the spectrographs here?

pypeit/scripts/fit_ramp.py
```diff
@@ -0,0 +1,133 @@
...
+        parser.add_argument('files', type=str, nargs='+',
+                            help='One or more raw multi-read cubes')
```
kbwestfall, line 48: You might think about this taking a pypeit file instead.  I.e., the workflow could be:
```
pypeit_setup ...
pypeit_fit_ramp blah.pypeit
run_pypeit blah.pypeit
```
where you just skip the 2nd step if you don't want to preprocess the ramp files.

pypeit/scripts/fit_ramp.py
```diff
@@ -0,0 +1,133 @@
...
+                            help='Re-fit and overwrite existing up-to-date '
+                                 'preprocessed images')
+        return parser
+
+    @classmethod
+    def main(cls, args: argparse.Namespace) -> None:
+        from pathlib import Path
+
+        from pypeit import io, log, PypeItError
+        from pypeit.spectrographs import mmt_mmirs
```
kbwestfall, line 70: Can this be refactored now in a way that doesn't require the `mmt_mmirs` module?

pypeit/scripts/fit_ramp.py
```diff
@@ -0,0 +1,133 @@
...
+from __future__ import annotations
+
+import argparse
+
+from pypeit.scripts import scriptbase
+
+#: Spectrographs for which up-the-ramp preprocessing is implemented.  Adding a
+#: new instrument here is the only change needed to enable it for this script,
+#: provided the spectrograph implements the ramp-fitting interface used below.
+RAMP_SPECTROGRAPHS = ('mmt_mmirs',)
```
kbwestfall, line 28: I feel like this is unnecessary.  Can we instead have a `Spectrograph` base method that raises `NotImplementedError` for spectrographs that can't process (or don't provide) ramp files?

pypeit/scripts/fit_ramp.py
```diff
@@ -0,0 +1,133 @@
...
+                f'supported spectrographs: {", ".join(RAMP_SPECTROGRAPHS)}.')
+
+        if args.sig is not None:
+            # Seeds the sigma cache: used for every frame
+            spec._ramp_sigma = float(args.sig)
+            log.info(f'Using forced single-read noise: {args.sig:.2f} e-')
+        elif args.dark is not None:
+            dark = Path(args.dark)
+            with io.fits_open(dark) as dhdu:
+                n_dark_reads = mmt_mmirs.mmirs_count_reads(dhdu)
```
kbwestfall, line 88: It would be better if this was a `spec` method.  Similar request for the `mmt_mmirs` specific functions below.

pypeit/spectrographs/mmirs_maskfile.py
```diff
@@ -0,0 +1,158 @@
...
+    slits['target'] = np.array([r['target'] for r in rows], dtype=int)
+    slits['object'] = np.array([r['object'] for r in rows], dtype=object)
+    slits['type'] = np.array([r['type'] for r in rows], dtype=object)
+    slits['height_mm'] = np.array([r['height_mm'] for r in rows], dtype=float)
+    slits['width_mm'] = np.array([r['width_mm'] for r in rows], dtype=float)
+    slits['offset_mm'] = np.array([r['offset_mm'] for r in rows], dtype=float)
+    slits['theta_deg'] = np.array([r['theta_deg'] for r in rows], dtype=float)
+    slits['bbox'] = np.array([r['bbox'] for r in rows], dtype=float)
+    slits['polygon'] = np.array([r['polygon'] for r in rows], dtype=float)
+    return header, slits
```
kbwestfall, line 158: Can all of this be put into `mmt_mmirs.py`?  And can you use `pypeit.spectrographs.slitmask.SlitMask`?

_Edit:_ I see you are using `SlitMask` in the `get_slitmask` method in `mmt_mmirs.py`.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -81,8 +149,246 @@ def compound_meta(self, headarr, meta_key):
             time = headarr[1]['DATE-OBS']
             ttime = Time(time, format='isot')
             return ttime.mjd
+        if meta_key == 'dithoff':
+            # Along-slit dither offset in arcsec: projection of the telescope
+            # pointing minus the catalog target (CAT-RA/CAT-DEC) onto the slit
+            # PA (POSANGLE).
+            hdr = headarr[1]
+            try:
```
kbwestfall, line 157: This try block is too long.  Can we isolate the specific lines we're trying to handle?

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -81,8 +149,246 @@ def compound_meta(self, headarr, meta_key):
...
+                The class holding the metadata for all the frames.
+        """
+        self._ramp_output_dir = Path(fitstbl.par['rdx']['redux_path'])
+        # Let the user override the ramp-fit threading/chunking from the
+        # [rdx] block of the pypeit file; unset (None) keeps the class default.
+        if fitstbl.par['rdx']['ramp_fit_cores'] is not None:
+            self.ramp_fit_workers = fitstbl.par['rdx']['ramp_fit_cores']
+        if fitstbl.par['rdx']['ramp_fit_chunk_rows'] is not None:
+            self.ramp_fit_chunk_rows = fitstbl.par['rdx']['ramp_fit_chunk_rows']
+        tbl = fitstbl.table
```
kbwestfall, line 227: Use the `PypeItMetaData` methods: `keys`, `find_frames`, `find_frame_files`, `frame_paths`, etc.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -81,8 +149,246 @@ def compound_meta(self, headarr, meta_key):
...
+            self._ramp_sigma_cache = {}
+        cache_key = (exptime, ron_floor)
+        if cache_key in self._ramp_sigma_cache:
+            return self._ramp_sigma_cache[cache_key]
+        lo = self.ramp_sig_range[0] if ron_floor is None \
+            else max(self.ramp_sig_range[0], float(ron_floor))
+        hi = self.ramp_sig_range[1]
+        match_exptime = exptime if self._ramp_match_dark_exptime else None
+        darks = self._ramp_dark_sigmas(exptime=match_exptime)
+        if darks:
```
kbwestfall, line 356: Make the condition explicit.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -249,9 +573,244 @@ def config_specific_par(
...
+            The input passed to :func:`config_specific_par`.
+
+        Returns
+        -------
+        `Path`_ or :obj:`None`
+            Path to the mask file if found, else ``None``.
+        """
+        try:
+            label = self.get_meta_value(inp, 'decker')
+        except Exception:
```
kbwestfall, line 617: Catch a specific exception.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -249,9 +573,244 @@ def config_specific_par(
...
+            directory = Path(inp).parent
+        elif isinstance(inp, Table):
+            if 'directory' in inp.colnames:
+                directory = Path(str(inp['directory'][0]))
+            elif 'filename' in inp.colnames:
+                directory = Path(str(inp['filename'][0])).parent
+        if directory is None:
+            return None
+        maskfile = directory / f'{label}.msk'
+        return maskfile if maskfile.exists() else None
```
kbwestfall, line 636: Is there a cleaner or more centralized location for this path determination code?  And, I may have missed it, but what happens when the user sets `maskdesign_filename`?

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -249,9 +573,244 @@ def config_specific_par(
...
+        onsky = np.column_stack([
+            np.array(slits['ra_deg'], dtype=float),
+            np.array(slits['dec_deg'], dtype=float),
+            np.asarray(slits['height_mm'], dtype=float) * arcsec_per_mm,  # length arcsec
+            np.asarray(slits['width_mm'], dtype=float) * arcsec_per_mm,   # width arcsec
+            np.full(n, header['pa'], dtype=float)])
+
+        self.slitmask = SlitMask(corners,
+                                 slitid=np.array(slits['slit'], dtype=int),
+                                 align=~is_target,
```
kbwestfall, line 730: Nit-picking: We generally prefer `np.logical_not` instead of the `~`.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -249,9 +573,244 @@ def config_specific_par(
...
+        if not Path(_fname).exists():
+            raise PypeItError(f'The mask design file {_fname} does not exist.')
+
+        bin_spat = 1
+        if binning is not None:
+            _, bin_spat = parse.parse_binning(binning)
+        platescale = self.get_detector_par(det=det)['platescale']
+
+        header, slits = mmirs_maskfile.read_mmirs_maskfile(_fname)
+        self.get_slitmask(_fname)
```
kbwestfall, line 793: I assume you need to read the file twice so that you can get the `header` object?  Is there a cleaner approach?

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -288,6 +847,123 @@ def check_frame_type(self, ftype, fitstbl, exprng=None):
...
+                    bkgid[j] = combid[i]
+                    paired[i] = paired[j] = True
+                    break
+            fitstbl['bkg_id'][idx] = bkgid
+
+            # Informational A/B (+prime) labels: within each nod side, distinct
+            # dithoff values (rounded to 0.1") get a prime suffix by first
+            # appearance in time.
+            dithpos = np.array(['None'] * idx.size, dtype=object)
+            for is_A, base in [(True, 'A'), (False, 'B')]:
```
kbwestfall, line 941: This `is_A` approach feels convoluted.  Can you use the `dithoff > midpoint` distinction directly?

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -361,31 +1037,80 @@ def get_rawimage(self, raw_file, det):
...
         """
         fil = utils.find_single_file(f'{raw_file}*', required=True)
 
         # Read
         log.info(f'Reading MMIRS file: {fil}')
         hdu = io.fits_open(fil)
-        head1 = fits.getheader(fil,1)
+
+        redux_path = self._ramp_output_dir if self._ramp_output_dir is not None \
+                else Path.cwd()
```
kbwestfall, line 1061: More bespoke path stuff that will be hard to maintain.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -361,31 +1037,80 @@ def get_rawimage(self, raw_file, det):
...
-        # First read over the header info to determine the size of the output array...
-        datasec = head1['DATASEC']
-        x1, x2, y1, y2 = np.array(parse.load_sections(datasec, fmt_iraf=False)).flatten()
-
-        # ToDo: I am currently using the standard double correlated frame, that is a difference between
-        # the first and final read-outs. In the future need to explore up-the-ramp fitting.
-        if len(hdu)>2:
-            data = mmirs_read_amp(hdu[1].data.astype('float64')) - mmirs_read_amp(hdu[2].data.astype('float64'))
+        # Need the exposure time
+        exptime = hdu[self.meta['exptime']['ext']].header[self.meta['exptime']['card']]
```
kbwestfall, line 1082: Use `self.get_meta_value`.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -361,31 +1037,80 @@ def get_rawimage(self, raw_file, det):
...
+            # Up-the-ramp fitting with jump detection
+            rate, sig, eff_ronoise = self._ramp_fit_image(hdu, detector_par)
+            detector_par['ronoise'] = np.atleast_1d(eff_ronoise)
+            array = rate * exptime / gain
+            # Persist the fit so later loads (and other scripts) reuse it
+            rampfit_file = mmirs_rampfit_path(fil, redux_path)
+            try:
+                mmirs_write_rampfit(rampfit_file, rate, hdu, sig, eff_ronoise,
+                                    Path(fil).stat().st_mtime, raw_file=fil)
+                log.info(f'Wrote preprocessed ramp image: {rampfit_file}')
```
kbwestfall, line 1099: Move this logging line into an `else` block.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -440,4 +1203,413 @@ def mmirs_read_amp(img, namps=32):
...
+        Path to the raw MMIRS cube.
+    redux_path : :obj:`str`, `Path`_
+        Path to the reduction directory.
+
+    Returns
+    -------
+    `Path`_
+        ``<redux_path>/RampFit/<raw filename>``
+    """
+    return Path(redux_path) / 'RampFit' / Path(raw_file).name
```
kbwestfall, line 1283: Should the names of the files exactly match the raw file name?  They're in a different directory, but it would seem useful to give them a different name reflecting they've been processed in some way.

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -440,4 +1203,413 @@ def mmirs_read_amp(img, namps=32):
...
+    :obj:`bool`
+        True if the preprocessed image can be used in place of the cube.
+    """
+    rampfit_file = Path(rampfit_file)
+    if not rampfit_file.exists():
+        return False
+    try:
+        header = fits.getheader(rampfit_file)
+        mtime = float(header['RAWMTIME'])
+        raw_mtime = Path(raw_file).stat().st_mtime
```
kbwestfall, line 1312: Isolate each line.  Which one's need a try block?

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -440,4 +1203,413 @@ def mmirs_read_amp(img, namps=32):
...
+    raw_mtime : :obj:`float`
+        Modification time of the source raw cube.
+
+    Raises
+    ------
+    OSError
+        If the output directory cannot be created or the file cannot be
+        written.
+    """
+    rampfit_file = Path(rampfit_file)
```
kbwestfall, line 1369: Is it safe to change the input object type?  I.e., does this change the object in the calling function?

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -440,4 +1203,413 @@ def mmirs_read_amp(img, namps=32):
...
+    out = fits.HDUList([fits.PrimaryHDU(header=prihead),
+                        fits.ImageHDU(data=rate.astype(np.float32),
+                                      header=head1)])
+    rampfit_file.parent.mkdir(parents=True, exist_ok=True)
+    tmp_file = rampfit_file.with_name(rampfit_file.name + f'.tmp{os.getpid()}')
+    try:
+        out.writeto(tmp_file, overwrite=True)
+        tmp_file.replace(rampfit_file)
+    finally:
+        tmp_file.unlink(missing_ok=True)
```
kbwestfall, line 1393: Why is this `tmp_file` stuff needed?

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -440,4 +1203,413 @@ def mmirs_read_amp(img, namps=32):
...
+
+    Returns
+    -------
+    diffs : `numpy.ndarray`_
+        ``(reads[i+1] - reads[i]) / covar.delta_t[i]``, shape
+        ``(ngroups-1, ny, nx)``, in e-/s.
+    """
+    diffs = np.diff(reads, axis=0)
+    diffs /= np.asarray(covar.delta_t)[:, None, None]
+    return diffs
```
kbwestfall, line 1415: Make this a one-liner?
```
return np.diff(reads, axis=0) / np.asarray(covar.delta_t)[:, None, None]
```

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -440,4 +1203,413 @@ def mmirs_read_amp(img, namps=32):
...
+    def worker(rng):
+        r0, r1 = rng
+        flat = sub[:, r0:r1, :].reshape(ndiffs, (r1 - r0) * nx)
+        sig_row = np.full(flat.shape[1], sig_guess, dtype=np.float64)
+        with np.errstate(divide='ignore', invalid='ignore'):
+            result = fitramp.fit_ramps(flat, covar, sig_row)
+            guess = result.countrate * (result.countrate > 0)
+            result = fitramp.fit_ramps(flat, covar, sig_row,
+                                       countrateguess=guess)
+        chisq[r0:r1] = result.chisq.reshape(r1 - r0, nx)
```
kbwestfall, line 1528: I'd prefer not to have nested function definitions.  Can we restructure?

pypeit/spectrographs/mmt_mmirs.py
```diff
@@ -440,4 +1203,413 @@ def mmirs_read_amp(img, namps=32):
...
+        r0, r1 = rng
+        flat = np.ascontiguousarray(diffs[:, r0:r1, :]).reshape(ndiffs,
+                                                                (r1 - r0) * nx)
+        sig_row = np.full(flat.shape[1], sig, dtype=np.float64)
+        with np.errstate(divide='ignore', invalid='ignore'):
+            diffs2use, guess = fitramp.mask_jumps(flat, covar, sig_row)
+            result = fitramp.fit_ramps(flat, covar, sig_row,
+                                       diffs2use=diffs2use,
+                                       countrateguess=guess * (guess > 0))
+        countrate[r0:r1] = result.countrate.reshape(r1 - r0, nx)
```
kbwestfall, line 1589: Same comment about this nested function.

pypeit/tests/test_collate_1d.py
```diff
@@ -345,6 +345,99 @@ def test_build_coadd_file_name():
...
+    assert rows[0][0] != base
+
+    # With no assigned name (e.g. a dry run), it falls back to the base name.
+    source.coaddfile = None
+    rows2, _ = get_report_metadata(['MJD'], ['MASKDEF_OBJNAME'], source)
+    assert rows2[0][0] == base
+
+    # Non-SourceObject inputs are unaffected.
+    assert get_report_metadata(['MJD'], ['MASKDEF_OBJNAME'], 'not a source') \
+        == (None, None)
```
kbwestfall, line 438: Add messages to all asserts.

pypeit/tests/test_fitramp.py
```diff
@@ -0,0 +1,45 @@
...
+    sig_arr = np.full(npix, sig)
+    diffs2use, countrates = fitramp.mask_jumps(diffs, covar, sig_arr)
+    # The jumped diff (index 9) must be masked for all affected pixels
+    assert np.all(diffs2use[9, :50] == 0)
+    # Unaffected pixels should be almost entirely unmasked
+    assert np.mean(diffs2use[:, 50:]) > 0.95
+    # Refit with the mask recovers the true rate for the hit pixels
+    result = fitramp.fit_ramps(diffs, covar, sig_arr, diffs2use=diffs2use,
+                               countrateguess=countrates * (countrates > 0))
+    assert np.abs(np.median(result.countrate[:50]) - rate) < 2.0
```
kbwestfall, line 45: Assert messages.

pypeit/tests/test_metadata.py
```diff
@@ -210,3 +210,63 @@ def test_get_row_for_filename():
...
+    spec = load_spectrograph('shane_kast_blue')
+    par = spec.default_pypeit_par()
+    called = []
+    # Instance-level monkeypatch records the argument
+    spec.cache_metadata = lambda fitstbl: called.append(fitstbl)
+    fitstbl = PypeItMetaData(spec, par=par,
+                             data=Table({'filename': ['a.fits'],
+                                         'directory': ['/tmp']}))
+    assert len(called) == 1
+    assert called[0] is fitstbl
```
kbwestfall, line 272: Assert messages.

pypeit/tests/test_mmt_mmirs.py
```diff
@@ -0,0 +1,1502 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Mask-definition (.msk) support
+# ---------------------------------------------------------------------------
+
+from pypeit import dataPaths
+from pypeit.spectrographs import mmirs_maskfile
+
+MSK = dataPaths.tests.get_file_path('nep.as1.msk')
```
kbwestfall, line 1260: Imports at the top.  Can we avoid the global?

pypeit/collate.py
```diff
@@ -269,7 +286,9 @@ def get_report_metadata(object_header_keys, spec_obj_keys, file_info):
     if not isinstance(file_info, SourceObject):
         return (None, None)
 
-    coaddfile = build_coadd_file_name(file_info)
+    # Prefer the disambiguated basename assigned when the coadd was written;
+    # fall back to recomputing it (e.g. for a dry run that never wrote a file).
+    coaddfile = getattr(file_info, 'coaddfile', None) or build_coadd_file_name(file_info)
```
kbwestfall, line 291: I dislike this `or` construction, even though it's perfectly valid python.  I think it's unclear.

pypeit/collate.py
```diff
@@ -508,16 +527,44 @@ def flux(par, spectrograph, spec1d_files, failed_fluxing_log):
...
+    Args:
+        name (:obj:`str`): The raw name.
+
+    Returns:
+        :obj:`str`: A filename-safe component (never empty; ``'_'`` if the
+        input reduces to nothing).
+    """
+    safe = re.sub(r'[^A-Za-z0-9.+_-]', '_', str(name).strip())
+    # Avoid a component that is empty or only dots (e.g. '', '.', '..').
+    return safe if safe.strip('.') else '_'
```
kbwestfall, line 549: Explicit condition.  This also feels like it should go somewhere more general, like `core/parse.py`.

pypeit/collate.py
```diff
@@ -529,14 +576,61 @@ def build_coadd_file_name(source_object):
 
     date_portion = f"{start_date_portion}_{end_date_portion}"
 
-    if source_object.match_type == 'ra/dec':
-        coord_portion = 'J' + source_object.coord.to_string('hmsdms', sep='', precision=2).replace(' ', '')
-    else:
-        coord_portion = source_object.spec_obj_list[0]['NAME'].split('_')[0]
+    # Prefer the slitmask-design object name if requested and available
+    coord_portion = None
+    if getattr(source_object, 'outfile_from', 'coord') == 'maskdef_objname':
```
kbwestfall, line 581: Do we need to use `getattr`?

pypeit/metadata.py
```diff
@@ -223,29 +237,37 @@ def _build(self, files, strict=True, usrdata=None):
...
-            headarr = self.spectrograph.get_headarr(_ifile, strict=strict)
-
-            # Grab Meta
-            for meta_key in self.spectrograph.meta.keys():
-                value = self.spectrograph.get_meta_value(headarr, meta_key, 
-                                                         required=strict,
-                                                         usr_row=usr_row, 
-                        ignore_bad_header = (
+            log.info(f'Adding metadata for {_ifile.name}')
+            try:
```
kbwestfall, line 244: try block is too big.

pypeit/metadata.py
```diff
@@ -223,29 +237,37 @@ def _build(self, files, strict=True, usrdata=None):
...
+                    raise
+                log.warning(f'Unable to read metadata for {_ifile.name}: The file does not contain '
+                            'an expected FITS extension.  Skipping this file.')
+                self.skipped_file_indices.append(idx)
+                continue
+
+            directory = str(_ifile.parent)
+            data['directory'].append(directory if directory else '.')
+            data['filename'].append(_ifile.name)
+            for meta_key, value in row_data.items():
```
kbwestfall, line 270: Can you say more about what is driving the need for these changes?

kbwestfall (changes requested):
Thanks, @tepickering .  Lot's of comments for you!

doc/releases/2.1.0dev.rst
tepickering: yeah, i was thinking of doing that, but then didn't get around to it. thanks for the nudge.

tepickering: fixed in [00a09e575](https://github.com/tepickering/PypeIt/commit/00a09e5759606c5cae4a8355b8915837697feccc).

doc/spectrographs/mmt_mmirs.rst
tepickering: that's worth spelling out, a paragraph has been added to `mmt_mmirs.rst` to cover it and i reworded things for better clarity. what's handed downstream is the fitted image, `rate * exptime`, and a detector `ronoise` that is the effective read noise, `sig * sqrt(12(N-1)/(N(N+1)))`. `sig` is ideally calibrated on darks so the added variance from Poisson noise is minimal. what the growth in the per-read noise is mostly reflecting is correlated, low-frequency noise increasing with exposure time. folding that into `ronoise` is appropriate.

doc/spectrographs/mmt_mmirs.rst
tepickering: well, i personally tripped it accidentally when my disk was full so that's one way. of course, in that case even though the ramp fitting working, the whole thing died a little bit further along.

tepickering: given that, i've removed that catch for consistency and greater simplicity.

doc/spectrographs/mmt_mmirs.rst
tepickering: yup, it was hard-coded and it's a good call to make it configurable like other directories via an `rdx` parameter. done in bb1e7ee3e.

doc/spectrographs/mmt_mmirs.rst
tepickering: the pypeit file always lists the raw data. `run_pypeit` will key off that to look for preprocessed data in `[rdx] rampfit_dir`. if it's there, it'll use it. if not, it'll make it itself. it can be useful to manually preprocess the data cubes to make the fitted images easier to inspect before doing a full reduction. as long as the directory you put them in matches `[rdx] rampfit_dir`, `run_pypeit` will find and use them.

tepickering: docs fixed for clarity in 7e9214b3c.

doc/spectrographs/mmt_mmirs.rst
tepickering: i have not personally used the setup gui at all. i'll give that a go. so far it's done the right thing automatically with all the datasets i've through at it. testing infrastructure for the setup gui would be a nice goal.

doc/spectrographs/mmt_mmirs.rst
tepickering: `SpecObj` is right in this case. the catalog identity (MASKDEF_OBJNAME, MASKDEF_ID, RA, DEC) is stamped per extracted object — i.e. per `SpecObj` — and a `spec1d` file holds many of them (all the targets on the mask), so it's not the `spec1d` file that carries "the real target identity," it's each `SpecObj` within it. the section uses `SpecObj` consistently for that per-object structure (the `SpecObj` name stays SPATxxxx-SLITyyyy-DET, with the catalog identity in its attributes).

doc/spectrographs/mmt_mmirs.rst
tepickering: good idea — done. put the params in proper pypeit-file form (the full set: both `[calibrations][[slitedges]]` and `[reduce][[slitmask]]`), framed as "equivalent to" what PypeIt sets automatically, since the user doesn't normally type these. and added your second point: decker (the mask/APERTURE) is a configuration key, so masks that change through the night get split into separate setups by `pypeit_setup` — one mask-design file per pypeit file. in 78006c3bc.

doc/spectrographs/mmt_mmirs.rst
tepickering: same as the previous one — the pypeit-file code block (78006c3bc) replaces the inline notation for these params too, and per your suggestion I added a pointer to `:ref:parameters` for the full description of each (50b36a8f7).

doc/spectrographs/mmt_mmirs.rst
tepickering: `pypeit_show_1dspec` does handle masked wavelengths correctly. it does not (currently) handle `OneSpec` files, though. `lt_xspec` handles `OneSpec` files, but does not handle masks properly. which is why i went and made `specterm1d` which handles both formats and properly handles masks via `specutils`. `pypeit_show_1dspec` should be fixed to handle `OneSpec` and the rendering glitches need to be fixed, too. out of scope for this PR...

pypeit/scripts/fit_ramp.py
tepickering: oops. should be limited to spectrographs that support ramp-fitting. currently only `mmt_mmirs`, but there will likely be more in the future. fixed in f3c65eada.

pypeit/scripts/fit_ramp.py
tepickering: ah! very good idea that simplifies things very nicely. implemented in d54633024.

pypeit/scripts/fit_ramp.py
tepickering: this was an oversight now fixed in 80ce59c0f. moved the per-file preprocessing into a `Spectrograph.preprocess_ramp_file()` method (base raises `NotImplementedError`; MMIRS implements it), so `pypeit_fit_ramp` no longer imports `mmt_mmirs` — it just dispatches through the spectrograph interface. adding a new up-the-ramp instrument just requires that instrument's class to implement that method.

pypeit/scripts/fit_ramp.py
tepickering: good call — that list is gone now. the base `Spectrograph.preprocess_ramp_file()` raises `NotImplementedError`, and `pypeit_fit_ramp` now checks whether the spectrograph overrides it rather than consulting a hardcoded list, rejecting unsupported ones up front before any frame is read. in 6191498b8.

pypeit/scripts/fit_ramp.py
tepickering: yeah, i had gone back and forth whether to generalize now or when we have another example to implement. however, the border between general and MMIRS-specific is clear enough it does make more sense to generalize it now. pretty big refactor, but internal to this PR. moves the core ramp-fitting to `pypeit/core/ramp.py` and orchestration into a `RampSpectrograph` mixin class. an instrument opts in by inheriting `(RampSpectrograph, Spectrograph)` and implementing two hooks — `_load_ramp` and `_count_reads`. committed in 422c829cb.

pypeit/spectrographs/mmirs_maskfile.py
tepickering: not sure why that got put into a separate file and i had meant to clean that up. silly AI. cleaned up and committed in 3904b2046.

pypeit/spectrographs/mmt_mmirs.py
tepickering: good call. split into two much tighter blocks around the things that can actually raise. committed in 30e6dd7bd.

pypeit/spectrographs/mmt_mmirs.py
tepickering: good call — done. `cache_metadata` no longer scans the table by hand; it just keeps the `fitstbl`, and the dark lookup for the read-noise calibration now uses `find_frame_files('dark')`. one wrinkle worth noting: `find_frames/find_frame_files` need `framebit`, which isn't set until `get_frame_types/finalize_usr_build` runs — i.e. after `PypeItMetaData.__init__`, so it's not available inside the `cache_metadata` hook itself. so i defer the dark lookup to when the ramp is actually fit (`get_rawimage` / `preprocess_ramp_file`), by which point the same table is typed. committed in cd355c8fe.

pypeit/spectrographs/mmt_mmirs.py
tepickering: done — `if darks:` is now `if len(darks) > 0:`. i also made the same-idiom `if not cat_cands:` in the dithoff branch explicit (`if len(cat_cands) == 0:`) for consistency. committed in 6908f9353.

pypeit/spectrographs/mmt_mmirs.py
tepickering: narrowed to `except PypeItError` — that's what `get_meta_value` raises on unresolvable input (bad type / unopenable file); a missing decker just returns None, no exception. bf1a88c70.

pypeit/spectrographs/mmt_mmirs.py
tepickering: if `maskdesign_filename` is set, that is what's used, period. however, having the fallback path determination happen behind the scenes was a bit lazy on my part. it allowed me to just copy a downloaded mask and have it work while i was testing. the right approach, i think, is to rather have `pypeit_setup` do that work and configure the `.pypeit` file accordingly. drawback to that is it requires a few new hooks. this is all working around the fact that MMIRS does not append mask info as FITS extensions like every other MOS spectrograph. since MMIRS files get pre-processed, anyway, we could add the masks to those outputs. that seems a little hacky to me so i went with the `pypeit_setup` approach in b8b03ecd4.

pypeit/spectrographs/mmt_mmirs.py
tepickering: bad AI habits. fixed in d6a2514e9.

pypeit/spectrographs/mmt_mmirs.py
tepickering: ope, good catch. split the SlitMask build out of `get_slitmask` into a private `_build_slitmask(header, slits)`, so `get_maskdef_slitedges` reads the `.msk` once and hands the parsed `(header, slits)` straight to it. b272699e1.

pypeit/spectrographs/mmt_mmirs.py
tepickering: yeah, agreed. now iterates the `side` mask (`dithoff > midpoint`) directly — [`('A', side), ('B', logical_not(side))`] — and gets each side's frames in time order by indexing the mjd sort order (`order[on_side[order]]`) instead of the double argsort. same output. ff5ff8603.

pypeit/spectrographs/mmt_mmirs.py
tepickering: more path stuff is somewhat unavoidable since ramp-fitting introduces a new output file to handle. however, this bit belongs in the `RampSpectrograph` mixin. added `RampSpectrograph.rampfit_path(raw_file)` which owns the recorded-redux-dir-or-cwd fallback and the `rampfit_dir` name — it's the reduction-time counterpart of `preprocess_ramp_file` (which gets the dir explicitly from `pypeit_fit_ramp`). `get_rawimage` now calls it once instead of resolving the path inline and calling `rampfit_path` twice. committed in a1cab5de0.

pypeit/spectrographs/mmt_mmirs.py
tepickering: fixed in f7002cbc5.

pypeit/spectrographs/mmt_mmirs.py
tepickering: this got refactored already and moved to `pypeit.core.ramp`. the try/except is gone and it simply does:
```
ramp.write_rampfit(...)
log.info(f'Wrote preprocessed ramp image: {rampfit_file}')
```
the exception handling is all internal to `write_rampfit` so the logging isn't reached if something goes wrong.

pypeit/spectrographs/mmt_mmirs.py
tepickering: fair point. adding a `_rampfit` marker in ff01ee5fe. also audited to make sure there are no other dangling hard-coded references to `RampFit` and none remain.

pypeit/spectrographs/mmt_mmirs.py
tepickering: narrowed it. only two lines actually touch I/O, each now in its own `except OSError:` reading the sidecar header (`fits.getheader`) and stat-ing the raw cube. the `RAWMTIME` lookup became an explicit `if 'RAWMTIME' not in header` membership check, and `float(header['RAWMTIME'])` is now outside any try. both `OSError` guards are needed — a corrupt sidecar and a raw source that's been deleted must each report not-fresh. committed in e47cff6ab.

pypeit/spectrographs/mmt_mmirs.py
tepickering: this is safe because is rebinds the local name to a new `Path` object. `write_rampfit` doesn't mutate any inputs.

pypeit/spectrographs/mmt_mmirs.py
tepickering: it protects the freshness cache and guards against a header being updated, but the data getting truncated for some reason. i looked for, but didn't find, and atomic-write helper in `pypeit.io`. i think there should be one and i think there are other places in the code where it may be useful.

pypeit/spectrographs/mmt_mmirs.py
tepickering: done. b59d7d6a2.

pypeit/spectrographs/mmt_mmirs.py
tepickering: removed both nested defs. the two worker closures are now module-level functions — `_calibrate_sigma_block` and `_fit_ramp_block` — that take their arrays as explicit args; the call sites bind them with `functools.partial(...)` and hand the result to the existing `_fit_rows` dispatcher. behavior is identical (same in-place block writes). commited in 6ffe3a836.

pypeit/spectrographs/mmt_mmirs.py
tepickering: fixed in 6ffe3a836. see previous comment.

pypeit/tests/test_collate_1d.py
tepickering: added explanatory messages to every assert in this test. 0b9409b0d

pypeit/tests/test_fitramp.py
tepickering: added explanatory messages to every assert in this test. 0b9409b0d

pypeit/tests/test_metadata.py
tepickering: added explanatory messages to every assert in this test. 0b9409b0d

pypeit/tests/test_mmt_mmirs.py
tepickering: both done. moved `from pypeit import dataPaths` up to the top import block, and replaced the module-level `MSK` constant with a module-scoped `msk` fixture — `get_file_path` can reach into the data cache (it calls `fetch_remote_file`), so as a module global it ran that at import/collection time; the fixture defers it to test time and only for the tests that need it. the six tests now take `msk`. 0b9409b0d

pypeit/collate.py
tepickering: agreed — replaced it with an explicit if/else, and dropped the getattr since SourceObject always sets .coaddfile in __init__ and we've already returned early for non-SourceObject input. ae2175b91

pypeit/collate.py
tepickering: no — outfile_from is always set in SourceObject.__init__, so it's now a direct attribute access. ae2175b91

pypeit/collate.py
tepickering: moved it to `pypeit.core.parse.safe_name_component` (good call, it's general), and made the empty/dots guard an explicit length check. ae2175b91

pypeit/metadata.py
tepickering: narrowed it — pulled the per-file read (get_headarr + the meta-key loop) into a helper `_read_meta_row`, so the `except IndexError` now wraps a single call. get_headarr stays inside the guard on purpose: a single-HDU file raises the IndexError while its headers are read, which is exactly the skip case. bdcbcbb6b

pypeit/metadata.py
tepickering: the driving case is MMIRS: its raw-data directories mix the multi-extension science cubes with single-HDU acquisition/guider frames. when pypeit_setup scans the directory and hits one of those single-HDU frames, reading the expected image extension raises an IndexError. before this change that either crashed setup (strict) or produced a metadata row of Nones. the change makes non-strict setup skip the file instead — warn, record the index so __init__ drops the matching usrdata row, and only append the table columns on a successful read. that last part is why `data` went from index-assigned pre-sized arrays to append-on-success lists: a skipped file must leave no gap or the user-metadata columns misalign. covered by test_setup_warns_and_skips_single_hdu_file and test_skip_invalid_file_with_usrdata. bdcbcbb6b

doc/spectrographs/mmt_mmirs.rst
tepickering: added.

doc/spectrographs/mmt_mmirs.rst
tepickering: another way to deal with `OneSpec` is https://github.com/pypeit/PypeIt/issues/2060.

doc/spectrographs/mmt_mmirs.rst
tepickering: ah, glad i did check. the GUI and CLI take different paths when writing out a `.pypeit` file. the bits i added to write out the mask location and information only triggered on the CLI path. i have now refactored that to work with both paths by moving the merge upstream a bit into `Spectrograph.merge_setup_cfg_lines`. a headless `pytest-qt` test has been added to cover this, the first such GUI test in the CI suite. committed in 41facf23a.

doc/spectrographs/mmt_mmirs.rst
kbwestfall: I'm not following.  What I mean is that [here](https://pypeit.readthedocs.io/en/stable/out_spec1D.html#specutils-interface), we recommend, e.g.:
```
from pypeit.specutils import Spectrum
spec = Spectrum.read(onespec_file)
```
and we note that the import statement above is essentially identical to
```
from specutils import Spectrum
from pypeit.specutils import pypeit_loaders
```
I'm asking you to update this code block, or the code in `doc/include/specutils_usage.rst` so that the documentation is consistent.

tepickering: i think i crossed reply streams amidst the deluge. anyway, this is cleaned up and fixed in 7332ca9fe.

PR #2167, opened 2026-07-20
Centralize definition of output paths

kbwestfall:
The goal of this PR is to clean-up the definition of the output paths.

If you want to see all the nitty gritty, I will next submit a dev-suite PR with the development docs.  In particular, it's worth browsing the `pypeit_output_paths.md` doc to see the assessment of the state of affairs before this development work.

There are two main things this PR does:
 - It adds a new `PypeItOutputPaths` class that sits alongside the other global classes, `PypeItDataPaths` and `PypeItLogger`.  The code should only carry a single instantiation of this object, created in the package-level `__init__`.  Configuration of the object beyond it's defaults happens early in any workflow.  There are definitely ways we can clean this up in the longer term, but the main places the configuration happens is in a calling script or very high-level classes like `PypeIt`.  All lower level classes (like `Calibrations`) import the instantiated object at the package level and use the provided directories, without allowing for class/function argument overrides.  Any file written by `core/` methods require a function argument directly.
 - It tries to complete the transition from use of `os.path` to `pathlib`.

I'm still running tests, but I wanted to put this on everyone's radar.  In particular, I think this may help clean up some of the pathing issues in the KCWI/datacube PRs.


kbwestfall:
Lots of the afterburn tests fail.  Investigating...

MNBrod:
I've started taking a deeper look into this. I'm still not 100% sure that the singleton & imported object is the way to go here, rather than having something like `pypeit.py` or some other core script create an object and pass it down into child function calls, or just stapling it to the parset so we don't need to adjust the function definitions all the way down. I'm also not sure that importing the object is the _wrong_ idea either.

I will do a proper dive into the code shortly.

## My not well organized thoughts:

- We already mix importable singletons (e.g. the logger) and passed objects (e.g. ParSet), so picking one or the other is not inherently wrong. There is no anti-pattern here.

- I am worried about the possibility of accidentally preserving the configuration object across multiple executions of pypeit using the same import. I use multiprocessing pools to process large sets of data. Each processing instance invokes `import pypeit` only once - the execution code I wrote uses the same `pypeit` import for many runs. If we go with the importable singleton option, we need to be _very_ careful to make sure that we:

1. Clear the configuration at the end of each run and/or
2. Make sure any `if not outputPaths.configured: outputPaths.configure(par, caller='CoAdd1DSpec.main')` blocks are carefully vetted to properly overwrite a prior configuration when appropriate and/or
3. Are confident that a re-trigger the outputPaths.configure call outside of normal execution propagates properly, with no weird caching behaviors

- The `if not outputPaths.configured` piece feels like a lot of repeated code, rather than have `outputPaths` just do that check internally, ala `logger.getLogger()`. Maybe we could do something along the lines of

```
import pypeit.outputPaths

outputPath = outputPaths.getConfiguration(caller='CoAdd1DSpec.main')
```

At the top of each file? That may be more bloat than we want to add, though. Just a thought.

- With regards to making parts of PypeIt easier to import for other uses, is this adding a layer of complexity we should avoid? If we instead pass in an object as a function argument, defining the "PypeIt API" becomes a lot easier

pypeit/pkg/outputpaths.py
```diff
@@ -0,0 +1,341 @@
...
+                collate_outdir is None and 'collate1d' in par.keys()
+                and par['collate1d']['outdir'] is not None
+            ):
+                collate_outdir = par['collate1d']['outdir']
+
+        if dryrun is not None:
+            self._dryrun = dryrun
+
+        self._apply(
+            redux_path,
```
MNBrod, line 297: If `par = None` and `redux_path = None`, this will create an outputPath object with no reduction path, is that desired?

pypeit/pkg/outputpaths.py
```diff
@@ -0,0 +1,341 @@
...
+            instantiated more than once in the same process (e.g.,
+            :mod:`~pypeit.scripts.ql` builds a calibration ``PypeIt`` and a
+            science ``PypeIt`` in one run; a notebook or driver script may
+            loop over several reductions). Constructing a new ``PypeIt``
+            object is a deliberate signal that a new, independent reduction
+            context is starting, which is categorically different from
+            some other function accidentally calling :func:`configure`
+            mid-run. Do not pass ``force=True`` from anywhere other than
+            :class:`~pypeit.pypeit.PypeIt`'s ``__init__``; every other
+            caller should rely on the default once-only behavior.
```
MNBrod, line 263: What is the alternative strategy here, if not forcing? This indicates that some things should be forcing (e.g. `ql`), but then asserts that nothing but `PypeIt.__init__` should be doing it

MNBrod:
As a general question, should `outputPaths` and `outputfiles` be merged into one file?

pypeit/scripts/setup_coadd2d.py
```diff
@@ -209,7 +209,7 @@ def main(cls, args):
             tbl['filename'] = [f.name for f in files]
             ofile_name = f'{spec_name}_{obj}.coadd2d' if args.pypeit_file is None \
                 else Path(args.pypeit_file).name.replace('.pypeit', f'_{obj}.coadd2d')
-            ofile = str(Path(cfg['rdx']['redux_path']) / ofile_name)
+            ofile = Path(cfg['rdx']['redux_path']) / ofile_name
```
MNBrod, line 212: Should this not be a reference to outputPaths? Or is there context I'm missing in this case

pypeit/tests/test_outputpaths.py
```diff
@@ -0,0 +1,157 @@
...
+from pypeit import PypeItPathError
+from pypeit import outputfiles
+from pypeit.pkg.outputpaths import PypeItOutputPaths
+from pypeit.par.pypeitpar import PypeItPar
+
+
+def test_default_resolution(tmp_path):
+    # An unconfigured instance still resolves every property to the
+    # expected Path, but nothing is created on disk and nothing is ready.
+    op = PypeItOutputPaths(redux_path=tmp_path)
```
MNBrod, line 17: Same as an earlier comment, there's nothing to keep someone from setting `redux_path=None`. I don't see a reason to make `redux_path` an optional argument

MNBrod: And more to the point for this function, this test would not catch that edge case

pypeit/pypeit.py
```diff
@@ -87,6 +88,21 @@ def __init__(
...
+        # docstring in PypeItOutputPaths.configure(). It exists narrowly so
+        # that constructing a *new* PypeIt object (e.g. pypeit_ql building a
+        # calibration PypeIt and a science PypeIt in one run, or a notebook
+        # looping over several reductions in one process) always starts a
+        # fresh, correctly-resolved set of output paths, rather than being
+        # blocked by an earlier PypeIt object having already configured
+        # `outputPaths` in the same process. No other caller should use
+        # force=True.
+        outputPaths.configure(self.par, redux_path=redux_path, caller='PypeIt.__init__',
+                              force=True)
```
MNBrod, line 104: This gets to a more general point, but we need to be very clear that adding this means each new script we write that invokes `PypeIt` directly needs to be very deliberate in what it is doing

pypeit/sensfunc.py
```diff
@@ -248,9 +248,9 @@ def __init__(self, spec1dfiles, sensfile, par, par_fluxcalib=None, debug=False,
         self.algorithm = self.__class__._algorithm
 
         # QA and throughput plot filenames
-        self.qafile = sensfile.replace('.fits', '') + '_QA.pdf'
-        self.thrufile = sensfile.replace('.fits', '') + '_throughput.pdf'
-        self.fstdfile = sensfile.replace('.fits', '') + '_fluxed_std.pdf'
+        self.qafile = self.sensfile.parent / f'{self.sensfile.stem}_QA.pdf'
+        self.thrufile = self.sensfile.parent / f'{self.sensfile.stem}_throughput.pdf'
+        self.fstdfile = self.sensfile.parent / f'{self.sensfile.stem}_fluxed_std.pdf'
```
MNBrod, line 253: In the style of `coadd_output_file`, should this be moved to outputfiles as a function that returns a tuple of output names? Might not be directly in scope of this PR

doc/outputs.rst
MNBrod, line 1: I think a warning for developers about the potential pitfalls of not properly utilizing this object is warranted (e.g. misuse with multi processing, overwriting parts of `par['rdx']` and expecting that to work, etc)

MNBrod (changes requested):

PR #2168, opened 2026-07-21
p200_ngps: support He dome-lamp exposures in the u-channel wavelength solution

cfremling:
Adds support for He dome-lamp exposures (`IMGTYPE=DOMEHE`) as a second arc lamp for the P200/NGPS `u` channel.

## Why

The NGPS internal ThAr lamp has no detectable lines in `u` below ~3490 Å. Everything blueward of that — 3050–3490 Å, roughly a quarter of the channel — is polynomial extrapolation with nothing holding it down. He dome-lamp exposures show He I 3188.67 Å in the central image slice and provide the missing anchor.

## What the PR does

Adds two spectrograph attributes, `_he_idname` and `_he_split_wave`, naming a secondary arc lamp and the wavelength below which it replaces the primary one. Where they are set, `get_rawimage` blanks each arc frame outside the band its own lamp illuminates, so the master arc is built from He frames alone below 3400 Å and ThAr frames alone above it.

The blanking writes a value below the detector `mincounts`, so the pixels are flagged and dropped by the existing masked combine. No combine, clipping or detection parameter is changed — the arc is assembled entirely by stock code.

The split wavelength is converted through the channel's own template and divided by the frame's `BINSPEC`, and the band is oriented from `specaxis`/`specflip`, so it works at any binning and readout geometry. 3400 Å sits in a 3200–3490 Å gap with no lines from either lamp.

Also for `u`:
- **line lists** — a one-entry `HeI_NGPS` list for the anchor, and `ThAr_NGPS_ublue`, the stock ThAr list with entries below 3400 Å dropped. The lamp has no lines there, but the stock list carries ~5 entries/Å, four within 0.3 Å of the anchor, which would win the nearest-entry match. Identical to stock ThAr redward of 3400 Å.
- **reid_arxiv** — the existing `u` template with its blue wave grid refit through the anchor and that line injected into the flux, so `reidentify` has something to match against.

## Why not just stack the two lamps

The master arc is a sigma-clipped mean, and clipping centres on the median, so the lamp that is a minority of the stack is rejected as outliers. With 3 ThAr + 2 He frames, He I 3889.75 lands in the master arc at 12.8 ADU against 4349 ADU in the raw frames; with 3 ThAr + 4 He the ThAr lines are destroyed instead. Turning the clipping off avoids that but removes cosmic-ray rejection from ordinary ThAr-only stacks. Splitting the bands sidesteps the question — neither lamp is ever a minority, and neither contributes noise where it has no signal.

## Scope

Inert unless a channel sets `_he_idname`/`_he_split_wave` *and* frames for that lamp exist at the binning being reduced. Only `u` sets them; `g`/`r`/`i` are untouched. Nights with no DOMEHE frames reduce bit-for-bit identically to before — same lines, same RMS, same solution on every slice.

Tested on two commissioning nights at both 2×1 and 2×2 binning. `test_spectrographs.py`, `test_arcimage.py`, `test_wvcalib.py` and `test_setups.py` pass.

One note for reviewers: `_mincounts` for NGPS goes from -1e10 to -1e4 e⁻ so the blanking sentinel trips `MINCOUNTS`. Real data never approaches that — arc noise is tens of ADU — and ThAr-only reductions were verified unchanged.


pypeit/spectrographs/p200_ngps.py
```diff
@@ -317,6 +315,79 @@ def pypeit_file_keys(self):
...
+    _he_split_wave = None
+    _he_cache: dict = {}
+
+    @classmethod
+    def _he_split_pix(cls, binspec):
+        """``_he_split_wave`` as a binned pixel, from this channel's template."""
+        key = ('pix', cls.name)
+        if key not in cls._he_cache:
+            from astropy.table import Table
+            from pypeit import dataPaths
```
kbwestfall: Imports at the top.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -317,6 +315,79 @@ def pypeit_file_keys(self):
...
+        return int(round(cls._he_cache[key] / binspec))
+
+    @classmethod
+    def _he_present(cls, raw_file, binspec):
+        """True if the raw directory holds He-lamp frames at this binning."""
+        d = Path(raw_file).parent
+        key = ('has', str(d), cls._extname, binspec)
+        if key not in cls._he_cache:
+            found = False
+            for p in sorted(d.glob('ngps_*.fits')):
```
kbwestfall: Instead of searching for NGPS files, is it possible to pass the full list of files to parse?  I.e., what if there are files in the parent directory that the user does not want/need to include in the reduction?

pypeit/spectrographs/p200_ngps.py
```diff
@@ -317,6 +315,79 @@ def pypeit_file_keys(self):
...
+                        if str(h[0].header.get('IMGTYPE', '')).strip().upper() != cls._he_idname:
+                            continue
+                        e = next((x.header for x in h[1:]
+                                  if str(x.header.get('EXTNAME', '')).strip().upper()
+                                  == cls._extname), None)
+                        if e is not None and int(e['BINSPEC']) == binspec:
+                            found = True
+                            break
+                except Exception:
+                    continue
```
kbwestfall: Can we limit the number of lines in the try block and catch a specific exception?  Limiting the number of lines may mean we need multiple try blocks, but that kind of granularity will help isolate/diagnose issues.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -317,6 +315,79 @@ def pypeit_file_keys(self):
...
+        binspec = int(hdu[idx].header['BINSPEC'])
+        if not self._he_present(raw_file, binspec):
+            return
+        n = self._he_split_pix(binspec)
+        blue = slice(-n, None) if self._specflip else slice(0, n)
+        red = slice(None, -n) if self._specflip else slice(n, None)
+        sl = [slice(None), slice(None)]
+        sl[self._specaxis] = red if imgtype == self._he_idname else blue
+        raw_img[tuple(sl)] = self._offband_adu
+
```
kbwestfall, line 456: For all the new functions, add input parameter and output return descriptions to each docstring.

pypeit/spectrographs/p200_ngps.py
```diff
@@ -317,6 +315,79 @@ def pypeit_file_keys(self):
...
+                                  == cls._extname), None)
+                        if e is not None and int(e['BINSPEC']) == binspec:
+                            found = True
+                            break
+                except Exception:
+                    continue
+            cls._he_cache[key] = found
+        return cls._he_cache[key]
+
+    def _mask_offband_arc(self, raw_img, raw_file, hdu, idx):
```
kbwestfall: Do we need all of these arguments?  I.e., what is the relationship between `raw_file`, `raw_img`, and `hdu`?

pypeit/spectrographs/p200_ngps.py
```diff
@@ -983,9 +1056,21 @@ class P200NGPSSpectrograph_u(P200NGPSSpectrograph):
...
-    _wave_lamps = ['ThAr']
-    _arc_idnames = ('THAR',)
+    #: As wvarxiv_p200_ngps_u_thar_central.fits, but with the wave grid
+    #: blueward of 3400 A refit through the He I 3188.67 anchor and that line
+    #: injected into the flux; identical redward of 3500 A.
+    _wvarxiv = 'wvarxiv_p200_ngps_u_thar_central_blueanchor.fits'
+    #: ThAr_NGPS_ublue is the stock ThAr list with entries blueward of 3400 A
+    #: removed: the lamp has no detectable u lines there, but the stock list
+    #: carries ~5 entries/A, which would win the nearest-entry match against
+    #: the He anchor.  Identical to stock ThAr redward of 3400 A.
```
kbwestfall: Is there a programmatic way of achieving this?  I.e., instead having a copy of a large file, can we instead limit the lines used from the existing file to be between a set of provided wavelengths?

kbwestfall (changes requested):
Thanks, @cfremling !  I have a few requests.

pypeit/spectrographs/p200_ngps.py
cfremling: Done — `Table` and `dataPaths` are now imported at the top of the module.

pypeit/spectrographs/p200_ngps.py
cfremling: Done. The file parsing is split into a separate `_he_frame()` classmethod whose `try` wraps only the `fits.open` block and catches `(OSError, KeyError, ValueError)`. It also now logs a warning naming the file instead of silently continuing.

pypeit/spectrographs/p200_ngps.py
cfremling: Done — all four new methods now have full `Args`/`Returns` sections.

pypeit/spectrographs/p200_ngps.py
cfremling: Good catch, they were redundant. `raw_file` was just `hdu.filename()` and `idx` was `self._extname_index(hdu)`, both derivable from `hdu`. The signature is now `_mask_offband_arc(self, raw_img, hdu)`; `raw_img` stays a separate argument because it is the possibly-squeezed array rather than `hdu[idx].data` itself. It also returns a bool now for whether it modified the image.

pypeit/spectrographs/p200_ngps.py
cfremling: Agreed in principle, but I could not find a way to do this without editing code outside the NGPS module, so I have left the trimmed list in place for now and would like your steer.

There is no existing mechanism to restrict a line list by wavelength. `restrict_on_instr` filters on the `Instr` bitmask, not wavelength, and `load_line_lists` reads each `<lamp>_lines.dat` whole.

I prototyped the general version: a `lamps_wvrng` parameter, a list parallel to `lamps` where each element is a wavelength range, applied per lamp in `load_line_lists` before stacking. It works and gives a line set identical to the trimmed file, but it touches `pypeitpar.py`, `waveio.py` and `autoid.py`, which felt like more than an instrument PR should carry without asking.

I also tried keeping it inside the NGPS module by deriving the trimmed list at run time and writing it to the cache with `cache.write_file_to_cache`, the way `pypeit_install_linelist` does. The write succeeds but `dataPaths.linelist.get_file_path` cannot then retrieve it, because the `linelist` path is configured with `host: None` and the lookup falls through to `fetch_remote_file`. So that route does not work as-is either.

Happy to go either way:

1. add `lamps_wvrng` to this PR,
2. propose it as its own PR and rebase this one on it, or
3. keep the trimmed list.

Which would you prefer?

pypeit/spectrographs/p200_ngps.py
cfremling: You are right that this is fragile, and fixing it properly needs a change outside the NGPS module, so I have not made it.

`get_rawimage` only receives a file path. `buildimage_fromlist` does not pass the frame group down to `RawImage`, and caching the list from `check_frame_type` does not work either, because `get_frame_types` returns early when the frametypes come from the `.pypeit` file, so it never runs during `run_pypeit`.

So the directory search stands, but it is now documented with an explicit `.. warning::` naming the failure mode you describe: a DOMEHE frame sitting beside the data but excluded from the reduction still counts, and would cause the ThAr frames to be blanked below 3400 A. The search itself is narrowed to files whose `IMGTYPE` matches, that carry this channel's `EXTNAME`, and whose `BINSPEC`/`BINSPAT` match the frame being read.

The clean fix is to pass the file list, or the frame group, into `get_rawimage`, but that changes the signature for every spectrograph. Let me know if you would like that and I will do it.

kbwestfall: Thanks for the detail, both in terms of considering how to implement this and the multiple guards against a user hitting this issue.  I agree that we should hold off on changing the signature for all the spectrographs, even if that's what we decide is best in the long run.

pypeit/spectrographs/p200_ngps.py
kbwestfall: I think this is useful feature in general, so I'm happy for you to have edited files outside of the NGPS-specific files.  Thanks!

cfremling: Thanks — restored, so `lamps_wvrng` is back in this PR.

It is a list parallel to `lamps`, one entry per lamp, each `min:max` in Angstroms with either bound optional, or `None` to use the list in full. `load_line_lists` applies it per lamp before stacking, and it is passed at the three call sites where `par` is in scope. It uses `:` rather than `,` to separate the bounds, since ConfigObj splits list elements on commas.

The `u` channel now uses the stock `ThAr` list with `['3400:None', 'None']`, and the copied `ThAr_NGPS_ublue_lines.dat` is deleted. The resulting line set is identical to the copied file: 15316 of 17099 ThAr lines, the He anchor retained, and no ThAr entries below 3400 A.

pypeit/spectrographs/p200_ngps.py
cfremling: Agreed, thanks — leaving the signature alone. The directory search stays, with the warning documenting the failure mode.

cfremling:
Summary of where this stands, since the branch has been force-pushed a few times while working through the review.

**All six review points are addressed.** Five are done in the code; the sixth we agreed to leave.

| | Request | Resolution |
|---|---|---|
| 1 | Imports at the top | Done |
| 2 | Limit the try block, catch specific exceptions | Done — file parsing split into `_he_frame`, catching `(OSError, KeyError, ValueError)` and warning by name |
| 3 | Parameter and return descriptions | Done on all new methods |
| 4 | Redundant arguments | Done — `_mask_offband_arc(raw_img, hdu)`; `raw_file` and `idx` were derivable from `hdu` |
| 5 | Pass the file list rather than searching | Left as agreed; directory search documented with an explicit warning |
| 6 | Avoid copying a large line list | Done — new `lamps_wvrng` parameter; the copied list is deleted |

**What the PR now contains**

- `p200_ngps.py` — He dome-lamp support for the `u` channel: `_he_idname`/`_he_split_wave` name a secondary arc lamp and the wavelength below which it replaces the primary one, and `get_rawimage` blanks each arc frame outside the band its own lamp illuminates so the stock masked combine builds the arc from He frames alone below 3400 A and ThAr frames alone above it.
- `pypeitpar.py`, `waveio.py`, `autoid.py` — the `lamps_wvrng` parameter, restricting the lines used from each lamp to a wavelength range.
- Data — a one-entry `HeI_NGPS` list and a `u` `reid_arxiv` with its blue wave grid refit through the anchor.

**Verified on the current branch**

- Nights with no DOMEHE frames are bit-for-bit identical to before: same lines, same RMS, max |dlambda| = 0 on all three slices.
- With DOMEHE, the central slice matches the He anchor at pixel 506.56 with a fit residual of -0.0059 A; side slices are unchanged.
- `test_spectrographs`, `test_pypeitpar`, `test_wvcalib`, `test_arcimage`, `test_setups` all pass.

cfremling:
@kbwestfall Gentle ping — all six review points are addressed (summary table above), whenever you have a chance to take another look.

The two `test-alldeps` CI failures were download timeouts in `test_input_flux_file` (unrelated to this change); I've pushed an empty commit to re-trigger the checks.

We've also now verified the He frames against production data: the sequencer writes `IMGTYPE = DOMEHE` for the R/I-off He dome exposures (confirmed on 2026-08-09 frames), matching what this PR expects.

kbwestfall (approved):
Thanks for the ping, @cfremling , and apologies for the delay.  This all looks good!

PR #2172, opened 2026-07-28
Fix: Load reduce.trim_edge from PypeIt file in ReducePar.from_dic

ImPriyatam:
## Quick Summary

`ReducePar` defines `trim_edge` as a valid parameter with default `[3, 3]`, and `find_objects.py` already consumes it when calling `global_skysub`. However, `ReducePar.from_dict()` lists `trim_edge` in `allkeys` but never copies it into `kwargs`, so any user-specified value in the PypeIt file is silently ignored and the default is always used. This PR adds the missing `kwargs['trim_edge'] = cfg['trim_edge'] if 'trim_edge' in k else None` line so the parameter is properly initialized from the input dictionary. 

---

## Details

The bug is in `pypeit/par/pypeitpar.py` inside `ReducePar.from_dict()`. The method validates that `trim_edge` is an allowed key, but only populates `kwargs` for the nested `ParSet` keys (`findobj`, `skysub`, `extraction`, `cube`, `slitmask`).

Because `trim_edge` is never added to `kwargs`, `cls(**kwargs)` falls back to the default `[3, 3]` defined in `__init__`.

The parameter is already used correctly downstream in `find_objects.py`, where it is passed as `trim_edg` to `global_skysub`.

So the only missing piece is the dictionary initialization in `from_dict()`. The fix is a one-line addition:

```python
kwargs['trim_edge'] = cfg['trim_edge'] if 'trim_edge' in k else None
```

This makes user-specified `[reduce] trim_edge = 6, 6` values propagate into the final parameter object and the reduction.


kbwestfall (approved):
Looks good.  Thanks for the fix!

PR #2175, opened 2026-07-31
State update fixes

kbwestfall:
Likely because of an unintentionally bad merge, the `IFUCalibrations` class was still making calls to various state updates.  This doesn't match the pattern in the main `Calibrations` class, and they were causing the `mmt_binospec_ifu` tests to fail in the dev-suite.


profxj (approved):
sorry about that!

tepickering (approved):
so many overlapping PRs...

PR #2178, opened 2026-08-02
Mark tests that require remote data

### LINKED ISSUE #2176: Pypeit as Debian package and remote data
Hi,

I am [trying to package pypeit for Debian](https://bugs.debian.org/1142952). Pypeit originally comes with a huge data set in Github, with a significantly reduced data set in the package itself. My question here is whether this reduced set of data still provides a useful package without remote access? I am asking because Debian packages should work (at least for some use cases) locally without external access. 

This specifically is true to build the package, which should include running the tests that come with pypeit. I identified all tests that require remote access and could provide a patch that marks them with `@pytest.mark.remote_data`. Would that be a valuable contribution?

Also I found a few Python source files that are not (completely) covered by the main license : `pypeit/utils.py` and `pypeit/core/arc.py`. Would it be useful to mention the additional license statements somewhere? I could also make a Pull Request for this; what would be the preferred form then?

olebole:
This makes it possible to run a subset of the tests when no internet connection is available, like when building a Debian package.

Fixes #2176

### DIFF
--- a/pypeit/tests/test_archive.py
+++ b/pypeit/tests/test_archive.py
@@ -101,6 +101,7 @@ def test_archive_meta(tmp_path):
 
     assert test_meta3._metadata == expected_rows
 
+@pytest.mark.remote_data
 def test_archive_dir(tmp_path):
 
     archive_root = str(tmp_path)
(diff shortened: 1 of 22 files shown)


kbwestfall (changes requested):
Thanks @olebole .  I think this is generally good, but is it possible to have the tests require the remote data by default?

Also, please rebase this to our `develop` branch.

olebole:
I rebased the test to `develop` and added `--remote-data` as default pytest option.

kbwestfall (approved):
Thanks @olebole !

PR #2179, opened 2026-08-02
Write temporary test output to pytest tmp_path

### LINKED ISSUE #2177: Pypeit tests fail with preinstalled package
If I want to test pypeit as installed package, a number of tests fail because they try to write to `pypeit/data/tests/`, which is not writeable by a normal user:
```
FAILED tests/test_alignments.py::test_alignments - PermissionError: [Errno 13...
FAILED tests/test_arcimage.py::test_io - PermissionError: [Errno 13] Permissi...
FAILED tests/test_calibframe.py::test_io - PermissionError: [Errno 13] Permis...
FAILED tests/test_calibframe.py::test_parse_key_dir - PermissionError: [Errno...
FAILED tests/test_detector.py::test_io - PermissionError: [Errno 13] Permissi...
FAILED tests/test_fitting.py::test_pypeitfit - PermissionError: [Errno 13] Pe...
FAILED tests/test_flatfield.py::test_flatimages - PermissionError: [Errno 13]...
FAILED tests/test_mosaic.py::test_io - PermissionError: [Errno 13] Permission...
FAILED tests/test_onespec.py::test_io - PermissionError: [Errno 13] Permissio...
FAILED tests/test_pypeitfile.py::test_write_pypeit_file - PermissionError: [E...
FAILED tests/test_pypeitimage.py::test_full - PermissionError: [Errno 13] Per...
FAILED tests/test_pypeitimage.py::test_bitmaskarray_io - PermissionError: [Er...
FAILED tests/test_pypeitimage.py::test_calib_io - PermissionError: [Errno 13]...
FAILED tests/test_skysub.py::test_skyregions_io - PermissionError: [Errno 13]...
FAILED tests/test_slittrace.py::test_io - PermissionError: [Errno 13] Permiss...
FAILED tests/test_slittrace.py::test_io_single - PermissionError: [Errno 13] ...
FAILED tests/test_spec2dobj.py::test_spec2dobj_io - PermissionError: [Errno 1...
FAILED tests/test_spec2dobj.py::test_all2dobj_write - PermissionError: [Errno...
FAILED tests/test_spec2dobj.py::test_all2dobj_update_image - PermissionError:...
FAILED tests/test_specobj.py::test_io - PermissionError: [Errno 13] Permissio...
FAILED tests/test_specobj.py::test_iotwo - PermissionError: [Errno 13] Permis...
FAILED tests/test_specobjs.py::test_io - PermissionError: [Errno 13] Permissi...
FAILED tests/test_utils.py::test_yamlify - PermissionError: [Errno 13] Permis...
FAILED tests/test_wavetilts.py::test_wavetilts - PermissionError: [Errno 13] ...
FAILED tests/test_wvcalib.py::test_wavefit - PermissionError: [Errno 13] Perm...
FAILED tests/test_wvcalib.py::test_wavecalib - PermissionError: [Errno 13] Pe...
FAILED tests/test_wvcalib.py::test_wvcalib_no2d - PermissionError: [Errno 13]...
```
I guess these are files that are created during the test.

Regularly running the tests on the installed package is however a very good way to ensure that the package still works if other packages (like Astropy, Numpy, Scipy, or Python itself) were updated. In Debian this is used to ensure that no update will break the current set of packages.

Would it be possible to create these files in a temporary directory (pytests **tmp_path** or so) instead of a subdirectroy of the package dir?

olebole:
Use the pytest **tmp_path** fixture instead of writing generated files into `pypeit/data/tests`. This avoids writes to the installed package tree and follows pytest best practices for temporary test output.

This allows to test the installed package with the command `python3 -m pytest --pyargs pypeit`, without write access to the package directory tree. Also it ensures that no remnants of the tests remain in the installed package.

Alternatives could be

* Directly use **tmp_path instead of a **data_output_path** fixture. This would make the code simpler, but would require more changes.
* Instead of using the **tmp_path** fixture, use the **tempfile** package. This would result in a minimal change (only in `tstutils.py`, no changes in the test modules at all) but would have the disadvantage that it is more difficult to investigate the result in case of a failure (the **tmp_path** fixture uses a well-defined directory to store the temporary files).

Fixes #2177

### DIFF
--- a/pypeit/tests/test_alignments.py
+++ b/pypeit/tests/test_alignments.py
@@ -5,15 +5,16 @@ from pathlib import Path
 
 from IPython import embed
 
+import pytest
+
 import numpy as np
 
 from astropy.io import fits
 
 from pypeit import alignframe
-from pypeit.tests.tstutils import data_output_path
 
 
-def test_alignments():
+def test_alignments(tmp_path):
     nspat, nspec, nslits, nalign = 100, 1000, 10, 5
     tmp = np.ones((nspec, nspat)) * 10.
     traces = np.zeros((nspec, nalign, nslits))
@@ -26,7 +27,7 @@ def test_alignments():
                         spat_id=np.arange(nslits))
 
     alignments = alignframe.Alignments(**instant_dict)
-    alignments.set_paths(data_output_path(''), 'A', '1', 'DET01')
+    alignments.set_paths(str(tmp_path), 'A', '1', 'DET01')
     ofile = Path(alignments.get_path()).absolute()
 
     # I/O
--- a/pypeit/tests/test_arcimage.py
+++ b/pypeit/tests/test_arcimage.py
@@ -5,6 +5,8 @@ from pathlib import Path
 
 from IPython import embed
 
+import pytest
+
 import numpy as np
 
 from pypeit.images import pypeitimage
@@ -22,7 +24,7 @@ def test_init():
     arcImage = buildimage.ArcImage.from_pypeitimage(pypeitImage)
 
 
-def test_io():
+def test_io(tmp_path):
     # Instantiate a simple pypeitImage
     pypeitImage = pypeitimage.PypeItImage(np.ones((1000, 1000)))
     pypeitImage.reinit_mask()
@@ -30,12 +32,12 @@ def test_io():
     pypeitImage.PYP_SPEC = 'shane_kast_blue'
     # Now the arcimage
     arcImage = buildimage.ArcImage.from_pypeitimage(
-        pypeitImage, calib_dir=tstutils.data_output_path(''), setup='A', calib_id=['1'],
+        pypeitImage, calib_dir=str(tmp_path), setup='A', calib_id=['1'],
         detname='DET01'
     )
     # Set paths and check name
     ofile = Path(arcImage.get_path()).absolute()
-    assert str(ofile) == str(Path(tstutils.data_output_path('Arc_A_1_DET01.fits')).absolute()), \
+    assert str(ofile) == str(tmp_path / 'Arc_A_1_DET01.fits'), \
             'Calibration file name changed'
     
     # Write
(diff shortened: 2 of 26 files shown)


kbwestfall (changes requested):
Thanks, @olebole .  I think this is definitely an improvement.  Tests are failing, I presume because we're not allowed to treat fixtures as functions.  It's more effort, but I would be fine with removing the `data_output_path` function and having these test functions construct the relevant paths as needed from a `tmp_path` fixture argument.

olebole:
OK, I'll adjust this. May take a few days however.

kbwestfall:
Sorry, forgot to say in the review: Please also rebase this to our `develop` branch.

olebole:
I rebased this PR to develop and re-implemented it using `tmp_path` directly. Testing this directly 8and independently of #2178) uncovers a problem where I am unsure to proceed: Some tests, like `test_inputfiles.py::test_grab_rawfiles` run `tstutils.install_shane_kast_blue_raw_data()` to download additional test files into the standard `tests` subdir, but then check their existence in the `data_output_path`: https://github.com/pypeit/PypeIt/blob/c1ebf73d1336519ecf731a3a2c2435c14fa6838f/pypeit/tests/test_inputfiles.py#L46-L52

Aside from the inconsistency of the data output path vs. pypeit data hierarchiy, the major problem here (and maybe generally?) is that Pypeit seems to try to download directly into the package hierarchy, and not (like f.e. Astropy) into a local cache directory. This works for a single-user installation, but not for a central installation that is set to read-only when using (or testing) it by other users.

For the Debian tests, this does not matter (and I didn't experience that before) because tests that try to download files are switched off by #2178. However, I assume that this will seriously limit the usefulness of the Debian package because for many use cases the installed set of files is not enough. Is it possible to have a caching mechanism here that (optionally) downloads to a local cache dir instead of to the package `data` subdirectory?

PR #2181, opened 2026-08-06
Normalize calibration output paths in run state

tbowers7:
Convert `pathlib.Path` values assigned to `output_file` into strings inside `update_calib()`. This centralizes normalization for every calibration producer that records a `CalibFrame` path.

Keep the in-memory state consistent with the declared Pydantic field type and the JSON representation. This prevents `PydanticSerializationUnexpectedValue` warnings without changing the serialized filename.

Add regression coverage that verifies `Path` inputs are stored as strings and that `model_dump_json()` emits no warning.


profxj (approved):
Thanks for this!

Issue #2180, opened 2026-08-04
Recover manual flexure correction


kbwestfall:
Reported by @Adolfo1519  :

The main code block doing the implementation was in the old objfind_one() function in pypeit.py . The specific lines were 826-840 in version 1.18.1.

These were removed in 2.0.0, likely by the step-by-step processing refactor.

kbwestfall:
Addressed in #2182

PR #2182, opened 2026-08-09
Fix manual

profxj:
Addresses Issue #2180

There will be a corresponding PR in the Dev-Suite

---

> [!NOTE]
> **Medium Risk**
> Touches science-frame spatial alignment and user sky masks; behavior change for pypeit files that relied on `shift=0` meaning “unset” rather than a zero-pixel override.
> 
> **Overview**
> Restores **manual spatial flexure** for science/standard frames (Issue #2180), which was dropped in the 2.0.0 step-by-step refactor.
> 
> The pypeit-file **`shift`** column again overrides the auto-computed spatial flexure when set to any **finite** value (pixels); the unset sentinel is now **`NaN`** instead of integer `0`, so **`0.0`** is a valid explicit override. Metadata initialization (`PypeItMetaData`) defaults `shift` to `np.nan` in the relevant column setup paths.
> 
> Reduction applies the override in **`process_one_det`** via new **`get_manual_flexure()`**, sets **`sciImg.spat_flexure`**, and logs the value used. **`load_skyregions`** accepts **`spat_flexure`** again and **`instantiate_objfind`** passes **`sciImg.spat_flexure`** when building user sky-region masks.
> 
> Docs add **:ref:`flexure-spatial-manual`** (`pypeit_setup -f` / `--flexure`) and a 2.1.0dev bug-fix note.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26e96ecf909dae804043daf156995911fbe10a45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


profxj:
cursor review

pypeit/pypeit_steps.py
```diff
@@ -575,6 +618,10 @@ def load_skyregions(spectrograph, fitstbl, par, frame, det, caliBrate,
         Flag to use the initial slits before any tweaking based on the
         slit-illumination profile; see
         :func:`~pypeit.slittrace.SlitTraceSet.select_edges`.
+    spat_flexure : :obj:`float`, optional
+        Spatial flexure shift (in pixels) of the science frame, used to
+        shift the user-defined sky regions consistently with the data.
+        If None, no shift is applied.
```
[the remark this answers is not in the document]
profxj: This is intentional, longstanding behavior rather than a defect of this PR (which restores exact v1.18.1 parity — the 1.18.1 `load_skyregions` also returned the file-based mask unshifted):

- The SkyRegions file is **per science frame** (its filename embeds the science-file basename) and is created interactively by `pypeit_skysub_regions` **on that same spec2d frame**, so the saved mask is already in that frame's pixel coordinates.
- The GUI applies the frame's recorded spatial flexure **at creation time** via its `--flexure` option (`pypeit/scripts/skysub_regions.py` passes `spec2DObj.sci_spat_flexure` into `SkySubGUI`, which shifts the slit edges with `select_edges(flexure=...)`).
- Applying `spat_flexure` again at load time would therefore **double-shift** masks created with `--flexure`.

The percentage-format regions, by contrast, are defined relative to the slit edges and so do need the flexure applied when the mask is generated — which is what this PR restores.

Added a clarifying NOTE comment and docstring text at the load site in f2a6a2e52 so this intent is explicit.

doc/calibrations/flexure.rst
```diff
@@ -89,6 +89,37 @@ This will:
...
+Set the column to any finite value (in pixels, including ``0.0``) for a
+science or standard frame to override the automatic determination for that
+frame:
+
+.. code-block:: console
+
+    filename |                 frametype | ... | shift
+    b27.fits |                   science | ... |   2.5
+    b28.fits |                   science | ... |   nan
+     b1.fits |                  arc,tilt | ... |   nan
```
kbwestfall: Using `NaN`s like this is a pet peeve of mine.  IMHO, nans should avoided (basically always), not used as a mask value.  Is there a different approach?  Can the parser handle None values or just leaving the entry blank?

pypeit/pypeit_steps.py
```diff
@@ -32,6 +32,39 @@
...
+def get_manual_flexure(fitstbl, frame:int):
+    """
+    Return the user-supplied manual spatial flexure for a frame, if any.
+
+    The manual value is taken from the ``shift`` column of the pypeit
+    file, which defaults to NaN; any finite value (including 0.) is
+    interpreted as a user request to override the automatically
+    computed spatial flexure.  See Issue #2180.
+
+    Generated by JXP and Claude.
```
kbwestfall: - Remove this attribution line.
- Convert to numpy style docs.

pypeit/pypeit_steps.py
```diff
@@ -575,6 +618,16 @@ def load_skyregions(spectrograph, fitstbl, par, frame, det, caliBrate,
...
+    spat_flexure : :obj:`float`, optional
+        Spatial flexure shift (in pixels) of the science frame, used to
+        shift the user-defined sky regions consistently with the data.
+        If None, no shift is applied.  Only used for percentage-format
+        ``user_regions``; a SkyRegions *file* (``user_regions = user``)
+        is loaded as-is, because it is created per science frame by
+        ``pypeit_skysub_regions`` in that frame's own pixel coordinates
+        (the GUI applies the frame's flexure at creation time via its
+        ``--flexure`` option), so shifting it here would double-apply
+        the correction.
```
kbwestfall: I know this isn't the point of this PR, but:
 - this docstring is missing many arguments,
 - I don't understand why we're passing the full `par` (`PypeItPar`), when the only parameter the function uses is `par['reduce']['skysub']['user_regions']`.

I also think the docstring for `spat_flexure` should be self contained.  I.e., it refers to context outside of the purpose of this function.  While useful, it feels off to use the argument description to talk about this nuance.

kbwestfall (changes requested):
Thanks!  My main comment is that I'm hoping there's a way we can avoid using nan to signify no value.

doc/calibrations/flexure.rst
profxj: Agreed — done in d69cc56a1. The `shift` column is now a **string** entry that mirrors the existing `manual` column: it defaults to blank, and the parser handles blank (or a literal `None`) entries naturally, meaning "no manual value". Any numeric entry — including `0.0` — is a manual override, and anything unparseable raises a `PypeItError` with a clear message. No NaN anywhere. Verified the round trip: `pypeit_setup -f` writes a blank column, and a hand-edited value reads back correctly through `PypeItFile.from_file`; also re-ran the full Kast-blue validation (shift=2.5 on one frame → `sci_spat_flexure=2.5` in its spec2d, blank frames unaffected).

pypeit/pypeit_steps.py
profxj: Done in d69cc56a1 — attribution line removed and the docstring converted to numpy style.

pypeit/pypeit_steps.py
profxj: All three points addressed in d69cc56a1:
- The docstring now documents every parameter (numpy style).
- The function now takes the `user_regions` value directly instead of the full `PypeItPar` — the single caller passes `par['reduce']['skysub']['user_regions']`.
- The `spat_flexure` argument description is now brief and self-contained; the SkyRegions-file/double-shift nuance moved to a `Notes` section.

kbwestfall:
Looks good.  I'm going to set off a dev-suite run just to be safe.

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  682 passed, 9436 warnings in 222.90s (0:03:42) ---
--- PYTEST UNIT TESTS PASSED  159 passed, 2802 warnings in 765.17s (0:12:45) ---
--- PYTEST VET TESTS PASSED  81 passed, 291625 warnings in 4432.95s (1:13:52) ---
Total disk usage: 338.202 GiB
Testing Started at 2026-08-11T16:11:09.040485
Testing Completed at 2026-08-12T03:50:05.995324
Total Time: 11:38:56.954839
```

kbwestfall (approved):

PR #2183, opened 2026-08-11
Update GH action versions

kbwestfall:
As titled.  We were getting a Node.js deprecation warning.

### DIFF
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -21,11 +21,11 @@ jobs:
         toxenv: [test, test-alldeps]
     steps:
     - name: Check out repository
-      uses: actions/checkout@v6
+      uses: actions/checkout@v7
       with:
         ref: develop
     - name: Set up Python ${{ matrix.python }}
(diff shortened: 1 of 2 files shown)


tbowers7 (approved):
Approve.

We can also set up dependabot actions to keep these kinds of things up-to-date, if desired.  I know there are philosophical and policy discussions behind setting that up for a larger project like PypeIt, but I have found it invaluable for smaller projects at my institution.  Maybe @tepickering has some thoughts about this.

rcooke-ast (approved):
Looks good to me 👍

tepickering:
> Approve.
> 
> We can also set up dependabot actions to keep these kinds of things up-to-date, if desired. I know there are philosophical and policy discussions behind setting that up for a larger project like PypeIt, but I have found it invaluable for smaller projects at my institution. Maybe @tepickering has some thoughts about this.

i didn't know dependabot could update actions config. i've only used it for dependencies and for node projects. in my experience, though, it's worth being intentional about updating actions versions because they can break or change behavior between versions.

PR #2187, opened 2026-08-24
Launches Shane Hamspec into the PypeIt eco-system

profxj:
Giddy up

---

> [!NOTE]
> **High Risk**
> Shared echelle wavecal and `wavecalib` paths are changed (order-ID offset hack, hardcoded FWHM, reidentification CC range), which can affect non-Hamspec reductions—not only the new spectrograph module.
> 
> **Overview**
> Adds **Shane/Hamspec** as a supported echelle spectrograph (`shane_hamspec`), including Loral vs e2v detector handling, header-derived metadata (computed **airmass**, era-specific reidentification archives), frame typing, and tuned defaults (slit-edge tracing, echelle reidentification, sky-subtraction off for slit-filling targets).
> 
> For narrow slits where peak-based finding fails, introduces **`force_center_obj`** and **`ech_slit_center_objs`**, which place one object per order on the order center with slit-width apertures and per-order S/N for echelle extraction.
> 
> **Echelle wavelength calibration** is repaired and sped up on shared paths: leftover debug skips in reidentification are removed; **`identify_ech_orders`** now returns archive spectra aligned 1:1 with observed orders (fixing size mismatches in `echelle_wvcalib`); cross-correlation can skip synthetic arc generation (`do_xcorr_arc=False`, `cont_subtract=False` on CCF peaks); **`fit2darc`** fails early with a clear error when fewer than two orders calibrate. **`get_echelle_angle_files`** accepts optional arc metadata for era-matched archives.
> 
> **Caveat:** the branch still carries instrument-specific hacks in shared code—e.g. **`order_vec_guess += 2`** in `predict_ech_arcspec` and a flat **`measured_fwhms = 3`** in `wavecalib`—that reviewers should treat as follow-up cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d22d171a1fcbaae56902b9b0356893c9a5e9b85. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


profxj:
@cursor review

pypeit/core/wavecal/autoid.py
```diff
@@ -1448,6 +1449,9 @@ def echelle_wvcalib(spec, orders, spec_arxiv, wave_arxiv, lamps, par,
             wv_calib[str(iord)] = None
             all_patt_dict[str(iord)] = None
             continue
+        # NOTE: removed a leftover debug block here that skipped every order
```
rcooke-ast: Remove this?

pypeit/core/wavecal/autoid.py
```diff
@@ -1463,9 +1467,16 @@ def echelle_wvcalib(spec, orders, spec_arxiv, wave_arxiv, lamps, par,
         # get rms threshold for this slit
         rms_thresh = round(par['rms_thresh_frac_fwhm'] * fwhm, 3)
         log.info(f"Using RMS threshold = {rms_thresh} (pixels); RMS/FWHM threshold = {par['rms_thresh_frac_fwhm']}")
+
+        # NOTE: removed a leftover per-order debug block here (an embed(), a
```
rcooke-ast: Also remove this? It seems like nothing is actually removed here according to the GH diff?

pypeit/core/wavecal/autoid.py
```diff
@@ -1463,9 +1467,16 @@ def echelle_wvcalib(spec, orders, spec_arxiv, wave_arxiv, lamps, par,
...
+        # forced `debug_all = True`, and a blocking plt.show() of the arxiv vs
+        # observed arc) that halted the reduction and popped a figure for every
+        # order.  See Report 05.
         detections[str(iord)], spec_cont_sub[:, iord], all_patt_dict[str(iord)] = reidentify(
             spec[:, iord], spec_arxiv[:, iord], wave_arxiv[:, iord], tot_line_list, par['nreid_min'],
-            cont_sub=par['reid_cont_sub'], match_toler=par['match_toler'], cc_shift_range=par['cc_shift_range'],
+            cont_sub=par['reid_cont_sub'], match_toler=par['match_toler'], 
+            #cc_shift_range=par['cc_shift_range'],
+            cc_shift_range=(-200., 200.),
```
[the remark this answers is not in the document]
rcooke-ast: Agreed with this. The parameter is correctly set in `shane_hamspec.py`, so reinstate `cc_shift_range=par['cc_shift_range'],`

pypeit/core/wavecal/echelle.py
```diff
@@ -151,26 +153,35 @@ def predict_ech_arcspec(angle_fits_file, composite_arc_file, echangle, xdangle,
     gpm_composite = (hdu[4].data).astype(bool)
 
     order_vec_guess = predict_ech_order_coverage(angle_fits_params, xd_angle_coeffs, xdisp, xdangle, norders, pad=pad)
+
+    # Remove this hack by X
+    order_vec_guess += 2
```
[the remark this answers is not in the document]
rcooke-ast: Agreed with this.

pypeit/core/wavecal/echelle.py
```diff
@@ -151,26 +153,35 @@ def predict_ech_arcspec(angle_fits_file, composite_arc_file, echangle, xdangle,
...
+
     norders_guess = order_vec_guess.size
     wave_soln_guess = predict_ech_wave_soln(angle_fits_params, ech_angle_coeffs, echangle, order_vec_guess, nspec)
 
     order_min, order_max = angle_fits_params['order_min'], angle_fits_params['order_max']
 
+    #embed(header='165 of echelle')
```
rcooke-ast: Remove these embeds?

rcooke-ast: Remove these embeds?

pypeit/core/wavecal/echelle.py
```diff
@@ -240,11 +252,27 @@ def identify_ech_orders(arcspec, echangle, xdangle, dispname,
...
+        embed(header='246 of echelle')
+        # -40 of X's data is Order 95 centered at ~6000Ang
+
     # Cross correlate the data with the predicted arc spectrum
     # TODO Does it make sense for xcorr_shift to continuum subtract here?
     shift_cc, corr_cc = wvutils.xcorr_shift(
         arccen_pad.flatten('F'), arcspec_guess_pad.flatten('F'), 
-        percent_ceil=cc_percent_ceil, sigdetect=5.0, sig_ceil=10.0, fwhm=4.0, debug=debug)
+        percent_ceil=cc_percent_ceil, 
+        do_xcorr_arc=False, # JXP ADDED THIS FOR SPEED; iter_continuum in arc_detect_lines is super slow-- Make it optional!
```
rcooke-ast: Can we make this an option before merging?

pypeit/core/wavecal/wvutils.py
```diff
@@ -608,15 +607,19 @@ def xcorr_shift(inspec1, inspec2, percent_ceil=50.0, use_raw_arc=False, sigdetec
...
-    tampl_true, tampl, pix_max, twid, centerr, ww, arc_cont, nsig = arc.detect_lines(corr_norm, sigdetect=3.0,
-                                                                                     fit_frac_fwhm=1.5, fwhm=5.0,
-                                                                                     cont_frac_fwhm=1.0, cont_samp=30, 
-                                                                                     nfind=1)
+
+    # Find the peak
+    tampl_true, tampl, pix_max, twid, centerr, ww, arc_cont, nsig = arc.detect_lines(
+        corr_norm, sigdetect=3.0, fit_frac_fwhm=1.5, fwhm=5.0, 
+        cont_frac_fwhm=1.0, cont_samp=30, nfind=1,
+        cont_subtract=False) # JXP added this; hamspec runs way too slow otherwise
```
rcooke-ast: Should this be added as a parameter?

pypeit/core/meta.py
```diff
@@ -56,6 +57,65 @@ def convert_radec(ra, dec):
...
+    horizontal (alt-az) frame.
+
+    Generated by JXP and Claude.
+
+    Args:
+        ra (:obj:`float`):
+            Target right ascension in decimal degrees (ICRS).
+        dec (:obj:`float`):
+            Target declination in decimal degrees (ICRS).
+        obstime (:obj:`float`, :obj:`str`, or `astropy.time.Time`_):
```
rcooke-ast, line 80: Should we ensure that the `obstime` is at the midpoint of the exposure? (i.e. add `exptime/2`)

pypeit/wavecalib.py
```diff
@@ -634,19 +636,27 @@ def build_wv_calib(self, arccen, method, skip_QA=False,
 
         # Generate a map of the instrumental spectral FWHM
         # TODO nsample should be a parameter
+
```
rcooke-ast: A couple of TODOs throughout here to put code back to the way it was before.

pypeit/wavecalib.py
```diff
@@ -634,19 +637,27 @@ def build_wv_calib(self, arccen, method, skip_QA=False,
...
         for islit in range(arccen.shape[1]):
             if islit not in ok_mask_idx:
                 continue
             # Measure the spectral FWHM (in pixels) at the midpoint of the slit
             # (i.e. the midpoint in both the spectral and spatial directions)
             measured_fwhms[islit] = fwhm_map[islit].eval(self.msarc.image.shape[0]//2, 0.5)
+        '''
 
         # Save for redo's
+        measured_fwhms[:] = 3. # REMOVE THIS!!
```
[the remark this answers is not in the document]
rcooke-ast: Agreed here!

pypeit/wavecalib.py
```diff
@@ -732,8 +742,19 @@ def build_wv_calib(self, arccen, method, skip_QA=False,
...
                 self.arccen = arccen
         elif self.par['method'] == 'echelle':
-            # Echelle calibration files
-            angle_fits_file, composite_arc_file = self.spectrograph.get_echelle_angle_files()
+            # Echelle calibration files; the arc-frame metadata lets
+            # multi-era spectrographs select era-matched archives.
+            angle_fits_file, composite_arc_file = \
+                    self.spectrograph.get_echelle_angle_files(meta_dict=self.meta_dict)
+
+            # TODO - remove these lines
```
rcooke-ast: rm?

pypeit/wavecalib.py
```diff
@@ -744,11 +765,16 @@ def build_wv_calib(self, arccen, method, skip_QA=False,
...
                     composite_arc_file,
                     pad=self.par['echelle_pad'],
-                    cc_percent_ceil = self.par['cc_percent_ceil'], debug=False)
+                    cc_percent_ceil = self.par['cc_percent_ceil'], 
+                    debug=False)
             # Put the order numbers in the slit object
             self.slits.ech_order = order_vec
             log.info(f"The observation covers the following orders: {order_vec}")
 
+            # NOTE: leftover debugging calls disabled so the 'echelle' wavecal
```
rcooke-ast: rm?

rcooke-ast: rm?

pypeit/spectrographs/shane_hamspec.py
```diff
@@ -0,0 +1,555 @@
...
+        # (~100).  Leaving add_missed_orders on inserts ~140 spurious orders in
+        # the dark detector regions (Hamspec orders do not fill the detector),
+        # which corrupts the echelle order-number identification.  Disable it.
+        par['calibrations']['slitedges']['add_missed_orders'] = False
+        par['calibrations']['slitedges']['order_width_poly'] = 2
+        par['calibrations']['slitedges']['order_gap_poly'] = 3
+
+        # These are the defaults
+        par['calibrations']['tilts']['tracethresh'] = 15
+        par['calibrations']['tilts']['spat_order'] = 3
```
rcooke-ast: Quite short slits... is 3 really needed?

pypeit/spectrographs/shane_hamspec.py
```diff
@@ -0,0 +1,555 @@
...
+            object: Metadata value read from the header(s).
+        """
+        if meta_key == 'mjd':
+            return Time(self._obs_isot(headarr[0]), format='isot').mjd
+        elif meta_key == 'airmass':
+            # The raw headers have no AIRMASS card, so compute it from the
+            # pointing (RA/DEC), the readout time, and the Shane/Lick location
+            # via the reusable pypeit.core.meta.airmass helper.
+            ra, dec = meta.convert_radec(headarr[0]['RA'], headarr[0]['DEC'])
+            obstime = Time(self._obs_isot(headarr[0]), format='isot')
```
rcooke-ast: Should the obstime be at the midpoint of the exposure?

rcooke-ast: Should this instead be the midpoint of the exposure (i.e. add or subtract half the exposure time, depending on if ISOT is recorded at the beginning/end of the exposure?)

pypeit/spectrographs/shane_hamspec.py
```diff
@@ -0,0 +1,555 @@
...
+            return good_exp & self.lamps(fitstbl, 'Off')
+        if ftype == 'bias':
+            return good_exp # & (fitstbl['target'] == 'Bias')
+        # The aperture plate (PLATENAM) is stored under the PypeIt 'decker'
+        # meta key.  The wide plate floods the detector (pixelflat); the narrow
+        # science plate gives the order edges (trace) and slit illumination
+        # (illumflat).  The plate names differ by era: 4kx4k (2014) uses
+        # 800:6.0 wide / 640:5.0 narrow; the Loral (2010) uses 800:5.0 wide /
+        # 800:2.5 narrow.
+        wide_plates = ['800:6.0', '800:5.0']           # -> pixelflat
```
rcooke-ast: Should the wide/narrow plates be set by the slit length and not the slit width?  I think the first number is the slit width (in um), and the second number is the slit length (in arcsecs).

Alternatively, if there are simply some set deckers that are used for pixflat vs illumflat, then maybe rename these variables? I'm also not sure how standard these deckers are for users of HamSpec.

pypeit/spectrographs/shane_hamspec.py
```diff
@@ -0,0 +1,555 @@
...
+        PypeIt metadata keyword.
+
+        This list is used by :meth:`~pypeit.spectrographs.spectrograph.Spectrograph.subheader_for_spec`
+        to include additional FITS keywords in downstream output files.
+
+        Returns:
+            :obj:`list`: List of keywords from the raw data files that should
+            be propagated in output files.
+        """
+        #return ['GRISM_N', 'BSPLIT_N']
```
rcooke-ast: remove the commented out line?

pypeit/spectrographs/shane_hamspec.py
```diff
@@ -0,0 +1,555 @@
...
+                                self.telescope['elevation'])
+        elif meta_key == 'detector':
+            # Same frame-size test as get_detector_par: the Loral 2Kx2K
+            # frames are 2080 columns, the e2v 4Kx4K frames are 4160.
+            return 'Loral2Kx2K' if headarr[0]['NAXIS1'] < 3000 \
+                else 'e2v4Kx4K'
+        raise PypeItError("Not ready for this compound meta")
+
+    @staticmethod
+    def _obs_isot(hdr):
```
rcooke-ast: Since this is not in the base spectrograph class, I wonder if it's better as a separate function at the bottom of this file, instead of a `staticmethod`?

rcooke-ast (changes requested):
Several changes requested, mostly cleanup.

pypeit/core/wavecal/echelle.py
```diff
@@ -218,6 +223,13 @@ def identify_ech_orders(arcspec, echangle, xdangle, dispname,
         Passed to xcorr_shift
     cc_percent_ceil: float, optional
         The percent_ceil value to be used by xcorr_shift to set the percentile to which to normalize the CCF
+    direct_cc : bool, optional
+        If True, cross-correlate the stacked arc spectra directly: skip the
+        synthetic line-arc construction and the continuum subtraction of the
+        correlation function in :func:`~pypeit.core.wavecal.wvutils.xcorr_shift`.
+        This is significantly faster for large spectral formats (e.g., 4k
+        detectors with ~100 orders, such as Shane/Hamspec).  The default
+        (False) preserves the original behavior.
```
kbwestfall: Do we know the downside for this?  I.e., if this is significantly faster, why don't we always use it?

pypeit/core/wavecal/echelle.py
```diff
@@ -267,12 +283,21 @@ def identify_ech_orders(arcspec, echangle, xdangle, dispname,
...
+    # Build per-observed-order arxiv arrays aligned 1:1 with order_vec.  Orders
+    # in order_vec that are not in the predicted coverage / composite (e.g. the
+    # data extends beyond the archived orders) are left as zeros; downstream
+    # echelle_wvcalib skips all-zero orders.  (Previously this returned only the
+    # overlapping subset, whose size != norders, crashing echelle_wvcalib.)
+    wave_soln_arxiv = np.zeros((nspec, norders))
+    arcspec_arxiv = np.zeros((nspec, norders))
+    for jord, oord in enumerate(order_vec):
+        match = np.where(order_vec_guess == oord)[0]
+        if match.size:
```
kbwestfall: I'd prefer that we always be explicit with checks like this.  I.e., don't rely on the "truthiness" of a value; check the value explicitly.

pypeit/core/findobj_skymask.py
```diff
@@ -1381,6 +1381,159 @@ def ech_objfind(image, ivar, slitmask, slit_left, slit_righ, slit_spat_id, order
...
+          slit, so its profile is the slit profile);
+        - the boxcar radius set to half the median order width, so a
+          boxcar extraction spans the full order.
+
+    A quick boxcar extraction is performed for each order to assign
+    the per-order S/N (``ech_snr``), which the downstream echelle
+    extraction (:func:`~pypeit.core.skysub.ech_local_skysub_extract`)
+    uses to set the order in which the orders are reduced.
+
+    Generated by JXP and Claude.
```
kbwestfall: Remove the attribution.

pypeit/core/findobj_skymask.py
```diff
@@ -1381,6 +1381,159 @@ def ech_objfind(image, ivar, slitmask, slit_left, slit_righ, slit_spat_id, order
     return sobjs_ech
 
 
+def ech_slit_center_objs(image, ivar, slitmask, slit_left, slit_righ,
+                         slit_spat_id, order_vec, det='DET01',
+                         inmask=None, specobj_dict=None):
```
kbwestfall, line 1386: I think both @bpholden and @holtzmanjon will want to know this function exists for possible use with APF/Levy and APO/ARCES.

pypeit/core/meta.py
```diff
@@ -56,6 +57,73 @@ def convert_radec(ra, dec):
...
+        Returns `numpy.nan`_ if the target is below the horizon.
+    """
+    # Observatory location on the Earth
+    location = coordinates.EarthLocation.from_geodetic(
+        lon=longitude*units.deg, lat=latitude*units.deg,
+        height=elevation*units.m)
+    # Interpret the observation time: MJD float, parseable string, or Time
+    if isinstance(obstime, Time):
+        time = obstime
+    elif isinstance(obstime, (int, float, np.floating)):
```
kbwestfall: Should we add `np.integer`?  Another option I recently learned (via a Claude suggestion) is using the standard python package `numbers`:

```
import numbers
is_scalar = isinstance(obstime, numbers.Real)
```

Technically, this will pass booleans to so we might just stick with adding `np.integer`.

pypeit/spectrographs/keck_hires.py
```diff
@@ -829,9 +829,14 @@ def get_detector_par(self, det, hdu=None):
         detector_dicts = [detector_dict1, detector_dict2, detector_dict3]
         return detector_container.DetectorContainer( **detector_dicts[det-1])
 
-    def get_echelle_angle_files(self):
+    def get_echelle_angle_files(self, meta_dict=None):
```
kbwestfall: Does this need to go here?  I.e., what can't it go in `config_specific_par`?

pypeit/spectrographs/shane_hamspec.py
```diff
@@ -0,0 +1,564 @@
...
+        """
+        par = super().config_specific_par(scifile, inp_par=inp_par)
+        
+        #headarr = self.get_headarr(scifile)
+
+        # wavelength
+        #par['calibrations']['wavelengths']['fwhm'] = 8.0/bin_spec
+
+        # Return
+        return par
```
kbwestfall, line 488: I think we should be using this instead of adding `meta_dict` to `get_echelle_angle_files`.  But if that doesn't make sense, we can remove this function because it's only calling the base class method.

pypeit/spectrographs/shane_hamspec.py
```diff
@@ -0,0 +1,564 @@
...
+
+    Args:
+        hdr (`astropy.io.fits.Header`_): Primary header of the raw frame.
+
+    Returns:
+        :obj:`str`: ISOT UT date/time string.
+    """
+    if 'DATE' in hdr and hdr['DATE'] not in (None, ''):
+        return hdr['DATE']
+    return hdr['DATE-OBS']
```
kbwestfall: This function feels unnecessary.  It is used twice in one function, which could consolidated into being used once.  And, if used once, we could just replace the call with these lines.

pypeit/tests/test_findobj.py
```diff
@@ -0,0 +1,92 @@
...
+    # Traces at the order centers, apertures spanning the slit
+    for iord, sobj in enumerate(sobjs):
+        center = 0.5*(slit_left[:, iord] + slit_righ[:, iord])
+        assert np.allclose(sobj.TRACE_SPAT, center)
+        assert sobj.SPAT_FRACPOS == 0.5
+        assert np.isclose(sobj.FWHM, 8.0)
+        assert np.isclose(sobj.BOX_R_PIX, 4.0)
+        # Bright slit-filling source must register a positive S/N
+        assert sobj.ech_snr > 0.
+        assert sobj.NAME is not None
```
kbwestfall: Add messages to all the asserts indicating the expected behavior.

kbwestfall (changes requested):
Some minor comments from me.

The only structural one is that I'd like to know if we can avoid adding the `meta_dict` keyword to `get_echelle_angle_files`.  Previously, we would have used `config_specific_par` for this purpose.

pypeit/core/wavecal/echelle.py
profxj: The downside is robustness: without the synthetic (clipped, line-only) arcs from get_xcorr_arc and the continuum subtraction of the CCF, a few very strong lines, detector artifacts, or continuum/scattered-light structure can dominate the correlation and skew the order registration. It's also untested on the HIRES/UVES/NIRSPEC setups, so the default (False) preserves the validated behavior; the docstring now spells out this trade-off. If it proves robust more broadly we could consider flipping the default.

pypeit/core/wavecal/echelle.py
profxj: Done — now an explicit `match.size > 0`.

pypeit/core/findobj_skymask.py
profxj: Removed (here and from the other new-code docstrings in this PR).

pypeit/core/findobj_skymask.py
profxj: Good point — the routine is instrument-agnostic (any echelle can enable it with [reduce][findobj] force_center_obj = True), so it should drop in for APF/Levy and APO/ARCES, which have the same short slit-filling orders. Happy to help wire those up.

pypeit/core/meta.py
profxj: Added np.integer. Stuck with the explicit tuple (rather than numbers.Real) to keep booleans out, per your note.

pypeit/spectrographs/keck_hires.py
profxj: Resolved — this change is reverted. keck_hires (and vlt_uves, keck_nirspec, the base class, and the wavecalib.py call site) are back to the develop signatures; the era selection now happens in shane_hamspec.config_specific_par (see the thread there).

pypeit/spectrographs/shane_hamspec.py
profxj: Done as you suggest: config_specific_par now records the detector era on the instance (from the 'detector' compound meta of the frame it is passed), and get_echelle_angle_files keeps the base-class signature and reads that attribute (defaulting to the Loral era). The meta_dict keyword is removed everywhere. Verified end-to-end: both dev-suite setups (Loral and e2v eras) reduce cleanly with the correct era-matched archives selected.

pypeit/spectrographs/shane_hamspec.py
profxj: Done — the helper is gone: the DATE/DATE-OBS logic is inlined in the 'mjd' compound meta (its single use), and the airmass branch now reuses compound_meta(headarr, 'mjd').

pypeit/tests/test_findobj.py
profxj: Done — messages added to all the asserts.

pypeit/spectrographs/shane_hamspec.py
kbwestfall: Sorry, I didn't really understand how `get_echelle_angle_files` was actually being used.  I had assumed these file names were kept as parameters that could be set as part of the config-specific set.  While I think it's better that we've traded a one-off need for a function keyword argument for a bespoke class attribute, both break our established pattern of having different spectrograph classes for different detectors.  Whether or not that pattern is the best approach is a different question...  I'm not sure what to recommend because requesting you separate this out into different spectrograph classes feels like too much work, but the current approach adds a one-off structural exception that complicates the code (not the first time we've done this).  Sorry that this comes off a bit ranty...

profxj: Claude and I propose:

Add two WavelengthSolutionPar keys (e.g. ech_angle_fits_file, ech_composite_arc_file, default None); wavecalib.py uses them when set, else falls back to get_echelle_angle_files() (HIRES/UVES/NIRSPEC untouched). shane_hamspec.config_specific_par then sets the two parameters by era — exactly the sanctioned mechanism Kyle assumed was in use, no bespoke class attribute, and it adds genuinely useful functionality (any user can point the echelle wavecal at a custom archive, like reid_arxiv). Cost: ~2 new pars + a 3-line change in wavecalib.py + simplifying shane_hamspec; re-verified by the existing 2-setup dev-suite run. I'd also offer this to Kyle as the resolution of his thread.

rcooke-ast: Just be warned that users often choose a different windowing on the detector, so I'm not sure how the `ech_angle_fits_file` and `ech_composite_arc_file` files will work in practice, since the reddest order may be different. Since hamspec is a fixed-format spectrograph, I'm not quite following why these files are needed. Could we not just assume the reddest order detected is a fixed order +/- some more lenient `echelle_pad` value?

profxj: Understood, and thanks for talking it through. We went with the approach you originally assumed: the two archive filenames are now WavelengthSolutionPar parameters (`ech_angle_fits_file` / `ech_composite_arc_file`, analogous to `reid_arxiv`). `wavecalib.py` uses them when set and falls back to `get_echelle_angle_files()` otherwise (HIRES/UVES/NIRSPEC unchanged). shane_hamspec sets the Loral-era files as its defaults and `config_specific_par` switches them per era — the bespoke instance attribute and the `get_echelle_angle_files` override are both gone, so there's no structural exception anymore, and any echelle user can now point the wavecal at custom archives. Verified: both dev-suite setups (Loral + e2v) reduce cleanly with the era-matched archives selected via the parameters.

pypeit/core/meta.py
rcooke-ast: Thanks for addressing this, I've just added into the docstring that the obstime is assumed to be the START of the exposure.

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  685 passed, 9432 warnings in 165.68s (0:02:45) ---
--- PYTEST UNIT TESTS PASSED  161 passed, 2804 warnings in 644.37s (0:10:44) ---
--- PYTEST VET TESTS PASSED  83 passed, 302467 warnings in 3229.22s (0:53:49) ---
--- PYPEIT DEVELOPMENT SUITE PASSED 310/310 TESTS  ---
Total disk usage: 346.881 GiB
Testing Started at 2026-09-10T15:48:42.094789
Testing Completed at 2026-09-10T23:57:34.850922
Total Time: 8:08:52.756133
```

kbwestfall (approved):

rcooke-ast (approved):
Looks good to me! Thanks Claude/X  :-)

PR #2189, opened 2026-08-28
HOTFIX: broadcast weights needs a new option for mosaics

rcooke-ast:
Keck/HIRES mosaic needs an additional option for broadcast weights when mean combining calibration frames.


bpholden (approved):
This seems so straightforward that even I understand it.

PR #2184, opened 2026-08-12
Keck/KCWI: Extraction coords, OH wavelength calibration, tests

kbwestfall:
Replaces #2150 .  The primary addition w.r.t. that PR is new tests to check the coordinate frame assumptions are correct; see `test_datacube.py`.

For the gory details, see the development doc here:

https://github.com/pypeit/PypeIt-development-suite/blob/fix-cube-manual-coords-kbw/claude_prompts/kcwi_wcs/kcwi_wcs.md

I need input on the following items from the development doc.  See:

https://github.com/pypeit/PypeIt-development-suite/blob/fix-cube-manual-coords-kbw/claude_prompts/kcwi_wcs/kcwi_wcs.md#inconsistencies--questions-for-the-team

- Item 4: Should we try to share code between the cube building used by the SlicerIFU and Fiber pipelines?  My impression is no, or at least not yet.
- Item 10: Should a failed slit/slice cause the cube building to fail?
- Item 11: Treatment of per-slicer offsets
- Item 15: Nuances of the `astropy.wcs.WCS` class and how it treats units.  I don't think there's an action here, but the unit conversion will likely cause headaches.  We should probably bring this up with Astropy.
- Item 16: Should we allow users to provide multiple manual extraction positions.

I'm going to run the full dev-suite on this branch now, and then try to have Claude add new tests with real data.

PR #2192, opened 2026-09-02
Fix output file basename inconsistencies

kbwestfall:
In the construction of the basename used for output file, we were not accounting for the different extensions possible for each spectrograph.  I ran into this issue because #2184 adds a `setup_datacube` script that assumes all the raw files have `.fits` extensions.

This PR accomplishes two things:

- It extracts the function primarily used to construct the basename from the `PypeItMetaData` class into its own stand-alone function.  This makes it easier to construct the basename at any place in the code.
- It consolidates the treatment of the construction of the basename that respects the allowed extensions for each spectrograph.

As a by-product of these goals, I also extended the `io.remove_suffix` function to respect additional compression extensions, instead of just working with `.gz` files.  And, I removed the old use of `.split('.fits')` to remove the file suffix.  Instead of the latter, I use either `io.remove_suffix` or `Path.with_suffix`.

For the nit-gritty details, see [the claude doc in the dev-suite](https://github.com/pypeit/PypeIt-development-suite/blob/basename/claude_prompts/basename/basename_refactoring_plan.md).


kbwestfall:
Dev-suite running....

pypeit/scripts/chk_noise_2dspec.py
```diff
@@ -224,8 +224,8 @@ def main(cls, args):
             # Save?
             folder = None
             if args.mode == 'save':
-                folder = '{}_noisecheck'.format(file.split('.fits')[0])
-                if not os.path.exists(folder): os.makedirs(folder)
+                folder = Path(f'{Path(file).with_suffix("")}_noisecheck')
```
rcooke-ast, line 227: Is it a concern that we add `_noisecheck` for both 1D and 2D checks (see above)? Should we also add `1D` and `2D`?

pypeit/spectrographs/soar_goodman.py
```diff
@@ -29,7 +29,7 @@ class SOARGoodmanSpectrograph(spectrograph.Spectrograph):
     ndet = 1
     telescope = telescopes.SOARTelescopePar()
     url = 'https://noirlab.edu/science/programs/ctio/instruments/goodman-high-throughput-spectrograph'
-    allowed_extensions = [".fz"]
+    allowed_extensions = [".fits", ".fits.fz"]
```
rcooke-ast: Might want to think about this... You can download `.fits` and `.fz` files from the SOAR archive, and the headers are either 0-indexed on 1-indexed. So, I think we will want to explicitly keep `.fz` here.

pypeit/tests/test_spectrographs.py
```diff
@@ -180,7 +180,7 @@ def test_load_spectrograph():
 
     # Test the allowed extensions for an oddball spectrograph
     spec5 = load_spectrograph('soar_goodman_red')
-    assert spec5.allowed_extensions == [".fz"], 'Found wrong extensions'
+    assert spec5.allowed_extensions == [".fits", ".fits.fz"], 'Found wrong extensions'
```
rcooke-ast: Likewise , probably want to revert this.

pypeit/coadd2d.py
```diff
@@ -461,16 +463,29 @@ def default_basename(spec2d_files):
         frsthdr = fits.getheader(spec2d_files[0])
         lasthdr = fits.getheader(spec2d_files[-1])
         if 'FILENAME' not in frsthdr:
-            raise PypeItError(f'Missing FILENAME keyword in {spec2d_files[0]}.  Set the basename '
-                        'using the command-line option.')
+            raise PypeItError(
+                f'Missing FILENAME keyword in {spec2d_files[0]}.  Set the basename using the '
```
rcooke-ast, line 467: Hmmm... I'm not sure what the command-line option is (but then again, I've never used coadd2d). I just thought I'd flag it in case it's not obvious to the user what they need to do to get a basename from these messages.

pypeit/io.py
```diff
@@ -227,10 +227,17 @@ def remove_suffix(file):
         'dot.separated.file.name'
         >>> remove_suffix('gzipped_file.fits.gz')
         'gzipped_file'
+        >>> remove_suffix('bz2_file.fits.bz2')
+        'bz2_file'
+        >>> remove_suffix('fpacked_file.fits.fz')
+        'fpacked_file'
 
     """
+    _compression_suffixes = ('.gz', '.bz2', '.fz')
```
rcooke-ast, line 236: Not sure if we want to include the fz suffix here. Otherwise, the SOAR Goodman might not work.

pypeit/pypeit_steps.py
```diff
@@ -100,6 +100,9 @@ def get_sci_metadata(spectrograph, fitstbl, frame:int, det):
 
     # Set binning, obstime, basename, and objtype
     binning = fitstbl['binning'][frame]
+    # NOTE: This determination of obstime is exactly what is done inside of
+    # construct_basename when obstime is not provided.  The only reason this is
+    # done here is because obstime is also returned by this function.
     obstime  = fitstbl.construct_obstime(frame)
```
rcooke-ast, line 106: Is it worthwhile uniting these here, or not worth the effort?

rcooke-ast (changes requested):
Some changes request. Let's see what the dev suite outputs, but I've mentioned a possible issue concerning SOAR/Goodman. Everything else is pretty minor. Thanks!

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  695 passed, 9436 warnings in 202.44s (0:03:22) ---
--- PYTEST UNIT TESTS PASSED  166 passed, 2803 warnings in 734.02s (0:12:14) ---
--- PYTEST VET TESTS PASSED  83 passed, 290779 warnings in 6566.10s (1:49:26) ---
Total disk usage: 331.205 GiB
Testing Started at 2026-09-02T21:49:27.697102
Testing Completed at 2026-09-03T11:57:12.928783
Total Time: 14:07:45.231681
```

pypeit/scripts/chk_noise_2dspec.py
kbwestfall: Good question.  I think the 1d script can accept either spec1d files or output files from 1D coadds.  In the former, the `spec1d` is kept because it uses the original filename as the base.  That won't be true for the latter, but its name can't get mixed up with the 2D file.  For the 2D script, the output file will always be `spec2d*_noisecheck`.

This maintains the old behavior, so I think we should keep it unless you have a strong objection.

pypeit/spectrographs/soar_goodman.py
kbwestfall: Good point. I had forgotten this.  I'll restrict it back to `.fits.fz`.  It needs to be the full extension, not just `.fz` to properly construct the basename now.

pypeit/tests/test_spectrographs.py
kbwestfall: Done!

pypeit/coadd2d.py
kbwestfall: Good catch.  The `pypeit_coadd_2dspec` script does have a `--basename` command-line option that can be used to override the default.  So I think this message is still valid (even if it's placement is less than ideal).

pypeit/io.py
kbwestfall: All the `soar_goodman` files in the dev-suite have the `.fits.fz` extension, so using `.fz` here is right for those files at least.  Are files distributed that don't have the `.fits` part of the extension?  The reason to have the `.fz` in the compression suffix list is to make sure `*.fits.fz` file strings are returned as `*`, not `*.fits`.

pypeit/pypeit_steps.py
kbwestfall: I'm of two minds.  `obstime` is used by `pypeit_steps.extract_det` as an argument passed to `pypeit_steps.refframe_correct`, but every other time it is returned it is ignored.  So it seems like we could consolidate, but it feels like it's more trouble than its worth for now.

kbwestfall (commented):
Thanks for the quick review, @rcooke-ast !  I've responded to everything.  Two things to draw your attention to:

- Do you have strong feelings about adding the 1D/2D notation to the noise check?  My edit maintains the old behavior and I think the 1D/2D will be included for most files because the output uses the input `spec1d/2d` files as the root name.
- Are there SOAR/Goodman files that have the `.fz` extension without the preceding `.fits`?  If so, I'll need to adjust, but otherwise things should work fine.  All the dev-suite files include `.fits.fz`.

rcooke-ast:
Thanks, @kbwestfall!  Sounds good, and I'm fine leaving noisecheck as it is (I've never used it before, and if others are happy with it as is, then we should leave it).

The only thing to avoid with SOAR/Goodman is to not allow `.fits` as an extension. So, I think your latest changes address this, thanks!

rcooke-ast (approved):
Approved!

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  695 passed, 9436 warnings in 235.02s (0:03:55) ---
--- PYTEST UNIT TESTS PASSED  166 passed, 2803 warnings in 826.20s (0:13:46) ---
--- PYTEST VET TESTS PASSED  83 passed, 291624 warnings in 5089.76s (1:24:49) ---
Total disk usage: 338.105 GiB
Testing Started at 2026-09-04T16:28:54.848358
Testing Completed at 2026-09-05T05:25:55.562082
Total Time: 12:57:00.713724
```

PR #2193, opened 2026-09-03
Auto-assign Binospec IFU combination groups from pointing pattern

tepickering:
## Summary

`pypeit_setup` on MMT/Binospec IFU (`mmt_binospec_ifu`) data now assigns combination groups automatically from the science/standard **pointing pattern**, so users no longer have to edit `comb_id` by hand:

- Frames that share a pointing (to within `comb_pointing_tol`, default **0.1″**) are given a common `comb_id` and **coadd** into one `spec2d` for further analysis.
- Each **distinct dither position** gets its own `comb_id` and reduces to its own `spec2d` file; repeat visits to a position coadd together.
- A **large offset for sky subtraction** simply falls into its own cluster currently — no special sky handling (deferred until we have data for that mode).

Binospec writes the *commanded* RA/DEC, so repeated visits to a position are byte-identical and a tight tolerance cleanly separates even sub-fiber sampling dithers. The `Fiber` pipeline models the sky from dedicated sky fibers or by combining them with science fibers via a joint fit, so there is no nod partner and **`bkg_id` is not set**.

This mirrors the pointing/`get_comb_group` approach recently added for MMT/MMIRS, adapted to a fiber IFU (2-D pointing clustering instead of an along-slit nod projection).

## Implementation

`MMTBINOSPECIFUSpectrograph` (`pypeit/spectrographs/mmt_binospec.py`):

- **`get_comb_group()`** — within each `setup`, greedily clusters science/standard frames by 2-D pointing separation and hands each cluster a contiguous `comb_id`. Combination never crosses a setup.
- **`comb_pointing_tol = 0.1`** (arcsec) class attribute.
- Two informational columns seeded in `init_meta` and appended in `pypeit_file_keys`: **`dithoff`** (arcsec offset from the setup's earliest pointing) and **`dithpos`** (`A`/`B`/… group label).

Grouping is only applied when the user has **not** pre-set any `comb_id` (guarded by `set_combination_groups`, `metadata.py`), so hand-edited pypeit files are left untouched.

## Testing

- New `pypeit/tests/test_mmt_binospec_comb.py` (7 unit tests): same-pointing coadd, multi-position dithers with repeat visits, sub-tolerance jitter, large sky-offset frame, cross-setup isolation, standard at its own pointing, and untouched calibration frames.
- All existing Binospec unit tests pass (32 total).
- Validated end-to-end on `JADES_1031022`: all 8 same-pointing science frames collapse to a single `comb_id` (`bkg_id=-1`), reproducing the previous hand edit automatically.

## Docs

Release note added under Instrument-specific Updates in `doc/releases/2.1.0dev.rst`. MMT Binospec documentation page updated to reflect new `pypeit_setup` behavior and what it means.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pypeit/spectrographs/mmt_binospec.py
```diff
@@ -1168,6 +1175,12 @@ def init_meta(self):
         for DAR correction).
         """
         super().init_meta()
+        # Dither columns filled in by get_comb_group.  Seeded here (with card
+        # defaults) so the columns exist -- and pypeit_file_keys can list them
+        # -- even when get_comb_group is skipped because the user pre-set
+        # comb_id in the pypeit file.
+        self.meta['dithoff'] = dict(ext=1, card=None, default=0.0)
+        self.meta['dithpos'] = dict(ext=1, card=None, default='None')
```
kbwestfall, line 1183: This feels like a hack, but I don't know a different way...

pypeit/spectrographs/mmt_binospec.py
```diff
@@ -1365,6 +1378,113 @@ def config_specific_par(
 
         return par
 
+    @staticmethod
+    def _dither_label(index):
+        """
+        Map a 0-based cluster index to a short dither-position label
+        (``A``, ``B``, ... ``Z``, then ``P26``, ``P27``, ...).
+        """
+        return chr(ord('A') + index) if index < 26 else f'P{index}'
```
kbwestfall: Is this overly complicated?  Why not `P01`, `P02`, etc?

pypeit/spectrographs/mmt_binospec.py
```diff
@@ -1365,6 +1378,113 @@ def config_specific_par(
...
+            `astropy.table.Table`_: The modified table.
+        """
+        # Make sure the derived columns exist and can hold the written values.
+        # init_meta seeds dithpos from the 4-char default 'None', giving a
+        # width-4 unicode column; coerce it to object dtype so multi-character
+        # labels are never truncated.
+        if 'dithoff' not in fitstbl.colnames:
+            fitstbl['dithoff'] = np.zeros(len(fitstbl), dtype=float)
+        if 'dithpos' not in fitstbl.colnames:
+            fitstbl['dithpos'] = np.full(len(fitstbl), 'None', dtype=object)
```
kbwestfall: Are these lines and the additions to `self.meta` above redundant?

pypeit/spectrographs/mmt_binospec.py
```diff
@@ -1365,6 +1378,113 @@ def config_specific_par(
...
+        elif fitstbl['dithpos'].dtype.kind in ('U', 'S'):
+            fitstbl['dithpos'] = np.asarray(fitstbl['dithpos'], dtype=object)
+
+        sci_std = np.array(['science' in ft or 'standard' in ft
```
kbwestfall, line 1418: Use `PypeItMetaData.find_frames`?

pypeit/spectrographs/mmt_binospec.py
```diff
@@ -1365,6 +1378,113 @@ def config_specific_par(
...
+                            for ft in fitstbl['frametype']])
+        if not np.any(sci_std):
+            return fitstbl
+
+        setup_col = np.asarray(fitstbl['setup']) if 'setup' in fitstbl.colnames \
+            else np.full(len(fitstbl), 'A')
```
kbwestfall: This feels dangerous.  `get_comb_group` is only ever called by `PypeItMetaData.set_combination_groups`, which follows a series of calls in a specific order to build the metadata table (one of which is `PypeItMetaData.set_configurations`).  I.e., if this function is called and `setup` isn't in `fitstbl` something has gone very wrong.  I would recommend raising an error instead.

pypeit/spectrographs/mmt_binospec.py
```diff
@@ -1365,6 +1378,113 @@ def config_specific_par(
...
+
+        # Cluster within each configuration and hand out contiguous comb_ids
+        # across the whole table.  Combination never crosses a setup.
+        next_comb = 1
+        for s in np.unique(setup_col[sci_std]):
+            idx = np.where(sci_std & (setup_col == s))[0]
```
kbwestfall, line 1435: Use `PypeItMetaData.find_configuration`?

pypeit/spectrographs/mmt_binospec.py
```diff
@@ -1365,6 +1378,113 @@ def config_specific_par(
...
+            ra = np.asarray(fitstbl['ra'][idx], dtype=float)
+            dec = np.asarray(fitstbl['dec'][idx], dtype=float)
```
kbwestfall: I think `Spectrograph.get_meta_value` always normalizes the RA/Dec coordinates using `pypeit.core.meta.convert_radec`.  Why do you need to force the type?

pypeit/spectrographs/mmt_binospec.py
```diff
@@ -1365,6 +1378,113 @@ def config_specific_par(
...
+                members = idx[c['members']]
+                fitstbl['comb_id'][members] = next_comb
+                fitstbl['dithpos'][members] = self._dither_label(ci)
+                next_comb += 1
+
+            # Offset (arcsec) of each frame from this setup's earliest pointing,
+            # for a quick read of the dither throw in the pypeit file.
+            ref = order[0]
+            off = np.hypot((ra - ra[ref]) * np.cos(np.radians(dec[ref])) * 3600.0,
+                           (dec - dec[ref]) * 3600.0)
```
kbwestfall: I feel like we should be using `astropy.coordinates` for all this offset stuff.

pypeit/tests/test_mmt_binospec_comb.py
```diff
@@ -0,0 +1,157 @@
+"""
+Unit tests for automatic combination-group assignment on MMT/Binospec IFU
+data.
```
kbwestfall: This will now be the 5th test file specific to MMT/Binospec.  Can we consolidate?

pypeit/tests/test_mmt_binospec_comb.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def test_same_pointing_shares_comb_id():
+    # Eight science frames at one identical pointing must coadd: a single
+    # shared comb_id, no background pairing, ~zero dither offset.
+    spec = load_spectrograph('mmt_binospec_ifu')
+    rows = [('science', RA0, DEC0, 60819.14 + 0.01 * i) for i in range(8)]
+    t = spec.get_comb_group(_comb_table(rows))
+    comb = np.asarray(t['comb_id'])
+    assert np.all(comb == comb[0])
```
kbwestfall: Add messages to all the assert statements explaining the expected behavior.

pypeit/spectrographs/mmt_binospec.py
kbwestfall: Just realized that the input to this function is the underlying Astropy table attribute of `PypeItMetaData`, not the object itself.  I'll leave this and other similar comments here because I think we should be using the object itself, but that's outside the scope of this PR.

pypeit/spectrographs/mmt_binospec.py
kbwestfall: Ignore.  `fitstbl` is not a `PypeItMetaData` object in this function...

kbwestfall (changes requested):
Thanks!

I have a number of comments.  All are minor, and some actually aren't relevant, but I left in anyway.

One primary request is that you add some discussion of this functionality to the `doc/spectrographs/mmt_binospec.rst` file.

tepickering:
> One primary request is that you add some discussion of this functionality to the `doc/spectrographs/mmt_binospec.rst` file.

that was an oversight on my part. added now.

pypeit/spectrographs/mmt_binospec.py
tepickering: hack as in how i'm initializing the metadata or hack as in how pypeit manages dither offsets?

pypeit/spectrographs/mmt_binospec.py
tepickering: Good call — you're right, it was overcomplicated. Dropped the `_dither_label` helper and its A..Z-then-`P26` fallback; `dithpos` is now a uniform, zero-padded `P01`/`P02`/... inlined at the assignment site. Fixed in a80439b84.

pypeit/spectrographs/mmt_binospec.py
tepickering: You're right — there's genuine overlap in the pipeline path. `self.meta` and `fitstbl` are different objects (a definition dict vs. the built table), but the `init_meta` additions aren't a no-op: because `dithoff`/`dithpos` are registered in the core meta data model (`core/meta.py`), `PypeItMetaData._build` creates both columns (with their card defaults, 0.0 / 'None') for every frame. So by the time `get_comb_group` runs in the real pipeline the columns already exist, and the two `if ... not in colnames` guards only ever fired for a bare hand-built table.

I kept `init_meta` as the single source of the columns — it's also what guarantees they exist for `pypeit_file_keys` when `get_comb_group` is skipped (user pre-set `comb_id`) — and dropped the redundant existence guards from `get_comb_group`, keeping only the `dithpos` object-dtype widening (which `_build` doesn't do). Tests now seed the two columns the way `_build` does. Fixed in bbe82fd6e.

pypeit/spectrographs/mmt_binospec.py
tepickering: Agreed — I confirmed the call order (`set_configurations` → `set_calibration_groups` → `set_combination_groups` in both `PypeItSetup.run` and `finalize_usr_build`), so `setup` is guaranteed present by the time `get_comb_group` runs. Replaced the `'A'` default with a `PypeItError` so a missing `setup` surfaces the broken build order instead of being masked. Fixed in 6e6770301.

pypeit/spectrographs/mmt_binospec.py
tepickering: You're right, the type forcing was unnecessary — `get_meta_value` already normalizes ra/dec to float degrees via `convert_radec`. I dropped the `dtype=float` casts and, while there, replaced the bespoke `cos(dec)` small-angle separation math (both the cluster-membership test and the `dithoff` offset) with `astropy.coordinates.SkyCoord.separation`, which is exact and self-documenting. Fixed in f916a17c1.

pypeit/tests/test_mmt_binospec_comb.py
tepickering: Consolidated. Merged all five Binospec test files (`test_binospec_bpm`, `test_binospec_ifu_extract`, `test_binospec_overscan`, `test_binospec_setup`, and this new comb file) into a single `test_mmt_binospec.py`, grouped by topic and named to match the spectrograph module (`mmt_binospec.py`). This also fixes the inconsistent naming — the new file was the only `test_mmt_binospec_*`; the rest were `test_binospec_*`. 57 tests pass. Fixed in 89392476f.

pypeit/tests/test_mmt_binospec_comb.py
tepickering: Done — added explanatory messages to every assert in the combination-group tests (e.g. 'all same-pointing science frames must share one comb_id so they coadd', 'a large sky-acquisition offset must get its own comb_id', 'bkg_id must stay unset: IFU sky comes from sky fibers, not a nod partner'). Fixed in 89392476f (in the consolidated `test_mmt_binospec.py`).

pypeit/spectrographs/mmt_binospec.py
tepickering: should we file an issue to track this idea?

pypeit/spectrographs/mmt_binospec.py
kbwestfall: Hack as in you're using this approach to add columns to the `PypeItMetaData` object, whereas this `init_meta` function is really about mapping FITS header keywords to pypeit metadata.  Having said that, I would have to dig in more to understand if there's a different way of adding spectrograph-specific metadata columns.  And, of course, all of this is muddied by the long-standing use of `compound_meta`, but even that is meant to produce metadata that is derived from FITS header data.

pypeit/spectrographs/mmt_binospec.py
kbwestfall: Done: https://github.com/pypeit/PypeIt/issues/2196

kbwestfall:
Tests pass
```
Test Summary
--- PYTEST PYPEIT UNIT TESTS PASSED  691 passed, 9432 warnings in 198.94s (0:03:18) ---
--- PYTEST UNIT TESTS PASSED  159 passed, 2802 warnings in 770.13s (0:12:50) ---
--- PYTEST VET TESTS PASSED  81 passed, 291704 warnings in 4340.18s (1:12:20) ---
Total disk usage: 364.528 GiB
Testing Started at 2026-09-04T19:40:00.767924
Testing Completed at 2026-09-05T07:28:59.775451
Total Time: 11:48:59.007527
```

I will merge assuming CI tests pass.

kbwestfall (approved):

